PR #15, opened 2025-06-23
Set up CI

vinistock:
This PR configures CI with some rake tasks, so that we can move forward more confidently while collaborating. I also fixed existing violations.

Note that for the Rust code where we're dereferencing pointers, clippy forces us to mark the function as unsafe and to document why it is unsafe. For now, I just went ahead with a generic comment since I'm not sure we'll even need to keep these functions around.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,7 @@ Rake::ExtensionTask.new("index", GEMSPEC) do |ext|
   ext.lib_dir = "lib/index"
 end
 
-Rake::TestTask.new(test: :compile_all) do |t|
+Rake::TestTask.new(test: :compile) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.ruby_opts << ["--enable=frozen_string_literal"]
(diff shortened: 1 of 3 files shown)


ext/index/rustbindings.h
```diff
@@ -15,8 +15,22 @@ typedef void CEntry;
 
 CRepository *get_repository(void);
 
+/**
```
ChallaHalla: Since this file is generated, should we add it to the .gitignore altogether?

vinistock: Yes, good point.

Rakefile
```diff
@@ -25,7 +20,25 @@ end
...
+end
+
 # Enhance the clean task to also clean Rust artifacts
 Rake::Task[:clean].enhance([:clean_rust])
 
-task default: [:rubocop, :test]
+desc "Compile both Rust and the C extension"
+task :compile_all do
+  Rake::Task[:compile_rust].invoke
+  Rake::Task[:compile].invoke
```
peterzhu2118: I'm not sure this will work if built as a gem. rake-compiler is only used in development and the Rake tasks are not executed at all when installing as a gem. I think you'll need to make sure that the Makefile generated by `extconf.rb` also builds the Rust gem.

vinistock: Moved all of the Rust compilation logic inside the Makefile. Thanks, Peter!

PR #17, opened 2025-06-24
Update extconf.rb to use dev build for Rust library when possible

ChallaHalla:
When building a Rust library, the command `cargo build --release` can be used to build an optimized version of the library. This however comes at a performance cost during compilation time. During development, these bits of time will add up significantly.

This PR enables the Ruby build script to call `cargo build` without the release flag. This will generate a dev build for the Rust lib whenever possible, saving us some time when developing.

### DIFF
--- a/ext/index/extconf.rb
+++ b/ext/index/extconf.rb
@@ -2,13 +2,18 @@ require "mkmf"
 
 # Get the absolute path to the target directory
 root_dir = File.expand_path("../..", __dir__)
-target_dir = File.join(root_dir, "target", "release")
+release = ENV["RELEASE"]
+if release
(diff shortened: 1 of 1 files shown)


ext/index/extconf.rb
```diff
@@ -2,13 +2,18 @@
 
 # Get the absolute path to the target directory
 root_dir = File.expand_path("../..", __dir__)
-target_dir = File.join(root_dir, "target", "release")
+release = ENV["RELEASE"]
```
ChallaHalla, line 5: Using an ENV variable seems like the best way to toggle between build strategies from what I can tell. I am not sure if there's some config Rake compiler has where we can pass an arg when calling `ruby extconf.rb`.

PR #24, opened 2025-06-26
Connect Rust to the Ruby VM through a registry

### LINKED ISSUE #13: Investigate different approaches for sharing index repository data across native code and Ruby layers
Sharing information across layers of the Indexer will be important. Some potential strategies include:
- Wrapping typed data from native code in to Ruby objects
- Instantiating new Ruby objects as needed
- Using an external data store to write/read index data without wrapping or instantiating memory

Each of these approaches should be explored and validated before moving forward.

vinistock:
Closes #13

## Context

A more traditional approach to building a native extension would be to back Ruby objects directly with C structs that contain the hidden data we want. Unfortunately, many of the great things we get with Rust aren't available in C and so we would lose the ability to use them or even read the data properly on the C side (e.g.: Rust strings, hashmaps or vectors).

It is possible to back the Ruby objects with a void pointer instead of a struct. In this case, we are basically telling C "ignore the structure of this data, just know that the object is backed by a package of foreign data".

Unfortunately, this option does not play well with Rust either because of its built-in free after leaving scope. If we transform the void pointer into a Rust object at any point (which is guaranteed to be necessary as we want to use the objects on the Rust side), then Rust will automatically free the pointer as it goes out of scope, leaving the Ruby VM with an object that is backed by a freed pointer - leading to all sorts of corruption.

For example:

```rust
// A function on the Rust side that would try to do something using the object already
// registered on the Ruby VM
#[no_mangle]
pub extern "C" fn do_something(pointer: *c_void) {
  // We use Box::from_raw to transform the pointer we stored in the Ruby VM back into
  // a Rust object
  let rust_version_of_the_object = unsafe { Box::from_raw(pointer) };

  // Once this function body ends, `rust_version_of_the_object` goes out of scope and
  // gets automatically freed by Rust, but that is way too early for Ruby, which still
  // depends on the pointer
}
```

I discussed this in detail with Peter and we landed on two decisions:

1. The Ruby VM needs to be able to control the life cycle of Ruby objects. This seems obvious, but with that void pointer approach it's easy to mistakenly pass control to the Rust side
2. We also signed off on the implementation below

## Implementation

The idea behind this implementation is to have a registry of repository objects. Basically, all of the persistent data for the index will live inside of them and the static registry allows us to maintain the Rust data fully independent from the Ruby VM side, while still allowing the VM to control when objects get freed.

In the C side we:

1. Define a custom allocator for the repository objects that will automatically register them in the Rust side. We also take the opportunity to register a free function for the cleanup
2. The free function frees the Ruby object after deregistering its data from the Rust side

## Thoughts

This approach seems a bit odd at first, but it has advantages and I think it's the way to go:

1. Full separation of the data. The Ruby objects are extra simple. They are connected by an ID and nothing else. On the Rust side, we can do whatever we want and use all of the modern data structures to our benefit. As long as we return simple data to the C side, it doesn't care
2. Because of 1, we minimize the surface area where we need to translate data types
3. At first glance, keeping a static registry might seem like a memory concern, but notice that this is mainly only to keep Ruby semantics. In real usage of the index, you would not have more than one `Repository` object and thus the normal mode of operation would be for the hashmap to have a single entry at all times. The only reason to have the map is so that the Ruby API works as expected (i.e.: you can have multiple repositories for testing and that's okay).


src/lib.rs
```diff
@@ -1,2 +1,28 @@
+use std::{collections::HashMap, sync::LazyLock, sync::Mutex};
+
 pub mod c_interface;
-pub mod internal;
+
+static REGISTRY: LazyLock<Mutex<HashMap<usize, Repository>>> = LazyLock::new(|| Mutex::new(HashMap::new()));
```
ChallaHalla: So from my understanding, the underlying purpose of this static `REGISTRY` is mainly to make sure that the memory allocated in Rust isn't prematurely freed as Ruby is using it. Is that correct?

ext/index/repository.c
```diff
@@ -0,0 +1,73 @@
+#include "ruby.h"
+#include "rustbindings.h"
+
+static VALUE cRepository;
+
+// The repository struct only holds the registry ID that connects to the Rust data structures
+typedef struct Repository {
+    size_t registry_id;
```
ChallaHalla: Ahh I see so passing an ID like this allows us to circumvent needing to constantly pass back and forth the repository across layers. Would we essentially do the same thing for other entities we need across both layers like `Entry`?

src/lib.rs
vinistock: Yes, that's correct. The only other way to share data would be through:

1. Only using C compatible data structures on the Rust side, which makes us lose many of the benefits of using Rust
2. Using void pointers, but then Rust ends up freeing them prematurely

ext/index/repository.c
vinistock: Yes, that's right. We would use whatever uniquely identifies an object to communicate with the Rust registry. If it's an entry's name, for example, then we would have an API that looks something like this:

```rust
#[no_mangle]
pub extern "C" fn get_entry(registry_id: usize, name: &mut &mut char) {
  // Get repository object from hashmap
  // Find entry based on the name
}
```

ChallaHalla: Okay cool that makes sense!

Mainly thinking aloud now -- I'm trying to wrap my head around how these would work:
- Chains of references (e.g for ownership)
- Sharing strings with the Ruby VM


For example, if we are looking up chained references for ownership, we could have an chain of entries that point to each other via an `owner` attribute. I suppose that we can implement fetching the owner via a method in C/Rust where we take the owner's unique identifier and perform a look up in the Registry. The downside is that we need to cross the Ruby VM layer in order to get the chained reference, but we maintain the separation of data across layers. Another possible solution could be implementing actions that require chained reference lookups within the Indexer instead of the language server, but that would only cover these sorts of operations that exist within Ruby LSP and not other tools using the shared indexer.

For sharing strings, it seems to me like we have to eat the cost of converting the Rust/C strings to Ruby strings when wrapping the memory using `TypedData_Wrap_Struct` since afaik we can't pre-convert to Ruby strings at the Rust layer.

vinistock: I think for the references thing, we will need to use IDs indeed. We'll need quick lookups both on the Rust and on the Ruby side.

For strings, there's nothing we can do, we need to convert.

ext/index/repository.c
```diff
@@ -0,0 +1,73 @@
...
+    }
+    free(converted_file_paths);
+    rb_gc_unregister_address(&file_paths);
+    return Qnil;
+}
+
+void initialize_repository(VALUE mIndex) {
+    cRepository = rb_define_class_under(mIndex, "Repository", rb_cObject);
+
+    rb_define_alloc_func(cRepository, rb_repository_alloc);
```
ChallaHalla, line 59: https://blog.peterzhu.ca/ruby-c-ext-part-7/#:~:text=can%20use%20the-,rb_define_alloc_func,-function.%20It%20accepts

This is neat!

ext/index/repository.c
```diff
@@ -0,0 +1,61 @@
+#include "ruby.h"
+#include "rustbindings.h"
+
+static VALUE cRepository;
+
+static void repository_free(void *ptr) {
```
Morriar, line 6: Should this be moved to Rust? Something like this:

```rust
#[no_mangle]
pub extern "C" fn repository_free(ptr: *mut Repository) {
    if !ptr.is_null() {
        unsafe {
            Box::from_raw(ptr);
        }
    }
}
```

And then C can call it: 

```c
struct Repository; // Opaque type

extern struct Repository* repository_new(int data);
extern void repository_free(struct Repository* repo);

void use_repo() {
    struct Repository* repo = repository_new(...);
    // ... use it
    repository_free(repo);
}
```

This would give us more control on the Rust side.

src/c_interface.rs
```diff
@@ -1,105 +1,34 @@
...
 }
 
 #[unsafe(no_mangle)]
-pub extern "C" fn dealloc_repository(repository: *mut CRepository) {
-    if !repository.is_null() {
-        unsafe {
-            let _ = Box::from_raw(repository as *mut Repository);
-        }
-    }
+pub extern "C" fn new_repo() -> *mut c_void {
```
Morriar, line 32: Does it need to be `c_void`? Can't we return an opaque extern struct?

src/c_interface.rs
```diff
@@ -1,105 +1,34 @@
...
 }
 
 #[unsafe(no_mangle)]
-pub extern "C" fn delloc_entry(entry: *mut CEntry) {
-    if !entry.is_null() {
-        unsafe {
-            let _ = Box::from_raw(entry as *mut Entry);
-        }
-    }
+pub extern "C" fn index_all_c(pointer: *mut c_void, file_paths: &mut &mut c_char, count: usize) {
```
Morriar, line 25: Same question here, does it need to be `c_void`?

PR #25, opened 2025-06-27
Add a pool implementation

vinistock:
This PR adds a pool implementation to keep data avoiding duplication, the need of cloning (other than reference cloning) and type-safe IDs.

We can use a data structure like this one for URIs, names or whatever else we need to store and it will not allow for duplicate entries to be inserted. Other data structures can store usize IDs exclusively and not have to keep cloned strings around.

Additionally, by setting an empty type as the ID used for each pool, we can avoid simple mistakes during development like accidentally using an ID from one pool to index another one.

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -1,56 +1,99 @@
 use std::collections::HashMap;
-use std::rc::Rc;
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::marker::PhantomData;
+use std::sync::Arc;
 
 /// A pool of unique items stored by indexed to avoid duplication. Pools can be used to store unique names, like
 /// `Foo::Bar#method`, file paths or any unique value we need to store avoiding cloning and minimizing memory usage.
 ///
-/// We use two hashes with reference counters to guarantee uniqueness without cloning and fast lookups both ways
-pub struct Pool<T: Eq + std::hash::Hash> {
-    id_to_entry: HashMap<usize, Rc<T>>,
-    entry_to_id: HashMap<Rc<T>, usize>,
-    next_id: usize,
+/// We use two hashes with reference counters to guarantee uniqueness without cloning and fast lookups both ways. We
+/// also require defining an ID type for the pool, so that IDs belonging to different pools cannot be mixed up and a
+/// type error will be thrown during compilation.
+///
+/// ```
+/// # #[derive(PartialEq, Debug)]
+/// # pub struct NameId;
+/// # #[derive(PartialEq, Debug)]
+/// # pub struct UriId;
+/// #
+/// # let mut name_pool = Pool::<NameId, String>::new();
+/// # let mut uri_pool = Pool::<UriId, String>::new();
+/// # let name_id = name_pool.add("Foo::Bar#method".into());
+/// # uri_pool.get(&name_id) // Not possible. This will fail to compile because of the ID types
+/// ```
+pub struct Pool<I, T: Eq + Debug + Hash> {
+    id_to_entry: HashMap<u32, Arc<T>>,
+    entry_to_id: HashMap<Arc<T>, u32>,
+    next_id: u32,
+    _phantom: PhantomData<I>,
 }
 
-impl<T: Eq + std::hash::Hash> Pool<T> {
+#[derive(PartialEq, Debug)]
+pub struct PoolId<I: PartialEq> {
+    value: u32,
+    _marker: PhantomData<I>,
+}
+
+impl<I: PartialEq> From<u32> for PoolId<I> {
+    fn from(id: u32) -> Self {
+        Self {
+            value: id,
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<I: PartialEq> From<PoolId<I>> for u32 {
+    fn from(id: PoolId<I>) -> Self {
+        id.value
+    }
+}
+
+impl<I: PartialEq, T: Eq + Debug + Hash> Pool<I, T> {
     pub fn new() -> Self {
         Pool {
             id_to_entry: HashMap::new(),
             entry_to_id: HashMap::new(),
             next_id: 0,
+            _phantom: PhantomData,
         }
     }
 
-    pub fn add(&mut self, item: T) -> usize {
+    pub fn add(&mut self, item: T) -> PoolId<I> {
         // If the item already exists, skip adding it and just return its ID
         if let Some(&id) = self.entry_to_id.get(&item) {
-            return id;
+            return id.into();
         }
 
-        let item_rc = Rc::new(item);
+        let item_arc = Arc::new(item);
 
         let id = self.next_id;
-        self.id_to_entry.insert(id, item_rc.clone());
-        self.entry_to_id.insert(item_rc, id);
+        self.id_to_entry.insert(id, item_arc.clone());
+        self.entry_to_id.insert(item_arc, id);
         self.next_id += 1;
-        id
+        id.into()
     }
 
-    pub fn get(&self, id: usize) -> Option<&T> {
-        self.id_to_entry.get(&id).map(|rc| rc.as_ref())
+    pub fn get(&self, id: &PoolId<I>) -> Option<&T> {
+        self.id_to_entry.get(&id.value).map(|rc| rc.as_ref())
     }
 
-    pub fn remove(&mut self, id: usize) {
-        if let Some(entry) = self.id_to_entry.remove(&id) {
+    pub fn remove(&mut self, id: &PoolId<I>) -> Option<T> {
+        if let Some(entry) = self.id_to_entry.remove(&id.value) {
             self.entry_to_id.remove(&entry);
+            return Some(Arc::try_unwrap(entry).unwrap());
         }
+
+        None
     }
 
-    pub fn size(&self) -> usize {
-        self.id_to_entry.len()
+    pub fn size(&self) -> u32 {
+        self.id_to_entry.len().try_into().unwrap()
     }
 }
 
-impl<T: Eq + std::hash::Hash> Default for Pool<T> {
+impl<I: PartialEq, T: Eq + Debug + Hash> Default for Pool<I, T> {
     fn default() -> Self {
         Self::new()
     }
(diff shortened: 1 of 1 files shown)


src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+/// # let mut name_pool = Pool::<NameId, String>::new();
+/// # let mut uri_pool = Pool::<UriId, String>::new();
+/// # let name_id = name_pool.add("Foo::Bar#method".into());
+/// # uri_pool.get(&name_id) // Not possible. This will fail to compile because of the ID types
+/// ```
+pub struct Pool<I, T: Eq + std::hash::Hash> {
+    id_to_entry: HashMap<usize, Rc<T>>,
+    entry_to_id: HashMap<Rc<T>, usize>,
+    next_id: usize,
+    _phantom: PhantomData<I>,
```
ChallaHalla, line 29: What's the purpose of the `PhantomData` here?

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+/// # pub struct UriId;
+/// #
+/// # let mut name_pool = Pool::<NameId, String>::new();
+/// # let mut uri_pool = Pool::<UriId, String>::new();
+/// # let name_id = name_pool.add("Foo::Bar#method".into());
+/// # uri_pool.get(&name_id) // Not possible. This will fail to compile because of the ID types
+/// ```
+pub struct Pool<I, T: Eq + std::hash::Hash> {
+    id_to_entry: HashMap<usize, Rc<T>>,
+    entry_to_id: HashMap<Rc<T>, usize>,
```
ChallaHalla: I haven't taken a look at the parallel indexing PR yet. But for parallelizing access to the Pool, would we need to replace the `Rc` with `Arc` so that they are atomic ref counters?

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+        let mut pool: Pool<UniqueNameId, String> = Pool::new();
+        let id = pool.add("Foo::Bar#method".into());
+        let item = pool.get(&id).unwrap();
+        assert_eq!(*item, "Foo::Bar#method".to_string());
+    }
+
+    #[test]
+    fn test_adding_the_same_entry_just_returns_id() {
+        let mut pool: Pool<UniqueNameId, String> = Pool::new();
+        let id = pool.add("Foo::Bar#method".into());
```
ChallaHalla, line 130: Adding this more to solidify my own Rust knowledge so feel free to ignore.

Is the `.into()` here essentially cloning the String literal on the stack to the heap? Would it be more explicit to call `.clone()` in such situations?

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+/// ```
+pub struct Pool<I, T: Eq + std::hash::Hash> {
+    id_to_entry: HashMap<usize, Rc<T>>,
+    entry_to_id: HashMap<Rc<T>, usize>,
+    next_id: usize,
+    _phantom: PhantomData<I>,
+}
+
+#[derive(PartialEq, Debug)]
+pub struct PoolId<I: PartialEq> {
```
Morriar, line 33: Do we really need this bound?

```suggestion
pub struct PoolId<I> {
```

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+/// # uri_pool.get(&name_id) // Not possible. This will fail to compile because of the ID types
+/// ```
+pub struct Pool<I, T: Eq + std::hash::Hash> {
+    id_to_entry: HashMap<usize, Rc<T>>,
+    entry_to_id: HashMap<Rc<T>, usize>,
+    next_id: usize,
+    _phantom: PhantomData<I>,
+}
+
+#[derive(PartialEq, Debug)]
```
Morriar, line 32: I wonder if we should add a few more derives here:

```suggestion
#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
```

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+        self.entry_to_id.insert(item_rc, id);
+        self.next_id += 1;
+        id.into()
+    }
+
+    pub fn get(&self, id: &PoolId<I>) -> Option<&T> {
+        self.id_to_entry.get(&id.value).map(|rc| rc.as_ref())
+    }
+
+    pub fn remove(&mut self, id: &PoolId<I>) {
```
Morriar: Should we make the `remove` return the value or `None`?

```suggestion
    pub fn remove(&mut self, id: &PoolId<I>) -> option<T> {
```

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+/// # #[derive(PartialEq, Debug)]
+/// # pub struct UriId;
+/// #
+/// # let mut name_pool = Pool::<NameId, String>::new();
+/// # let mut uri_pool = Pool::<UriId, String>::new();
+/// # let name_id = name_pool.add("Foo::Bar#method".into());
+/// # uri_pool.get(&name_id) // Not possible. This will fail to compile because of the ID types
+/// ```
+pub struct Pool<I, T: Eq + std::hash::Hash> {
+    id_to_entry: HashMap<usize, Rc<T>>,
```
Morriar: Do we expect more than 4 billions entries in the same pool? Could we use a `u32` instead?

```suggestion
    id_to_entry: HashMap<u32, Rc<T>>,
```

src/pool.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub struct OtherId;
+
+    #[test]
+    fn test_adding_and_fetching_entries() {
+        let mut pool: Pool<UniqueNameId, String> = Pool::new();
+        let id = pool.add("Foo::Bar#method".into());
+        let item = pool.get(&id).unwrap();
+        assert_eq!(*item, "Foo::Bar#method".to_string());
+    }
+
```
Morriar, line 119: Let's add a test showing retrieving an invalid ID returns `None`?

src/pool.rs
Morriar: I agree, `Arc` may be better.

Or I wonder if we could make this generic: 

```rust
pub struct Pool<I, T, S>
where
    T: Eq + Hash,
    S: Clone + Deref<Target = T> + Eq + Hash,
{
    id_to_entry: HashMap<usize, S>,
    entry_to_id: HashMap<S, usize>,
    next_id: usize,
    _phantom: PhantomData<I>,
}
```

So the client can choose:

```rust
Pool<NameId, String, Rc<String>>
```

And provide type aliases:

```rust
pub type RcPool<I, T> = Pool<I, T, Rc<T>>;
pub type ArcPool<I, T> = Pool<I, T, Arc<T>>;
```

src/pool.rs
Morriar, line 29: I wonder if `_marker` would be more idiomatic?

Morriar: Good read: https://doc.rust-lang.org/nomicon/phantom-data.html

vinistock: ☝️ highly recommend the read, but the idea is that if you simply add a type parameter `T` to a struct without using it, Rust is not happy. The `PhantomData` thing allows you to "use" the parameter `T` without actually using it.

It's basically a trick for the type checker to allow us differentiating based on `T`, but without actually having to add more data to the struct.

src/pool.rs
vinistock: That's a good question and I'm not completely sure. I assumed that this would be equivalent to `to_string()`, since we don't accept string literals, but only strings in the heap.

src/pool.rs
vinistock: I switched to using `Arc` for now. It's a bit early to tell if we're going to need an `Rc` and an `Arc` version of the pool, so I'd vote for us to revisit later if we realize there's a need.

src/pool.rs
vinistock: It's needed to compare the IDs, otherwise Rust doesn't know that they can be checked with `==`.

src/pool.rs
vinistock: I'd vote for us to add more as we realize we need them.

src/pool.rs
vinistock: Not really, I was just trying to avoid conversions because vectors use `usize` for indices. I changed to u32.

src/pool.rs
vinistock: I don't know to be honest, but I switched to `_marker`.

src/pool.rs
vinistock: I don't know if we'll need it, but I added it as it better matches the behaviour of hashmaps and vectors.

src/pool.rs
```diff
@@ -0,0 +1,145 @@
...
+/// ```
+pub struct Pool<I, T: Eq + Debug + Hash> {
+    id_to_entry: HashMap<u32, Arc<T>>,
+    entry_to_id: HashMap<Arc<T>, u32>,
+    next_id: u32,
+    _phantom: PhantomData<I>,
+}
+
+#[derive(PartialEq, Debug)]
+pub struct PoolId<I: PartialEq> {
```
ChallaHalla, line 33: Is there a reason we wrap the u32 ID value in a struct VS just copying around the u32 ID as needed? My guess is that there's some advantage to passing around the reference instead of copying the value constantly -- is that correct?

vinistock: The only goal is to leverage the type system to avoid accidentally trying to access a pool using an index that belongs to a different pool.

For example:

```rust
// Say we need to remember unique entry names and unique URIs we have indexed
let mut name_pool = Pool::<NameId, String>::new();
let mut uri_pool = Pool::<UriId, String>::new();

// Entries and other entities will store IDs and not the strings that they refer to
let name_id = name_pool.add("Foo::Bar#method".into());

// The empty structs that wrap the u32 values provide us type safety that we cannot
// accidentally use an ID from one pool to index into another pool. This will fail
// type checking and prevent compilation:
uri_pool.get(&name_id)
```

src/pool.rs
Morriar: We need the _derive_ on `PartialEq` but not the _bound_ right?

PR #27, opened 2025-07-04
Add script to count symbols from Ruby files

Morriar:
Useful to get a high level picture of what's in a codebase and how our things scale in comparison.

Usage:

    $ ./utils/symbol-metrics <path> [<path>...]

### DIFF
--- a/utils/symbol-metrics
+++ b/utils/symbol-metrics
@@ -1,0 +1,117 @@
+#! /usr/bin/env ruby
+# frozen_string_literal: true
+
+# This script counts the number of symbols found in Ruby files.
+
+if ARGV.empty?
+  puts "Usage: #{$PROGRAM_NAME} <path> [<path>...]"
+  exit 1
+end
+
+require "bundler/inline"
+
+gemfile do
+  source "https://rubygems.org"
+
+  gem "prism"
+end
+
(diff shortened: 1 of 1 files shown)


utils/symbol-metrics
```diff
@@ -0,0 +1,112 @@
...
+
+  def visit_class_variable_write_node(node)
+    @counters[:class_variables] += 1
+    super
+  end
+
+  def visit_local_variable_write_node(node)
+    @counters[:local_variables] += 1
+    super
+  end
```
vinistock, line 75: We're missing global variables

utils/symbol-metrics
```diff
@@ -0,0 +1,112 @@
...
+end
+
+paths = ARGV.map do |path|
+  if File.directory?(path)
+    Dir.glob(File.join(path, "**/*.{rb,rbi}")).select { |file| File.file?(file) }
+  else
+    [path]
+  end
+end.flatten
```
vinistock:
```suggestion
paths = ARGV.flat_map do |path|
  if File.directory?(path)
    Dir.glob(File.join(path, "**/*.{rb,rbi}")).select { |file| File.file?(file) }
  else
    [path]
  end
end
```

PR #28, opened 2025-07-04
Add script to generate random corpus of different sizes

Morriar:
Script that generates a corpus of random Ruby files with a given scale.

Useful when testing thing like speed or memory usage against a large amount of files/classes/methods.

Usage:

```bash
$ ./generate-corpus tmp/corpus/tiny -s 0.1
$ ./generate-corpus tmp/corpus/small -s 1
$ ./generate-corpus tmp/corpus/medium -s 10
$ ./generate-corpus tmp/corpus/large -s 100
$ ./generate-corpus tmp/corpus/huge -s 1000
```

I'd avoid going over a scale of 5000: 500K files, 5M classes, 50M methods, 50M variables and around 10Go on disk 🤯

### DIFF
--- a/utils/generate-corpus
+++ b/utils/generate-corpus
@@ -1,0 +1,265 @@
+#! /usr/bin/env ruby
+# frozen_string_literal: true
+
+# Generate a corpus of random Ruby files with a given scale.
+#
+# Usage:
+#   utils/generate-corpus tmp/corpus/tiny -s 0.1
+#   utils/generate-corpus tmp/corpus/small -s 1
+#   utils/generate-corpus tmp/corpus/medium -s 10
+#   utils/generate-corpus tmp/corpus/large -s 100
+#   utils/generate-corpus tmp/corpus/huge -s 1000
+
+require "bundler/inline"
+
+gemfile do
+  source "https://rubygems.org"
+
+  gem "thor"
+end
+
+require "fileutils"
+require "stringio"
+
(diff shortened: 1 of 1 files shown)


utils/generate-corpus
```diff
@@ -0,0 +1,265 @@
...
+# frozen_string_literal: true
+
+# Generate a corpus of random Ruby files with a given scale.
+#
+# Usage:
+#   ./generate-corpus tmp/corpus/tiny -s 0.1
+#   ./generate-corpus tmp/corpus/small -s 1
+#   ./generate-corpus tmp/corpus/medium -s 10
+#   ./generate-corpus tmp/corpus/large -s 100
+#   ./generate-corpus tmp/corpus/huge -s 1000
```
alexcrocha:
```suggestion
#   ./utils/generate-corpus tmp/corpus/tiny -s 0.1
#   ./utils/generate-corpus tmp/corpus/small -s 1
#   ./utils/generate-corpus tmp/corpus/medium -s 10
#   ./utils/generate-corpus tmp/corpus/large -s 100
#   ./utils/generate-corpus tmp/corpus/huge -s 1000
```
Making it easy to copy paste and run from the project root. What do you think? 
This way the corpus lands in the root `tmp/` directory, which is covered by `.gitignore`.

Morriar: Good idea, updated 👍

PR #33, opened 2025-07-08
Benchmarks for crossing VM boundary

ChallaHalla:


ext/index/index.c
```diff
@@ -85,4 +159,19 @@ void Init_index(void) {
...
+    cPoint = rb_define_class_under(mIndex, "Point", rb_cObject);
+    rb_define_singleton_method(cPoint, "new", rb_create_point, 2);
+    rb_define_attr(cPoint, "x", 1, 0);  
+    rb_define_attr(cPoint, "y", 1, 0);  
+
+    cMessage = rb_define_class_under(mIndex, "Message", rb_cObject);
+    // NOTE: Choosing to use definition via singleton method instead of 
+    // defining an alloc function becasue rb_define_alloc_func doesn't 
+    // take any arguments after the klass.
+    rb_define_singleton_method(cMessage, "new", rb_create_message, 1);
```
ChallaHalla, line 175: Could potentially redefine the alloc function here, and then update the initialize method to populate the struct data.

PR #34, opened 2025-07-10
Introduce `StringPool` implementation

Morriar:
Alternative implementation for a string pool.

This one is a bit more efficient than #25 because we avoid duplicating string objects and use a single string buffer for interning.

I tried multiple implementation deriving from #25 including a comparison with [`string_interner`] (https://github.com/Robbepop/string-interner) against `StringPool` (proposed implementation in this PR).

Here's the Maximum Resident Set Size (memory) comparison in MB:

|      | tiny | small | medium | large  | huge     |
| :-------------- | :--: | :---: | :----: | :----: | :------: |
| #25            | 3.09 | 10.42 | 48.59  | 469.68 | 3,350.78 |
| vec\<String\>      | 2.81 | 7.65  | 40.89  | 434.65 | 3,279.26 |
| vec\<box\<str\>\>  | 2.60 | 7.64  | 37.55  | 328.62 | 2,533.56 |
| [`string_interner`](https://github.com/Robbepop/string-interner) | 2.26 | 6.07  | 25.82  | 167.14 | 959.93   |
| StringPool        | 2.25 | 5.93  | 20.21  | 153.32 | 815.92   |

We went from 3.3GB on `huge` to 815MB!

I also compared insertion time (s):

|      | tiny | small | medium | large | huge  |
| :------------- | :--: | :---: | :----: | :---: | :---: |
| #25             | 0.17 | 0.23  | 0.55   | 4.75  | 56.73 |
| vec String      | 0.01 | 0.02  | 0.28   | 3.13  | 49.19 |
| vec box\<str\>  | 0.01 | 0.02  | 0.27   | 3.14  | 49.23 |
| [`string_interner`](https://github.com/Robbepop/string-interner) | 0.01 | 0.02  | 0.24   | 2.71  | 43.48 |
| StringPool        | 0.01 | 0.02  | 0.24   | 2.57  | 42.14 |

We went from 56s on `huge` to 42s (single thread).

The main drawback is that this approach doesn't support deletion. We just continuously add to the internal buffer. We _could_ implement some kind of compaction if needed but, considering we need to front-load the codebase and do deletions only as the user edit the code, I think the memory gains on insertion outbalance the loss on deletion.

Closes #25.

### DIFF
--- a/src/string_pool.rs
+++ b/src/string_pool.rs
@@ -60,12 +60,24 @@ use hashbrown::hash_map::RawEntryMut;
 ///         self.0
 ///     }
 /// }
-pub trait PoolId {
+pub trait PoolId: From<u32> + Into<u32> {
     /// Creates a new pool identifier from a u32 value.
     fn from_u32(id: u32) -> Self;
 
     /// Converts this pool identifier to a u32 value.
     fn to_u32(&self) -> u32;
+}
+
+impl<T> PoolId for T
+where
+    T: Copy + From<u32> + Into<u32>,
+{
+    fn from_u32(id: u32) -> Self {
+        Self::from(id)
+    }
+    fn to_u32(&self) -> u32 {
+        (*self).into()
+    }
 }
 
 /// A string pool implementation with efficient string interning.
@@ -162,7 +174,7 @@ impl<I: PoolId> StringPool<I> {
         };
 
         // Step 4: Convert the internal u32 ID to the user's ID type
-        I::from_u32(id)
+        id.into()
     }
 
     /// Retrieves a string from the pool by its ID.
@@ -215,20 +227,11 @@ impl<I: PoolId> StringPool<I> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use derive_more::{From, Into};
 
     // Test ID type for testing the Pool
-    #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+    #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, From, Into)]
     struct TestId(u32);
-
-    impl PoolId for TestId {
-        fn from_u32(id: u32) -> Self {
-            Self(id)
-        }
-
-        fn to_u32(&self) -> u32 {
-            self.0
-        }
-    }
 
     #[test]
     fn pool_new_creates_empty_pool() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
+name = "derive_more"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "093242cf7570c207c83073cf82f79706fe7b8317e98620a47d5be7c3d8497678"
+dependencies = [
+ "derive_more-impl",
+]
+
+[[package]]
+name = "derive_more-impl"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bda628edc44c4bb645fbe0f758797143e4e07926f7ebf4e9bdfbd3d2ce621df3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -197,6 +217,7 @@ version = "0.1.0"
 dependencies = [
  "ahash",
  "cbindgen",
+ "derive_more",
  "hashbrown 0.14.5",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 ahash = "0.8"
+derive_more = { version = "2.0.1", features = ["from", "into"] }
 hashbrown = "0.14"
 
 [build-dependencies]


src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
+//! String pooling and interning implementation.
+//!
+//! This module provides a `StringPool` that implements efficient string interning,
+//! where identical strings are stored only once and referenced by unique IDs.
+//!
+//! This implementation is strongly inspired by the `string_interner` crate.
+//! See <https://github.com/Robbepop/string-interner>.
```
vinistock, line 7: Do we need any license notices for this? I know compiled projects are different than something like Ruby and you end up having licenses like APACHE which has different requirements on the final linked product.

Just asking, I have no idea and maybe we need to check with legal.

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+//! use index::string_pool::StringPool;
+//!
+//! // Define a custom ID type
+//! #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+//! struct NameId(u32);
+//!
+//! impl PoolId for NameId {
+//!     fn from_u32(id: u32) -> Self { NameId(id) }
+//!     fn to_u32(&self) -> u32 { self.0 }
+//! }
```
vinistock, line 22: Do these have to be implemented every time?

Can we add a default implementation in the PoolId trait instead and then we just need to do

```rust
impl PoolId for MyId;
```

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+//! assert_eq!(pool.get_str(&id1), Some("Foo"));
+//! assert_eq!(pool.get_str(&id4), Some("Baz"));
+//! ```
+
+use core::fmt::Debug;
+use core::marker::PhantomData;
+
+use ahash::RandomState; // For faster hashing
+use hashbrown::hash_map::RawEntryMut;
+use hashbrown::HashMap; // This is what is used to implement std::collections::HashMap but we need access to the RawEntryMut trait
```
vinistock: Nit: I'm okay with removing these comments if people agree.

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+            hasher: RandomState::new(),
+            _marker: PhantomData,
+        }
+    }
+
+    /// Creates a new string pool with the specified capacity.
+    ///
+    /// This pre-allocates memory for the expected number of unique strings,
+    /// which can improve performance when the approximate size is known.
+    pub fn with_capacity(cap: usize) -> Self {
```
vinistock: Just out of curiosity, do you have an idea of how we could estimate the number of strings?

Maybe on first-ever indexing we can't, but after it gets cached we could have a pretty close number.

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+    /// string interning - identical strings share the same ID.
+    ///
+    /// # Arguments
+    ///
+    /// * `item` - The string to add to the pool
+    ///
+    /// # Returns
+    ///
+    /// The ID of the string in the pool
+    pub fn add(&mut self, item: String) -> I {
```
vinistock: My understanding is that any method that accepts `&str` can also handle a `String`. Do we need this method? Am I misunderstanding?

I'd vote for a single add method if we can get away with it.

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+                vacant.insert_with_hasher(hash, id as u32, (), |id| {
+                    // Rehashing function: reconstruct string from buffer and hash it
+                    let span_str = Self::span_to_str(&self.buffer, &self.ends, *id);
+                    self.hasher.hash_one(span_str)
+                })
+            }
+        };
+
+        // Step 4: Convert the internal u32 ID to the user's ID type
+        I::from_u32(id)
```
vinistock: Would `id.into()` work here?

src/string_pool.rs
```diff
@@ -0,0 +1,765 @@
...
+    /// Retrieves a string from the pool by its ID.
+    ///
+    /// # Arguments
+    ///
+    /// * `id` - The ID of the string to retrieve
+    ///
+    /// # Returns
+    ///
+    /// `Some(String)` if the ID exists in the pool, `None` otherwise
+    pub fn get(&self, id: &I) -> Option<String> {
```
vinistock: Similarly, I would vote for us to start with a smaller API and expand if we see the need. Callers can decide if they need a string instead of a slice.

src/string_pool.rs
Morriar: I believe this comment is enough in the sense that we didn't copy the code, only the idea but we do attribute ownership of the idea.

It also seems that `string_interner` got the idea from another project: https://github.com/Robbepop/string-interner/blob/master/src/backend/string.rs#L12-L13.

src/string_pool.rs
Morriar: I was thinking we could approximate based on the number of files we'll need to index? 

I can remove the method for now 👍

src/string_pool.rs
Morriar: This was to avoid the copy when handling string slices but I can remove.

src/string_pool.rs
Morriar: Agreed I'll remove 👍

src/string_pool.rs
Morriar: I couldn't find an optimal solution for this.

I pushed another commit with a proposal: https://github.com/Shopify/index/pull/34/commits/a404e6f.

We provide a blanket implementation for `PoolId` on anything that can `From<u32>` and `Into<u32>` allowing us to create ids just by deriving from `From` and `Into`:

```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, From, Into)]
struct NameId(u32);
```

But... it requires `derive_more`.

src/string_pool.rs
Morriar: Yes if we agree on the direction proposed in the last commit

src/string_pool.rs
vinistock: I would vote to avoid the dependency honestly. The trait strategy doesn't work? Even if a default implementation doesn't work, then I still vote to avoid the dependency and implement from and into every time.

PR #36, opened 2025-07-10
Add basic parallel indexing logic

vinistock:
Add _example_ Rust to Ruby VM bindings to the repository, so that people can use this while working on their prototypes. The keyword here is _example_. 

This PR demonstrates how we can share an `Arc` of `Repository` with a Ruby object and how we can return a boxed value to produce a new Ruby `Declaration` object. That doesn't mean we'll actually need this exactly as is. Until we have settled on the data structures, we have no way of knowing exactly what we'll need. This is useful nonetheless so that people can test their prototypes with Ruby scripts and see if they actually do what they want.

The PR includes:

- Parallel indexing
- A basic visitor that captures classes and modules (notice that the name part is incorrect, it doesn't handle nesting yet)
- The ability to add or get declarations from Ruby

I didn't stress over adding a bunch of tests since this code is a temporary example and will eventually be substituted by the real implementation.

The most important part of this PR, and what I would love for people to focus on, is the documentation of the `c_interface` file. Do you agree with everything in there? Is it clear enough for you to feel empowered to add more bindings to the Ruby VM? How can we ensure that the details are as easy to follow as possible?

**Unfortunate note**

Windows CI will be broken until https://github.com/ruby/prism/pull/3594 is merged (provided it actually fixes the issue, which I'm only between 60-70% sure).


cbindgen.toml
```diff
@@ -7,3 +7,6 @@ autogen_warning = "/* Warning, this file is autogenerated by cbindgen. Don't mod
 tab_width = 4
 
 usize_is_size_t = true
+
+[enum]
+rename_variants = "ScreamingSnakeCase"
```
vinistock, line 12: Enums in C are usually all in caps, so I added this to make sure cbindgen will transform them for us.

ext/index/index.c
```diff
@@ -1,88 +1,10 @@
 #include "ruby.h"
 #include "rustbindings.h"
-#include <stdlib.h>
-#include <string.h>
+#include "repository.h"
 
-// Store the class references
```
vinistock, line 6: C extensions are usually organized the same way as CRuby: each file contains everything related to a specific class/module and the file name is the same as the class/module name.

I moved everything to `repository.c`.

lib/index/declaration.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Index
+  # Represents the base declaration class for all indexed items. This class is partially defined in C to connect with
+  # the Rust layer
+  class Declaration
```
vinistock, line 6: Not strictly necessary, but a good example of how we can mix C and Ruby for defining the API. In reality, we only need C to access the Rust side of things. Simple things that aren't performance critical can be defined in Ruby.

src/lib.rs
```diff
@@ -1,2 +1,45 @@
...
+use crate::{repository::Repository, ruby_indexer::RubyIndexer};
+
 pub mod c_interface;
-pub mod internal;
+mod declaration;
+mod repository;
+mod ruby_indexer;
+
+// Index all of the given file paths in parallel, populating the repository with the results
+pub fn index_in_parallel(repository: &Arc<Mutex<Repository>>, file_paths: &Arc<Mutex<Vec<String>>>) {
```
vinistock, line 14: Credit to Sid, who actually implemented this originally.

src/ruby_indexer.rs
```diff
@@ -0,0 +1,62 @@
...
+
+impl<'a> RubyIndexer<'a> {
+    pub const fn new(repository: &'a Arc<Mutex<Repository>>) -> Self {
+        Self {
+            nesting: Vec::new(),
+            repository,
+        }
+    }
+
+    pub fn index(&mut self, file_path: String) {
```
Morriar, line 19: I wonder if we should return a `Result<(), std::io::Error>` in case we can't read the file?

ext/index/repository.c
```diff
@@ -0,0 +1,122 @@
...
+}
+
+static VALUE rb_repository_index_all(VALUE self, VALUE file_paths) {
+    rb_gc_register_address(&file_paths);
+    long length = RARRAY_LEN(file_paths);
+    char **converted_file_paths = str_array_to_char(file_paths, length);
+
+    void* repository;
+    TypedData_Get_Struct(self, void*, &repository_type, repository);
+    idx_index_all_c(repository, converted_file_paths, length);
```
Morriar, line 47: If the indexing panics, is there a change we're leaking `converted_file_paths`?

src/c_interface.rs
```diff
@@ -1,105 +1,148 @@
...
+type RepositoryPointer = *mut c_void;
+type DeclarationPointer = *const Declaration;
 
-#[unsafe(no_mangle)]
-pub extern "C" fn get_repository() -> *mut CRepository {
-    let repository = Repository::new();
-    Box::into_raw(Box::new(repository)) as *mut CRepository
+/// Converts an array of C strings (const **char) to a Vec<String>
+unsafe fn convert_double_pointer_to_vec(data: &mut &mut c_char, len: size_t) -> Result<Vec<String>, Utf8Error> {
+    unsafe {
```
Morriar, line 43: Do we need the `unsafe` block if the function itself is already marked `unsafe`?

ext/index/repository.c
Morriar: Maybe we should also raise an exception back to Ruby is the indexing failed? For example if we can't open one of the file in the given paths?

PR #39, opened 2025-07-11
Introduce `Offset`

Morriar:
This pull request introduces two main concepts:

* `Offset`: represents byte-based position in a file
* `Location`: represents line/column based position in a file

`Offset` is ment to be memory efficient. It only store two u32 for the start and end offset plus the file id within the path string pool. `Offset` can be stored in our symbols (Class, Method, etc.) to represent their position. 

`Location` is using much more memory: 4 u32 for the position and a full string for the file path. It is not ment to be stored, only used when presented to the user.

### DIFF
--- a/src/offset.rs
+++ b/src/offset.rs
@@ -1,0 +1,79 @@
+//! Offset handling for byte-based file positions.
+//!
+//! This module provides the [`Offset`] struct which represents a span of bytes
+//! within a file. It can be used to track positions in source code and convert
+//! between byte offsets and line/column positions.
+
+use crate::pools::uri_pool::{UriId, UriPool};
+
+/// Represents a byte offset range within a specific file.
+///
+/// An `Offset` tracks a contiguous span of bytes from `start_offset` to `end_offset`
+/// within a file identified by `uri_id`. This is useful for representing the location
+/// of tokens, AST nodes, or other text spans in source code.
+///
+/// # Examples
+///
+/// ```rust
+/// use index::offset::Offset;
+/// use index::pools::uri_pool::{UriId, UriPool};
+///
+/// let mut uri_pool = UriPool::new();
+/// let uri_id = uri_pool.add("file://path/to/file.txt".to_string());
+///
+/// let offset = Offset::new(uri_id, 10, 20);
+///
+/// // Print the source covered by the offset
+/// println!("{}", offset.source(&uri_pool));
+///
+/// // Create a location from the offset
+/// let location = offset.to_location(&uri_pool);
+/// println!("{}", location);
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Offset {
+    /// The ID of the uri this offset refers to (see `UriPool`)
+    pub uri_id: UriId,
+    /// The starting byte offset (inclusive)
+    pub start_offset: u32,
+    /// The ending byte offset (exclusive)
+    pub end_offset: u32,
+}
+
+impl Offset {
+    /// Creates a new `Offset` with the specified file ID and byte range.
+    ///
+    /// # Arguments
+    ///
+    /// * `uri_id` - The ID of the uri this offset refers to (see `UriPool`)
+    /// * `start_offset` - The starting byte position (inclusive)
+    /// * `end_offset` - The ending byte position (exclusive)
+    #[must_use]
+    pub const fn new(uri_id: UriId, start_offset: u32, end_offset: u32) -> Self {
+        Self {
+            uri_id,
+            start_offset,
+            end_offset,
+        }
+    }
+
+    /// Returns a string representation of this offset in the format "URI:start-end".
+    ///
+    /// # Arguments
+    ///
+    /// * `uri_pool` - The uri pool to resolve the uri from the uri ID
+    ///
+    /// # Returns
+    ///
+    /// A string in the format "URI:start_offset-end_offset"
+    ///
+    /// # Panics
+    ///
+    /// Panics if the uri ID is not found in the uri pool.
+    #[must_use]
+    pub fn to_string(&self, uri_pool: &UriPool) -> String {
+        let uri = uri_pool.get(self.uri_id).unwrap();
+
+        format!("{}:{}-{}", uri, self.start_offset, self.end_offset)
+    }
+}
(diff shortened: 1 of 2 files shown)


src/file_pool.rs
```diff
@@ -0,0 +1,37 @@
...
+use derive_more::{From, Into};
+
+use crate::string_pool::StringPool;
+
+/// A unique identifier for a file path in a [`FilePool`].
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, From, Into)]
+pub struct FileId(u32);
+
+/// A specialized string pool for efficient file path storage and retrieval using [`FileId`].
+pub type FilePool = StringPool<FileId>;
```
vinistock: This should be URIs instead of file paths.

src/offset.rs
```diff
@@ -0,0 +1,594 @@
...
+    /// * `file_pool` - The file pool to resolve the file path from the file ID
+    ///
+    /// # Returns
+    ///
+    /// A string in the format "path:start_offset-end_offset"
+    ///
+    /// # Panics
+    ///
+    /// Panics if the file ID is not found in the file pool.
+    pub fn to_string(&self, file_pool: &FilePool) -> String {
```
vinistock: Would it be more idiomatic to implement the `Display` trait instead?

src/file_pool.rs
Morriar: Good call 👍

ChallaHalla: Difference here is that a URI contains the type of the resource it points to whereas a file path is implicitly a location for a file right?

src/location.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// * `file_pool` - A mutable reference to the file pool for managing file IDs
+    ///
+    /// # Returns
+    ///
+    /// An `Offset` representing the same location using byte offsets.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the file cannot be read.
+    pub fn to_offset(&self, file_pool: &mut FilePool) -> Offset {
```
ChallaHalla: We shouldn't be converting from `Location` to `Offset` too much right? From my understanding we will:
1. During the initial indexing, store entries along with `Offsets`
2. Get `Locations` computed on demand from an `Offset` when doing things that require line info

src/file_pool.rs
vinistock: Yes. When you have files on disk, they are file URIs (e.g.: `file:///some/path/to/file.rb`). When you have files opened in the editor's UI, before saving, their URI is `untitled:Untitled-1`.

We need to be able to index unsaved files too, otherwise we can't provide any features on them.

ChallaHalla: Ahhh okay that makes sense why this is important then! Ty for explaining

src/location.rs
Morriar: Indeed, we don't even do 1. because Prism already represent locations as offsets so we create them directly.

src/offset.rs
Morriar: The issue is that we need the `FilePool` to display the file URI properly or else we just get the FileId.

src/file_pool.rs
```diff
@@ -0,0 +1,45 @@
...
+        Self(symbol)
+    }
+
+    fn to_symbol(&self) -> SymbolId {
+        self.0
+    }
+}
+
+/// A specialized string pool for efficient file path storage and retrieval using [`FileId`].
+pub type FilePool = StringPool<FileId>;
```
vinistock: Should all of these be named Uri instead of file?

src/offset.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// The source text as a `String` for the byte range covered by this offset
+    ///
+    /// # Panics
+    ///
+    /// Panics if:
+    /// - The file ID is not found in the file pool
+    /// - The file cannot be read from disk
+    /// - The byte offsets are out of bounds for the file content
+    #[must_use]
+    pub fn source(&self, file_pool: &FilePool) -> String {
```
vinistock: Should this method be in `Offset` or in `Location`? Also, it still reads contents from disk, which only works for `file` URIs.

Morriar: Removed

src/file_pool.rs
Morriar: Yeah we can if you prefer 👍

src/pools/uri_pool.rs
```diff
@@ -0,0 +1,45 @@
...
+//!
+//! ```rust
+//! use index::pools::uri_pool::{UriId, UriPool};
+//!
+//! let mut pool = UriPool::new();
+//!
+//! // Add uri strings to the pool
+//! let uri1 = pool.add("file://path/to/file1.txt".to_string());
+//! let uri2 = pool.add("file://path/to/file2.txt".to_string());
+//! let uri3 = pool.add("file://path/to/file1.txt".to_string()); // Same uri
```
vinistock: Otherwise, the URI is invalid.

```suggestion
//! let uri1 = pool.add("file:///path/to/file1.txt".to_string());
//! let uri2 = pool.add("file:///path/to/file2.txt".to_string());
//! let uri3 = pool.add("file:///path/to/file1.txt".to_string()); // Same uri
```

src/pools/uri_pool.rs
```diff
@@ -0,0 +1,45 @@
...
+//!
+//! // String interning: same uris get the same ID
+//! assert_ne!(uri1, uri2);
+//! assert_eq!(uri1, uri3);
+//!
+//! // Retrieve uris by ID
+//! assert_eq!(pool.get(&uri1), Some("file://path/to/file1.txt".to_string()));
+//! assert_eq!(pool.get(&uri2), Some("file://path/to/file2.txt".to_string()));
```
vinistock:
```suggestion
//! assert_eq!(pool.get(&uri1), Some("file:///path/to/file1.txt".to_string()));
//! assert_eq!(pool.get(&uri2), Some("file:///path/to/file2.txt".to_string()));
```

src/offset.rs
```diff
@@ -0,0 +1,79 @@
...
+            start_offset,
+            end_offset,
+        }
+    }
+
+    /// Returns a string representation of this offset in the format "URI:start-end".
+    ///
+    /// # Arguments
+    ///
+    /// * `file_pool` - The file pool to resolve the file URI from the file ID
```
vinistock:
```suggestion
    /// * `uri_pool` - The URI pool to resolve the file URI from the URI ID
```

PR #44, opened 2025-07-18
Create Claude and Copilot instructions

vinistock:
Similar to what we did for the Ruby LSP, this PR adds some Claude/Copilot instructions so that generation can be more useful.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,49 @@
+# Instructions
+
+## Project Overview
+
+This Ruby gem and companion Rust crate provide a modern, high performance and low memory usage code indexing and
+static analysis tools for hyper scale Ruby projects. The Rust crate is a library that implements all of the indexing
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,49 @@
...
+
+The entire indexing and static analysis logic implementation. This crate aims to be optimized to achieve maximum
+performance in large codebases while maintaining memory usage to a minimum.
+
+The crate's goal is to provide all indexing and static analysis capabilities to power tools such as language servers,
+type checkers, linting and other code analysis features.
+
+### Commands
+
+When necessary, commands can be executed for the Ruby code.
```
alexcrocha, line 43:
```suggestion
When necessary, commands can be executed for the Rust code.
```

alexcrocha: Didn't realize Graphite doesn't show from the GitHub side that auto-merge is enabled 🙃

PR #46, opened 2025-07-21
Add a `StringPool` backed by `StringInterner`

Morriar:
A much simpler implementation than #34: we just delegate all the interning to `string-interner`.

Closes https://github.com/Shopify/index/pull/34.

### DIFF
--- a/src/string_pool.rs
+++ b/src/string_pool.rs
@@ -1,0 +1,411 @@
+//! String pooling and interning implementation.
+//!
+//! This module provides a `StringPool` that implements efficient string interning,
+//! where identical strings are stored only once and referenced by unique IDs.
+//!
+//! This implementation is backed by the `string_interner` crate.
+//! See <https://github.com/Robbepop/string-interner> for more details.
+//!
+//! # Example
+//!
+//! ```rust
+//! use index::string_pool::{StringPool, PoolId, SymbolId};
+//!
+//! // Define custom ID types directly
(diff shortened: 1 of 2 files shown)


src/string_pool.rs
```diff
@@ -0,0 +1,411 @@
...
+///
+/// This is the type of the symbols used by the string interner.
+/// It is used to identify unique strings in the pool.
+///
+/// This is a type alias to avoid having to import the `symbol` module everywhere.
+pub type SymbolId = SymbolU32;
+
+/// A trait for types that can serve as pool identifiers.
+///
+/// Types implementing this trait should be newtype wrappers around `SymbolId`
```
vinistock:
```suggestion
/// Types implementing this trait should be new type wrappers around `SymbolId`
```

src/string_pool.rs
```diff
@@ -0,0 +1,411 @@
...
+    #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+    struct TestId(SymbolId);
+
+    impl PoolId for TestId {
+        fn from_symbol(symbol: SymbolId) -> Self {
+            Self(symbol)
+        }
+
+        fn to_symbol(&self) -> SymbolId {
+            self.0
```
ChallaHalla, line 186: Interesting I somehow totally missed that `.` followed by an Int could be used to refer to an item in a tuple by index. Neat syntax

PR #48, opened 2025-07-23
Introduce `NamePool`

Morriar:
A specialized pool to store names. Like `FilePool` but... for names 😅 

I also moved all the pools together under the same module.

### DIFF
--- a/src/pools/name_pool.rs
+++ b/src/pools/name_pool.rs
@@ -1,0 +1,22 @@
+//! Name management with efficient string interning.
+//!
+//! See [`StringPool`].
+
+use crate::pools::string_pool::{PoolId, StringPool, SymbolId};
+
+/// A unique identifier for a name in a [`NamePool`].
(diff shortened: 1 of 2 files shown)


src/pools/name_pool.rs
```diff
@@ -0,0 +1,22 @@
+//! Name management with efficient string interning.
```
ChallaHalla, line 1: What are the names in the `NamePool` going to be used for?

ChallaHalla: Ahh this must be for Entry names

Morriar: The names of the entities like `::Foo::Bar` or `Baz::foo`. It's coming in the next PR with the hierarchy of definitions.

ChallaHalla: Ahh makes sense. Ty for explaining!

PR #49, opened 2025-07-23
Add dev.yml

vinistock:
This morning dev seemed to have removed Rust from my machine. This PR adds a `dev.yml`, so that we can dev up and get the expected setup.

Also, formatting is failing for me with the `--edition` flag, but I'm not 100% sure that's actually necessary. Everything works without that flag.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,15 +2,6 @@
     "rust-analyzer.check.command": "clippy",
     "rust-analyzer.check.allTargets": true,
     "rust-analyzer.diagnostics.enable": true,
-    "rust-analyzer.rustfmt.extraArgs": [
-        "--edition",
-        "2024"
-    ],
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -2,15 +2,6 @@
     "rust-analyzer.check.command": "clippy",
     "rust-analyzer.check.allTargets": true,
     "rust-analyzer.diagnostics.enable": true,
-    "rust-analyzer.rustfmt.extraArgs": [
```
Morriar, line 5: I guess `dev` makes sure we have the latest version?

vinistock: I think rustfmt can figure out the edition based on `Cargo.toml`, no?

alexcrocha: It will read the edition from `rustfmt.toml`, which we have set to 2024.

PR #50, opened 2025-07-23
Introduce symbol definitions stubs

Morriar:
A small contained PR to show the direction and agree on naming.

This only handles creating classes, modules and singleton classes by visiting the Prism AST.

I also added a small `main` so we can run it on files.

Closes https://github.com/Shopify/index/pull/51.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,0 +1,93 @@
+use glob::glob;
+use ruby_prism::Visit;
+use std::collections::HashMap;
+use std::env;
+use std::path::Path;
+use std::process;
+
+use indexing::ruby_indexer::RubyIndexer;
+use model::declaration::Declaration;
+use pools::name_pool::{NameId, NamePool};
+use pools::uri_pool::UriPool;
+
+mod indexing;
+mod model;
+mod offset;
+mod pools;
+
+fn glob_files(dir_path: &str) -> Vec<String> {
+    let mut files = Vec::new();
+
+    let patterns = vec![format!("{}/**/*.rb", dir_path)];
+
+    for pattern in patterns {
+        match glob(&pattern) {
+            Ok(paths) => {
+                for entry in paths {
+                    match entry {
+                        Ok(file_path) => {
+                            let path_str = file_path.to_string_lossy().to_string();
+                            if !files.contains(&path_str) && Path::new(&path_str).is_file() {
+                                files.push(path_str);
+                            }
+                        }
+                        Err(e) => eprintln!("Error reading glob entry: {e}"),
+                    }
+                }
+            }
+            Err(e) => eprintln!("Failed to read glob pattern {pattern}: {e}"),
+        }
+    }
+
+    files
+}
+
+fn collect_files(paths: &[String]) -> Vec<String> {
+    let mut files = Vec::new();
+    for path in paths {
+        if Path::new(path).is_dir() {
+            let mut dir_files = glob_files(path);
+            files.append(&mut dir_files);
+        } else {
+            files.push(path.to_string());
+        }
+    }
+    files.sort();
+    files.dedup();
+    files
+}
+
+fn process_files(paths: &[String], uri_pool: &mut UriPool, name_pool: &mut NamePool) -> HashMap<NameId, Declaration> {
+    let mut declarations: HashMap<NameId, Declaration> = HashMap::new();
+
+    for path in paths {
+        let uri_id = uri_pool.add(format!("file://{path}"));
+
+        let source = match std::fs::read_to_string(path) {
+            Ok(content) => content,
+            Err(err) => {
+                eprintln!("Error reading file {path}: {err}");
+                continue;
+            }
+        };
+
+        let result = ruby_prism::parse(source.as_ref());
+        RubyIndexer::new(uri_id, name_pool, &mut declarations).visit(&result.node());
+    }
+
+    declarations
+}
+
+fn main() {
+    let args: Vec<String> = env::args().skip(1).collect();
+    let files = collect_files(&args);
+    println!("  Found {} files.", files.len());
+
+    let mut uri_pool = UriPool::new();
+    let mut name_pool = NamePool::new();
+
+    let definitions = process_files(&files, &mut uri_pool, &mut name_pool);
+    println!("  Found {} definitions.", definitions.len());
+
+    process::exit(0);
+}


src/model/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+use ruby_prism::Visit;
+
+/// The builder for the symbol definitions.
+///
+/// It implements the `Visit` trait from `ruby_prism` to visit the AST and build the symbol definitions.
+#[derive(Debug)]
+pub struct Builder<'a> {
+    uri_id: UriId,
+    name_pool: &'a mut NamePool,
+    symbol_definitions: &'a mut Vec<SymbolDefinition>,
```
vinistock: I'd vote to call this declarations. Symbol is an actual concept in Ruby, so I'd be afraid that this could be confusing.

src/model/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// ```
+    /// let mut name_pool = NamePool::new();
+    /// let mut symbol_definitions = Vec::new();
+    /// let builder = Builder::new(uri_id, &mut name_pool, &mut symbol_definitions);
+    ///
+    /// let result = ruby_prism::parse("class Foo; end");
+    /// builder.visit(&result.node());
+    ///
+    /// assert_eq!(symbol_definitions.len(), 1);
+    /// ```
```
vinistock: Should this be on top of the struct instead?

src/model/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// # Arguments
+    ///
+    /// * `location` - The Prism location of the string to intern
+    ///
+    /// # Returns
+    ///
+    /// The name ID of the interned string.
+    fn intern_slice(&mut self, location: &ruby_prism::Location) -> NameId {
+        self.name_pool
+            .add(String::from_utf8_lossy(location.as_slice()).to_string())
```
vinistock: Note for posterity: using `slice` produces incorrect results in a few cases, but we don't have the same convenience APIs in Rust that we do in Ruby to do the right thing.

Notably, this example provides the incorrect name:

```ruby
Foo:: # some comment
  Bar
```

The complete constant path node spans from the beginning of `Foo` to the end of `Bar` and the slice includes that comment. We have `ConstantPathNode#full_name` in the Ruby API, but we need that in Rust too.

src/model/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// * `location` - The Prism location to convert
+    ///
+    /// # Returns
+    ///
+    /// The offset of the location.
+    fn prism_location_to_offset(&self, location: &ruby_prism::Location) -> Offset {
+        Offset::new(
+            self.uri_id,
+            location.start_offset().try_into().unwrap(),
+            location.end_offset().try_into().unwrap(),
```
vinistock: Should we use expect so that we know what went wrong if this fails?
```suggestion
            location.start_offset().try_into().expect("Expected usize start_offset to fit in u32"),
            location.end_offset().try_into().expect("Expected usize end_offset to fit in u32"),
```

src/model/builder.rs
```diff
@@ -0,0 +1,274 @@
...
+use crate::offset::Offset;
+use crate::pools::name_pool::{NameId, NamePool};
+use crate::pools::uri_pool::UriId;
+use ruby_prism::Visit;
+
+/// The builder for the symbol definitions.
+///
+/// It implements the `Visit` trait from `ruby_prism` to visit the AST and build the symbol definitions.
+#[derive(Debug)]
+pub struct Builder<'a> {
```
vinistock: I vote for calling this `RubyIndexer`, so that we can easily differentiate between the future `RBSIndexer`.

src/model/mod.rs
vinistock, line 1: What is the `model` module supposed to contain here?

src/model/builder.rs
Morriar: Good idea.

src/model/builder.rs
Morriar: I'm not a fan of "declaration". For me a "declaration" _introduces_ something, while a "definition" gives the actual content.

In Ruby we never really introduce anything we just define them and the execution order will decide which one is the introduction and which one is the reopening.

We can drop the `Symbol` prefix and just have definitions: `Vec<Definition>`.

src/model/mod.rs
Morriar: TIL: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs

src/model/builder.rs
Morriar: I was trying to not repeat the document too much.

Let's say we always comment the structs and not the methods unless necessary?

src/model/mod.rs
Morriar: Replaced the `mod.rs` by `model.rs`.

Do we prefer `index`? I feel it's too reductive considering what we want to store/provide

src/model/builder.rs
Morriar: Good idea 👍

src/model/builder.rs
Morriar: Good call, I added a test showing the wrong behavior, we'll have to fix it.

src/model/definitions.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+}
+
+/// The definition of a class in the source code.
+///
+/// A class can have multiple definitions for the same name.
+#[derive(Debug)]
+pub struct ClassDefinition {
+    pub name_id: NameId,
```
vinistock: Let's move forward like this for now, but I think the name ID would be stored by the global version of the definition and not these structs that are tied to a specific file and offset. Right?

src/model/builder.rs
vinistock: Yeah, definition is fine 👍

src/model/builder.rs
vinistock: Yeah, sure. We also need to pay attention to the `//` vs `///` vs `//!`, which I totally ignored for now. The triple version is only meant for the public API of the crate that may be used externally.

Internal comments about the implementation are always double slash.

src/model/mod.rs
vinistock: Well, what goes inside `model`? What group of things are we trying to tie together? The word model for me suggested the definitions, but not the indexer part.

If it's the entire logic of indexing, then I'd vote for `indexing`, but I feel like you meant for `model` to house a subset of indexing.

src/model/builder.rs
Morriar: My understanding is that:

* `//!` is for crate/module level documentation and used at the top of the file
* `///`is for public API documentation, used on top of items like structs, enums, fns, etc.
* `//` is for internal comment

I've been trying to always add a `//!` comment on modules and `///` on top of public enums and structs. Sometimes a `///` on functions that deserve it.

src/model/definitions.rs
Morriar: You're right, but since we don't have the global view yet, we need somekind of name info to make it easier to debug.

src/model/mod.rs
Morriar: Maybe something like this:

* `model`: all the structures that represent definitions, global views, etc.
* `indexing`: all the `ruby`, `rbs`, `c` indexers

vinistock: Sure, let's try that. Not a big deal, we can change this easily too.

src/model/builder.rs
vinistock: Yes, that's correct (afaik). I think it depends on what we would consider as the crate's public API. We can revisit this later.

.cargo/config.toml
vinistock, line 1: We may want to take a look at this one separately

PR #54, opened 2025-07-29
Add global index

vinistock:
Add our prototype of the global index graph, so that we can build on top of it.


rust/index/src/indexing.rs
```diff
@@ -1 +1,94 @@
...
+            loop {
+                // Pop a file path from the shared queue. This needs to happen in a small scope to avoid holding the
+                // lock for too long, which would cause other threads to slow down
+                let maybe_document = {
+                    let mut queue = queue_clone.lock().unwrap();
+                    queue.pop()
+                };
+
+                if let Some(document) = maybe_document {
+                    if let Ok(uri) = Url::parse(&document.uri) {
```
ChallaHalla, line 52: I found a bug here where file paths are being passed in instead of URIs resulting in a parse failure. This causes the indexer to run but store no data about the discovered files.

```suggestion
                    if let Ok(uri) = Url::parse(&format!("file://{}", document.uri.to_string())) {
```

PR #55, opened 2025-07-29
Organize Rust code into a workspace under `rust`

vinistock:
We discovered that the Rust convention for FFI crates is that there should always be two:

- the-crate (pure Rust implementation)
- the-crate-sys (crate that provides the FFI)

While we're still in the early stages of the implementation, I think we should re-organize all of this. This PR:

- Separates the C interface and the pure Rust of the implementation into `index` and `index-sys`
- Puts both crates under the `rust` directory as a cargo workspace
- Updates whatever else was necessary


rust/index-sys/build.rs
```diff
@@ -3,5 +3,5 @@ extern crate cbindgen;
 fn main() {
     cbindgen::generate(".")
         .expect("Unable to generate bindings")
-        .write_to_file("ext/index/rustbindings.h");
+        .write_to_file("../../ext/index/rustbindings.h");
```
alexcrocha: Since `index-sys` is its own crate, I think this should be self-contained.

vinistock: What do you mean by self contained? We generate these bindings so that the C extension side can hook to the Rust crate. We need the header file to be a part of the `ext` directory.

alexcrocha: As I understand, being its own crate means other projects should be able to use `index-sys` independently. Currently it's too integrated with our project. Shouldn't the dependency be the other way around? 

This may be a limit of my understanding of C, but is it not possible to use the header file if it lives in `rust/index-sys/rustbindings.h`?

alexcrocha: Oh, hold on. I think I found the source of my confusion. I need to think about this a bit more.

vinistock: You're making a good point. The crate is indeed very tied to the gem's structure. We might be able to do better 👍

vinistock: Okay, I started generating the `rustbindings.h` file inside the crate and I changed the Makefile to copy the file over when compiling the native extension.

ext/index/extconf.rb
```diff
@@ -48,6 +50,7 @@
 
   .rust_built: $(RUST_SRCS)
   \t#{cargo_command} || (echo "Compiling Rust failed" && exit 1)
+  \t$(COPY) #{bindings_path} #{__dir__}
```
ChallaHalla, line 53: Can we just directly generate the bindings to be placed in the sys crate?

ChallaHalla: Within rust/index-sys/build.rs

rust/index-sys/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "index-sys"
+version = "0.1.0"
+edition = "2024"
+
+[lib]
+crate-type = ["cdylib", "staticlib"]
+
+[dependencies]
+index = { path = "../index" }
```
ChallaHalla, line 10: Shouldn't the dependency be the other way around? E.g `index` depends on `index-sys`?

rust/index-sys/rustbindings.h
```diff
@@ -0,0 +1,44 @@
+#ifndef RUSTBINDINGS_H
```
ChallaHalla: Can add this file to the git ignore since it's generated

rust/index-sys/Cargo.toml
ChallaHalla: Ahh nevermind I understand why the dependency works this way. We need bits of `index` to write the glue layer within the sys create

ext/index/extconf.rb
vinistock: Are you referring to this https://github.com/Shopify/index/pull/55#discussion_r2240888355?

We made the change to allow the crate to be used by other Rust projects without tying it to the gem structure.

rust/index-sys/rustbindings.h
vinistock: Yes

ext/index/extconf.rb
ChallaHalla: Ahhh I see. I missed that thread 😅 makes sense ty

PR #61, opened 2025-08-01
Add SQLite DB for declarations

ChallaHalla:
Implements SQLite persistent storage for the index to reduce memory usage and enable cross-session data persistence.

### Key Changes

- Added `rusqlite` dependency and `Db` struct for database operations
- Added `dump_to_db()` method to persist declarations and clear memory
- Replace`get_declaration()` with `load_declaration()` to fetch with lazy loading from database
- Modified `Index::new()` to accept a database

### Technical Details

- Database schema with indexed `declarations` table for optimized lookups
- Batch database operations for efficiency
- Fallback from memory to database storage
- Test coverage for DB dump/retrieval

🤖 Generated in part by [Claude Code](https://claude.ai/code)

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -9,6 +9,8 @@ use crate::model::{
     ids::{DefinitionId, NameId, UriId},
 };
 
+const EXPECTED_TABLE_COUNT: i64 = 3;
+
 #[derive(Debug)]
 pub struct Db {
     path: String,
@@ -50,12 +52,21 @@ impl Db {
             Connection::open(&self.path)?
         };
 
-        let schema = std::fs::read_to_string(
-            std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
-                .join("src/db")
-                .join("schema.sql"),
-        )?;
-        conn.execute_batch(&schema)?;
+        // Only load schema if tables don't exist
+        let table_count: i64 = conn.query_row(
+            "SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name IN ('documents', 'names', 'definitions')",
+            [],
+            |row| row.get(0),
+        ).unwrap_or(0);
+
+        if table_count < EXPECTED_TABLE_COUNT {
+            let schema = std::fs::read_to_string(
+                std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                    .join("src/db")
+                    .join("schema.sql"),
+            )?;
+            conn.execute_batch(&schema)?;
+        }
 
         match self.conn.set(conn) {
             Ok(()) => self


rust/index/src/model/index.rs
```diff
@@ -9,14 +10,20 @@ use crate::{indexing::ruby_indexer::DefinitionCollection, model::declaration::De
 pub struct Index {
     declarations: HashMap<DeclarationId, Declaration>,
     uri_pool: HashMap<UriId, String>,
+    db: Db,
 }
 
 impl Index {
-    #[must_use]
-    pub fn new() -> Self {
+    pub fn new(db_path: &str) -> Self {
```
ChallaHalla: Cargo run and cargo test run from different directories, so relatively referencing `tmp/` was breaking. This structure allows for more flexible specification of what db path to use.

ChallaHalla: Opted to remove a default option. Any caller of Index::new should specify a DB path.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -40,8 +40,8 @@ impl DefinitionCollection {
 /// ```
 /// use index::indexing::ruby_indexer::RubyIndexer;
 /// use index::model::index::Index;
-///
-/// let mut index = Index::new();
+/// let db_path = std::env::temp_dir().join("index.db");
```
ChallaHalla: Was getting a weird CI failure when this was unchanged. Probably still a good idea to update though since it keeps the example accurate.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        conn.execute(
+            "CREATE INDEX IF NOT EXISTS declarations_name_index ON declarations (name)",
+            (),
+        )?;
+
+        conn.execute(
+            "CREATE INDEX IF NOT EXISTS declarations_declaration_id_index ON declarations (declaration_id)",
+            (),
+        )?;
```
vinistock: Would it be possible to put the schema in a separate file and load it all in one go?

.gitignore
```diff
@@ -18,3 +18,4 @@ rust/index-sys/rustbindings.h
 ext/index/rustbindings.h
 ext/index/compile_commands.json
 ext/index/.cache
+.DS_Store
```
vinistock: I'm not necessarily opposed to ignoring this, but you should probably ignore this with a global gitignore.

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,67 @@
...
+        let mut rows = stmt
+            .query_map([&declaration_id_str], |row| row.get::<_, String>(0))
+            .ok()?;
+
+        if let Some(Ok(name)) = rows.next() {
+            return Some(Declaration::new(name, None));
+        }
+
+        None
+    }
```
vinistock: Let's move this to the `Db` implementation. The `Index` should not interface with db queries directly.

rust/index/src/model/index.rs
```diff
@@ -41,10 +76,159 @@ impl Index {
             }
         }
     }
+
+    pub fn dump_to_db(&mut self) {
+        if let Err(e) = self.db.batch_insert_declaration(&self.declarations) {
+            eprintln!("Warning: Failed to dump declarations to database: {e}");
```
vinistock: Let's make this method return a `Result` instead of printing, so that consumers can decide what to do. Also, if this fails, we should probably not clear the in-memory declarations.

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,67 @@
...
+
+        // First, ensure the declaration is loaded (either from memory or DB)
+        if !self.declarations.contains_key(&declaration_id) {
+            if let Some(declaration) = self.get_declaration_from_db(&declaration_id) {
+                self.declarations.insert(declaration_id, declaration);
+            }
+        }
+
+        // Now we can safely return the reference
+        self.declarations.get(&declaration_id)
```
vinistock: Would it be possible to simplify this by mapping the option? Something like this

```rust
self.declarations
  .get(&declaration_id)
  .unwrap_or_else(|| {
    self.db.get_declaration(&declaration_id)
  })
```

.gitignore
ChallaHalla: Hmm weirdly this is already in my global git ignore. I'll take a look at why the global ignore isn't working

rust/index/src/model/index.rs
ChallaHalla: Good point I'll separate this out

rust/index/src/model/index.rs
ChallaHalla: I was having some trouble getting the return types to play nicely when doing something similar to this. Gonna give it another whirl.

ChallaHalla: The root issue is that returning `Option<&Declaration>` from `self.db.get_declaration(..)` is not possible because the reference points to a value that will not live after the method completes. The only way to return `Option<&Declaration>` is to fetch the declaration from the `declarations` hash map. That's why I implemented this as a hash map insertion followed by a hash map `.get`

vinistock: Oh, I see. Maybe something with `self.declarations.entry(declaration_id)`? If there's no other way, it's fine. Just wondering if this can be simplified.

ChallaHalla: Ahh that return type looks easier to work with!

rust/index/src/model/index.rs
```diff
@@ -41,10 +60,161 @@ impl Index {
...
+        let test_declarations = vec![
+            ("Foo::Bar", "class_definition"),
+            ("Foo::Baz", "module_definition"),
+            ("SomeMethod", "method_definition"),
+            ("AnotherClass::NestedClass", "class_definition"),
+            ("Utils::Helper", "module_definition"),
+        ];
+
+        for (name, _desc) in &test_declarations {
+            let declaration_ = DeclarationId::new(name);
```
Morriar: Should this be `declaration_id`?

ChallaHalla: oh weird this looks like a typo. I wonder if the tests are still passing

ChallaHalla: They were indeed failing 😌 nice catch!

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
...
     declarations: HashMap<DeclarationId, Declaration>,
     uri_pool: HashMap<UriId, String>,
+    db: Db,
 }
 
 impl Index {
     #[must_use]
-    pub fn new() -> Self {
+    pub fn new(db_path: &str) -> Self {
+        let db = Db::new(db_path).expect("Failed to initialize database");
```
Morriar: I don't think the `index` should be responsible to open the connexion to the database. Can't we just take a `Db` and open it in the main?

Morriar: Also I _think_ we should avoid `expect`s in the lib part of the code and return proper errors.

rust/index/src/model/index.rs
```diff
@@ -41,10 +60,161 @@ impl Index {
...
-impl Default for Index {
-    fn default() -> Self {
-        Self::new()
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn insert_retrieve() {
+        let test_db_path = std::env::temp_dir().join("index_test.db").to_string_lossy().to_string();
```
Morriar: Should we cleanup the database after each test?

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
-use std::collections::HashMap;
-use std::collections::hash_map::Entry;
-
+use crate::model::db::Db;
+use crate::model::declaration;
```
Morriar: Looks redundant with the `crate::model::declaration::Declaration` import line 4?

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
...
-
     #[must_use]
     pub fn intern_uri(&mut self, uri: String) -> UriId {
         let uri_id = UriId::new(&uri);
         self.uri_pool.insert(uri_id, uri);
         uri_id
     }
 
+    #[must_use]
+    pub fn get_declaration(&mut self, fully_qualified_name: &str) -> Option<&Declaration> {
```
Morriar: I feel like a method called `get_` that takes a `&mut self` may be a smell.

We should communicate that the declaration will be loaded from the DB if not already in memory. What do you think about `load_declaration` instead?

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
...
+        let declaration_id = DeclarationId::new(fully_qualified_name);
+
+        // First, ensure the declaration is loaded (either from memory or DB)
+        if !self.declarations.contains_key(&declaration_id) {
+            if let Some(declaration) = self.db.get_declaration(&declaration_id) {
+                self.declarations.insert(declaration_id, declaration);
```
Morriar: Can't we return it right away instead of accessing again the map line 46?

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
...
+            }
+        }
```
Morriar: Could we already return `None` in the else branch instead of trying the map again?

rust/index/src/model/index.rs
```diff
@@ -1,32 +1,51 @@
...
+
+        // Now we can safely return the reference
+        self.declarations.get(&declaration_id)
+    }
```
Morriar: Considering we access the same hash on the same key a few times I wonder if that wouldn't be the right place for a `Entry` pattern: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

Morriar: Maybe something like this:

```rust
pub fn get_declaration(&mut self, fully_qualified_name: &str) -> Option<&Declaration> {
    let declaration_id = DeclarationId::new(fully_qualified_name);
    
    use std::collections::hash_map::Entry;
    
    match self.declarations.entry(declaration_id) {
        Entry::Occupied(entry) => Some(entry.get()),
        Entry::Vacant(entry) => {
            if let Some(declaration) = self.db.get_declaration(&declaration_id) {
                Some(entry.insert(declaration))
            } else {
                None
            }
        }
    }
}
```

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl Db {
+    pub fn new(path: &str) -> Result<Self> {
+        let conn = Connection::open(path)?;
+
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS declarations (
+                id INTEGER PRIMARY KEY,
+                name TEXT NOT NULL,
+                declaration_id INTEGER NOT NULL
```
Morriar: Can't we use this as the primary key directly? 🤔

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,67 @@
+use std::collections::HashMap;
+
+use rusqlite::{Connection, Result};
+
+use crate::model::{declaration::Declaration, ids::DeclarationId};
+
+pub struct Db {
+    pub conn: Connection,
```
Morriar: I wonder if this shouldn't be private for now?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        tx.commit()?;
+        Ok(())
+    }
+
+    pub fn get_declaration(&self, declaration_id: &DeclarationId) -> Option<Declaration> {
+        let mut stmt = self
+            .conn
+            .prepare("SELECT name FROM declarations WHERE declaration_id = ?")
+            .ok()?;
```
Morriar: Does this mean we'll fail silently is there is an error with the query?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,67 @@
...
+        let tx = self.conn.unchecked_transaction()?;
+        {
+            let mut stmt = tx.prepare("INSERT INTO declarations (name, declaration_id) VALUES (?, ?)")?;
+
+            for (declaration_id, declaration) in declarations {
+                stmt.execute([&declaration.name, &declaration_id.to_string()])?;
+            }
+        }
+
+        tx.commit()?;
```
Morriar: Same question, will we fail silently if we can't save?

rust/index/src/model/db.rs
Morriar: Yes, let's move it into a `.sql` file?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,67 @@
...
+            (),
+        )?;
+
+        Ok(Self { conn })
+    }
+
+    pub fn batch_insert_declaration(&self, declarations: &HashMap<DeclarationId, Declaration>) -> Result<()> {
+        let tx = self.conn.unchecked_transaction()?;
+        {
+            let mut stmt = tx.prepare("INSERT INTO declarations (name, declaration_id) VALUES (?, ?)")?;
```
Morriar: https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.prepare_cached

rust/index/src/model/index.rs
ChallaHalla: The return value for `.entry` is unfortunately not fit for this use case.

Chatted with @jesse-shopify and aligned that cloning is probably the easiest way around the issue. I added the `Clone` trait to relevant types and simplified the method for getting declarations.

rust/index/src/model/index.rs
ChallaHalla: This seems good actually. The match on the Entry explicitly outlines all the cases we need to worry about quite well

rust/index/src/model/index.rs
vinistock: I agree regarding `expect`, we should avoid it as much as possible (unless it indicates a bug in our own code and not some uncontrollable error like IO problems).

> I don't think the index should be responsible to open the connexion to the database

Can you elaborate why not? Are you concerned about error handling scenarios or just that the `Index` shouldn't conceptually own the `Db`?

Morriar: Yes error handling is one thing, I'm not sure the `Index` should be responsible of knowing how to report or handle them. And who is responsible of closing the connection?

There is also the notion of configuration. Should the database information such as the path and maybe other options live in the `Index` itself?

I'm also not sure about the life cycle. Let's say I'm running a bunch of tests about indexing, each test creating its own index, do we want to pay the price of opening/closing the connection for each test? Or keep the connection alive, reset the db and just initialize a new index? Also passing the `Db` object may enable mocking?

On one hand by encapsulating the `db` into the index we avoid people messing with the connexion while we use it. On the other hand we transfer a lot of responsibilities to the index.

vinistock: I like the idea of more easily mocking or reusing the same DB for tests. Concerning configuration, there are certain aspects that do related to the index, like which gems to ignore or which paths to ignore from indexing.

I'm also not sure how this connects to the Ruby API. If we have two instances of `Index`, do we allow them both to connect to the same DB?

I don't know what's the right answer here. Maybe we can just give it a try and see how it looks.

rust/index/src/model/db.rs
ChallaHalla: Yeah now that db interaction is just within this Struct I'll make it private

rust/index/src/model/db.rs
ChallaHalla: I think this is the issue w/ respect to generating a unique ID that is persistent across runs. We need to change our strategy for DB writes if we want to go the route of using the DB primary keys to identify entities across runs.

ChallaHalla: Or is the suggestion to make the declaration ID the primary key?

rust/index/src/model/index.rs
ChallaHalla: Sure I'll try making the database connection static and passing that reference to the index. So multiple `Index` instances should be able to use the same db connection. Should make mocking easier since we could potentially pass a mocked db value to the index as needed since the connection is extracted out

rust/index/src/model/db.rs
Morriar: Yes, let's make `declaration_id` the primary key and remove `id`

rust/index/src/model/db.rs
ChallaHalla: The `?` operator here would wrap and error that occurs in the `Error` enum and return early from the function. It's basically shorthand for something like:

``` rust
let result = match  tx.commit() {
    Ok(result) => result,
    Err(error) => return Err(error),
};

```

rust/index/src/model/db.rs
ChallaHalla: Same as above

Morriar: Wouldn't that return `None` and silence the `Err`?

```rust
fn some_err_result() -> Result<i32, &'static str> {
    Err("oh no".into())
}

fn result_to_option_err() -> Option<i32> {
    // Here, some_err_result() will be Err, so .ok()? returns None
    let s = some_err_result().ok()?;
    Some(s)
}

fn main() {
   println!("from err result: {:?}", result_to_option_err());
}
```

ChallaHalla: Ahh yeah nice catch you're def right. We'd be returning `None` for both a declaration not being found as well as a DB error when querying which isn't great. I'll separate out the error case so we don't silently swallow it

rust/index/src/model/ids.rs
```diff
@@ -41,6 +42,23 @@ impl DeclarationId {
     pub fn new(id: &str) -> Self {
         Self(create_hash(id))
     }
+
+    /// Convert the first 8 bytes of the hash to a 64-bit signed integer for database storage
+    /// This provides excellent uniqueness while being compatible with SQLite INTEGER type
+    #[must_use]
+    pub fn to_i64(&self) -> i64 {
```
ChallaHalla: Choosing to convert the blake3 hash to an integer. This should still provide 2^64 diff possibilities for key values which is sufficient for uniqueness.

rust/index/src/model/index.rs
ChallaHalla: Using load seems good to me! Will update

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,11 @@
+-- SQLite schema for the index database
+-- Contains tables and indexes for storing code declarations and their relationships
+
+-- Table for storing code declarations (classes, modules, methods, etc.)
+CREATE TABLE IF NOT EXISTS declarations (
```
ChallaHalla: Will need to update the DB schema to reflect how we are representing declarations and definitions now

rust/index/src/model/index.rs
```diff
@@ -53,22 +80,149 @@ impl Index {
...
     }
-}
 
-impl Default for Index {
-    fn default() -> Self {
-        Self::new()
+    /// # Errors
+    /// Will return `Err` if  there is a failure to dump declarations in to the DB.
+    pub fn dump_to_db(&mut self) -> rusqlite::Result<()> {
+        self.db.batch_insert_uris(&self.uri_pool)?;
```
ChallaHalla: Place this all within a tx

rust/index/src/model/index.rs
```diff
@@ -1,24 +1,51 @@
...
+    pub fn intern_uri(&mut self, uri: String) -> UriId {
+        let uri_id = UriId::new(&uri);
+        self.uri_pool.insert(uri_id, uri);
+        uri_id
+    }
+
+    /// # Errors
+    /// Will return `Err` if there is a database connection error or query failure.
+    /// Returns `Ok(Some(declaration))` if found, `Ok(None)` if not found.
+    pub fn load_declaration(&mut self, fully_qualified_name: &str) -> rusqlite::Result<Option<Declaration>> {
```
ChallaHalla: Let's remove the loading logic from this PR and add it in a follow up. 

We want to take the approach of loading from the DB given a URI so that we can load stuff in to memory when files in the editor are opened.

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,32 @@
+-- SQLite schema for the index database
+-- Contains tables and indexes for storing code declarations and their relationships
+
+-- Table for storing unique URIs (file paths)
+CREATE TABLE IF NOT EXISTS uris (
+    id INTEGER PRIMARY KEY,  -- Blake3 hash converted to i64
+    path TEXT NOT NULL UNIQUE
+);
```
vinistock: Let's use the document name, since we already know we'll need more information such as content hashes.
```suggestion
-- Table for storing unique documents
CREATE TABLE IF NOT EXISTS documents (
    id INTEGER PRIMARY KEY,  -- Blake3 hash converted to i64
    uri TEXT NOT NULL UNIQUE
);
```

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,32 @@
...
+-- Table for storing code declarations (classes, modules, methods, etc.)
+CREATE TABLE IF NOT EXISTS declarations (
+    id INTEGER PRIMARY KEY,  -- Blake3 hash converted to i64
+    name TEXT NOT NULL
+);
+
+-- Table for storing definitions that make up declarations
+CREATE TABLE IF NOT EXISTS definitions (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    declaration_id INTEGER NOT NULL,  -- References declarations.id
```
vinistock: Can we make this the ID? Or do we need both `id` and `declaration_id`?

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,32 @@
...
+    definition_type TEXT NOT NULL CHECK(definition_type IN ('Class', 'Module')),
```
vinistock: Let's make this an enum.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// an error openining or using the DB connection.
+    pub fn new(path: &str) -> Result<Self, Box<dyn std::error::Error>> {
+        let conn = Connection::open(path)?;
+
+        let schema = std::fs::read_to_string(
+            std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("src/db")
+                .join("schema.sql"),
+        )?;
+        conn.execute_batch(&schema)?;
```
vinistock: Do we need to load the schema every time we create a new DB instance? Maybe we can have a function for creating the DB if it does not exist and then that one will load the schema.

If it's already there, we can probably skip loading the schema.

rust/index/src/model/definitions.rs
```diff
@@ -25,7 +25,7 @@
 
 use crate::offset::Offset;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
vinistock: Why do these need to be cloned? If possible, I think we should avoid adding clone to these as we want to try to ensure that we're maintaining a single entry for each definition.

rust/index/src/offset.rs
```diff
@@ -11,9 +11,9 @@
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Offset {
     /// The starting byte offset (inclusive)
-    start_offset: u32,
+    pub start_offset: u32,
```
vinistock: Instead of making fields public, let's add just getters to read the information.

ChallaHalla: Is the main reason for this to ensure that the values are readable but not writable?

rust/index/src/model/definitions.rs
ChallaHalla: Removed

rust/index/src/db/schema.sql
ChallaHalla: Updated to name_id. But yeah we'd need something like that here since it's a FK

rust/index/src/model/db.rs
ChallaHalla: Good call, I added a check so that we only load the schema if we don't have all the tables.

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,32 @@
...
+    end_offset INTEGER NOT NULL,
+    FOREIGN KEY (name_id) REFERENCES names (id) ON DELETE CASCADE,
+    FOREIGN KEY (document_id) REFERENCES documents (id) ON DELETE CASCADE
+);
+
+-- Indexes for fast lookups
+CREATE INDEX IF NOT EXISTS names_name_index ON names (name);
+CREATE INDEX IF NOT EXISTS definitions_name_id_index ON definitions (name_id);
+CREATE INDEX IF NOT EXISTS definitions_document_id_index ON definitions (document_id);
+CREATE INDEX IF NOT EXISTS definitions_type_index ON definitions (definition_type);
```
vinistock, line 32: Do we need the index on definition type? What query would it be used for?

rust/index/src/model/graph.rs
```diff
@@ -1,40 +1,67 @@
...
 }
 
 impl Graph {
-    #[must_use]
-    pub fn new() -> Self {
-        Self {
+    /// # Errors
+    /// Will return an error if db connection fails
+    pub fn new() -> Result<Self, Box<dyn std::error::Error>> {
+        let db = Db::new(DB_PATH)?;
```
Morriar, line 37: Shouldn't the db path be passed as an option? Or maybe add a `new_with_db_path`?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[derive(Debug)]
+pub struct Db {
+    path: String,
+    conn: OnceLock<Connection>,
+}
+
+impl Db {
+    /// # Errors
+    /// Will return `Err` if  there is a failure to load the schema file or if there is
```
Morriar, line 22: Does it? Looks like this moved to `get_connection` now?

rust/index/src/db/schema.sql
```diff
@@ -0,0 +1,32 @@
...
+
+-- Table for storing definitions that make up declarations
+CREATE TABLE IF NOT EXISTS definitions (
+    id TEXT PRIMARY KEY,  -- Blake3 hash converted to hex
+    name_id TEXT NOT NULL,  -- References names.id
+    definition_type INTEGER NOT NULL CHECK(definition_type IN (1, 2)), -- 1=Class, 2=Module
+    document_id TEXT NOT NULL,          -- References documents.id
+    start_offset INTEGER NOT NULL,
+    end_offset INTEGER NOT NULL,
+    FOREIGN KEY (name_id) REFERENCES names (id) ON DELETE CASCADE,
```
Morriar, line 24: We may need to pass `PRAGMA foreign_keys = ON` to the connection then?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+        Ok(Self {
+            path: path.to_string(),
+            conn: OnceLock::new(),
+        })
+    }
+
+    /// This method should return a new memory DB that can be used for tests
+    ///
+    /// # Errors
+    /// Will return `Err` if  there is a failure to open the DB memory connection.
```
vinistock, line 34:
```suggestion
    /// Will return `Err` if there is a failure to open the DB memory connection.
```

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+        ).unwrap_or(0);
+
+        if table_count < EXPECTED_TABLE_COUNT {
+            let schema = std::fs::read_to_string(
+                std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                    .join("src/db")
+                    .join("schema.sql"),
+            )?;
+            conn.execute_batch(&schema)?;
+        }
```
vinistock, line 69: Instead of having to check which tables exist whenever we open a connection, is it possible to know if the DB was freshly created?

If there's anything indicating that we just created the database, we can immediately load the schema whenever that's the case and skip having to verify which tables are in there.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(Debug)]
+pub struct Db {
+    path: String,
+    conn: OnceLock<Connection>,
+}
+
+impl Db {
+    /// # Errors
+    /// Will return `Err` if  there is a failure to load the schema file or if there is
+    /// an error openining or using the DB connection.
```
Morriar, line 23:
```suggestion
    /// an error opening or using the DB connection.
```

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+                "INSERT INTO definitions (id, name_id, definition_type, document_id, start_offset, end_offset) VALUES (?, ?, ?, ?, ?, ?)"
+            )?;
+
+        for (definition_id, definition) in definitions {
+            let definition_type = definition.definition_type_value();
+
+            let (start_offset, end_offset) = match definition {
+                Definition::Class(class) => (class.offset.start_offset(), class.offset.end_offset()),
+                Definition::Module(module) => (module.offset.start_offset(), module.offset.end_offset()),
+            };
```
vinistock, line 123: This should probably be defined in `Definition`, so that we can just do `definition.start_offset()`.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+            let uri_id = definition_to_uri[definition_id];
+            let name_id = definition_to_name[definition_id];
+
+            stmt.execute(rusqlite::params![
+                definition_id.to_string(),
+                name_id.to_string(),
+                definition_type,
+                uri_id.to_string(),
+                i64::from(start_offset),
+                i64::from(end_offset)
```
vinistock, line 133: Why do we need to convert to i64? Can't Sqlite store u32?

rust/index/src/model/definitions.rs
```diff
@@ -50,7 +58,7 @@ impl Definition {
 /// ```
 #[derive(Debug)]
 pub struct ClassDefinition {
-    offset: Offset,
+    pub offset: Offset,
```
vinistock, line 61: Let's not make these public and instead add a getter for it.

rust/index/src/model/graph.rs
```diff
@@ -1,40 +1,67 @@
...
+
 // The `Graph` is the global representation of the entire Ruby codebase. It contains all declarations and their
 // relationships
-#[derive(Default, Debug)]
+#[derive(Debug)]
 pub struct Graph {
     // *** Graph nodes: the following represent possible nodes in our graph ***
     // Map of fully qualified names. These represent global declarations, like `Foo`, `Foo#bar` or `Foo.baz`
-    names: HashMap<NameId, String>,
+    pub names: HashMap<NameId, String>,
```
vinistock, line 15: Same here. Let's not make any of these public as we don't want other entities to be able to mutate the Graph.

Let's add getters that return immutable references to the data.

rust/index/src/model/graph.rs
```diff
@@ -1,40 +1,67 @@
...
     // *** Graph edges: the following represent relationships between nodes ***
     // Map of a fully qualified name to all definitions we discovered for it
     name_to_definitions: HashMap<NameId, HashSet<DefinitionId>>,
     // Reverse map of a definition to its unique fully qualified name
-    definition_to_name: HashMap<DefinitionId, NameId>,
+    pub definition_to_name: HashMap<DefinitionId, NameId>,
     // Map of URI to all definitions discovered in that document
     uris_to_definitions: HashMap<UriId, HashSet<DefinitionId>>,
+    // Reverse map of a definition to the URI of its document
+    pub definition_to_uri: HashMap<DefinitionId, UriId>,
```
vinistock, line 29: Why do we need this new edge?

rust/index/src/model/ids.rs
```diff
@@ -1,6 +1,6 @@
 //! This module contains stable ID representations that composed the `Graph` global representation
-
 use blake3::Hash;
+use std::fmt;
```
vinistock, line 3: I agree with this change, but can we move it to a separate PR to make this one smaller?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+pub struct Db {
+    path: String,
+    conn: OnceLock<Connection>,
+}
+
+impl Db {
+    /// # Errors
+    /// Will return `Err` if  there is a failure to load the schema file or if there is
+    /// an error openining or using the DB connection.
+    pub fn new(path: &str) -> rusqlite::Result<Self> {
```
vinistock, line 24: Is this still supposed to return a Result? I think it's no longer doing anything that may error, right?

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,157 @@
...
+        Ok(Self {
+            path: path.to_string(),
+            conn: OnceLock::new(),
```
vinistock, line 27: If we use a `OnceLock` here, what happens if the user deletes the Sqlite file after we established the connection?

rust/index/src/offset.rs
vinistock: Yes, to prevent someone from mutating these externally.

Morriar: Another option would be to make it `pub(crate)` or `pub(super)`

rust/index/src/model/db.rs
ChallaHalla: I think for detecting if a DB is fresh, we'd just need to check if the file has already been created. That isn't great for situations where we are adding a new table for a new release of the indexer though.

Claude says we could set a version like this:

```
  4. Use user_version pragma:
  // Set version when creating schema
  conn.execute("PRAGMA user_version = 1", [])?;

  // Check version to detect fresh DB
  let version: i64 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
  let is_fresh = version == 0;

```

We could set a version in the code and check to see if the version matches what's in the DB. If it doesn't, then we know it's either freshly created and needs to have the schema loaded, or it's an old version which needs to have new tables created.

rust/index/src/model/db.rs
ChallaHalla: Need to confirm what happens here. I chose this route so that we don't have the overhead of establishing the connection each time we want to run a query. It's not a huge overhead but it would still require reading some piece of the DB file before executing any queries.

ChallaHalla: Might be easier for now to just establish a new connection each time we query though and change that if we see the overhead is too large

PR #65, opened 2025-08-06
Remove URI interning lock

vinistock:
This PR removes the interning of URIs from the index, so that the threads don't have to compete on any locks and everything is merged at the end.

I also switched to using rayon for parallelism, which makes the implementation much simpler.

### DIFF
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -1,87 +1,100 @@
 use std::{
+    cmp, fs,
     sync::{Arc, Mutex},
-    thread,
 };
 
+use crate::{
+    indexing::{
+        errors::{IndexingError, MultipleErrors},
+        ruby_indexer::RubyIndexer,
+    },
+    model::index::Index,
+};
+use rayon::prelude::*;
 use url::Url;
-
-use crate::{indexing::ruby_indexer::RubyIndexer, model::index::Index};
-
+mod errors;
+pub mod indexed_data;
 pub mod ruby_indexer;
 
+// Represents a document to be indexed. If `source` is `Some(String)`, we're indexing a file state that's not committed
+// to disk yet and should then use the given `source` instead of reading from disk
 pub struct Document {
-    uri: String,
+    uri: Url,
     source: Option<String>,
 }
 
 impl Document {
+    /// # Errors
+    ///
+    /// Creating a new document fails on invalid URIs
+    pub fn new(uri: &str, source: Option<String>) -> Result<Self, url::ParseError> {
+        Ok(Self {
+            uri: Url::parse(uri)?,
+            source,
+        })
+    }
+
     #[must_use]
-    pub fn new(uri: String, source: Option<String>) -> Self {
-        Self { uri, source }
+    pub fn path(&self) -> &str {
+        self.uri.path()
     }
 }
 
 /// Indexes the given items, reading the content from disk and populating the given `Index` instance.
 ///
+/// # Errors
+///
+/// Returns Ok if indexing succeeded for all given documents or a vector of errors for all failures
+///
 /// # Panics
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
-pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &Arc<Mutex<Vec<Document>>>) {
-    let num_threads = thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4);
-    let mut threads = Vec::with_capacity(num_threads);
+pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &[Document]) -> Result<(), MultipleErrors> {
+    let chunk_size = cmp::max(10, documents.len() / rayon::current_num_threads());
 
-    for _ in 0..num_threads {
-        let queue_clone = Arc::clone(documents);
-        let index_clone = Arc::clone(index_arc);
-
-        // Each thread creates its own indexer instance, which they use to visit ASTs. The threads pop from a shared
-        // queue to avoid being idle. Definitions are collected for all of the files that got processed and then
-        // returned at the end
-        let handle = thread::spawn(move || {
+    let indexers: Vec<RubyIndexer> = documents
+        .par_chunks(chunk_size)
+        .map(|chunk| {
             let mut ruby_indexer = RubyIndexer::new();
 
-            loop {
-                // Pop a file path from the shared queue. This needs to happen in a small scope to avoid holding the
-                // lock for too long, which would cause other threads to slow down
-                let maybe_document = {
-                    let mut queue = queue_clone.lock().unwrap();
-                    queue.pop()
-                };
-
-                if let Some(document) = maybe_document {
-                    if let Ok(uri) = Url::parse(&document.uri) {
-                        let uri_id = index_clone.lock().unwrap().intern_uri(document.uri.to_string());
-
-                        if let Some(source) = document.source {
-                            ruby_indexer.index(uri_id, &source);
-                        } else if let Ok(source) = std::fs::read_to_string(uri.path()) {
-                            ruby_indexer.index(uri_id, &source);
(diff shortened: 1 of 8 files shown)


rust/index/src/indexing/errors.rs
```diff
@@ -0,0 +1,12 @@
...
+    FileReadError(String),
+}
+
+impl std::fmt::Display for IndexingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            IndexingError::FileReadError(msg) => write!(f, "File read error: {msg}"),
+        }
+    }
+}
```
Morriar, line 26: I wonder if we should also implement `Error`? See https://doc.rust-lang.org/std/error/trait.Error.html.

```suggestion
}

impl std::error::Error for IndexingError {}
```

rust/index/src/main.rs
```diff
@@ -18,14 +18,21 @@ fn main() {
...
+                }
+            })
             .collect::<Vec<_>>()
     };
 
-    let file_queue = Arc::new(Mutex::new(documents));
-
     let index = Arc::new(Mutex::new(Index::new()));
-    index_in_parallel(&index, &file_queue);
+    let errors = index_in_parallel(&index, &documents);
```
Morriar: I guess there is some overhead to run the parallel work. As a followup, I wonder if we should have a threshold, like under 5-10 files we just run in one thread?

rust/index/src/indexing.rs
```diff
@@ -26,62 +42,48 @@ impl Document {
 /// # Panics
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
-pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &Arc<Mutex<Vec<Document>>>) {
-    let num_threads = thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4);
-    let mut threads = Vec::with_capacity(num_threads);
+pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &[Document]) -> Vec<IndexingError> {
+    let chunk_size = documents.len() / rayon::current_num_threads();
```
Morriar: What if we have fewer documents than threads?

```suggestion
    let chunk_size = std::cmp::max(1, documents.len() / rayon::current_num_threads());
```

Morriar: Is there any risk `current_num_threads` is ever 0?

rust/index/src/indexing/indexed_data.rs
```diff
@@ -0,0 +1,81 @@
+use std::collections::HashMap;
+
+use crate::model::{
+    definitions::Definition,
+    ids::{DeclarationId, UriId},
+};
+
+// A view of a declaration that may be incomplete since it is indexer specific and not global. This struct contains
+// everything we discovered about a given declaration inside a single indexing thread
+pub struct IncompleteDeclaration {
```
Morriar:
```suggestion
pub struct PartialDeclaration {
```

rust/index/src/indexing/indexed_data.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+// A collection of declarations specific for this indexer. This is not the global representation of declarations, just a
+// list of all definitions discovered that will be merged into the global index. It helps us avoid having to hash fully
+// qualified names into DeclarationIds multiple times.
+//
+// This is similar to a temporary global index instance, but we prioritize performance since we will throw away this
+// state once it gets merged globally
+pub struct IndexedData {
```
Morriar: I think this whole logic only makes sense when we have > 1 thread otherwise it's a lot of overhead.

Maybe we can name accordingly with something like `IndexingThreadData` or `IndexerThreadData`?

rust/index/src/indexing.rs
vinistock: Good point. I don't believe it can ever be zero. The default is to just return the maximum available parallelism and setting that to zero actually means "determine the maximum parallelism automatically".

rust/index/src/indexing/indexed_data.rs
vinistock: I renamed the struct 👍. I guess on single edits on the editor, we would have a distinct indexing flow that doesn't spawn threads, but other than that scenario I can't think of any others where we wouldn't want to index in parallel.

It'll be good to prototype synchronization and get a better sense of how everything fits together.

rust/index/src/main.rs
vinistock: I set the minimum to 10 for the time being. I played around with some different numbers, but the interesting find is that if you're indexing a very low number of things, it's so fast that it's hard to even make any differences with the chunk size.

For example, indexing the Prism codebase (which has very few Ruby files), with a min chunk size of 10 finishes in 0.06 seconds. With a min chunk size of 50, it finishes in 0.09 seconds.

Maybe this will be more relevant for synchronization, where the number of files being reindexed can be smaller.

rust/index/src/main.rs
```diff
@@ -18,14 +18,21 @@ fn main() {
...
+    for error in errors {
+        eprintln!("Indexing error: {error}");
```
Morriar: Maybe as a follow up we should change the `main` signature so it returns an `Result`:

```rust
fn main() -> Result<(), Box<dyn std::error::Error>>
```

And then return either a `Err()` or `Ok()`.

rust/index/src/indexing/indexed_data.rs
```diff
@@ -0,0 +1,81 @@
...
+            .or_insert_with_key(|name| {
+                let id = DeclarationId::new(name);
+                PartialDeclaration::new(id)
+            })
+            .add_definition(definition);
+    }
+
+    // Returns all the owned data, so that we can merge it into the global index
+    #[must_use]
+    pub fn into(self) -> (HashMap<String, PartialDeclaration>, HashMap<UriId, String>) {
```
Morriar: Is this some kind of custom [`Into`](https://doc.rust-lang.org/std/convert/trait.Into.html) implementation?

rust/index/src/indexing/indexed_data.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use crate::model::{
+    definitions::Definition,
+    ids::{DeclarationId, UriId},
+};
+
+// A view of a declaration that may be incomplete since it is indexer specific and not global. This struct contains
+// everything we discovered about a given declaration inside a single indexing thread
+pub struct PartialDeclaration {
+    declaration_id: DeclarationId,
```
Morriar, line 11: What do you think of returning the declaration full name as a string here as to keep concepts better separated and delegate the id creation to the global state merging when we actually create the declaration?

Morriar: On the other hand we may bloat the memory because all the strings will have to stay alive until we apply the merging 🤔

vinistock: I had originally tried that, but then hashing all fully qualified names into IDs in a single thread was pretty slow. The only reason why I separated this was so that they can be computed in parallel and then inserting the already computed ID is much much faster.

rust/index/src/indexing/indexed_data.rs
vinistock: Not really, maybe we can rename this. Not sure if there's a convention for transferring the ownership.

rust/index/src/indexing/indexed_data.rs
Morriar: Good for me 👍

rust/index/src/main.rs
vinistock: Switched to using a `Result`. It required creating a `MultipleErrors` error struct otherwise we can't return a vector of errors that we accumulate during indexing.

rust/index/src/indexing/indexed_data.rs
vinistock: Renamed to `into_parts` which I think is one of the conventions for transferring ownership.

PR #67, opened 2025-08-06
Add deterministic string hashing using Blake3

Morriar:
I compared a bunch of hashes and `blake3` seemed to perform quite well: 

image

When it comes to collisions using a `[u8; 38]` (256 bits) gives us some leeway:

| Number of entries      | Chance of collision                |
|------------------------|------------------------------------|
| 1 million (10⁶)        | ~10⁻⁷¹ (basically 0)               |
| 1 billion (10⁹)        | ~10⁻⁶⁵ (even more 0)               |
| 1 trillion (10¹²)      | ~10⁻⁵⁸ (still zero in real terms)   |
| 1 septillion (10²⁴)    | ~10⁻³² (still negligible)           |
| 10²⁰                   | ~10⁻¹⁹ (incredibly tiny)            |
| 10³⁸                   | ~50% (this is the "birthday bound")|

Nonetheless, I added a check so we would panic if we index two different names behind the same ID.

Since we plan to aggressively cache data on disk I think we can allow some overhead on the ID size with 256 bits instead of 32 but we should revisit as we progress.

### DIFF
--- a/rust/index/src/model/index.rs
+++ b/rust/index/src/model/index.rs
@@ -24,7 +24,17 @@ impl Index {
 
     fn add_declaration(&mut self, id: DeclarationId, name: String, definitions: Vec<Definition>) {
         match self.declarations.entry(id) {
-            Entry::Occupied(mut _occ) => {
+            Entry::Occupied(existing_entry) => {
+                let existing_declaration = existing_entry.get();
+
+                // Check for hash collision: same hash but different name
+                assert_eq!(
+                    existing_declaration.name, name,
+                    "Hash collision detected! DeclarationId {:?} maps to both '{}' and '{}'",
+                    id, existing_declaration.name, name
+                );
+
+                // Same name, so this is a legitimate duplicate declaration
                 // todo synchronize modifications to the existing declarations
             }
             Entry::Vacant(vac) => {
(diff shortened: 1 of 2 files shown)


rust/index/src/model/index.rs
```diff
@@ -21,20 +22,33 @@ impl Index {
         }
     }
 
+    fn add_declaration(&mut self, id: DeclarationId, name: String, definitions: Vec<Definition>) {
+        match self.declarations.entry(id) {
+            Entry::Occupied(existing_entry) => {
+                let existing_declaration = existing_entry.get();
+
+                // Check for hash collision: same hash but different name
+                assert_eq!(
```
vinistock, line 31: Is this erased in release mode?

Morriar: I don't think so, maybe you're thinking about `debug_assert_eq!`?

I don't think it should be disabled, it will most likely happen in a production setup and we'll want to know about it as soon as possible.

vinistock: I was just thinking about not paying the performance cost of doing this check every time, but maybe it's not significant. I agree that if we start hitting collisions, we need to know about it as soon as possible.

PR #69, opened 2025-08-08
Make the index a graph

### LINKED ISSUE #11: Investigate data structures used to store entries
Investigate which data structures are most adequate for storing index entries. We want to minimize the amount of memory being used, while at the same time allowing for fast operations.

For example, if we know the fully qualified name of something we are searching for, then using a hash map ensures we can quickly access it based on that name. For completion, we need to quickly access with partial names (which is done with a prefix tree in the Ruby LSP).

Can we find a way to structure both a hash map and a prefix tree in a way that the memory used for all entries is shared? Is there some other structure that allows both types of access (full and partial names)?

Should we store all known names in a pool and then use unsigned integers to refer to entries everywhere?

vinistock:
Closes #11

Make the index implementation a graph structure, supporting synchronizing newly indexed definitions. There are a few changes in this PR:

1. Now that the Index is all implemented in a top level graph, with no sub-graphs inside of nodes, merging different `Index` instances becomes much simpler. I made each `RubyIndexer` create its own local index and we merge everything at the end
2. I got rid of the `Declaration` concept, at least for now. Right now, we have global fully qualified names that have relationships with all definitions that compose that name. This effectively moves the sub-graph we had inside of the `Declaration` struct into the top level graph - which is the key change in this PR

Here's a visual representation of what the graph looks like and where each part is stored.

image

As we advance in the implementation, we are likely going to add more edges (relationships) to the graph. For example, inheritance can be represented with new inheritance edges between name IDs.


rust/index/src/offset.rs
```diff
@@ -4,17 +4,13 @@
 //! within a file. It can be used to track positions in source code and convert
 //! between byte offsets and line/column positions.
 
-use crate::model::ids::UriId;
-
 /// Represents a byte offset range within a specific file.
 ///
 /// An `Offset` tracks a contiguous span of bytes from `start_offset` to `end_offset`
 /// within a file identified by `uri_id`. This is useful for representing the location
```
Morriar: This comment will need to be updated

rust/index/src/model/index.rs
```diff
@@ -1,91 +1,243 @@
...
+            uris_to_definitions: HashMap::new(),
         }
     }
 
-    fn add_declaration(&mut self, id: DeclarationId, name: String, definitions: Vec<Definition>) {
-        match self.declarations.entry(id) {
-            Entry::Occupied(existing_entry) => {
-                let existing_declaration = existing_entry.get();
+    #[must_use]
+    pub fn get(&self, name: &str) -> Option<Vec<&Definition>> {
```
Morriar, line 41: I wonder if we really want to allocate the vector here? Should we return an iterator instead?

rust/index/src/model/ids.rs
```diff
@@ -25,24 +25,45 @@ impl UriId {
...
     }
 }
 
+#[derive(Hash, Eq, PartialEq, Debug, Copy, Clone)]
+pub struct DefinitionId([u8; 32]);
+
+impl DefinitionId {
+    #[must_use]
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn new(fully_qualified_name: &str, uri_id: UriId, start_offset: u32) -> Self {
```
Morriar: I wonder if we should include `end_offset` as well?

rust/index/src/model/ids.rs
```diff
@@ -25,24 +25,45 @@ impl UriId {
     }
 }
 
-// A DeclarationId is the hashed version of the unique name representing a declaration. There cannot be two entities
-// that shared the exact same name and so we can guarantee that IDs are both unique and stable, making it easy to load
-// things from cache.
+impl std::fmt::Display for UriId {
```
Morriar, line 27: Should we also implement `Display` for the other ids?

rust/index/src/main.rs
```diff
@@ -48,10 +48,11 @@ fn collect_files_recursive(directory: &PathBuf, uris: &mut Vec<String>) {
                         .is_some_and(|name| !excluded_dirs.contains(&name.to_str().unwrap()))
```
Morriar: Maybe `excluded_dirs` should be a `HashSet`?

rust/index/src/indexing.rs
```diff
@@ -84,14 +83,16 @@ pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &[Document])
 
     // Insert all discovered definitions into the global index
     let mut all_errors = Vec::new();
-    let mut index = index_arc.lock().unwrap();
+    let mut other = Index::new();
 
     for indexer in indexers {
```
Morriar, line 90: Why don't we create one `Indexer` per document and map its index instead? So we just get one index per document and merge everything at the end.

This will also make `uri_id` be an `UriId` instead of an `Option<UriId>`.

rust/index/src/model/index.rs
vinistock: For now, I added this just to make testing a bit easier. I think we need to return a proper `Declaration` struct here that can answer questions including all definitions.

If you're okay with it, I'd leave it as is until we improve the returned struct. It's temporary code.

rust/index/src/model/ids.rs
vinistock: You can't have two definitions with the same start offset, so that's why I went without the end. Is there a situation where we'd need to include it?

rust/index/src/model/ids.rs
vinistock: Sure.

rust/index/src/main.rs
vinistock: This `main` implementation is temporary just to be able to test things out. Let's improve it once we implement a proper CLI with option parsing and what not.

rust/index/src/indexing.rs
vinistock: Switched to doing that.

rust/index/src/model/ids.rs
Morriar: I'm not sure how we would model stuff like `A = B = 42` or `a, b = foo`

rust/index/src/indexing.rs
```diff
@@ -54,44 +53,48 @@ pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &[Document])
...
+                    ruby_indexer
+                })
+                .collect::<Vec<_>>()
         })
         .collect();
 
     // Insert all discovered definitions into the global index
     let mut all_errors = Vec::new();
-    let mut index = index_arc.lock().unwrap();
+    let mut other = Index::new();
```
Morriar: Why do we need `other`? Can't we lock and directly extend `index_arc`?

rust/index/src/model/ids.rs
ChallaHalla: We could actually get rid of the `fully_qualified_name` since `UriId` and `start_offset` should already be unique for each definition

vinistock: > I'm not sure how we would model stuff like A = B = 42 or a, b = foo

Those all have different start offsets. For multi-assigns, each target has their own location with a different start offset byte.

Same thing for assigning to another assignment, the name loc is different for `A` and `B`.

rust/index/src/indexing.rs
vinistock: I think I originally needed this, but with the implementation changes, it doesn't make sense anymore. I removed it.

rust/index/src/indexing.rs
```diff
@@ -54,41 +53,43 @@ pub fn index_in_parallel(index_arc: &Arc<Mutex<Index>>, documents: &[Document])
...
+                    ruby_indexer
+                })
+                .collect::<Vec<_>>()
         })
         .collect();
 
     // Insert all discovered definitions into the global index
     let mut all_errors = Vec::new();
-    let mut index = index_arc.lock().unwrap();
+    let mut index_guard = index_arc.lock().unwrap();
```
Morriar, line 88: Do we still need it to be a `&Arc<Mutex<Index>>`?

vinistock: I think we should keep it in a mutex for now. I would assume that other phases of analysis, like resolving constant references, can happen in parallel and they would need to read from a shared index instance.

We can always revisit later.

PR #73, opened 2025-08-12
Add index integrity verification for tests

vinistock:
When constructing our graph representation of the code, one thing we need to avoid at all costs is any form of data corruption.

For example, we cannot have a fully qualified name registered as a node, but no definitions associated to that name. We also can't a definition that is not associated to any name. In general, any code we index must result in a graph that makes sense and that has the expected nodes and connections.

This PR adds a method to verify the integrity of the index. This method simply follows the entire graph from URI -> Definition -> Fully qualified name, ensuring that all expected relationships exist and panicking otherwise. For now, I only used it in tests, but I made it public so that we can attach it to our CLI in the future.

### DIFF
--- a/rust/index/src/model/graph.rs
+++ b/rust/index/src/model/graph.rs
@@ -132,6 +132,58 @@ impl Graph {
             }
         }
     }
+
+    /// # Panics
+    ///
+    /// This method iterates through the entire graph and panics if there are any inconsistencies. For example, if a
+    /// relationship is created for URI to definition, but that definition is not registered. Or if a name is
+    /// registered, but has no relationship with any definitions
+    pub fn verify_integrity(&self) {
+        for (uri_id, definition_ids) in &self.uris_to_definitions {
+            // Each URI must be registered as a node in the graph
+            let uri = self.uri_pool.get(uri_id).unwrap_or_else(|| {
+                panic!("URI id {uri_id} is registered in URI to definition map, but not present in the URI pool")
+            });
+
+            for definition_id in definition_ids {
+                // Each definition has to be registered as a node in the graph
+                self.definitions.get(definition_id).unwrap_or_else(|| {
+                    panic!(
+                        "Definition id {definition_id} is registered for {uri}, but not present in the definitions map"
+                    )
+                });
+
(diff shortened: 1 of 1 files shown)


rust/index/src/model/index.rs
```diff
@@ -132,6 +132,41 @@ impl Index {
...
         }
     }
+
+    /// # Panics
+    ///
+    /// This method iterates through the entire graph and panics if there are any inconsistencies. For example, if a
+    /// relationship is created for URI to definition, but that definition is not registered. Or if a name is
+    /// registered, but has no relationship with any definitions
+    pub fn verify_integrity(&self) {
+        for (uri_id, definition_ids) in &self.uris_to_definitions {
```
Morriar: If we only check this list, wouldn't we miss an orphaned definitions that is not associated to an URI anymore?

rust/index/src/model/index.rs
```diff
@@ -132,6 +132,41 @@ impl Index {
...
+            let uri = self.uri_pool.get(uri_id).unwrap_or_else(|| {
```
Morriar: For each loop, let's add a comment to explain what we are validating:

```rust
// Each name should have at least one definition
```

etc.

rust/index/src/model/index.rs
vinistock: Good catch! We don't have a reverse map, so for now we need to go through the values, which I guess is fine for the integrity check.

Morriar: I wonder if, as the index grow, this code is going to become horrible to maintain. I suggested another approach in https://github.com/Shopify/index/pull/99. LMKWYT

vinistock: Honestly, I would rather wait to see what patterns and needs emerge before investing in a more elaborate framework to create validation rules.

It's not clear to me that we'll all of that or if that is generic enough to apply to other stages of the analysis.

Maybe other folks feel differently, but I would favour a more simplistic approach and we can evolve the complexity as the need emerges.

Morriar: > all of that

It's basically an array of procs..

Issue #70, opened 2025-08-11
Generate visual representation of the index graph


Morriar:
With https://github.com/Shopify/index/pull/69 we're moving the implementation of the index to a graph structure:

* Nodes: names, definitions, URIs
* Edges: name -> definitions, definition -> name, uri -> definition

Vini provided a visualization of the structure:

<img width="2659" height="1198" alt="Image" src="https://github.com/user-attachments/assets/4286d20c-fd53-4755-ada0-0760a3832532" />

Let's provide a way to automatically generate this kind of representation directly backed into the index. 

So for example this:

```bash
$ index_cli path/to/files --dot graph.dot
```

Would directly generate the visual representation of the graph in [DOT format](https://graphviz.org/doc/info/lang.html). One can take inspiration from how we do it in  Spoom's POSet: https://github.com/Shopify/spoom/blob/main/lib/spoom/poset.rb#L101-L129

Of course as we aim to index quite large codebases, we should find ways to make the output actually readable:

* Should we be able to filter the edges we want to show?
* Or the kind of nodes?
* Or to focus on a node and only show n-parents/children?
* Or is there a way to have an interactive map of the graph?
* Should we also have a [`--mermaid` output](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/)?

PR #74, opened 2025-08-12
Add CLI stub

Morriar:
Make it easier to pass options to `index_cli`, useful when testing things or to pass options like for #70.

I think this is enough for now, we can make this better later.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -74,6 +74,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "assert_cmd"
+version = "2.0.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bd389a4b2970a01282ee455294913c0a43724daedcd1a24c3eb0ec1c1320b66"
+dependencies = [
+ "anstyle",
+ "bstr",
+ "doc-comment",
+ "libc",
+ "predicates",
+ "predicates-core",
+ "predicates-tree",
+ "wait-timeout",
+]
+
+[[package]]
+name = "autocfg"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+
+[[package]]
 name = "bindgen"
 version = "0.66.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


rust/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
vinistock: Why do we need to specify these?

rust/Cargo.toml
```diff
@@ -13,12 +13,12 @@ codegen-units = 1
 
 [workspace.lints.clippy]
 # See: https://doc.rust-lang.org/clippy/lints.html
-correctness = "forbid"
-suspicious = "forbid"
-perf = "forbid"
-complexity = "forbid"
-style = "forbid"
+correctness = "deny"
```
vinistock, line 16: What's the difference between `forbid` and `deny`? Why do we need to change it?

rust/rust-toolchain.toml
Morriar: Not needed since #94

rust/Cargo.toml
Morriar: From https://doc.rust-lang.org/rustc/lints/levels.html, the `forbid` level is the same as `deny`. The only difference is that it can't be overridden.

We need it because of `clap`'s derive. 

Clap comes with two flavors, a [`derive`](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html) (the one we use) and ` builder`.

With the `derive` mode, options are defined in a struct using the clap DSL:

```rust
#[derive(Parser)]
#[command(version, about, long_about = None)]
struct Cli {
    /// Optional name to operate on
    name: Option<String>,

    /// Sets a custom config file
    #[arg(short, long, value_name = "FILE")]
    config: Option<PathBuf>,

    /// Turn debugging information on
    #[arg(short, long, action = clap::ArgAction::Count)]
    debug: u8,

    #[command(subcommand)]
    command: Option<Commands>,
}

#[derive(Subcommand)]
enum Commands {
    /// does testing things
    Test {
        /// lists test values
        #[arg(short, long)]
        list: bool,
    },
}

fn main() {
    let cli = Cli::parse();
    # ...
}
```

The `builder` mode require us to build the CLI manually:

```rust
fn main() {
    let matches = command!() // requires `cargo` feature
        .arg(arg!([name] "Optional name to operate on"))
        .arg(
            arg!(
                -c --config <FILE> "Sets a custom config file"
            )
            // We don't have syntax yet for optional options, so manually calling `required`
            .required(false)
            .value_parser(value_parser!(PathBuf)),
        )
        .arg(arg!(
            -d --debug ... "Turn debugging information on"
        ))
        .subcommand(
            Command::new("test")
                .about("does testing things")
                .arg(arg!(-l --list "lists test values").action(ArgAction::SetTrue)),
        )
        .get_matches();
    # ...
```

I like the `derive` approach better. It's more concise, it separate the CLI options from the actual main logic and it's easier to read/maintain.

But the derive uses a bunch of macros that try to disable a few clippy rules and it errors since we use `forbid` level.

vinistock: Got it! Thanks

PR #94, opened 2025-08-13
Pin minimal Rust version to 1.89

Morriar:
The let chains feature was stabilized in 1.88 so we have a strong requirement on it in `index.rs`.

https://github.com/Shopify/index/blob/main/rust/index/src/model/index.rs#L121-L122

See https://github.com/rust-lang/rust/issues/53667.

While at it, let's use the latest version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,12 @@ jobs:
         with:
           submodules: "recursive"
 
-      - name: Install Rust tools
-        run: |
-          rustup update --no-self-update stable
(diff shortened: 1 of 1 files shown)


rust/index/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "index"
 version = "0.1.0"
 edition = "2024"
+rust-version = "1.88.0"
```
vinistock:
```suggestion
rust-version = "1.89.0"
```

PR #96, opened 2025-08-13
Separate out local VS global indexes

ChallaHalla:
## Separate Local and Global Index structures

This PR introduces a clear architectural separation between local and global indexing concerns by creating a dedicated `LocalIndex` struct for single-file indexing operations.

### Key Changes
- **New `LocalIndex` struct**: Lightweight index for collecting definitions during single-file indexing
- **Updated `RubyIndexer`**: Now uses `LocalIndex` instead of the full `Index` struct
- **Refined merge operations**: `Index::update_from_local()` and `Index::extend_from_local()` replace generic merge methods
- **Cleaner API boundaries**: Local indexing operations are now isolated from global state management

### Why This Matters
This separation is crucial for upcoming database work. The global `Index` will need to interact with persistent storage while `LocalIndex` operations remain in-memory. This refactoring eliminates the need for conditional logic throughout the codebase to distinguish between local vs global operations.

### DIFF
--- a/rust/index/src/model/local_index.rs
+++ b/rust/index/src/model/local_index.rs
@@ -46,47 +46,4 @@ impl LocalIndex {
             .or_default()
             .insert(definition_id);
     }
-
-    #[must_use]
-    pub fn uri_id(&self) -> UriId {
-        self.uri_id
-    }
-
-    #[must_use]
-    pub fn definitions(&self) -> &HashMap<DefinitionId, Definition> {
-        &self.definitions
-    }
-
-    #[must_use]
-    pub fn names(&self) -> &HashMap<NameId, String> {
-        &self.names
-    }
-
-    #[must_use]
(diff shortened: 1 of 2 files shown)


rust/index/src/model/local_index.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::model::definitions::Definition;
+use crate::model::ids::{DefinitionId, NameId, UriId};
+
+/// A lightweight index for collecting definitions during single-file indexing.
+/// Used by RubyIndexer to accumulate definitions before merging into the global Index.
+/// Contains the same hashmaps as the global Index for consistent structure.
+#[derive(Debug)]
+pub struct LocalIndex {
+    pub uri_id: UriId,
```
ChallaHalla, line 11: One assumption being made here is that a local index will just relate to a single URI. This assumption influenced the new `update_from_local` method logic

PR #97, opened 2025-08-13
Add command to generate graph representation

jenny-codes:
## Summary

Adds DOT format visualization to debug the indexer's graph structure. Useful for understanding how classes, modules, and their definitions are connected.

## Usage

```
# Visualize a directory
cargo run --bin index_cli -- ./src -o graph.dot

# view the generated dot file
dot -Tsvg graph.dot -o graph.svg
# or, interactively (but beware that it doesn't run on bigger input)
xdot graph.dot

```

#### dot

```
digraph {
    rankdir=TB;

    "Name:RubyLsp" [label="RubyLsp",shape=hexagon];
    "Name:RubyLsp::Server" [label="RubyLsp::Server",shape=hexagon];

    "def_6659466995e20096b7beb50fc33de0d901655c992405d18ac937b94a87e53d9e" [label="Class(RubyLsp::Server)",shape=ellipse];
    "def_3d5da1207c50cb438f7788658a1b1ce83b8e98da779045b0d67aff460bc0f44d" [label="Module(RubyLsp)",shape=ellipse];

    "file:///Users/jennyshih/src/github.com/Shopify/ruby-lsp/lib/ruby_lsp/server.rb" [label="server.rb",shape=box];

    "Name:RubyLsp::Server" -> "def_6659466995e20096b7beb50fc33de0d901655c992405d18ac937b94a87e53d9e" [dir=both];
    "Name:RubyLsp" -> "def_3d5da1207c50cb438f7788658a1b1ce83b8e98da779045b0d67aff460bc0f44d" [dir=both];
    "def_6659466995e20096b7beb50fc33de0d901655c992405d18ac937b94a87e53d9e" -> "file:///Users/jennyshih/src/github.com/Shopify/ruby-lsp/lib/ruby_lsp/server.rb";
    "def_3d5da1207c50cb438f7788658a1b1ce83b8e98da779045b0d67aff460bc0f44d" -> "file:///Users/jennyshih/src/github.com/Shopify/ruby-lsp/lib/ruby_lsp/server.rb";
}

```

#### svg

server_unfiltered

## Future Iterations

Filtering capabilities will be added in a different pr

## Testing

```
cargo test --package index
```

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -539,6 +539,7 @@ dependencies = [
  "clap",
  "predicates",
  "rayon",
+ "regex",
  "rmp-serde",
  "ruby-prism",
  "rusqlite",
--- a/rust/index/src/main.rs
+++ b/rust/index/src/main.rs
@@ -8,6 +8,7 @@ use clap::Parser;
 use index::{
     indexing::{self, errors::MultipleErrors},
     model::graph::Graph,
+    visualization::dot,
 };
 
 #[derive(Parser, Debug)]
@@ -21,6 +22,9 @@ struct Args {
         help = "Run integrity checks on the index after processing"
     )]
     check_integrity: bool,
+
+    #[arg(long = "visualize")]
+    visualize: bool,
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
@@ -50,6 +54,17 @@ fn main() -> Result<(), Box<dyn Error>> {
             }
             std::process::exit(1);
         }
+    }
+
+    // Generate visualization or print statistics
+    let graph_lock = graph_arc.lock().unwrap();
+    if args.visualize {
+        println!("{}", dot::generate(&graph_lock));
+    } else {
+        println!("Indexed {} files", documents.len());
+        println!("Found {} names", graph_lock.names().len());
+        println!("Found {} definitions", graph_lock.definitions().len());
+        println!("Found {} URIs", graph_lock.uri_pool().len());
     }
 
     Ok(())
--- a/rust/index/src/model/definitions.rs
+++ b/rust/index/src/model/definitions.rs
@@ -73,6 +73,22 @@ impl Definition {
             Definition::Method(it) => it.offset.end(),
         }
     }
+
+    #[must_use]
+    pub fn kind(&self) -> &'static str {
+        match self {
+            Definition::Class(_) => "Class",
+            Definition::Module(_) => "Module",
+            Definition::Constant(_) => "Constant",
+            Definition::Method(_) => "Method",
+            Definition::AttrAccessor(_) => "AttrAccessor",
+            Definition::AttrReader(_) => "AttrReader",
+            Definition::AttrWriter(_) => "AttrWriter",
+            Definition::GlobalVariable(_) => "GlobalVariable",
+            Definition::InstanceVariable(_) => "InstanceVariable",
+            Definition::ClassVariable(_) => "ClassVariable",
+        }
+    }
 }
 
 /// A class definition
--- a/rust/index/src/visualization/dot.rs
+++ b/rust/index/src/visualization/dot.rs
@@ -1,0 +1,194 @@
+//! DOT format generator for Graphviz visualization of the graph structure.
+
+use std::fmt::Write;
+
+use crate::model::graph::Graph;
+
+const NAME_NODE_SHAPE: &str = "hexagon";
+const DEFINITION_NODE_SHAPE: &str = "ellipse";
+const URI_NODE_SHAPE: &str = "box";
+
+/// Escapes a string for use in DOT format labels and identifiers.
+fn escape_dot_string(s: &str) -> String {
+    if !s.contains('"') {
+        return s.to_string();
+    }
+
+    let mut result = String::with_capacity(s.len());
+    for c in s.chars() {
+        match c {
+            '"' => result.push_str("\\\""),
+            _ => result.push(c),
+        }
+    }
+    result
+}
+
+#[must_use]
+pub fn generate(graph: &Graph) -> String {
+    let mut output = String::new();
+    output.push_str("digraph {\n");
+    output.push_str("    rankdir=TB;\n\n");
+
+    write_name_nodes(&mut output, graph);
+    write_definition_nodes(&mut output, graph);
+    write_uri_nodes(&mut output, graph);
+    write_name_to_definition_edges(&mut output, graph);
+    write_definition_to_uri_edges(&mut output, graph);
+
+    output.push_str("}\n");
+    output
+}
+
+fn write_name_nodes(output: &mut String, graph: &Graph) {
+    let mut names: Vec<_> = graph.names().values().collect();
+    names.sort();
+
+    for name in names {
+        let escaped_name = escape_dot_string(name);
+        let node_id = format!("Name:{name}");
+        let _ = writeln!(
+            output,
+            "    \"{node_id}\" [label=\"{escaped_name}\",shape={NAME_NODE_SHAPE}];"
+        );
+    }
+    output.push('\n');
+}
+
+fn write_definition_nodes(output: &mut String, graph: &Graph) {
+    let mut definitions: Vec<_> = graph.definitions().iter().collect();
+    definitions.sort_by_key(|(def_id, _)| def_id.to_string());
+
+    for (def_id, definition) in definitions {
+        if let Some(name_id) = graph.definition_to_name().get(def_id)
+            && let Some(name) = graph.names().get(name_id)
+        {
+            let def_type = definition.kind();
+            let escaped_name = escape_dot_string(name);
+
+            let _ = writeln!(
+                output,
+                "    \"def_{def_id}\" [label=\"{def_type}({escaped_name})\",shape={DEFINITION_NODE_SHAPE}];"
+            );
+        }
+    }
+    output.push('\n');
+}
+
+fn write_uri_nodes(output: &mut String, graph: &Graph) {
+    let mut uris: Vec<_> = graph.uri_pool().values().collect();
+    uris.sort();
+
+    for uri in uris {
+        let label = uri.rsplit('/').next().unwrap_or(uri);
+        let escaped_uri = escape_dot_string(uri);
+        let escaped_label = escape_dot_string(label);
+        let _ = writeln!(
+            output,
+            "    \"{escaped_uri}\" [label=\"{escaped_label}\",shape={URI_NODE_SHAPE}];"
+        );
+    }
+    output.push('\n');
+}
+
+fn write_name_to_definition_edges(output: &mut String, graph: &Graph) {
+    let mut name_to_def_edges: Vec<_> = Vec::new();
+
+    for (name_id, definition_ids) in graph.name_to_definitions() {
+        if let Some(name) = graph.names().get(name_id) {
+            for def_id in definition_ids {
+                if graph.definitions().contains_key(def_id) {
+                    name_to_def_edges.push((name.clone(), def_id.to_string()));
+                }
+            }
+        }
+    }
+    name_to_def_edges.sort();
+
+    for (name, def_id) in name_to_def_edges {
+        let name_node = format!("Name:{}", escape_dot_string(&name));
+        let _ = writeln!(output, "    \"{name_node}\" -> \"def_{def_id}\" [dir=both];");
+    }
+}
+
+fn write_definition_to_uri_edges(output: &mut String, graph: &Graph) {
+    let mut uri_edges: Vec<_> = graph.uris_to_definitions().iter().collect();
+    uri_edges.sort_by_key(|(uri_id, _)| uri_id.to_string());
+
+    for (uri_id, definition_ids) in uri_edges {
+        if let Some(uri) = graph.uri_pool().get(uri_id) {
+            let escaped_uri = escape_dot_string(uri);
+            let mut sorted_def_ids: Vec<_> = definition_ids.iter().collect();
+            sorted_def_ids.sort_by_key(std::string::ToString::to_string);
+
+            for def_id in sorted_def_ids {
+                if graph.definitions().contains_key(def_id) {
+                    let _ = writeln!(output, "    \"def_{def_id}\" -> \"{escaped_uri}\";");
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::GraphTest;
+
+    fn create_test_graph() -> Graph {
+        let mut graph_test = GraphTest::new();
+        graph_test.index_uri(
+            "file:///test.rb",
+            "
+                class TestClass
+                end
+
+                module TestModule
+                end
+            ",
+        );
+        graph_test.graph
+    }
+
+    #[test]
+    fn test_dot_generation() {
+        let graph = create_test_graph();
+        let dot_output = generate(&graph);
+
+        let class_def_id = graph
+            .definitions()
+            .iter()
+            .find(|(_, def)| matches!(def, crate::model::definitions::Definition::Class(_)))
+            .map(|(id, _)| id.to_string())
+            .unwrap();
+
+        let module_def_id = graph
+            .definitions()
+            .iter()
+            .find(|(_, def)| matches!(def, crate::model::definitions::Definition::Module(_)))
+            .map(|(id, _)| id.to_string())
+            .unwrap();
+
+        let expected = format!(
+            r#"digraph {{
+    rankdir=TB;
+
+    "Name:TestClass" [label="TestClass",shape=hexagon];
+    "Name:TestModule" [label="TestModule",shape=hexagon];
+
+    "def_{module_def_id}" [label="Module(TestModule)",shape=ellipse];
+    "def_{class_def_id}" [label="Class(TestClass)",shape=ellipse];
+
+    "file:///test.rb" [label="test.rb",shape=box];
+
+    "Name:TestClass" -> "def_{class_def_id}" [dir=both];
+    "Name:TestModule" -> "def_{module_def_id}" [dir=both];
+    "def_{module_def_id}" -> "file:///test.rb";
+    "def_{class_def_id}" -> "file:///test.rb";
+}}
+"#
+        );
+
+        assert_eq!(dot_output, expected);
+    }
+}
--- a/rust/index/Cargo.toml
+++ b/rust/index/Cargo.toml
@@ -25,6 +25,7 @@ rmp-serde = "1.3.0"
 tempfile = "3.0"
 assert_cmd = "2.0"
 predicates = "3.1"
+regex = "1.10"
 
 [lints]
 workspace = true
--- a/rust/index/src/lib.rs
+++ b/rust/index/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod indexing;
 pub mod model;
 pub mod offset;
+pub mod visualization;
 
 #[cfg(test)]
 pub mod test_utils;
--- a/rust/index/src/visualization.rs
+++ b/rust/index/src/visualization.rs
@@ -1,0 +1,6 @@
+//! Graph visualization module for the Index structure.
+//!
+//! This module provides functionality to generate visual representations of the Index's
+//! internal graph structure, showing the relationships between Names, Definitions, and URIs.
+
+pub mod dot;
--- a/rust/index/tests/cli.rs
+++ b/rust/index/tests/cli.rs
@@ -1,5 +1,7 @@
 use assert_cmd::prelude::*;
 use predicates::prelude::*;
+use regex::Regex;
+use std::fs;
 use std::process::Command;
 
 #[test]
@@ -8,7 +10,9 @@ fn prints_help() {
     cmd.arg("--help");
     cmd.assert()
         .success()
-        .stdout(predicate::str::contains("A Ruby code indexer"));
+        .stdout(predicate::str::contains("A Ruby code indexer"))
+        .stdout(predicate::str::contains("Usage:"))
+        .stdout(predicate::str::contains("--visualize"));
 }
 
 #[test]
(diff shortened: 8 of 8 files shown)


rust/Cargo.lock
```diff
@@ -176,6 +176,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "be92d32e80243a54711e5d7ce823c35c41c9d929dc4ab58e1276f625841aadf9"
 dependencies = [
  "clap_builder",
+ "clap_derive",
```
Morriar: You can rebase on main to get clap and the args handling

rust/index/examples/visualize.rs
```diff
@@ -0,0 +1,56 @@
+//! Example program to visualize the Index structure
```
Morriar: Do we really need these examples files?

rust/index/src/bin/visualize.rs
```diff
@@ -0,0 +1,48 @@
+//! Minimal CLI for generating DOT visualization of Ruby code
```
Morriar: Let's merge this into the current main: https://github.com/Shopify/index/blob/main/rust/index/src/main.rs

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,229 @@
...
+#[must_use]
+pub fn generate(index: &Index) -> String {
+    let mut output = String::new();
+    writeln!(output, "digraph {{").unwrap();
+    writeln!(output, "    rankdir=BT;").unwrap();
+    writeln!(output).unwrap();
+
+    write_nodes(&mut output, index);
+    writeln!(output).unwrap();
+    writeln!(output, "    // All edges").unwrap();
```
Morriar: Let's move this inside `write_edges`?

Also, should we add a comment about `// All nodes` for parity?

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,229 @@
...
+use std::fmt::Write;
+
+use crate::model::definitions::Definition;
+use crate::model::index::Index;
+
+/// Generate DOT format representation directly from the Index
+#[must_use]
+pub fn generate(index: &Index) -> String {
+    let mut output = String::new();
+    writeln!(output, "digraph {{").unwrap();
```
Morriar: Why do we need two `{{`?

rust/index/src/model/graph.rs
```diff
@@ -132,6 +132,37 @@ impl Graph {
...
+    }
+
+    #[must_use]
+    pub fn definition_to_name(&self) -> &HashMap<DefinitionId, NameId> {
+        &self.definition_to_name
+    }
+
+    #[must_use]
+    pub fn uris_to_definitions(&self) -> &HashMap<UriId, HashSet<DefinitionId>> {
+        &self.uris_to_definitions
```
jenny-codes: Check if the HashSet can be mutated. If it can, it is not very safe and we would probably want to explore a safer option than directly exposing this property

jenny-codes: Idea: Implement iterator. May be more idiomatic.

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,267 @@
...
+    for (def_id, definition) in index.definitions() {
+        if let Some(name_id) = index.definition_to_name().get(def_id)
+            && let Some(name) = index.names().get(name_id)
+        {
+            let def_type = match definition {
+                Definition::Class(_) => "Class",
+                Definition::Module(_) => "Module",
+            };
+
+            let node_id = format!("{def_type}:{name}");
```
jenny-codes: Here we want to have separate nodes for each definition that came from different URI, instead of merging them all into one if they share the same name. Hence we should use the `def_id` instead (hash of the uri and offset of the definition) as the node_id. But we still want the def_type and name be the label.

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,267 @@
...
+        if let Some(name) = index.names().get(name_id) {
+            let name_node = format!("Name:{name}");
+
+            for def_id in definition_ids {
+                if let Some(definition) = index.definitions().get(def_id) {
+                    let def_type = match definition {
+                        Definition::Class(_) => "Class",
+                        Definition::Module(_) => "Module",
+                    };
+                    let def_node = format!("{def_type}:{name}");
```
jenny-codes: Also need to change here

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,267 @@
...
+            for def_id in definition_ids {
+                if let Some(definition) = index.definitions().get(def_id)
+                    && let Some(name_id) = index.definition_to_name().get(def_id)
+                    && let Some(name) = index.names().get(name_id)
+                {
+                    let def_type = match definition {
+                        Definition::Class(_) => "Class",
+                        Definition::Module(_) => "Module",
+                    };
+                    let def_node = format!("{def_type}:{name}");
```
jenny-codes: Also have to change here

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,267 @@
...
+            '\\' => result.push_str("\\\\"),
+            '"' => result.push_str("\\\""),
+            '\n' => result.push_str("\\n"),
+            '<' => result.push_str("\\<"),
+            '>' => result.push_str("\\>"),
+            _ => result.push(ch),
+        }
+    }
+    result
+}
```
jenny-codes: YANGI–based on what we know of the string inputs (the URIs and the names) this escaping is most likely not needed as the input won't include those special characters. We should remove these two functions for now and we will add them back if future proves that we do need them

rust/index/src/model/graph.rs
jenny-codes: Looks like it's immutable after all. Test code: 

```rust
use std::collections::{HashMap, HashSet};

struct Graph {
    data: HashMap<String, HashSet<i32>>,
}

impl Graph {
    fn new() -> Self {
        let mut data = HashMap::new();
        let mut set = HashSet::new();
        set.insert(1);
        set.insert(2);
        set.insert(3);
        data.insert("key1".to_string(), set);

        Graph { data }
    }

    // This returns an immutable reference
    pub fn get_data(&self) -> &HashMap<String, HashSet<i32>> {
        &self.data
    }
}

fn main() {
    let graph = Graph::new();
    let data_ref = graph.get_data();

    if let Some(hashset) = data_ref.get("key1") {
        // Try mutating
        hashset.insert(4);
    }
}
```

Result of running the file

```
error[E0596]: cannot borrow `*hashset` as mutable, as it is behind a `&` reference
  --> /tmp/hashset_mutation_attempt.rs:31:9
   |
29 |     if let Some(hashset) = data_ref.get("key1") {
   |                 ------- consider changing this binding's type to be: `&mut HashSet<i32>`
30 |         // Try mutating
31 |         hashset.insert(4);
   |         ^^^^^^^ `hashset` is a `&` reference, so the data it refers to cannot be borrowed as mutable

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0596`.

```

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use std::fmt::Write;
+
+use crate::model::definitions::Definition;
+use crate::model::graph::Graph;
+
+/// Generate DOT format representation directly from the Graph
+#[must_use]
+pub fn generate(index: &Graph) -> String {
+    let mut output = String::with_capacity(estimate_output_size(index));
```
vinistock: I'd probably start simpler with just `String::new()` and then optimize if we conclude it's necessary.

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,213 @@
+//! DOT format generator for Graphviz visualization of the Index structure.
+
+use std::fmt::Write;
+
+use crate::model::definitions::Definition;
+use crate::model::graph::Graph;
+
+/// Generate DOT format representation directly from the Graph
+#[must_use]
+pub fn generate(index: &Graph) -> String {
```
vinistock:
```suggestion
pub fn generate(graph: &Graph) -> String {
```

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Estimate output size for pre-allocation to reduce reallocations
+fn estimate_output_size(index: &Graph) -> usize {
+    // Rough estimate: 100 bytes per node + 50 bytes per edge + overhead
+    let node_count = index.names().len() + index.definitions().len() + index.uri_pool().len();
+    let edge_count = index.name_to_definitions().len() + index.uris_to_definitions().len();
+    node_count * 100 + edge_count * 50 + 1000
+}
+
+/// Write a legend explaining the visual encoding
+fn write_legend(output: &mut String) {
```
vinistock: I wonder if we could get away without a legend if we made the node's contents really obvious.

For example, if the content is a file name, then there's no need for the legend since we know it's a URI.

Maybe we could have:

- URI nodes: the file name
- Name nodes: the name itself
- Definition nodes: the name prefixed with the type of definition. This would help us visualize even codebases with errors (e.g.: a user reopens the same entity as a class, but it was a module originally)

Something like this

```
(Foo::Bar) -> (Module(Foo::Bar)) -> [bar1.rb]
           -> (Class(Foo::Bar)) -> [bar2.rb]
```

rust/index/src/main.rs
```diff
@@ -28,15 +29,30 @@ struct Args {
...
     ])]
     exclude: Vec<String>,
+
+    /// Generate graph visualization
+    #[arg(short, long = "visualize")]
+    visualize: bool,
+
+    /// Output file for visualization
+    #[arg(short, long = "output", value_name = "FILE")]
+    output: Option<String>,
```
vinistock: What do you think about renaming `output` to something like `visualization_output` and then using the presence of this argument as indication of whether we should do it or not?

Thus removing the `visualize` boolean.

rust/index/src/main.rs
```diff
@@ -28,15 +29,30 @@ struct Args {
...
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::parse();
     let excluded_dirs_set: HashSet<&str> = args.exclude.iter().map(String::as_str).collect();
 
-    let documents = {
+    let documents = if PathBuf::from(&args.dir).is_file() {
```
vinistock: This is addressed in #104, so hopefully we can get that one shipped and you won't need to worry about this.

rust/index/src/main.rs
jenny-codes: I like it 👍🏻

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+                output,
+                "    \"def_{def_id}\" [label=\"{def_type}({name})\",shape={DEFINITION_NODE_SHAPE}];"
+            );
+        }
+    }
+    output.push('\n');
+
+    // Write URI nodes
+    for uri in graph.uri_pool().values() {
+        let label = uri.rsplit('/').next().unwrap_or(uri);
```
jenny-codes, line 83: Currently this will make it so that the graph is only showing the file's base name. IMO ideally we'd want to display the relative path from the project root. I will do it in a follow up

rust/index/src/main.rs
```diff
@@ -28,15 +29,26 @@ struct Args {
         "fixtures",
     ])]
     exclude: Vec<String>,
+
+    /// Generate graph visualization and write to file (use "-" for stdout)
+    #[arg(short = 'o', long = "visualization-output", value_name = "FILE")]
```
Morriar: Let's avoid the short options for now until we have a better idea of the options with want.

rust/index/src/main.rs
```diff
@@ -50,6 +62,27 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    // Generate visualization or print statistics
+    if let Some(output_path) = args.visualization_output {
+        let dot_output = dot::generate(&index);
+
+        if output_path == "-" {
+            println!("{dot_output}");
+        } else {
+            fs::write(&output_path, &dot_output)?;
+            println!("Visualization written to {output_path}");
+        }
```
Morriar: Let's always print the output to stdout and let the user redirect to a file if needed

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    // Write Definition nodes
+    for (def_id, definition) in graph.definitions() {
+        if let Some(name_id) = graph.definition_to_name().get(def_id)
+            && let Some(name) = graph.names().get(name_id)
+        {
+            let def_type = match definition {
+                Definition::Class(_) => "Class",
+                Definition::Module(_) => "Module",
+            };
```
Morriar: I think this can be moved to the `Definition` itself

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    output.push_str("}\n");
+    output
+}
+
+fn write_nodes_and_edges(output: &mut String, graph: &Graph) {
+    // Write Name nodes
+    for name in graph.names().values() {
+        let node_id = format!("Name:{name}");
+        let _ = writeln!(output, "    \"{node_id}\" [label=\"{name}\",shape={NAME_NODE_SHAPE}];");
```
Morriar: I wonder if we should reopen the implementation of `Name`, `Definition` and `Uri` in this file to add a `to_dot()` method?

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::indexing::ruby_indexer::RubyIndexer;
```
Morriar: Let's use https://github.com/Shopify/index/blob/main/rust/index/src/test_utils.rs instead

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+        indexer.index(source);
+        indexer.into_parts().0
+    }
+
+    #[test]
+    fn test_dot_generation_basic() {
+        let graph = create_test_graph();
+        let dot_output = generate(&graph);
+
+        assert!(dot_output.contains("digraph {"));
```
Morriar: The output should be small, can't we compare the full strings instead of doing `contains`?

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,173 @@
+use std::fs;
+use std::io::Write;
+use std::process::Command;
+use tempfile::NamedTempFile;
+
+fn get_binary_path() -> String {
```
Morriar: Can'we use `cargo_bin` like in https://github.com/Shopify/index/blob/main/rust/index/tests/cli.rs?

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+    output.push('\n');
+
+    // Write edges between Names and Definitions (bidirectional)
+    for (name_id, definition_ids) in graph.name_to_definitions() {
+        if let Some(name) = graph.names().get(name_id) {
+            let name_node = format!("Name:{name}");
+
+            for def_id in definition_ids {
+                if graph.definitions().contains_key(def_id) {
+                    let _ = writeln!(output, "    \"{name_node}\" -> \"def_{def_id}\" [dir=both];");
```
Morriar: Should we escape all the strings that end up in between `"`?

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,173 @@
...
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join("index_cli")
+        .to_str()
+        .unwrap()
+        .to_string()
+}
+
+fn create_test_ruby_file(content: &str) -> NamedTempFile {
```
Morriar: Could we instead create a tempdir and create file into it so we can specify the name of these files and assert on the output for URIs?

rust/index/src/visualization/dot.rs
jenny-codes: Can you expand on that? Is there a string value here that is not escaped?

Morriar: If I create a name like `"foo"`, it will break the output:

```dot
    ""foo"" -> "def_0000" [dir=both];
```

rust/index/src/main.rs
jenny-codes: I ended up consolidating the two options into one `---visualize` that prints the output to stdout which we can then pipe elsewhere as per Alex's suggestion

rust/index/src/visualization/dot.rs
jenny-codes: Ah I see thanks. Will update

jenny-codes: We actually thought about handling it and decided to remove it for now https://app.graphite.dev/github/pr/Shopify/index/97/Add-command-to-generate-graph-representation?utm_source=gt-slack-notif&panel=timeline#comment-PRRC_kwDOML-q9M6H2Iqk

I just looked into it more, and it looks like Ruby does accept special characters in method definition (when using `define_method`) so we actually need to support it

rust/index/src/model/definitions.rs
```diff
@@ -73,6 +73,22 @@ impl Definition {
             Definition::Method(it) => it.offset.end(),
         }
     }
+
+    #[must_use]
+    pub fn type_name(&self) -> &'static str {
```
Morriar: `type` is a bit overloaded in our context, let's use `kind` instead:

```suggestion
    pub fn kind(&self) -> &'static str {
```

rust/index/src/main.rs
```diff
@@ -21,6 +22,19 @@ struct Args {
...
+
+    #[arg(long = "exclude", value_name = "NAME", num_args = 0.., default_values = [
+        "tmp",
+        "test",
+        "vendor",
+        ".vscode",
+        "vscode",
+        "fixtures",
+    ])]
+    exclude: Vec<String>,
```
Morriar: I don't think this belongs in this PR?

rust/index/src/main.rs
```diff
@@ -52,5 +66,17 @@ fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 
+    // Generate visualization or print statistics
+    let graph_lock = graph_arc.lock().unwrap();
+    if args.visualize {
+        let dot_output = dot::generate(&graph_lock);
+        println!("{dot_output}");
```
Morriar:
```suggestion
        println!("{}", dot::generate(&graph_lock));
```

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,84 @@
...
+        .arg("--help")
+        .output()
+        .expect("Failed to execute command");
+
+    assert!(output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("Usage:"));
+    assert!(stdout.contains("--visualize"));
+    assert!(stdout.contains("--exclude"));
+}
```
Morriar: Let's move this to the `prints_help` test in `tests/cli.rs`

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,84 @@
+use assert_cmd::prelude::*;
```
Morriar: Can we rename this test file to make it about visualization?

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,84 @@
...
+        .arg(temp_dir.path())
+        .output()
+        .expect("Failed to execute command");
+
+    assert!(output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("Indexed 2 files"));
+    assert!(stdout.contains("Found 2 names"));
+    assert!(stdout.contains("Found 2 definitions"));
+}
```
Morriar: Let's move this to `tests/cli.rs`

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,84 @@
...
+        .unwrap()
+        .arg(temp_dir.path())
+        .arg("--visualize")
+        .output()
+        .expect("Failed to execute command");
+
+    assert!(output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+
+    // Parse the actual output to extract the dynamic parts
```
Morriar: Maybe we can simplify this by replacing the dynamic strings.

Something like this:

```rb
actual = String::from_utf8_lossy(&output.stdout);

actual
  .gsub!(/def_[0-9]+/, "def_<ID>")
  .gsub!(/file:\/\/.*/simple\.rb/, "file://<PATH>/simple.rb")

expected = "digraph {{\n    rankdir=TB;\n\n    \"Name:SimpleClass\" [label=\"SimpleClass\",shape=hexagon];\n\n    \"def_<ID>\" [label=\"Class(SimpleClass)\",shape=ellipse];\n\n    \"file/<PATH>/simple.rb\" [label=\"simple.rb\",shape=box];\n\n    \"Name:SimpleClass\" -> \"def_<ID>\" [dir=both];\n    \"def_<ID>\" -> \"file/<PATH>/simple.rb\";\n}}\n\n"

assert_equal(expected, actual)
```

rust/index/tests/cli_integration_test.rs
```diff
@@ -0,0 +1,84 @@
...
+        .find(|line| line.starts_with("    \"file://") && line.contains("[label=\"simple.rb\""))
+        .expect("Should find URI line");
+    let uri_start = uri_line.find('"').unwrap() + 1;
+    let uri_end = uri_line.find("\" [label=").unwrap();
+    let uri = &uri_line[uri_start..uri_end];
+
+    // Build expected output with the extracted values
+    let expected_output = format!(
+        "digraph {{\n    rankdir=TB;\n\n    \"Name:SimpleClass\" [label=\"SimpleClass\",shape=hexagon];\n\n    \"def_{def_id}\" [label=\"Class(SimpleClass)\",shape=ellipse];\n\n    \"{uri}\" [label=\"simple.rb\",shape=box];\n\n    \"Name:SimpleClass\" -> \"def_{def_id}\" [dir=both];\n    \"def_{def_id}\" -> \"{uri}\";\n}}\n\n"
+    );
```
Morriar: Could we extract and reuse https://github.com/Shopify/index/blob/main/rust/index/src/test_utils.rs#L22-L65?

So the output would look like this:

```rb
let expected_output = normalize_indentation(
    "
        digraph {
            rankdir=TB;

            \"Name:SimpleClass\" [label=\"SimpleClass\",shape=hexagon];
            \"def_{def_id}\" [label=\"Class(SimpleClass)\",shape=ellipse];
            \"{uri}\" [label=\"simple.rb\",shape=box];
            \"Name:SimpleClass\" -> \"def_{def_id}\" [dir=both];
            \"def_{def_id}\" -> \"{uri}\";
        }
    "
    );
```

Morriar: btw, where it the second `{` coming  from? I don't see where we generate it in `dot.rs`: `digraph {{`?

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+        // Test with a simulated name containing quotes
+        // Note: Ruby parser won't actually accept quotes in class names,
+        // so we'll test the escaping function directly and verify the DOT structure
+        graph_test.index_uri(
+            "file:///test.rb",
+            r"
+class TestClass
+end
+",
+        );
```
Morriar: You should be able to indent the text cleanly now that you use `GraphTest`

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let dot_output = generate(&graph_test.graph);
+
+        // Verify the DOT is still valid
+        assert!(dot_output.starts_with("digraph {"));
+        assert!(dot_output.ends_with("}\n"));
+
+        // Test escaping directly for names with quotes
+        assert_eq!(escape_dot_string("Name\"WithQuote"), "Name\\\"WithQuote");
```
Morriar: Since you have access to the graph, you can use the actual URI and IDs. Let's compare the full string like you do in the cli_test

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        let uri_edges: Vec<&&str> = lines.iter().filter(|l| l.contains("-> \"file:///test.rb\"")).collect();
+
+        assert_eq!(uri_edges.len(), 2, "Expected 2 edges to URI");
+
+        for edge in &uri_edges {
+            assert!(edge.trim().starts_with("\"def_"));
+        }
+    }
```
Morriar: Same here, let's compare complete output instead of asserting on small pieces

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+        ];
+
+        for expected in &expected_elements {
+            assert!(
+                lines.iter().any(|line| line.contains(expected)),
+                "Expected to find '{expected}' in output:\n{dot_output}"
+            );
+        }
+    }
+
```
Morriar: Same here, let's compare outputs

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+        let mut graph_test = GraphTest::new();
+        graph_test.index_uri(
+            "file:///test.rb",
+            r"
+class TestClass
+end
+
+module TestModule
+end
+",
```
Morriar: You should be able to indent this properly with `GraphTest`

rust/index/src/visualization/dot.rs
```diff
@@ -0,0 +1,205 @@
...
+            output,
+            "    \"{escaped_uri}\" [label=\"{escaped_label}\",shape={URI_NODE_SHAPE}];"
+        );
+    }
+    output.push('\n');
+
+    // Write edges between Names and Definitions (bidirectional)
+    for (name_id, definition_ids) in graph.name_to_definitions() {
+        if let Some(name) = graph.names().get(name_id) {
+            let name_node = format!("Name:{name}");
```
Morriar:
```suggestion
            let name_node = format!("Name:{}", escape_dot_string(name));
```

rust/index/tests/cli_integration_test.rs
jenny-codes: I decided to remove the file and have all the tests in tests/cli

rust/index/tests/cli_integration_test.rs
jenny-codes: It is an escape sequence in `format!` function that produces a single `{` in the end

PR #99, opened 2025-08-13
Alternative proposal for integrity checks

Morriar:
Alternative to #73.

### DIFF
--- a/rust/index/src/main.rs
+++ b/rust/index/src/main.rs
@@ -15,6 +15,12 @@ use index::{
 struct Args {
     #[arg(value_name = "DIR", default_value = ".")]
     dir: String,
+
+    #[arg(
+        long = "check-integrity",
+        help = "Run integrity checks on the index after processing"
+    )]
+    check_integrity: bool,
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
@@ -29,5 +35,22 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let graph_arc = Arc::new(Mutex::new(graph));
     indexing::index_in_parallel(&graph_arc, &documents)?;
+
+    // Run integrity checks if requested
+    if args.check_integrity {
+        let graph_lock = graph_arc.lock().unwrap();
+        let errors = Graph::integrity_checker().apply(&graph_lock);
+
+        if errors.is_empty() {
+            println!("✓ Index integrity check passed");
+        } else {
+            eprintln!("✗ Index integrity check failed with {} errors:", errors.len());
+            for error in &errors {
+                eprintln!("  - {error}");
+            }
+            std::process::exit(1);
+        }
+    }
+
     Ok(())
 }


rust/index/src/model/index.rs
```diff
@@ -9,19 +9,19 @@ use crate::model::ids::{DefinitionId, NameId, UriId};
 pub struct Index {
     // *** Graph nodes: the following represent possible nodes in our graph ***
     // Map of fully qualified names. These represent global declarations, like `Foo`, `Foo#bar` or `Foo.baz`
-    names: HashMap<NameId, String>,
+    pub names: HashMap<NameId, String>,
```
vinistock: Let's make these getters for immutable references, so that the data can't be modified externally.

rust/index/src/model/integrity.rs
```diff
@@ -0,0 +1,229 @@
...
+/// An integrity rule that can be applied to an Index to validate its consistency
+pub struct IntegrityRule {
+    /// Human-readable description of what this rule checks
+    pub description: String,
+    /// The validation function that takes an Index and appends errors to the error list
+    pub validator: ValidatorFn,
+}
+
+impl IntegrityRule {
+    pub fn new<F>(description: impl Into<String>, validator: F) -> Self
```
vinistock: Do we need flexibility for the descriptions? How about we only accept string literals for them?

rust/index/src/model/integrity.rs
```diff
@@ -0,0 +1,229 @@
...
+            rules: Self::default_rules(),
+        }
+    }
+
+    pub fn add_rule(&mut self, rule: IntegrityRule) {
+        self.rules.push(rule);
+    }
+
+    #[must_use]
+    pub fn default_rules() -> Vec<IntegrityRule> {
```
vinistock: What do you think about not having default rules and just adding these from the Index?

Or alternatively, in what scenario would we use `add_rule` instead of just adding another entry to the list of default rules?

rust/index/src/model/integrity.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Panics if any integrity rule fails
+    pub fn assert_integrity(&self, index: &Index) {
+        let errors = self.apply(index);
+        assert!(
+            errors.is_empty(),
+            "Integrity check failed with {} errors:\n{}",
+            errors.len(),
+            errors.join("\n")
+        );
+    }
```
vinistock: IMO, we can merge all of these and just have `assert_integrity`.

rust/index/src/model/index.rs
jenny-codes: Ah I made these changes at https://app.graphite.dev/github/pr/Shopify/index/97/Add-command-to-generate-graph-representation?utm_source=gt-slack-notif#file-rust/index/src/model/graph.rs as well

rust/index/src/model/integrity.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn check(&self, index: &Index) -> bool {
+        self.apply(index).is_empty()
+    }
+
+    /// Applies all rules and panics if any fail (for debugging/testing)
+    ///
+    /// # Panics
+    ///
+    /// Panics if any integrity rule fails
+    pub fn assert_integrity(&self, index: &Index) {
```
jenny-codes: What are the debugging situations where we'd want to reach for this function instead of just using the `apply` method and check the returned errors? Can we keep only one of two functions given that they seem to serve a very similar purpose?

rust/index/src/model/integrity.rs
```diff
@@ -0,0 +1,229 @@
...
+            rule.apply(index, &mut errors);
+        }
+        errors
+    }
+
+    /// Applies all rules to the given index and returns true if no errors found
+    #[must_use]
+    pub fn check(&self, index: &Index) -> bool {
+        self.apply(index).is_empty()
+    }
```
jenny-codes: This function is only used in tests–do we need it?

rust/index/src/model/integrity.rs
jenny-codes: We should also add rules on the uri nodes and `uris_to_definitions` edges?

rust/index/src/model/integrity.rs
Morriar: Now taking a `&str` 👍

rust/index/src/model/integrity.rs
Morriar: Removed it, all tests are now using `assert_integrity` 👍

rust/index/src/model/integrity.rs
Morriar: It's a nice shortcut so you can write tests with just `checker.assert_integrity(&index)` instead of having to check if the array is empty all the time.

But this is a good point, I marked it as `#[cfg(test)]` 👍

rust/index/src/model/integrity.rs
Morriar: I kept `apply` which is the official non-test API.

`assert_integrity` is now marked `#[cfg(test)]`.

rust/index/src/model/integrity.rs
Morriar: Yeah good idea, moved to the index 👍

Morriar: Added a bunch of missing rules

PR #102, opened 2025-08-14
Hook up Ruby API for the global graph

### LINKED ISSUE #77: Hook the current Index implementation to a Ruby API
Similar to what we had in https://github.com/Shopify/index/pull/36, let's hook the current Rust Index implementation to a simple Ruby object.

For now, this is the extent of the API:

```ruby
# We need to be able to instantiate a new graph. This needs to use a custom allocator (see #36)
index = Index::Graph.new

# We need to be able to index a list of documents
documents = Dir.glob("**/*.rb").map do |path|
  {
    uri: "file://#{path}", # not actually correct! Need to use real URI parsing to account for Windows
    source: nil, # Optional source for files not committed to disk. `nil` means read from disk, a string means use it
  }
end

index.index_all(documents)

# We need to be able to verify the index graph's integrity
index.verify_integrity
```

While you're at it, please clean up the old `Repository` code.

vinistock:
Closes #77

This PR hooks up the current `Graph` implementation to a Ruby API.

I would like to have this foundation pushed forward since decisions that we make in the Rust design can actually impact the Ruby API too. I find useful to be able to reason about all parts.

Additionally, I want to push forward the implementation to index a workspace including all of its dependencies, which requires integrations with Bundler. That way, when we run integrity verification checks or just manual testing, we can index an entire codebase including its dependencies.


rust/index/src/indexing.rs
```diff
@@ -26,13 +26,24 @@ impl Document {
...
-            uri: Url::parse(uri)?,
+            uri: Url::parse(uri).map_err(|e| IndexingError::InvalidUri(format!("Failed to parse URI '{uri}': {e}")))?,
             source,
         })
     }
 
+    /// # Errors
+    ///
+    /// Errors if the file path cannot be turned into a URI
+    pub fn from_file_path(path: &str) -> Result<Self, IndexingError> {
```
Morriar, line 39: Should we reuse this for https://github.com/Shopify/index/blob/main/rust/index/src/main.rs#L72?

ext/index/graph.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Graph#index_all: (Array[String] file_paths) -> String?
+// Returns the error messages concatenated as a single string if anything failed during indexing or `nil`
+static VALUE rb_graph_index_all(VALUE self, VALUE file_paths) {
+    // Convert the given file paths into a char** array, so that we can pass to Rust. Also, mark the object so that it
+    // cannot be moved or garbage collected while Rust still points to it
+    rb_gc_register_address(&file_paths);
+    long length = RARRAY_LEN(file_paths);
+    char **converted_file_paths = str_array_to_char(file_paths, length);
```
Morriar, line 52: Should we validate the array contains strings? What if I pass `[1.42]` for example? Does it crash?

ext/index/index.c
```diff
@@ -1,88 +1,12 @@
...
-
-    cRepository = rb_define_class_under(mIndex, "Repository", rb_cObject);
-    rb_define_singleton_method(cRepository, "new", rb_index_new, 0);
-    rb_define_method(cRepository, "add_entry", rb_repository_add_entry, 2);
-    rb_define_method(cRepository, "get_entry", rb_repository_get_entry, 1);
-
-    cEntry = rb_define_class_under(mIndex, "Entry", rb_cObject);
-    rb_define_method(cEntry, "name", rb_entry_name, 0);
-    rb_define_method(cEntry, "value", rb_entry_value, 0);
+    rb_ext_ractor_safe(true);
```
Morriar, line 7: Are we really?

ext/index/graph.c
```diff
@@ -0,0 +1,80 @@
+#include "ruby.h"
+#include "ruby/internal/intern/string.h"
```
Morriar: This looks.. _internal_? Do we need this include?

ext/index/index.c
```diff
@@ -1,88 +1,12 @@
 #include "ruby.h"
+#include "ruby/internal/intern/load.h"
```
Morriar: Same question about including internal stuff?

test/graph_test.rb
```diff
@@ -0,0 +1,17 @@
...
+  def test_indexing_a_list_of_file_paths
+    graph = Index::Graph.new
+    assert_nil(graph.index_all([__FILE__]))
+  end
+
+  def test_indexing_returns_errors
+    graph = Index::Graph.new
+    message = graph.index_all(["/non_existing_file.rb"])
+
+    assert_equal("File read error: Failed to read /non_existing_file.rb: No such file or directory (os error 2)", message)
```
Morriar, line 15: Is this error message stable across platforms?

ext/index/index.c
Morriar: Do we actually need if?

ext/index/graph.c
```diff
@@ -0,0 +1,80 @@
...
+
+    return converted_array;
+}
+
+// Graph#index_all: (Array[String] file_paths) -> String?
+// Returns the error messages concatenated as a single string if anything failed during indexing or `nil`
+static VALUE rb_graph_index_all(VALUE self, VALUE file_paths) {
+    // Convert the given file paths into a char** array, so that we can pass to Rust. Also, mark the object so that it
+    // cannot be moved or garbage collected while Rust still points to it
+    rb_gc_register_address(&file_paths);
```
Morriar: Do we need to mark this value? It looks like `file_paths` is on the C stacks and we deep copy it right away with `str_array_to_char`?

rust/index/src/indexing.rs
vinistock: That code goes away in the next PR in the stack.

ext/index/index.c
vinistock: Yup. We're not holding references to any Ruby objects.

ext/index/graph.c
vinistock: I think the clangd LSP made a bad include completion and I didn't notice. We definitely don't need it, although I'm surprised it didn't crash compilation. AFAIK, you shouldn't be able to include internal headers.

ext/index/index.c
vinistock: Same thing as the other one.

test/graph_test.rb
vinistock: I think so given that tests are passing. If we find it's not when we bring back Windows we can just update.

ext/index/graph.c
```diff
@@ -0,0 +1,80 @@
...
+    long length = RARRAY_LEN(file_paths);
+    char **converted_file_paths = str_array_to_char(file_paths, length);
+
+    // Get the underying graph pointer and then invoke the Rust index all implementation
+    void* graph;
+    TypedData_Get_Struct(self, void*, &graph_type, graph);
+    const char* error_messages = idx_index_all_c(graph, converted_file_paths, length);
```
Morriar: Looks like length is a `long` here and we pass it to `idx_index_all_c` that expects a `usize`. Could we use `size_t` instead of `long`?

ext/index/graph.c
vinistock: I guess not. Removing it didn't cause any trouble.

ext/index/graph.c
vinistock: Yes, I added the type guards.

ext/index/graph.c
vinistock: Switched to using `size_t`.

ext/index/graph.c
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return converted_array;
+}
+
+// Graph#index_all: (Array[String] file_paths) -> String?
+// Returns the error messages concatenated as a single string if anything failed during indexing or `nil`
+static VALUE rb_graph_index_all(VALUE self, VALUE file_paths) {
+    // Convert the given file paths into a char** array, so that we can pass to Rust
+    size_t length = RARRAY_LEN(file_paths);
```
Morriar, line 51: Shouldn't `Check_Type(array, T_ARRAY);` happen before this?

rust/index-sys/src/conversions.rs
```diff
@@ -0,0 +1,30 @@
...
+///
+/// # Safety
+///
+/// This function is unsafe because it attempts to instantiate a Vec<String> from a raw char** pointer
+///
+/// # Errors
+///
+/// This function errors if any of the strings inside the array contain invalid UTF-8 data
+pub unsafe fn convert_double_pointer_to_vec(data: *mut *mut c_char, len: size_t) -> Result<Vec<String>, Utf8Error> {
+    unsafe {
```
Morriar: Do we need the double `unsafe` here? Isn't the one in the function declaration enough?

rust/index-sys/src/conversions.rs
```diff
@@ -0,0 +1,30 @@
...
+            .iter()
+            .map(|arg| CStr::from_ptr(*arg).to_str().map(ToString::to_string))
+            .collect()
+    }
+}
+
+/// Frees a `CString` allocated on the Rust side
+#[unsafe(no_mangle)]
+pub extern "C" fn free_c_string(ptr: *const c_char) {
+    unsafe {
```
Morriar, line 39: Similarly, could we move the `unsafe` to the declaration?

rust/index-sys/src/index_api.rs
```diff
@@ -0,0 +1,79 @@
...
+#[unsafe(no_mangle)]
+pub extern "C" fn idx_graph_new() -> GraphPointer {
+    let repo_arc = Arc::new(Mutex::new(Graph::new()));
+    Arc::into_raw(repo_arc) as GraphPointer
+}
+
+/// Frees a Graph through its pointer
+#[unsafe(no_mangle)]
+pub extern "C" fn idx_graph_free(pointer: GraphPointer) {
+    unsafe {
```
Morriar, line 23: Same question about `unsafe` here and the other functions in this file

ext/index/graph.c
```diff
@@ -0,0 +1,81 @@
+#include "ruby.h"
+#include "rustbindings.h"
```
Morriar: I see we're using `malloc`, `free`, `strlen` and friends, should we include `stdlib.h` and `string.h`?

ext/index/graph.h
```diff
@@ -0,0 +1,6 @@
+#ifndef GRAPH_H
+#define GRAPH_H
```
Morriar, line 2: Shouldn't we move the `ruby.h` include here since we declare using `VALUE`?

ext/index/graph.c
```diff
@@ -0,0 +1,81 @@
...
+// Custom allocator for the Graph class. Calls into Rust to create a new `Arc<Mutex<Graph>>` that gets stored internally
+// as a void pointer
+static VALUE rb_graph_alloc(VALUE klass) {
+    void* graph = idx_graph_new();
+    return TypedData_Wrap_Struct(klass, &graph_type, graph);
+}
+
+// Convert a Ruby array of strings into a double char pointer so that we can pass that to Rust. This copies the data
+// so it must be freed
+static char** str_array_to_char(VALUE array, long length) {
```
Morriar: Shouldn't this be a `size_t`? We're passing a `size_t` lower.

rust/index-sys/src/index_api.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Will panic if the given array of C string file paths cannot be converted to a Vec<String>
+///
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers coming from C. The caller has to ensure that the Ruby
+/// VM will not free the pointers related to the string array while they are in use by Rust
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn idx_index_all_c(
+    pointer: GraphPointer,
+    file_paths: *mut *mut c_char,
```
Morriar: Do we actually mutate these guys?

rust/index-sys/src/conversions.rs
vinistock: Unfortunately, it isn't. The `unsafe` block inside allows us to call other functions marked as unsafe, like `slice::from_raw_parts`.

However, in addition to calling unsafe functions, we also dereference the `data` raw pointer and then Rust enforces that that we need to mark the function as unsafe for consumers.

If we weren't trying to dereference the pointer, then we could get away with just using the unsafe block and not marking the function as unsafe (making a safe abstraction on top of an unsafe call), but this one we need both.

rust/index-sys/src/conversions.rs
vinistock: Not in this case because we're wrapping the unsafe call in a safe abstraction. We don't want consumers of `free_c_string` to need to use `unsafe` to call this function and it's not required because we're not dereferencing the raw pointer.

rust/index-sys/src/index_api.rs
vinistock: Same case as the others.

ext/index/graph.c
vinistock: Yes, good point.

ext/index/graph.c
vinistock: Added those.

ext/index/graph.h
vinistock: I totally forgot to `include "graph.h"`. Moved all includes there.

rust/index-sys/src/index_api.rs
vinistock: We don't. It has to be mutable on the C side in order to copy and free memory, but not for the Rust API.

I changed all of them to `*const` and started casting before passing the pointers to the Rust side.

PR #103, opened 2025-08-14
Add test DSL for index

Morriar:
Proposes an approach to write index tests that work around the lack of heredocs in Rust:

```rust
    #[test]
    fn test_index_uri_with_multiple_lines() {
        let mut context = IndexTest::new();

        context.index_uri("file://method.rb", {
            "
            class Foo
              class Bar; end
            end
            "
        });

        let foo_defs = context.index.get("Foo").unwrap();
        assert_eq!(foo_defs.len(), 1);
        assert_eq!(foo_defs[0].start_offset(), 0);
        assert_eq!(foo_defs[0].end_offset(), 30);

        let bar_defs = context.index.get("Foo::Bar").unwrap();
        assert_eq!(bar_defs.len(), 1);
        assert_eq!(bar_defs[0].start_offset(), 12);
        assert_eq!(bar_defs[0].end_offset(), 26);
    }
```

### DIFF
--- a/rust/index/src/model/graph.rs
+++ b/rust/index/src/model/graph.rs
@@ -137,58 +137,64 @@ impl Graph {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::indexing::ruby_indexer::RubyIndexer;
-
-    // Index the given source and URI and return the resulting `Graph`
-    fn index_source(uri: &str, source: &str) -> Graph {
-        let mut indexer = RubyIndexer::new(uri.to_string());
-        indexer.index(source);
-        indexer.into_parts().0
-    }
+    use crate::test_utils::GraphTest;
 
     #[test]
     fn deleting_a_uri() {
-        let mut global = index_source("file:///foo.rb", "module Foo; end");
-        global.delete_uri("file:///foo.rb");
+        let mut context = GraphTest::new();
 
-        assert!(global.definitions.is_empty());
-        assert!(global.names.is_empty());
-        assert!(global.definition_to_name.is_empty());
-        assert!(global.name_to_definitions.is_empty());
-        assert!(global.uris_to_definitions.is_empty());
-        assert!(global.uri_pool.is_empty());
+        context.index_uri("file:///foo.rb", "module Foo; end");
+        context.delete_uri("file:///foo.rb");
+
+        assert!(context.graph.definitions.is_empty());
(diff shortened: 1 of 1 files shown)


rust/index/src/test_utils.rs
```diff
@@ -0,0 +1,133 @@
+use crate::indexing::ruby_indexer::RubyIndexer;
+use crate::model::index::Index;
+
+pub struct IndexTest {
```
vinistock: So we don't need the manual definition below.

```suggestion
#[derive(Default)]
pub struct IndexTest {
```

rust/index/src/test_utils.rs
```diff
@@ -0,0 +1,133 @@
...
+        assert_eq!(bar_defs.len(), 1);
+        assert_eq!(bar_defs[0].start_offset(), 12);
+        assert_eq!(bar_defs[0].end_offset(), 26);
+    }
+
+    #[test]
+    fn test_index_uri_with_new_lines() {
+        let mut context = IndexTest::new();
+
+        context.index_uri("file://method.rb", "\n\nclass Foo; end");
```
vinistock, line 121:
```suggestion
        context.index_uri("file:///method.rb", "\n\nclass Foo; end");
```

rust/index/src/test_utils.rs
```diff
@@ -0,0 +1,133 @@
...
+        assert_eq!(foo_defs.len(), 1);
+        assert_eq!(foo_defs[0].start_offset(), 0);
+        assert_eq!(foo_defs[0].end_offset(), 14);
+    }
+
+    #[test]
+    fn test_index_uri_with_multiple_lines() {
+        let mut context = IndexTest::new();
+
+        context.index_uri("file://method.rb", {
```
vinistock, line 98:
```suggestion
        context.index_uri("file:///method.rb", {
```

rust/index/src/test_utils.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_index_uri_with_single_line() {
+        let mut context = IndexTest::new();
+
+        context.index_uri("file://method.rb", "class Foo; end");
```
vinistock, line 86:
```suggestion
        context.index_uri("file:///method.rb", "class Foo; end");
```

Issue #75, opened 2025-08-12
Create a proper CLI for the indexer


vinistock:
Right now, we have our test CLI written in the Rust crate in `main.rs`. I think it makes more sense to have our actual CLI be written in Ruby - since we need the Bundler integrations.

For example, if we wanted to run the integrity verification from #73 in the current workspace - something like this:

```shell
index --verify-integrity
```

We would like to have that executable hook to Bundler to also index dependencies. That can only be achieved with a Ruby executable and not a pure Rust executable.

I propose we get rid of our Rust executable and create one for the Ruby gem, integrating with the Rust crate.

Depends on #77.

PR #104, opened 2025-08-14
Add ability to index an entire workspace

vinistock:
This PR adds the ability to index an entire workspace, including its dependencies, in a single API.

For now, I removed the ability to exclude directories. I want to create a proper configuration abstraction to propagate that through indexing.

Also, this gets us one step closer to #75, where we can hook the our CLI at the Ruby level, enabling things like:

```
index --verify-integrity
```

To index an entire workspace and verify integrity.

**Note**: one great aspect of using `Bundler.locked_gems&.specs` is that, even if `Bundler.setup` failed due to a gem not being installed, it still returns information for all of the gems that are installed. This allows us to partially index dependencies even if bundle install didn't fully succeed.

### DIFF
--- a/lib/index/graph.rb
+++ b/lib/index/graph.rb
@@ -5,11 +5,13 @@ module Index
   #
   # Note: this class is partially defined in C to integrate with the Rust backend
   class Graph
+    #: (?workspace_path: String) -> void
     def initialize(workspace_path: Dir.pwd)
       @workspace_path = workspace_path
     end
 
     # Index all files and dependencies of the workspace that exists in `@workspace_path`
+    #: -> String?
     def index_workspace
       paths = workspace_dependency_paths
       paths.unshift(@workspace_path)
@@ -31,6 +33,7 @@ module Index
       end
 
       paths.flatten!
+      paths.uniq!
       paths
     end
   end


rust/index/src/main.rs
```diff
@@ -18,61 +15,17 @@ use index::{
...
-            })
-            .collect::<Vec<_>>()
-    };
+    if !errors.is_empty() {
+        return Err(Box::new(MultipleErrors(errors)));
+    }
 
-    let index = Arc::new(Mutex::new(Graph::new()));
-    index_in_parallel(&index, &documents)?;
+    indexing::index_in_parallel(&graph, &documents)?;
```
Morriar: We should check for errors here as well

lib/index.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
+require "bundler"
```
Morriar, line 3: I wonder if we could avoid the hard dependency with something like this:

```rb
def workspace_dependency_paths
  begin
    require "bundler"
  rescue LoadError
    return []
  end
  # ...
end
```

🤔

lib/index/graph.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    private
+
+    # Gathers the paths we have to index for all workspace dependencies
+    def workspace_dependency_paths
+      specs = Bundler.locked_gems&.specs
+      return [] unless specs
+
+      paths = specs.filter_map do |lazy_spec|
+        spec = Gem::Specification.find_by_name(lazy_spec.name)
```
Morriar, line 29: Can we also pass the version?

```suggestion
        spec = Gem::Specification.find_by_name(lazy_spec.name, lazy_spec.version)
```

lib/index/graph.rb
```diff
@@ -0,0 +1,37 @@
...
+        spec.require_paths.map { |path| File.join(spec.full_gem_path, path) }
+      rescue Gem::MissingSpecError
+        nil
+      end
+
+      paths.flatten!
```
Morriar, line 35:
```suggestion
      paths.flatten!.uniq!
```

lib/index/graph.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Index
+  # The global graph representing all declarations and their relationships for the workspace
+  #
+  # Note: this class is partially defined in C to integrate with the Rust backend
+  class Graph
+    def initialize(workspace_path: Dir.pwd)
```
Morriar, line 9:
```suggestion
    #: (?workspace_path: String) -> void
    def initialize(workspace_path: Dir.pwd)
```

lib/index/graph.rb
```diff
@@ -0,0 +1,37 @@
...
+      @workspace_path = workspace_path
+    end
+
+    # Index all files and dependencies of the workspace that exists in `@workspace_path`
+    def index_workspace
```
Morriar, line 15: Isn't it returning a nilable error? Something like:

```suggestion
    #: -> String?
    def index_workspace
```

Morriar: We should make `index_all` raise a proper error as a follow up

rust/index/src/main.rs
```diff
@@ -18,61 +15,17 @@ use index::{
...
-        "fixtures",
-    ])]
-    exclude: Vec<String>,
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::parse();
-    let excluded_dirs_set: HashSet<&str> = args.exclude.iter().map(String::as_str).collect();
+    let graph = Arc::new(Mutex::new(Graph::new()));
+    let (documents, errors) = indexing::collect_documents_in_parallel(vec![args.dir]);
```
Morriar, line 24: Shouldn't we still exclude some directories?

lib/index/graph.rb
vinistock: No, that's not necessary since the project's gems are already scoped by Bundler setup.

lib/index/graph.rb
vinistock: What could cause the paths to not be unique? I don't think that can happen, can it?

lib/index/graph.rb
vinistock: Currently, it returns all indexing errors so that they can be reported, but we don't halt anything because we failed to index a document.

We can certainly raise the error, but we'd just rescue it immediately in the LSP.

rust/index/src/main.rs
vinistock: > For now, I removed the ability to exclude directories. I want to create a proper configuration abstraction to propagate that through indexing.

lib/index.rb
vinistock: Bundler is always present on every Ruby installation no? Would the require ever fail?

If we want to support alternative package managers, it would require some sort of configuration because I think requiring Bundler will always succeed.

Morriar: If it's always there then it's okay 👍

lib/index/graph.rb
Morriar: Another option would be to use the same approach than Prism with `index.success?` and return a partial index in case of error?

vinistock: Great idea. We can return an `IndexingResult` object with diagnostics. It could even tell the user which files took the longest to index.

lib/index/graph.rb
Morriar: Gems with multiple platforms will appear twice for example. You can try with `sorbet-static`

rust/index/src/main.rs
vinistock: Sorry, what do you mean? We're doing the early return in the case of errors with the `?` thing at the end

Morriar: Sorry I missed the `?`

PR #106, opened 2025-08-15
Add top gems check

### LINKED ISSUE #78: Add rake task and CI step to verify indexing against most popular gems
Migrate what we have in [the Ruby LSP](https://github.com/Shopify/ruby-lsp/blob/main/rakelib/index.rake) to run indexing on the top 100 gems.

We need to ensure that we:

1. Don't crash
2. End up with proper integrity (#73)

One modification we can do, is I would prefer if we could download the top 100 gems from the RubyGems API, rather than having a hard coded list.

Depends on #75

vinistock:
Closes #78

This PR adds a Rake task similar to the one we have in the Ruby LSP, where we download the top 100 most popular gems and verify that indexing succeeds for them.

When we have integrity verifications, we can also start verifying integrity in this check. For now, it just verifies we can index them successfully.

**Notes**

- I wanted to figure out the 100 most popular gems on the fly, but I didn't find an API to pull those from rubygems, so I kept the same mechanism of using the list which was initially compiled for Prism
- I really wanted to use Ractors instead of threads to speed up this check and have something constantly verifying that we can run inside a Ractor, but the built-in gems we need to use (net/http, rubygems and yaml) crash on Ractors


rakelib/top_gems.yml
```diff
@@ -0,0 +1,101 @@
+---
+- actioncable-7.0.4.3
```
Morriar: Does that mean we need ton update these versions manually? Can't we just pull the latest version all the time?

Morriar: Do we have a script to generate this file we should commit along with it?

rakelib/index_top_gems.rake
```diff
@@ -0,0 +1,80 @@
...
+          end
+        end
+      end
+
+      workers.each(&:join)
+    end
+  end
+
+  gems = YAML.safe_load_file(File.expand_path("top_gems.yml", __dir__))
+  gems_dir = File.expand_path(File.join("..", "tmp", "top_gems"), __dir__)
```
Morriar, line 67: Since we run the script from the root directory, does that mean we create `../tmp` outside of it?

rakelib/top_gems.yml
vinistock: Started pulling the latest version from the API.

> Do we have a script to generate this file we should commit along with it?

I don't have a script for it, Kevin generated the list for Prism and we just copied it along. There's no API to return the most popular gems AFAIK.

rakelib/index_top_gems.rake
vinistock: No, it's `index/tmp` because we're expanding the path based on `__dir__` (which is `rakelib`).

PR #110, opened 2025-08-15
Revert use of archived actions-rs/toolchain action

vinistock:
The `actions-rs/toolchain` action was [archived in 2023](https://github.com/actions-rs/toolchain). My understanding is that we shouldn't need any custom actions as GH runners already have `rustup` and `rustc` installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,11 @@ jobs:
         with:
           submodules: "recursive"
 
-      - name: Set up Rust
-        uses: actions-rs/toolchain@88dc2356392166efad76775c878094f4e83ff746 #v1.0.6
-        with:
-          toolchain: stable
-          components: rustfmt, clippy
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -10,12 +10,11 @@ jobs:
         with:
           submodules: "recursive"
 
-      - name: Set up Rust
-        uses: actions-rs/toolchain@88dc2356392166efad76775c878094f4e83ff746 #v1.0.6
-        with:
-          toolchain: stable
-          components: rustfmt, clippy
-          override: true
+      - name: Install Rust tools
```
Morriar, line 13: Why do we need it in this step but not the other one?

vinistock: Because the other one doesn't require additional components like clippy and rustfmt. It only runs tests.

PR #111, opened 2025-08-15
Add DB model with connection init

ChallaHalla:
This PR adds a DB model to the codebase that can be used for initializing DB connections. 

The important areas for review are:

- Db connection initialization and schema loading
  -  Assuming that going forward we can initialize a new DB connection for each query we need to run
  - Added schema versioning so that it's clear when the schema needs to be loaded
- Db type (memory vs disk) and path config
- `db` as a new attribute on `Graph`
- Database schema within `schema.sql`

Changes related to writing Index data to the DB will be added in a follow up PR.

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -3,15 +3,24 @@ use std::error::Error;
 
 const SCHEMA_VERSION: i32 = 1;
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
+enum ConnectionType {
+    Path(String),
+    #[allow(dead_code)]
+    Memory,
+}
+
+#[derive(Debug)]
 pub struct Db {
-    path: String,
+    connection_type: ConnectionType,
 }
 
 impl Db {
     #[must_use]
     pub fn new(path: &str) -> Self {
-        Self { path: path.to_string() }
+        Self {
+            connection_type: ConnectionType::Path(path.to_string()),
+        }
     }
 
     /// Initializes a fresh database with schema and configuration
@@ -38,36 +47,41 @@ impl Db {
 
     /// Creates a fresh file database by removing existing file and recreating
     fn recreate_database(&self) -> Result<Connection, Box<dyn Error>> {
-        if std::path::Path::new(&self.path).exists() {
-            std::fs::remove_file(&self.path)?;
+        match &self.connection_type {
+            ConnectionType::Path(path) => {
+                if std::path::Path::new(path).exists() {
+                    std::fs::remove_file(path)?;
+                }
+                let mut conn = Connection::open(path)?;
+                Self::initialize_database(&mut conn)?;
+                Ok(conn)
+            }
+            ConnectionType::Memory => Err("Cannot recreate in-memory database".into()),
         }
-        let mut conn = Connection::open(&self.path)?;
-        Self::initialize_database(&mut conn)?;
-        Ok(conn)
     }
 
     /// # Errors
     /// Will return an Error if we fail to establish or set a connection
     pub fn get_connection(&self) -> Result<Connection, Box<dyn Error>> {
-        let mut conn = if self.path == ":memory:" {
-            Connection::open_in_memory()?
-        } else {
-            Connection::open(&self.path)?
+        let conn = match &self.connection_type {
+            ConnectionType::Path(path) => {
+                let conn = Connection::open(path)?;
+                // Check if schema needs to be loaded using user_version pragma.
+                // The default value for user_version is 0.
+                let current_version: i32 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
+                if current_version < SCHEMA_VERSION {
+                    drop(conn);
+                    return self.recreate_database();
+                }
+                conn
+            }
+            ConnectionType::Memory => {
+                let mut conn = Connection::open_in_memory()?;
+                Self::initialize_database(&mut conn)?;
+                conn
+            }
         };
 
-        // Check if schema needs to be loaded using user_version pragma.
-        // The default value for user_version is 0.
-        let current_version: i32 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
-
-        if current_version < SCHEMA_VERSION {
-            if self.path == ":memory:" {
-                Self::initialize_database(&mut conn)?;
-            } else {
-                // For file-based databases, delete and recreate for clean schema
-                drop(conn);
-                return self.recreate_database();
-            }
-        }
         Ok(conn)
     }
 }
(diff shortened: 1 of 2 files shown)


rust/index/src/main.rs
```diff
@@ -48,7 +48,8 @@ fn main() -> Result<(), Box<dyn Error>> {
             .collect::<Vec<_>>()
     };
 
-    let index = Arc::new(Mutex::new(Graph::new()));
+    let db_path = std::env::temp_dir().join("index.db").to_string_lossy().to_string();
```
ChallaHalla, line 51: This path should be within the project being indexed

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,59 @@
...
+            PRAGMA foreign_keys = ON;
+        ",
+        )?;
+
+        // Check if schema needs to be loaded using user_version pragma.
+        // The default value for user_version is 0.
+        let current_version: i32 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
+
+        if current_version < SCHEMA_VERSION {
+            let schema = std::fs::read_to_string(
```
ChallaHalla: We should drop delete db and recreate if user_version doesn't match

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,59 @@
...
+        let conn = if self.path == ":memory:" {
+            Connection::open_in_memory()?
+        } else {
+            Connection::open(&self.path)?
+        };
+
+        // Configure SQLite pragmas for performance and consistency
+        conn.execute_batch(
+            "
+            PRAGMA journal_mode = WAL;
```
ChallaHalla, line 31: Check that these pragmas area on a connection level and not a DB level

ChallaHalla: These are at the DB level. Will move these to be with the schema loading steps

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,89 @@
...
+    #[must_use]
+    pub fn new_memory_db() -> Self {
+        Self {
+            path: ":memory:".to_string(),
+        }
+    }
+
+    /// # Errors
+    /// Will return an Error if we fail to establish or set a connection
+    pub fn get_connection(&self) -> Result<Connection, Box<dyn std::error::Error>> {
```
vinistock: Nit: let's bring `Error` into scope
```suggestion
    pub fn get_connection(&self) -> Result<Connection, Box<dyn Error>> {
```

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl Db {
+    #[must_use]
+    pub fn new(path: &str) -> Self {
+        Self { path: path.to_string() }
+    }
+
+    /// This method should return a new memory DB that can be used for local indexing or tests.
+    #[must_use]
+    pub fn new_memory_db() -> Self {
```
vinistock: Since the local indices don't interact with the Db and we will need to manually invoke something to dump the Graph to cache, let's remove this function for now and add it when the need for it appears.

Especially now that we're not creating the Db file unless `get_connection` is invoked, which does not happen automatically.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,80 @@
...
+            PRAGMA journal_mode = WAL;
+            PRAGMA synchronous = NORMAL;
+            PRAGMA foreign_keys = ON;
+        ",
+        )?;
+
+        let schema = std::fs::read_to_string(
+            std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("src/db")
+                .join("schema.sql"),
```
vinistock, line 40: I think this path might work in development, but in release mode we need a way to bundle the SQL file with the compilation results.

The user won't have the project with the Cargo manifest present. We can fix this in a later PR, but something to keep in mind.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,80 @@
...
+        )?;
+        tx.execute_batch(&schema)?;
+        tx.execute(&format!("PRAGMA user_version = {SCHEMA_VERSION}"), [])?;
+        tx.commit()?;
+        Ok(())
+    }
+
+    /// Creates a fresh file database by removing existing file and recreating
+    fn create_fresh_file_database(&self) -> Result<Connection, Box<dyn std::error::Error>> {
```
vinistock: Nit: this is also deleting the old one, so I'd reflect that in the name
```suggestion
    fn recreate_database(&self) -> Result<Connection, Box<dyn std::error::Error>> {
```

rust/index/src/model/db.rs
ChallaHalla: Sounds good, I'll add it in the follow up PR once it's needed for testing.

rust/index/src/main.rs
ChallaHalla: Will leave this for now and address in a follow up PR once we determine where temp files for the indexer should be placed in a project.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -22,7 +22,7 @@ pub struct RubyIndexer {
 impl RubyIndexer {
     #[must_use]
     pub fn new(uri: String) -> Self {
-        let mut local_index = Graph::new();
+        let mut local_index = Graph::new(String::new());
```
ChallaHalla, line 25: Choosing to simply pass an empty string here since this instance won't be using a DB connection

rust/index/src/model/db.rs
ChallaHalla: Makes sense, I'll see if there's a good way for us to bundle this when we compile.

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        let mut conn = Connection::open(&self.path)?;
+        Self::initialize_database(&mut conn)?;
+        Ok(conn)
+    }
+
+    /// # Errors
+    /// Will return an Error if we fail to establish or set a connection
+    pub fn get_connection(&self) -> Result<Connection, Box<dyn Error>> {
+        let mut conn = if self.path == ":memory:" {
```
vinistock: Let's remove the memory thing for now since we're not using it. If we were to keep it, I'd argue this should leverage an enum instead of a string, so that we make better use of the type system.

```rust
enum ConnectionType {
  Path(String),
  Memory
}
```

rust/index/src/model/db.rs
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+            Connection::open(&self.path)?
+        };
+
+        // Check if schema needs to be loaded using user_version pragma.
+        // The default value for user_version is 0.
+        let current_version: i32 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
+
+        if current_version < SCHEMA_VERSION {
+            if self.path == ":memory:" {
```
vinistock: Same here.

rust/index/src/model/db.rs
ChallaHalla: True the magic string comparison isn't great. Will add an enum we can check here

PR #112, opened 2025-08-15
Add filtering capability to visualization

jenny-codes:
## Summary

This PR adds filtering capability to the graph visualization, allowing users to focus on specific components and their direct connections in large codebases.

## Features

- Filter graph output to show only a specific name and its direct connections
- Shows the name node, its definitions (Class/Module/etc.), and containing documents

### Basic usage (unfiltered)

```bash
# Generate visualization for entire project
cargo run -- /path/to/project --visualize

# Generate visualization and save to file
cargo run -- /path/to/project --visualize > output.dot
```

### Filtered visualization

```bash
# Example: Show only RubyLsp::Server and its connections
cargo run -- /path/to/project --visualize --name "RubyLsp::Server"

# Filter to show only specific module
cargo run -- /path/to/project --visualize --name "RubyLsp"
```

### Convert to SVG

```bash
cargo run -- /path/to/project --visualize --name "RubyLsp" | dot -Tsvg > output.svg
```

### Filtered view with `--name "RubyLsp::Requests"`

```bash
cargo run -- $HOME/src/github.com/Shopify/ruby-lsp --visualize --name "RubyLsp::Requests" | \
  dot -Tsvg > /tmp/rubylsp_requests.svg && open /tmp/rubylsp_requests.svg
```

Shows only:

- The RubyLsp::Requests name node (hexagon)
- All definitions of RubyLsp::Requests (ellipse)
- Documents containing those definitions (box)

rubylsp_requests

### Error handling

```
$ cargo run -- /path/to/project --visualize --name "NonExistent"
Error: No declarations found for name 'NonExistent'
```

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -28,6 +28,9 @@ struct Args {
     #[arg(long = "visualize")]
     visualize: bool,
 
+    #[arg(long = "name", help = "Filter visualization to show only the specified name")]
+    name: Option<String>,
+
     #[arg(long = "use-db", help = "Store the graph in a database", default_value = "false")]
     use_db: bool,
 
@@ -40,6 +43,8 @@ struct Args {
 
 fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::parse();
+
+    validate_args(&args);
 
     if args.stats {
         Timer::set_global_timer(Timer::new());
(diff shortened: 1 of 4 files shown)


rust/index/src/main.rs
```diff
@@ -33,6 +33,10 @@ struct Args {
     /// Generate graph visualization and write to file (use "-" for stdout)
     #[arg(short = 'o', long = "visualization-output", value_name = "FILE")]
     visualization_output: Option<String>,
+
+    /// Filter visualization to show only the specified name and its connections
+    #[arg(short = 'f', long = "filter", value_name = "NAME")]
```
Morriar: Let's give the option a name that shows it's related to the visualization

rust/index/src/visualization/dot.rs
```diff
@@ -2,30 +2,65 @@
...
+/// - Its definition nodes (Class/Module)
+/// - The URI nodes containing those definitions
+/// - The edges connecting them
+///
+/// # Errors
+///
+/// Returns an error if:
+/// - The filter is empty
+/// - The filter name doesn't exist in the index
+pub fn generate(graph: &Graph, filter: Option<&str>) -> Result<String, String> {
```
Morriar: We'll most likely want more filters as we progress, let's make the name more specific like `filter_name`?

rust/index/src/visualization/dot.rs
```diff
@@ -72,13 +96,124 @@ fn write_nodes_and_edges(output: &mut String, graph: &Graph) {
...
-                    let _ = writeln!(output, "    \"def_{def_id}\" -> \"{uri}\";");
+                    write_definition_uri_edge(output, def_id, uri);
                 }
             }
         }
     }
 }
 
+/// Write the filtered graph content showing only direct connections to the input name
+fn write_filtered_graph_content(output: &mut String, graph: &Graph, filter_name: &str) {
```
Morriar: I'm wondering if the filtering logic shouldn't be extracted 🤔 

We could apply the filters on the graph itself to produced a filtered graph and then pass it to the dot renderer. That way we can build increasingly complex filtering but the dot logic stay always the same:


```rust
let graph = graph.lock().unwrap();
let filtered = putitsomewhere::filter(graph)
let dot = dot::generate(graph)
```

PR #113, opened 2025-08-18
Add memory leak CI check

### LINKED ISSUE #107: Add memory leak checks
Memory leaks are painful to hunt down. We need automated CI checks from the beginning to help us catch them early on.

Let's add checks for both the Rust crates and the C native extension:

1. For Rust, we can build the project using nightly and turn on sanitization
2. For the C extension, we can use https://github.com/Shopify/ruby_memcheck

vinistock:
Closes #107 

This PR adds a CI check for verify for memory leaks in both the Rust part of the code and the C layer that connects it to the Ruby VM.

Both levels of checks work similarly. They add runtime checks to the code so that if we allocate any memory and forget to free it, they can report that as a problem and let us know.

Rust offers [many types of sanitizers](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html) to check for problems and for the C layer we're using [ruby_memcheck](https://github.com/Shopify/ruby_memcheck) (which uses valgrind under the hood).

The CI check will:

1. Compile the crate with some form of sanitization
2. Run our Ruby test suite with ruby_memcheck
3. Report any issues

I added 3 Rust sanitizers that may be of interest to us:

- Leak checks for memory leaks
- Memory checks for uninitialized reads
- Thread checks search for data races


.github/workflows/memleak.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          bundler-cache: true
+
+      - run: sudo apt-get install -y valgrind
+
+      - name: Install Rust tools
+        run: |
+          rustup update --no-self-update stable
+          rustup default stable
+          rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```
Morriar, line 32: Why do we need the nightly?

ext/index/graph.c
```diff
@@ -41,11 +39,20 @@ static char** str_array_to_char(VALUE array, size_t length) {
...
+        Check_Type(item, T_STRING);
+    }
+}
+
 // Graph#index_all: (Array[String] file_paths) -> String?
 // Returns the error messages concatenated as a single string if anything failed during indexing or `nil`
 static VALUE rb_graph_index_all(VALUE self, VALUE file_paths) {
-    // Verify that the value passed is indeed an array
-    Check_Type(file_paths, T_ARRAY);
+    check_array_of_strings(file_paths);
```
Morriar, line 55: Why was this change needed in this PR? Did you find a memory leak?

.github/workflows/memleak.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Run tests capturing leaks with `memory` sanitizer
+        run: SANITIZER=memory bundle exec rake test:valgrind
```
Morriar, line 35: Are we trying to run the Rust sanitizers on top of valgrind here? Could this lead to false positives or even flakiness as one will interfere with the other?

ext/index/graph.c
Morriar: Ah it was answered in https://github.com/Shopify/index/pull/113#issuecomment-3214540399 👍 

Ideally this would be done in a separate commit.

.github/workflows/memleak.yml
vinistock: The sanitizers have not been stabilized yet, they only exist in nightly for now.

ext/index/graph.c
vinistock: Yeah, I thought the fix was so small I'd just include it with this one.

.github/workflows/memleak.yml
vinistock: I don't believe so, but I guess we'll find out as we start introducing memory leaks. Both the sanitizers and valgrind use the same approach: they add runtime tracing to memory operations so that they can verify everything checks out in the end.

I think the worst that could happen is getting the same problem reported twice.

If we find that they interfere with each other, we can make them separate steps.

.github/workflows/memleak.yml
Morriar: This means we need to also install the nightly locally if we want to run the sanitizers on our end?

vinistock: Yes, but the `cargo +nightly` syntax does that automatically. So you don't have to manually run `rustup add ...`.

Also, I should point out that valgrind only supports Linux, so we can't run that part of the check on our macs unfortunately.

Morriar: Yeah with both sanitizers and valgrind merged into one check it becomes confusing for the user how to run what locally.

Do you mind adding a comment either here or in the README about this?

.github/workflows/memleak.yml
Morriar: Let's yolo it then 👍

.github/workflows/memleak.yml
vinistock: Added a note to the readme. The readme itself is a bit outdated, but we can update the rest later.

README.md
```diff
@@ -1,17 +1,39 @@
...
+- Rust sanitizers for valid memory addresses, leaks and thread data races
+
+Here's how to run the checks locally:
+
+```shell
+# Run ruby_memcheck in combination with a Rust sanitization mode (Linux only)
+SANITIZER=leak bundle exec rake test:valgrind
+
+# Run only ruby_memcheck (Linux only)
+SANITIZER=leak bundle exec rake test:valgrind
```
Morriar: Do we need the sanitizer for this?

```suggestion
bundle exec rake test:valgrind
```

vinistock: Bad copy and paste

PR #122, opened 2025-08-22
Index variable definitions

Morriar:
Partial implementation for https://github.com/Shopify/index/issues/82, https://github.com/Shopify/index/issues/83 and https://github.com/Shopify/index/issues/84.

Some notes:

* For now we only create definition for `=` assign, not for `&=`, `&&=`, `||=`. It's still unclear if the should be definitions, references or both
* Names are simplified to always use `::` so `Foo::@foo` instead of `Foo.@goo`. We should revisit this later.

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -2,8 +2,8 @@
 
 use crate::indexing::errors::IndexingError;
 use crate::model::definitions::{
-    ClassDefinition, ConstantDefinition, Definition, GlobalVariableDefinition, InstanceVariableDefinition,
-    ModuleDefinition,
+    ClassDefinition, ClassVariableDefinition, ConstantDefinition, Definition, GlobalVariableDefinition,
+    InstanceVariableDefinition, ModuleDefinition,
 };
 use crate::model::graph::Graph;
 use crate::model::ids::UriId;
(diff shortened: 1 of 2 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -171,12 +174,91 @@ impl Visit<'_> for RubyIndexer {
                             .add_definition(indexer.uri_id, fully_qualified_name, definition);
                     });
                 }
+                ruby_prism::Node::GlobalVariableTargetNode { .. } => {
```
vinistock, line 177: Can a `multi_write_node` contain variables? I thought if we had

```ruby
$a, $b = ...
@a, @b = ...
@@a, @@c = ...
```

You'd have the `target` variants of the nodes (e.g.: `global_variable_target_node`). My previous understanding was that multi write was only for constants.

Morriar: I'm not sure I understand the question.

A `MultiWriteNode` is any write to multiple targets at once: constants, variables, globals, etc.

Depending on the kind of target the type of `target` varies:
* `A, B = ...` => `ConstantTargetNode`
* `A::B, C::D = ...` => `ConstantPathTargetNode`
* `$a, $b = ...` => `GlobalVariableTargetNode`
* `@a, @b = ...` => `InstanceVariableTargetNode`
* `@@a, @@b = ...` => `ClassVariableTargetNode`
* `a, b = ...` => `LocalVariableTargetNode`

And they can be mixed up:

```rb
Prism.parse("a, @b, $c, D = ARGV")
```
```
#<Prism::ParseResult:0x0000000136958a30
 @value=
  @ ProgramNode (location: (1,0)-(1,19))
  └── statements:
      @ StatementsNode (location: (1,0)-(1,19))
      └── body: (length: 1)
          └── @ MultiWriteNode (location: (1,0)-(1,19))
              │   ├── @ LocalVariableTargetNode (location: (1,0)-(1,1))
              │   │   └── name: :a
              │   ├── @ InstanceVariableTargetNode (location: (1,3)-(1,5))
              │   │   └── name: :@b
              │   ├── @ GlobalVariableTargetNode (location: (1,7)-(1,9))
              │   │   └── name: :$c
              │   └── @ ConstantTargetNode (location: (1,11)-(1,12))
              │       └── name: :D
              └── value:
                  @ ConstantReadNode (location: (1,15)-(1,19))
                  └── name: :ARGV,

PR #123, opened 2025-08-22
Adding logic to dump graph data to DB

ChallaHalla:
This PR extends the DB model to take graph data from `Graph` and batch insert `Definitions`, `Names`, and `Documents` to the database.

Main changes include:
- New methods on `Db` for batch inserting `Definitions`, `Names`, and `Documents`
- New methods on graph for writing graph data to the DB and for clearing graph data from memory
- New hash map of `definition_to_uri` to allow for fetching a definition's URI without iterating through all the entries of `uri_to_definition`.
  - This addition is important because there is no way to know a definition's URI when inserting it in to the DB without this map
- Tests for new additions

This PR is only meant to cover writing graph data to the DB. A follow up PR will introduce loading graph data from the db. Once loading is introduced, we can begin dumping to the database/clearing memory after indexing and loading data from the DB in to memory as needed.

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -1,10 +1,8 @@
-use crate::model::definitions::Definition;
-use crate::model::ids::{DefinitionId, NameId, UriId};
+use crate::model::graph::Graph;
 use rusqlite::Connection;
-use std::collections::HashMap;
 use std::error::Error;
 
-const SCHEMA_VERSION: i32 = 1;
+const SCHEMA_VERSION: u16 = 1;
 
 #[derive(Default, Debug)]
 enum ConnectionType {
@@ -38,9 +36,7 @@ impl Db {
         conn.execute_batch(
             "
             PRAGMA journal_mode = WAL;
-            PRAGMA synchronous = NORMAL;
-            PRAGMA foreign_keys = ON;
-        ",
+            ",
         )?;
 
         // Perform remaining DB init within at tx
@@ -66,15 +62,33 @@ impl Db {
         }
     }
 
+    /// Saves the full graph to the database. This is ONLY meant to be used for newly created database files and not
+    /// synchronization as it will not try to update existing records for better insertion performance
+    ///
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn save_full_graph(&self, graph: &Graph) -> Result<(), Box<dyn Error>> {
+        let mut conn = self.get_connection()?;
+        let tx = conn.transaction()?;
+
+        Self::batch_insert_documents(&tx, graph)?;
+        Self::batch_insert_names(&tx, graph)?;
+        Self::batch_insert_definitions(&tx, graph)?;
+
+        tx.commit()?;
+        Ok(())
+    }
+
     /// # Errors
     /// Will return an Error if we fail to establish or set a connection
-    pub fn get_connection(&self) -> Result<Connection, Box<dyn Error>> {
+    fn get_connection(&self) -> Result<Connection, Box<dyn Error>> {
         let conn = match &self.connection_type {
             ConnectionType::Path(path) => {
                 let conn = Connection::open(path)?;
                 // Check if schema needs to be loaded using user_version pragma.
                 // The default value for user_version is 0.
-                let current_version: i32 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
+                let current_version: u16 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
                 if current_version < SCHEMA_VERSION {
                     drop(conn);
                     return self.recreate_database();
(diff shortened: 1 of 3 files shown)


rust/index/src/model/db.rs
```diff
@@ -31,15 +34,17 @@ impl Db {
 
     /// Initializes a fresh database with schema and configuration
     fn initialize_database(conn: &mut Connection) -> Result<(), Box<dyn Error>> {
-        let tx = conn.transaction()?;
-        tx.execute_batch(
+        // Set Pragmas first
+        conn.execute_batch(
```
ChallaHalla, line 36: There is a DB error that occurs if Pragmas are set within a transaction. So this step must exist outside of the `tx`

rust/index/src/model/db.rs
```diff
@@ -85,4 +90,281 @@ impl Db {
...
+        documents: &HashMap<UriId, String>,
+        names: &HashMap<NameId, String>,
+        definitions: &HashMap<DefinitionId, Definition>,
+        definition_to_name: &HashMap<DefinitionId, NameId>,
+        definition_to_uri: &HashMap<DefinitionId, UriId>,
+    ) -> Result<(), Box<dyn Error>> {
+        let mut conn = self.get_connection()?;
+        let tx = conn.transaction()?;
+
+        Self::batch_insert_documents(&tx, documents)?;
```
ChallaHalla: I learned about the `Deref` trait with this PR which is pretty neat ([link](https://doc.rust-lang.org/std/ops/trait.Deref.html)).

This allows us to pass the tx in as an argument to the batch insert methods. The insert methods will deref the tx and get the underlying `Connection` which it can use to query as normal. The `Connection` used will still be part of the `Transaction`. This allows us flexibility to use the batch insert methods either with a `Transaction` or directly with a `Connection`.

ChallaHalla: Relevant code on `Transaction` from the rusqlite crate: 
```rust
impl Deref for Transaction<'_> {
    type Target = Connection;

    #[inline]
    fn deref(&self) -> &Connection {
        self.conn
    }
}
```

rust/index/src/db/schema.sql
```diff
@@ -16,9 +16,9 @@ CREATE TABLE IF NOT EXISTS names (
 -- Table for storing definitions that make up declarations
 CREATE TABLE IF NOT EXISTS definitions (
     id TEXT PRIMARY KEY,  -- Blake3 hash converted to hex
-    name_id TEXT NOT NULL,  -- References names.id
+    name_id TEXT NOT NULL REFERENCES names(id), -- References names.id
```
ChallaHalla, line 19: `REFERENCES` here will just enforce existence of the row that the FK points to. I think including it should be fine since we do insertion within a transaction for all the data.

vinistock: Yes, we should ensure that the database's state is always valid at the schema level.

rust/index/src/model/db.rs
vinistock: Did you confirm that these only have to be executed once for the database? I remember we chatted about whether this was per database or per connection.

rust/index/src/model/definitions.rs
```diff
@@ -25,6 +25,9 @@
 
 use crate::offset::Offset;
 
+const CLASS_ID: i32 = 1;
+const MODULE_ID: i32 = 2;
```
vinistock: Honestly, I think we can inline these. There's no need for them to be constants if we're always going to interact with them through `type_id`.

rust/index/src/model/definitions.rs
```diff
@@ -47,6 +50,15 @@ impl Definition {
             Definition::Module(it) => it.offset.end_offset(),
         }
     }
+
+    // Mapping of a definition's type to the definition_type enum value in the DB.
+    #[must_use]
+    pub fn type_id(&self) -> i32 {
```
vinistock: `i32` is quite a large number to store definitions. Also, we'd never have have a negative ID.

Can we use `u8` here?


```suggestion
    pub fn type_id(&self) -> u8 {
```

rust/index/src/model/graph.rs
```diff
@@ -24,6 +25,8 @@ pub struct Graph {
     name_to_definitions: HashMap<NameId, HashSet<DefinitionId>>,
     // Reverse map of a definition to its unique fully qualified name
     definition_to_name: HashMap<DefinitionId, NameId>,
+    // Reverse map of a definition to its URI
+    definition_to_uri: HashMap<DefinitionId, UriId>,
```
vinistock, line 29: Why is this new map necessary?

rust/index/src/model/db.rs
ChallaHalla: I checked and confirmed that this is actually set once for the DB, not per connection.

rust/index/src/model/graph.rs
vinistock: Talked to Sid: not strictly necessary for dumping the Graph, but we do need this to provide hover or go to definition anyway.

rust/index/src/model/db.rs
vinistock: For posterity, I double checked this and it actually depends on which pragma we're talking about.

The `journal_mode` is persistent per database, but `synchronous` and others are actually per connection and I was able to make it a bit faster by setting them on the connection.

There were also other interesting pragmas for us, like `locking_mode = EXCLUSIVE`, which prevents any other connections from existing. This is nice because it allows us to protect the database from unwanted modifications coming from other processes.

PR #124, opened 2025-08-22
Index attr_accessor, attr_reader, attr_writer definitions

Morriar:
Partial implementation for https://github.com/Shopify/index/issues/90.

Note: for now names are registered as `Foo::foo` instead of `Foo#foo`.

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -2,8 +2,8 @@
 
 use crate::indexing::errors::IndexingError;
 use crate::model::definitions::{
-    AttrAccessorDefinition, AttrReaderDefinition, ClassDefinition, ClassVariableDefinition, ConstantDefinition,
-    Definition, GlobalVariableDefinition, InstanceVariableDefinition, ModuleDefinition,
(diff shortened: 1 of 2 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -216,6 +216,77 @@ impl Visit<'_> for RubyIndexer {
...
+        let message_loc = node.message_loc();
+
+        if message_loc.is_none() {
+            return;
+        }
+
+        let message = Self::location_to_string(&message_loc.unwrap());
+        if message != "attr_accessor" && message != "attr_reader" && message != "attr_writer" {
+            return;
+        }
```
vinistock, line 229: We are matching on the name twice for the early return and then for deciding what to do.

This method will grow a lot since a lot of indexing cares about call nodes (e.g.: include, extend, prepend, module_function, private_constant...). Can we use a big match statement to handle each case and then have a catch all that does nothing?

I think that will look cleaner.

Morriar: https://github.com/Shopify/index/pull/136

PR #125, opened 2025-08-22
Index method definitions

Morriar:
Closes https://github.com/Shopify/index/issues/81.

Note, for now names are created as `Foo::foo` instead of `Foo#foo`.

### DIFF
--- a/rust/index/src/model/definitions.rs
+++ b/rust/index/src/model/definitions.rs
@@ -30,6 +30,7 @@ pub enum Definition {
     Class(Box<ClassDefinition>),
     Module(Box<ModuleDefinition>),
     Constant(Box<ConstantDefinition>),
+    Method(Box<MethodDefinition>),
     AttrAccessor(Box<AttrAccessorDefinition>),
     AttrReader(Box<AttrReaderDefinition>),
     AttrWriter(Box<AttrWriterDefinition>),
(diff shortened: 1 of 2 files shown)


rust/index/src/model/definitions.rs
```diff
@@ -143,6 +147,67 @@ impl ConstantDefinition {
...
+    pub fn is_singleton(&self) -> bool {
+        self.is_singleton
+    }
+}
+
+#[derive(Debug)]
+pub struct Parameter {
+    pub offset: Offset,
+    pub name: String,
+    pub kind: ParameterKind,
```
vinistock, line 189: Can we avoid making these public?

rust/index/src/model/definitions.rs
```diff
@@ -143,6 +147,67 @@ impl ConstantDefinition {
...
+
+impl Parameter {
+    #[must_use]
+    pub const fn new(offset: Offset, name: String, kind: ParameterKind) -> Self {
+        Self { offset, name, kind }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ParameterKind {
```
vinistock, line 200: We're missing forward arguments. Also, instead of having a `kind` in the parameters, can we lean into enums so that we can exhaustiveness at the parameter level?

We can even reuse the same struct for all variants.

PR #130, opened 2025-08-26
Add ability to load all data related to a URI

vinistock:
Allow loading all graph data related to a URI from the database.

In addition to adding the loading logic, I also made the connection persistent. The previous approach was really annoying because in-memory databases are destroyed immediately if there are no connections, which meant that we would need to over-expose the API for testing.

I think this approach will also make it easier to handle failures, like the user deleting or corrupting the database file - which we can recover from.

### DIFF
--- a/ext/index/graph.c
+++ b/ext/index/graph.c
@@ -86,7 +86,10 @@ static VALUE rb_graph_set_configuration(VALUE self, VALUE db_path) {
     void* graph;
     TypedData_Get_Struct(self, void*, &graph_type, graph);
 
-    idx_graph_set_configuration(graph, StringValueCStr(db_path));
+    if (!idx_graph_set_configuration(graph, StringValueCStr(db_path))) {
+        rb_raise(rb_eRuntimeError, "Failed to set the database configuration");
+    }
+
     return Qnil;
 }
 
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -1,33 +1,144 @@
+use crate::model::definitions::Definition;
 use crate::model::graph::Graph;
 use rusqlite::{Connection, params};
-use std::error::Error;
+use std::{cell::RefCell, error::Error, fs, path::Path};
 
 const SCHEMA_VERSION: u16 = 1;
 
-#[derive(Default, Debug)]
-enum ConnectionType {
-    Path(String),
-    #[default]
-    Memory,
+pub struct LoadResult {
+    pub name_id: String,
+    pub name: String,
+    pub definition_id: String,
+    pub definition: Definition,
 }
 
 #[derive(Default, Debug)]
 pub struct Db {
-    connection_type: ConnectionType,
+    connection: RefCell<Option<Connection>>,
 }
 
 impl Db {
     const SCHEMA: &str = include_str!("../db/schema.sql");
+    const LOAD_DOCUMENT: &str = include_str!("../db/load_document.sql");
 
     #[must_use]
     pub fn new() -> Self {
         Self {
-            connection_type: ConnectionType::Memory,
+            connection: RefCell::new(None),
         }
     }
 
-    pub fn set_db_path(&mut self, path: String) {
-        self.connection_type = ConnectionType::Path(path);
+    /// Initialize the database connection. To use an in memory database, pass `None`
+    ///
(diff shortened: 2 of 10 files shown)


ext/index/graph.c
```diff
@@ -86,7 +87,10 @@ static VALUE rb_graph_set_configuration(VALUE self, VALUE db_path) {
     void* graph;
     TypedData_Get_Struct(self, void*, &graph_type, graph);
 
-    idx_graph_set_configuration(graph, StringValueCStr(db_path));
+    if (!idx_graph_set_configuration(graph, StringValueCStr(db_path))) {
+        rb_raise(rb_eArgError, "Failed to set the database configuration");
```
Morriar:
```suggestion
        rb_raise(rb_eRuntimeError, "Failed to set the database configuration");
```

ext/index/graph.c
```diff
@@ -1,4 +1,5 @@
 #include "graph.h"
+#include "ruby/internal/globals.h"
```
Morriar: Can't we use `ruby.h` instead of reaching to internals?

rust/index/src/model/db.rs
```diff
@@ -68,50 +110,55 @@ impl Db {
...
+    {
+        let mut borrow = self.connection.borrow_mut();
+        let connection: &mut Connection = borrow
+            .as_mut()
+            .expect("Database connection has to be initialized ahead of time");
+
+        match f(connection) {
+            Ok(result) => Ok(result),
+            Err(e) => Err(e),
+        }
```
Morriar:
```suggestion
        f(connection)
```

rust/index/graph.db
Morriar, line 1: Do we want to commit this?

rust/index/src/model/definitions.rs
```diff
@@ -55,9 +69,9 @@ impl Definition {
     #[must_use]
     pub fn type_id(&self) -> u8 {
         match self {
-            Definition::Class(_) => 1,
```
Morriar: Let's extract the numbers as enum?

```
pub enum DefinitionKind {
    Class = 0,
    Module = 1,
    Constant = 2,
}
```

And reuse them in `type_id` and `from_db`

ext/index/graph.c
vinistock: Clangd is just trolling me at this point. We are not reaching for any internals, I have no idea why it keeps adding these includes.

rust/index/graph.db
vinistock: Nope

rust/index/src/model/definitions.rs
vinistock: This actually makes the implementation a bit less convenient because we need to instantiate the enum variants when getting the integer value from the database.

And since we need to pass the types as integers, we just end up returning a u8 anyway. We can give it another shot later, but my attempts made the code more convoluted.

Morriar: What about something like this?

```rust
pub enum DefinitionKind {
    Class = 0,
    Module = 1,
}

impl TryFrom<u8> for DefinitionKind {
    type Error = &'static str;
    fn try_from(value: u8) -> Result<Self, Self::Error> {
        match value {
            0 => Ok(DefinitionKind::Class),
            1 => Ok(DefinitionKind::Module),
            _ => Err("unknown definition kind"),
        }
    }
}

impl Definition {
    pub fn type_id(&self) -> u8 {
        match self {
            Definition::Class(_) => DefinitionKind::Class as u8,
            Definition::Module(_) => DefinitionKind::Module as u8,
        }
    }

    pub fn from_db(definition_type: u8, start_offset: u32, end_offset: u32) -> Definition {
        match DefinitionKind::try_from(definition_type).expect("invalid definition kind from DB") {
            DefinitionKind::Class => Definition::Class(Box::new(ClassDefinition::new(Offset::new(start_offset, end_offset)))),
            DefinitionKind::Module => Definition::Module(Box::new(ModuleDefinition::new(Offset::new(start_offset, end_offset)))),
        }
    }
}
```

vinistock: By properly encoding the definition's data as a blob, we no longer even need the concept of `definition_type`. I removed it and now there's no longer a need to sync anything between the schema and the definition implementations.

PR #136, opened 2025-08-26
Rework sends indexing

Morriar:
Adresses https://github.com/Shopify/index/pull/124#discussion_r2301923550

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -325,70 +325,111 @@ impl Visit<'_> for RubyIndexer {
         let message_loc = node.message_loc();
 
         if message_loc.is_none() {
+            // No message, we can't index this node
             return;
         }
 
         let message = Self::location_to_string(&message_loc.unwrap());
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -325,70 +325,111 @@ impl Visit<'_> for RubyIndexer {
...
+            }
+        }
+
+        if let Some(receiver) = node.receiver() {
+            self.visit(&receiver);
+        }
+
+        if let Some(arguments) = node.arguments() {
+            self.visit(&arguments.as_node());
+        }
```
vinistock, line 429: Do we need to visit the receiver and argument? Is there a case where we'd discover something in those? Maybe we can get away not visiting them

Morriar:
```rb
Class.new(Bar) do
  def foo
    puts FOO
  end
end.new
```

Even if we don't know how to handle the `Class.new` yet, the constant references to `Bar` and `FOO` require going through the receiver and args.

We can disable it but we'll need to put it back once we collect the references.

PR #137, opened 2025-08-26
Rework method parameters

Morriar:
Address comments from https://github.com/Shopify/index/pull/125.

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -230,18 +230,14 @@ impl Visit<'_> for RubyIndexer {
                     )));
                 }
                 for parameter in parameters_list.optionals().iter() {
-                    if let ruby_prism::Node::OptionalParameterNode { .. } = parameter
-                        && let Some(opt_param) = parameter.as_optional_parameter_node()
-                    {
-                        parameters.push(Parameter::OptionalPositional(ParameterStruct::new(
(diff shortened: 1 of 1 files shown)


rust/index/src/model/definitions.rs
```diff
@@ -183,29 +183,39 @@ impl MethodDefinition {
...
+}
+
+#[derive(Debug)]
+pub struct ParameterStruct {
+    offset: Offset,
+    name: String,
 }
 
-impl Parameter {
+impl ParameterStruct {
```
Morriar, line 206: Note a huge fan of this name

vinistock: Maybe just `Param`? Or make the enum `Param`, so that `Param::Post` becomes shorter?

Morriar: At the moment the public API is all without unabbreviated names: `Definition`, `Class`, `Module`, `Method`, `Variable`. The exception is `AttrAccessor` and friends. So maybe we should keep them unabbreviated and possibly fix `Attr`.

`ParameterStruct` is private-ish and doesn't need to be imported by the client unless they need to create them. But we could also add different constructors into `Parameter` to hide it: `Parameter::new_required_positional`.

I'm not convinced by `Param` vs. `Parameter`, it's just confusing which is which.

Morriar: I'll merge and rename if we find a better way.

PR #138, opened 2025-08-27
Ensure deleting documents clear the database

vinistock:
This PR is the second part of #130, adding the ability to unload a URI from memory. We actually already had that implemented, but for the `delete_uri` (which should handle a file being completely deleted).

I extracted the original implementation for only unloading from memory into `unload_uri` and made `delete_uri` handle the database portion of deletion as well.

With this, we are in a position where we can start benchmarking and testing what the memory usage looks like after we dump things to Sqlite and incrementally load/unload them.

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -1,5 +1,5 @@
-use crate::model::definitions::Definition;
 use crate::model::graph::Graph;
+use crate::model::{definitions::Definition, ids::UriId};
 use rusqlite::{Connection, params};
 use std::{cell::RefCell, error::Error, fs, path::Path};
 
@@ -123,6 +123,21 @@ impl Db {
             Self::batch_insert_definitions(&tx, graph)?;
 
             tx.commit()?;
+            Ok(())
+        })
+    }
+
+    /// Deletes all of the data related to the given URI ID. This method relies on the foreign key constraints to
+    /// cascade the deletion to definitions and names that are no longer referenced
+    ///
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn delete_data_for_uri(&self, uri_id: UriId) -> Result<(), Box<dyn Error>> {
+        self.with_connection(|connection| {
+            let id_as_string = uri_id.to_string();
+            let mut stmt = connection.prepare_cached("DELETE FROM documents WHERE id = ?")?;
+            stmt.execute([id_as_string])?;
             Ok(())
         })
     }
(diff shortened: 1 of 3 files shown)


rust/index/src/model/db.rs
```diff
@@ -127,6 +127,20 @@ impl Db {
...
+    /// Deletes all of the data related to the given URI ID. This method relies on the foreign key constraints to
+    /// cascade the deletion to definitions and names that are no longer referenced
+    ///
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn delete_data_for_uri(&self, uri_id: UriId) -> Result<(), Box<dyn Error>> {
+        self.with_connection(|connection| {
+            let id_as_string = uri_id.to_string();
+            connection.execute("DELETE FROM documents WHERE id = ?", [id_as_string])?;
```
Morriar: Should we first use `prepare_cached` like other queries in this file?

rust/index/src/model/graph.rs
```diff
@@ -117,12 +117,19 @@ impl Graph {
...
+    /// Handles when a document (identified by `uri`) is deleted. This removes the URI from the graph along with all
+    /// relationships and definitions that had been discovered in it
+    ///
+    /// # Errors
+    ///
+    /// Any database errors will prevent the data from being deleted
+    pub fn delete_uri(&mut self, uri: &str) -> Result<(), Box<dyn Error>> {
+        // Delete the data from memory
+        self.unload_uri(uri);
+        // Delete the data from the database
```
Morriar, line 129: I wonder if we shouldn't invert the order. So if the DB removal fails, the memory stays in a consistent state.

rust/index/src/model/graph.rs
```diff
@@ -172,6 +179,14 @@ impl Graph {
         Ok(())
     }
 
+    /// Removes all data related to the given URI from memory. This method should only be used for when a document is
+    /// closed. If the file is deleted, we also need to update the database
+    pub fn unload_uri(&mut self, uri: &str) {
```
Morriar, line 184: `load_uri` takes a String but `unload_uri` takes a `&str`?

Morriar: Should it also return a `Result<(), Box<dyn Error>>` like `load_uri`?

rust/index/src/model/db.rs
vinistock: I made that change, although I still want to benchmark and understand this thing better. I'm suspicious that these queries might be getting cached too many times.

rust/index/src/model/graph.rs
vinistock: I don't think so because this method is intended to handle a file being deleted. If the file was removed from the system by the user, there's no reason for us to keep the data still in memory.

If they restore the file, we'd then handle it as a newly created addition.

rust/index/src/model/graph.rs
vinistock: Yes because:

1. `load_uri` needs to insert the owned string into the `uri_pool`. Unload URI only needs to figure out the `UriId` to be able to follow the graph and delete stuff. It does not require ownership
2. It shouldn't return a result because it doesn't depend on the database - unlike `load_uri`. Loading will try to query the database, which may fail due to connection issues, deleted DB file and so on. Unloading only removes the data from memory without doing any DB communication, which means it cannot fail

PR #144, opened 2025-08-28
Ignore SQLite db files

Morriar:
I keep seeing this file in my git status

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@
 *.o
 *.a
 *.dSYM
+*.db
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,6 +11,7 @@
 *.o
 *.a
 *.dSYM
+*.db
```
vinistock, line 14: Let's also do `*.db-*` because depending on the operations you get some extra ones.


```suggestion
*.db
*.db-*
```

PR #145, opened 2025-08-28
Switch to u64 IDs

vinistock:
This PR changes our IDs from blake3 to xxhash for now. Blake3 seems to be faster to compute, but uses more memory and we can save a bunch by using u64.

I also made the Ids generic to avoid code duplication.


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-C", "target-cpu=native"]
```
Morriar: I think we're missing the final newline here

PR #147, opened 2025-08-28
Make it easier to test against real files

Morriar:
@jenny-codes I think it will make your CLI tests much easier to write?

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -214,31 +214,76 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
-
     use super::*;
+    use crate::test_utils::Context;
 
-    #[test]
-    fn collecting_documents_in_parallel() {
-        let cargo_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
-        let root = cargo_path.parent().unwrap().parent().unwrap();
+    fn collect_document_paths(context: &Context, paths: &[&str]) -> (Vec<String>, Vec<IndexingError>) {
+        let (documents, errors) = collect_documents(
+            paths
+                .iter()
+                .map(|p| context.absolute_path_to(p).to_string_lossy().into_owned())
+                .collect(),
+        );
 
-        // Paths including a specific file and a directory
-        let paths = vec![
-            root.join("lib/saturn.rb").to_str().unwrap().to_string(),
-            root.join("lib/saturn").to_str().unwrap().to_string(),
-        ];
+        let mut paths: Vec<String> = documents
+            .iter()
+            .map(|d| {
+                context
+                    .relative_path_to(d.path().ok().unwrap())
+                    .to_string_lossy()
+                    .into_owned()
+            })
+            .collect();
 
-        let (documents, errors) = collect_documents(paths);
-        assert!(errors.is_empty(), "Errors: {errors:#?}");
-        assert!(documents.len() > 2);
+        paths.sort();
+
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing.rs
```diff
@@ -156,31 +156,72 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
-
     use super::*;
+    use crate::test_utils::Context;
+
+    fn collecting_document_paths(context: &Context, paths: &[&str]) -> (Vec<String>, Vec<IndexingError>) {
```
vinistock: Nit: can we call this `collect` instead of `collecting`?

rust/index/src/indexing.rs
```diff
@@ -156,31 +156,72 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
...
+    }
+
+    #[test]
+    fn collecting_all_documents() {
+        let context = Context::new();
+        context.touch("bar/baz.rb");
+        context.touch("bar/qux.rb");
+        context.touch("foo/bar.rb");
+
+        let (paths, errors) = collecting_document_paths(&context, &["foo", "bar"]);
```
vinistock: Are these missing the `.rb` at the end? Or is this doing partial matches?

rust/index/src/test_utils/context.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        assert_eq!(
+            context.read("foo/bar.rb"),
+            normalize_indentation({
+                "
+                class Foo;
+                    def bar;
+                        puts 'baz';
+                    end
+                end
```
vinistock: I think the indentation here is using 4 instead of 2 spaces. Also, did you mean to have both line breaks and semi colons?

rust/index/src/indexing.rs
Morriar: It's matching on the directory, see the glob: `glob(&format!("{path}/**/*.rb"))`

rust/index/src/test_utils/context.rs
Morriar: Oh nice catch, I split the text with line breaks but forgot to remove the `;`. I also fixed the indent 👍

rust/index/src/test_utils/context.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// Returns the absolute path to the temp directory as a `PathBuf`
+    ///
+    /// # Panics
+    ///
+    /// Panics if the path cannot be canonicalized.
+    #[must_use]
+    pub fn absolute_path(&self) -> PathBuf {
+        fs::canonicalize(self.root.path()).unwrap()
```
vinistock: Do we ever need the relative path of the tmpdir? Maybe we can just store the absolute path directly on instantiation.

Like, I know we need relative paths in respect to the tmpdir, but the tmpdir itself could always be absolute, no?

rust/saturn/src/test_utils/context.rs
```diff
@@ -0,0 +1,191 @@
...
+        let path = self.absolute_path().join(relative);
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent).unwrap();
+        }
+        let content = normalize_indentation(content);
+        fs::write(path, content).unwrap();
+    }
+}
+
+impl Default for Context {
```
vinistock, line 106: Nit: let's use `#[derive(Default)]` in the struct instead

Morriar: We can't because `TempDir` doesn't implement `Default`

PR #150, opened 2025-08-29
Add benchmark script

vinistock:
Add a convenience script to benchmark against the huge corpus without having to supply any arguments or take any manual steps.

### DIFF
--- a/utils/bench
+++ b/utils/bench
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# A script to benchmark indexing against the huge corpus in release mode. Use this to check for any performance
+# improvements or regressions
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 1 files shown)


utils/bench
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# A script to benchmark indexing against the huge corpus in release mode. Use this to check for any performance
+# improvements or regressions
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+CORPUS_PATH="$(dirname "$PROJECT_ROOT")/huge_corpus"
```
Morriar, line 8: Should we avoid creating the corpus inside the project root?

vinistock: It's not inside, it's a sibling of the project. `dirname` gives us the parent directory.

PR #151, opened 2025-08-29
Proposal for internals timers

Morriar:
Running the Rust CLI with `--timers` shows internal timers:

```
$ index_cli corpus/huge --timers

Timers:
  ├─ 00.16s     collect_documents
  ├─ 25.03s     index_documents
  │  ├─ 42.20s  *parse_ruby
  │  ├─ 24.66s  *visit_ruby
  │  ├─ 00.01s  *add_uri
  │  ├─ 05.84s  *add_definition
  |  └─ 12.96s  extend_graph
  └─ 24.28s     clear_graph
```

Entries marked with a `*` are happening in parallel.

To accumulate inside a timer use the `start`/`stop` API:

```rust
    graph.timers().index_documents().start();
    indexing::index_in_parallel(&mut graph, &documents)?;
    graph.timers().index_documents().stop();
```

### DIFF
--- a/rust/index/src/timers.rs
+++ b/rust/index/src/timers.rs
@@ -101,3 +101,13 @@ impl Default for Timers {
         Self::new()
     }
 }
+
+#[macro_export]
+macro_rules! timer {
+    ($index:expr, $timer:ident, $block:block) => {{
+        $index.timers().$timer().start();
+        let __result = { $block };
+        $index.timers().$timer().stop();
+        __result
+    }};
+}
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -9,6 +9,7 @@ use crate::model::definitions::{
 use crate::model::graph::Graph;
 use crate::model::ids::UriId;
 use crate::offset::Offset;
+use crate::timer;
 
 use ruby_prism::Visit;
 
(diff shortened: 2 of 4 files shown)


rust/index/src/indexing.rs
```diff
@@ -90,6 +90,7 @@ pub fn index_in_parallel(graph: &mut Graph, documents: &[Document]) -> Result<()
 
     for indexer in indexers {
         let (local_graph, errors) = indexer.into_parts();
+        graph.timers().add(local_graph.timers_ref());
```
vinistock, line 93: Should this be a part of `update` so that we merge graphs in one single call?

rust/index/src/timers.rs
```diff
@@ -0,0 +1,113 @@
+use crate::timer::Timer;
+
+#[allow(clippy::cast_possible_truncation)]
+fn seconds(microseconds: u128) -> String {
+    format!("{:0>5.2}s", f64::from(microseconds as u32) / 1_000_000.0)
+}
+
+#[derive(Debug)]
+pub struct Timers {
```
vinistock, line 9: Right now, adding any new metric means we also need to make changes to the timers. Would it be possible to generalize this into a hashmap so that the name used by `timer!` knows where to accumulate?

rust/index/src/timers.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl Default for Timers {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[macro_export]
+macro_rules! timer {
```
vinistock, line 106: Do we see any different before and after this PR for our huge corpus benchmark? Do we need to make this available through some compilation flag or is it basically zero cost?

Issue #139, opened 2025-08-27
Experiment with moving edges inside nodes


vinistock:
We're currently keeping edges in top level hashmaps, which means that we end up duplicating the IDs that are used for the hash keys.

If we move the edges inside the nodes themselves, we can avoid that price and reduce memory usage. Let's give that a try.

PR #152, opened 2025-08-29
Put name_id and uri_id inside definitions

vinistock:
First step for #139

This PR moves the edges of `definition -> fully qualified name` and `definition -> uri` inside of the definitions themselves.

There are 3 commits because after my changes 2 indexing methods became too long and I had to refactor those:

1. The first commit is the actual change, putting `name_id` and `uri_id` inside of all definition structs and getting rid of the related hashmaps
2. This second one is to resolve the method too long violation for indexing method definitions. This one is straight forward. I moved the parameter handling into a separate method
3. This third one is for the same violation, but for call nodes. This one was a little trickier as each attribute type uses a different definition struct. I decided to go with a generic method to loop through any call node invoked directly on `self` with symbol or string literal arguments. This allows us to handle all attributes, but it's also useful for other indexing relevant call nodes, like `private`, `public`, `protected`, `private_constant`, `module_function` and so on

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -144,6 +144,38 @@ impl RubyIndexer {
         parameters
     }
 
+    // Runs the given closure if the given call `node` is invoked directly on `self` for each one of its string or
+    // symbol arguments
+    fn each_string_or_symbol_arg<F>(node: &ruby_prism::CallNode, mut f: F)
+    where
+        F: FnMut(String, ruby_prism::Location),
+    {
+        let receiver = node.receiver();
+
+        if (receiver.is_none() || receiver.unwrap().as_self_node().is_some())
+            && let Some(arguments) = node.arguments()
+        {
+            for argument in arguments.arguments().iter() {
+                match argument {
+                    ruby_prism::Node::SymbolNode { .. } => {
+                        let symbol = argument.as_symbol_node().unwrap();
+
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -147,9 +277,12 @@ impl Visit<'_> for RubyIndexer {
...
 
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
-            let definition = Definition::Constant(Box::new(ConstantDefinition::new(Offset::from_prism_location(
-                &node.target().location(),
-            ))));
+            let name_id = NameId::from(&fully_qualified_name);
+            let definition = Definition::Constant(Box::new(ConstantDefinition::new(
+                name_id,
+                indexer.uri_id,
+                Offset::from_prism_location(&node.target().location()),
```
Morriar, line 284: Here and the other places, should we move it inside like for the `offset`?

```suggestion
            let definition = Definition::Class(Box::new(ClassDefinition::new(
                NameId::from(&fully_qualified_name),
                indexer.uri_id,
                Offset::from_prism_location(&node.location()),
```

vinistock: Unfortunately, I can't inline most of them because of borrow checking rules. We pass the `name` into the `add_definition` calls, which moves the value and then we can't get a reference to it to create the ID.

We need to create the ID first using the reference and then move the name inside of the graph - which means we need the intermediate variables.

PR #153, opened 2025-08-29
Sort DOT output for deterministic visualization

jenny-codes:
The DOT visualization output was non-deterministic due to definitions being sorted by their hash-based IDs. This caused test failures when the same  definitions appeared in different orders across test runs.

### DIFF
--- a/rust/index/src/visualization/dot.rs
+++ b/rust/index/src/visualization/dot.rs
@@ -56,21 +56,28 @@ fn write_name_nodes(output: &mut String, graph: &Graph) {
 }
 
 fn write_definition_nodes(output: &mut String, graph: &Graph) {
-    let mut definitions: Vec<_> = graph.definitions().iter().collect();
-    definitions.sort_by_key(|(def_id, _)| def_id.to_string());
+    let mut definitions: Vec<_> = graph
+        .definitions()
+        .iter()
+        .filter_map(|(def_id, definition)| {
+            graph
+                .definition_to_name()
+                .get(def_id)
+                .and_then(|name_id| graph.names().get(name_id))
+                .map(|name| {
+                    let def_type = definition.kind();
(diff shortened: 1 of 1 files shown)


rust/index/src/visualization/dot.rs
```diff
@@ -112,20 +116,38 @@ fn write_name_to_definition_edges(output: &mut String, graph: &Graph) {
...
+                        .and_then(|name_id| graph.names().get(name_id))
+                        .unwrap_or(&String::new())
+                        .clone();
+
+                    let def_type = definition.kind();
+                    let line = format!("    \"def_{def_id}\" -> \"{escaped_uri}\";\n");
+
+                    edges_for_uri.insert((def_type.to_string(), name), line);
                 }
             }
```
jenny-codes: Actually we shouldn't need to use BTreeMap for the uris as they are already deterministic strings. Reverting

rust/index/src/visualization/dot.rs
```diff
@@ -56,22 +57,25 @@ fn write_name_nodes(output: &mut String, graph: &Graph) {
...
 
-    for (def_id, definition) in definitions {
+    for (def_id, definition) in graph.definitions() {
         if let Some(name_id) = graph.definition_to_name().get(def_id)
             && let Some(name) = graph.names().get(name_id)
         {
             let def_type = definition.kind();
             let escaped_name = escape_dot_string(name);
+            let line =
+                format!("    \"def_{def_id}\" [label=\"{def_type}({escaped_name})\",shape={DEFINITION_NODE_SHAPE}];\n");
```
Morriar: Why do we need a BTreeMap? Can't we just sort on the `escaped_name`?

jenny-codes: Good point. Updated. A small adjustment to your suggestion is that I added the def_type info too and not just the escaped name.

PR #154, opened 2025-09-02
Add identity maps

vinistock:
This PR creates `IdentityHashMap` and `IdentityHashSet` implementations. These are identical to a regular `HashMap` or `HashSet`, with the sole difference that it does not try to hash any of the inserted values as they have already been hashed externally by our IDs.

This improves performance by quite a bit as we avoid a lot of duplicate work.

### DIFF
--- a/rust/index/src/model/identity_maps.rs
+++ b/rust/index/src/model/identity_maps.rs
@@ -1,0 +1,38 @@
+//! This module contains identity maps that use externally hashed IDs as keys. They are used to avoid hashing the same
+//! value twice, simply using the given key directly
+
+use std::{
+    collections::{HashMap, HashSet},
+    hash::{BuildHasher, Hasher},
+};
+
+#[derive(Default)]
+pub struct IdentityHasher {
+    hash: u64,
+}
+
+impl Hasher for IdentityHasher {
+    fn write(&mut self, _bytes: &[u8]) {}
+
+    fn write_u64(&mut self, i: u64) {
+        self.hash = i;
+    }
+
+    fn finish(&self) -> u64 {
+        self.hash
+    }
+}
+
+#[derive(Default)]
+pub struct IdentityHashBuilder;
+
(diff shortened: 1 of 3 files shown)


rust/index/src/model/identity_maps.rs
```diff
@@ -0,0 +1,38 @@
...
+    hash::{BuildHasher, Hasher},
+};
+
+#[derive(Default)]
+pub struct IdentityHasher {
+    hash: u64,
+}
+
+impl Hasher for IdentityHasher {
+    fn write(&mut self, _bytes: &[u8]) {}
```
Morriar, line 15: Should we do something like this?

```suggestion
    fn write(&mut self, _bytes: &[u8]) {
      unreachable!("IdentityHasher#write should never be called");
    }
```

rust/index/src/model/identity_maps.rs
```diff
@@ -0,0 +1,38 @@
...
+impl BuildHasher for IdentityHashBuilder {
+    type Hasher = IdentityHasher;
+
+    fn build_hasher(&self) -> Self::Hasher {
+        IdentityHasher::default()
+    }
+}
+
+pub type IdentityHashMap<K, V> = HashMap<K, V, IdentityHashBuilder>;
+pub type IdentityHashSet<K> = HashSet<K, IdentityHashBuilder>;
```
Morriar, line 38:
```suggestion
pub type IdentityHashSet<T> = HashSet<T, IdentityHashBuilder>;
```

rust/index/src/model/identity_maps.rs
Morriar: Should we also disable all the other variants? http://doc.rust-lang.org/beta/std/hash/trait.Hasher.html#provided-methods

rust/index/src/model/identity_maps.rs
Morriar, line 1: Any way we can add tests for this?

PR #156, opened 2025-09-02
Move definition edges inside declaration

vinistock:
Second step for #139

This PR creates the concept of a `Declaration`, so that we can move the edges of `name_to_definitions` inside of it and have more memory savings on the hash keys.

For now, the struct only stores its name and the list of definition IDs it is related to.

### DIFF
--- a/rust/index/src/model/declaration.rs
+++ b/rust/index/src/model/declaration.rs
@@ -1,0 +1,48 @@
+use std::collections::HashSet;
+
+use crate::model::{
+    identity_maps::{IdentityHashBuilder, IdentityHashSet},
+    ids::DefinitionId,
+};
+
+#[derive(Debug)]
+pub struct Declaration {
+    name: String,
+    definition_ids: IdentityHashSet<DefinitionId>,
+}
+
+impl Declaration {
+    #[must_use]
+    pub fn new(name: String) -> Self {
+        Self {
+            name,
+            definition_ids: HashSet::with_hasher(IdentityHashBuilder),
+        }
+    }
+
+    // Extend this declaration with more definitions by moving `other.definition_ids` inside
+    pub fn extend(&mut self, other: Declaration) {
+        self.definition_ids.extend(other.definition_ids);
+    }
+
+    #[must_use]
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    #[must_use]
+    pub fn definitions(&self) -> &IdentityHashSet<DefinitionId> {
+        &self.definition_ids
+    }
+
+    // Deletes a definition from this declaration. Returns `true` if this declaration is now empty, which indicates that
+    // it must be removed from the graph
+    pub fn remove_definition(&mut self, definition_id: &DefinitionId) -> bool {
+        self.definition_ids.remove(definition_id);
+        self.definition_ids.is_empty()
+    }
+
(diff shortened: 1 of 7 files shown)


rust/index/src/model/declaration.rs
```diff
@@ -0,0 +1,50 @@
+use std::collections::HashSet;
+
+use serde::{Deserialize, Serialize};
```
st0012: Vini: Do we still need this?

rust/index/src/model/graph.rs
```diff
@@ -120,12 +114,12 @@ impl Graph {
...
 
-        self.names.insert(name_id, name);
-        self.definitions.insert(definition_id, definition);
-        self.name_to_definitions
+        self.declarations
             .entry(name_id)
-            .or_default()
-            .insert(definition_id);
+            .or_insert_with(|| Declaration::new(name))
+            .add_definition(definition_id);
```
st0012, line 120: Note: this function actually doesn't need to take `uri_id` as it's already stored in `Definition`

rust/index/src/model/declaration.rs
vinistock: We don't. I removed it.

rust/index/src/model/graph.rs
vinistock: I'll make that change in a new PR to avoid cluttering this one with another refactor.

rust/index/src/model/declaration.rs
```diff
@@ -0,0 +1,48 @@
...
+    name: String,
+    definition_ids: IdentityHashSet<DefinitionId>,
+}
+
+impl Declaration {
+    #[must_use]
+    pub fn new(name: String) -> Self {
+        Self {
+            name,
+            definition_ids: HashSet::with_hasher(IdentityHashBuilder),
```
Morriar, line 19: I wonder if we could hide this implementation detail under a helper like `IdentityHashSet::new`

rust/index/src/model/declaration.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[must_use]
+    pub fn definitions(&self) -> &IdentityHashSet<DefinitionId> {
+        &self.definition_ids
+    }
+
+    // Deletes a definition from this declaration. Returns `true` if this declaration is now empty, which indicates that
+    // it must be removed from the graph
+    pub fn remove_definition(&mut self, definition_id: &DefinitionId) -> bool {
+        self.definition_ids.remove(definition_id);
+        self.definition_ids.is_empty()
```
Morriar, line 42: This API is a bit unusual, `remove` methods generally return a boolean indicating if the entry was actually removed from the collection.

vinistock: It made it more convenient to use, but I changed it in https://github.com/Shopify/index/pull/182.

rust/index/src/model/declaration.rs
vinistock: It ends up requiring a bit more boilerplate code to account for the generics of `HashMap`. I would hold back on adding it for now unless you feel really strongly about it.

Morriar: Actually I think we can do even better: https://github.com/Shopify/index/pull/185

PR #159, opened 2025-09-03
Use channels to incrementally build global graph

vinistock:
We're still spending quite a bit of time sleeping because the main thread doesn't do anything while the others are indexing. We can get some more speed up by using channels to communicate partial results back to the main thread as they are ready to be merged into the global graph.

This eliminates most of the sleeping and further reduces our initial indexing time.

### DIFF
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -1,5 +1,3 @@
-use std::{fs, path::Path};
-
 use crate::{
     indexing::{
         errors::{IndexingError, MultipleErrors},
@@ -8,7 +6,12 @@ use crate::{
     model::graph::Graph,
 };
 use glob::glob;
-use rayon::prelude::*;
+use std::sync::{
+    Arc, Mutex,
+    mpsc::{Receiver, Sender},
+};
+use std::{fs, path::Path};
+use std::{sync::mpsc, thread};
 use url::Url;
 pub mod errors;
 pub mod ruby_indexer;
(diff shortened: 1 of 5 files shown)


rust/index/src/indexing.rs
```diff
@@ -8,7 +6,12 @@ use crate::{
     model::graph::Graph,
 };
 use glob::glob;
-use rayon::prelude::*;
```
st0012, line 11: Is this and the dependency changes related to the main change?

rust/index/src/indexing.rs
```diff
@@ -57,38 +60,54 @@ impl Document {
...
-                    }
-                    Err(e) => {
-                        ruby_indexer.add_error(IndexingError::FileReadError(format!(
-                            "Failed to read {}: {}",
-                            document.path(),
-                            e
-                        )));
+pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
+    let (tx, rx): (Sender<RubyIndexer>, Receiver<RubyIndexer>) = mpsc::channel();
+    let num_threads = thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4);
```
st0012, line 65: 1. nit: Can we have `use std::num::NonZero;` and just use `NonZero::get` here?
2. Is it common in Rust to define magic numbers (`4`) in a constant with comments? Like how is 4 decided?

PR #163, opened 2025-09-04
Rename to Saturn

alexcrocha:
This PR renames the project name references from `Index` to `Saturn` across the entire codebase. 

## What Changed
- Gem name: `index` → `saturn`
- Module name: `Index::` → `Saturn::`
- CLI executable: `index` → `sat`
- FFI functions: `idx_*` → `sat_*`
- Rust crates: `index` → `saturn`, `index-sys` → `saturn-sys`
- Directory structure: `lib/index/` → `lib/saturn/`, `ext/index/` → `ext/saturn/`, etc.

## Next Steps
- [ ] Rename repository from `index` to `saturn`
- [ ] Rename Codeowners from `ruby-indexer-maintainers` to `saturn-maintainers`


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @Shopify/ruby-indexer-maintainers
+* @Shopify/saturn-maintainers
```
alexcrocha, line 1: Should this line be updated only after the team name is changed?

rust/saturn/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
-name = "index"
+name = "saturn"
```
alexcrocha, line 2: Hmm, shouldn't it be `rb-saturn` instead?

.github/CODEOWNERS
vinistock: I'd merge it as part of this PR. Just make sure to update the team name right before you merge.

rust/saturn/Cargo.toml
vinistock: We haven't actually checked what's available in `crates.io` and I'm not even sure we will publish this as a crate yet. Maybe the crate will be local to the Ruby gem and the only thing we publish is the gem itself.

For now, just `saturn` is probably fine.

rust/saturn-sys/src/lib.rs
```diff
@@ -1,2 +1,2 @@
 pub mod conversions;
-pub mod index_api;
+pub mod saturn_api;
```
vinistock: Let's actually rename this file to `graph_api`. My idea was to break down the API for each object the Ruby gem can interact with in a separate file and calling it `index_api` was an initial mistake.

alexcrocha: Opted to make this change in a new commit. Let me know if a different approach would be better

PR #167, opened 2025-09-09
Source location compuation

### LINKED ISSUE #126: Collect line and code unit column information
To parse Ruby code, Prism uses only byte offset information from the start of the file. We only know how many bytes away from the beginning of the source code any node's position is.

To return results back to the editor, we need to translate the byte offsets into 4 pieces of information: start_line, end_line, start_code_unit_column, end_code_unit_column.

Language servers communicate positions based on code units instead of characters. This is critical because depending on which encoding got negotiated with the editor, the lengths of the code unit values will change.

Ensure that we are collecting all of the information we need taking into consideration the 3 encodings that all language servers have to support: UTF-8, UTF-16 and UTF-32.

**Note**: computing these can be expensive. Check Prism's implementation as inspiration for optimizing the performance of collecting these.

soutaro:
Closes #126 

This PR adds a component which implements the translation between the source location -- line and column -- and the byte offset. This supports three [position encodings](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind) defined in LSP.

It also adds the converter to RubyIndexer so that the positions are accessible during the indexing.

Todos:

* [x] Confirm the casts between `usize` and `u32` makes sense and we can assume they are safe
* [x] Clarify the conditions to return `None`, fail the conversion

### DIFF
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -1,9 +1,9 @@
 use crate::{
     indexing::{
         errors::{IndexingError, MultipleErrors},
-        ruby_indexer::RubyIndexer,
+        ruby_indexer::{RubyIndexer, IndexerParts},
     },
-    model::graph::Graph,
+    model::graph::Graph, source_location::{UTF8SourceLocationConverter},
 };
 use glob::glob;
 use std::sync::{
@@ -61,7 +61,7 @@ impl Document {
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
-    let (tx, rx): (Sender<RubyIndexer>, Receiver<RubyIndexer>) = mpsc::channel();
+    let (tx, rx): (Sender<IndexerParts>, Receiver<IndexerParts>) = mpsc::channel();
     let num_threads = thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4);
     let mut threads = Vec::with_capacity(num_threads);
     let document_queue = Arc::new(Mutex::new(documents));
@@ -72,29 +72,39 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
 
         let handle = thread::spawn(move || {
             while let Some(document) = { queue.lock().unwrap().pop() } {
-                let mut ruby_indexer = RubyIndexer::new(document.uri.to_string());
+                let (source, errors) = {
+                    let mut errors = Vec::new();
 
-                // If the document includes the source, use it directly to index (especially since the content may not
-                // be committed to disk yet). Otherwise, read it from disk
-                if let Some(source) = &document.source {
-                    ruby_indexer.index(source);
+                    let source: String =
+                        if let Some(source) = &document.source {
+                            source.clone()
+                        } else {
+                            fs::read_to_string(document.path()).unwrap_or_else(|e| {
+                                errors.push(IndexingError::FileReadError(format!(
+                                    "Failed to read {}: {}",
+                                    document.path(),
+                                    e
+                                )));
+                                String::new()
+                            })
+                        };
+                    
+                    (source, errors)
+                };
+                
+                let converter = UTF8SourceLocationConverter::new(&source);
+                let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter);
+
+                if errors.is_empty() {
+                    ruby_indexer.index(&source);
                 } else {
-                    match fs::read_to_string(document.path()) {
-                        Ok(source) => {
-                            ruby_indexer.index(&source);
-                        }
-                        Err(e) => {
-                            ruby_indexer.add_error(IndexingError::FileReadError(format!(
-                                "Failed to read {}: {}",
-                                document.path(),
-                                e
-                            )));
-                        }
+                    for error in errors {
+                        ruby_indexer.add_error(error);
                     }
                 }
 
                 thread_tx
-                    .send(ruby_indexer)
+                    .send(ruby_indexer.into_parts())
                     .expect("Receiver end should not be closed until all threads are done");
             }
         });
@@ -107,8 +117,7 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
     // Insert all discovered definitions into the global graph
     let mut all_errors = Vec::new();
 
-    for indexer in rx {
-        let (local_graph, errors) = indexer.into_parts();
+    for (local_graph, errors) in rx {
         graph.update(local_graph);
         all_errors.extend(errors);
     }
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -9,23 +9,27 @@ use crate::model::definitions::{
 use crate::model::graph::Graph;
 use crate::model::ids::{NameId, UriId};
 use crate::offset::Offset;
+use crate::source_location::SourceLocationConverter;
 
 use ruby_prism::Visit;
+
+pub type IndexerParts = (Graph, Vec<IndexingError>);
 
 /// The indexer for the definitions found in the Ruby source code.
 ///
 /// It implements the `Visit` trait from `ruby_prism` to visit the AST and create a hash of definitions that must be
 /// merged into the global state later.
-pub struct RubyIndexer {
+pub struct RubyIndexer<'a> {
     uri_id: UriId,
     local_index: Graph,
     nesting_stacks: Vec<Vec<String>>,
     errors: Vec<IndexingError>,
+    location_converter: &'a dyn SourceLocationConverter,
 }
 
-impl RubyIndexer {
+impl<'a> RubyIndexer<'a> {
     #[must_use]
-    pub fn new(uri: String) -> Self {
+    pub fn new(uri: String, location_converter: &'a dyn SourceLocationConverter) -> Self {
         let mut local_index = Graph::new();
         let uri_id = local_index.add_uri(uri);
 
@@ -34,11 +38,12 @@ impl RubyIndexer {
             nesting_stacks: vec![Vec::new()],
             local_index,
             errors: Vec::new(),
+            location_converter,
         }
     }
 
     #[must_use]
-    pub fn into_parts(self) -> (Graph, Vec<IndexingError>) {
+    pub fn into_parts(self) -> IndexerParts {
         (self.local_index, self.errors)
     }
 
@@ -230,16 +235,20 @@ impl RubyIndexer {
     }
 }
 
-impl Visit<'_> for RubyIndexer {
+impl Visit<'_> for RubyIndexer<'_> {
     fn visit_class_node(&mut self, node: &ruby_prism::ClassNode<'_>) {
         let name = Self::location_to_string(&node.constant_path().location());
-
+        
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
             let name_id = NameId::from(&fully_qualified_name);
+            let offset = Offset::from_prism_location(&node.location());
+            // Uses `location_converter` to evaluate the performance impact of the conversion
+            self.location_converter.offset_to_position(&offset).unwrap();
+
             let definition = Definition::Class(Box::new(ClassDefinition::new(
                 name_id,
                 indexer.uri_id,
-                Offset::from_prism_location(&node.location()),
+                offset,
             )));
 
             indexer.local_index.add_definition(fully_qualified_name, definition);
@@ -255,10 +264,13 @@ impl Visit<'_> for RubyIndexer {
 
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
             let name_id = NameId::from(&fully_qualified_name);
+            let offset = Offset::from_prism_location(&node.location());
+            // Uses `location_converter` to evaluate the performance impact of the conversion
+            self.location_converter.offset_to_position(&offset).unwrap();
             let definition = Definition::Module(Box::new(ModuleDefinition::new(
                 name_id,
                 indexer.uri_id,
-                Offset::from_prism_location(&node.location()),
+                offset,
             )));
             indexer.local_index.add_definition(fully_qualified_name, definition);
 
@@ -274,10 +286,13 @@ impl Visit<'_> for RubyIndexer {
 
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
             let name_id = NameId::from(&fully_qualified_name);
+            let offset = Offset::from_prism_location(&name_loc);
+            // Uses `location_converter` to evaluate the performance impact of the conversion
+            self.location_converter.offset_to_position(&offset).unwrap();
             let definition = Definition::Constant(Box::new(ConstantDefinition::new(
                 name_id,
                 indexer.uri_id,
-                Offset::from_prism_location(&name_loc),
+                offset,
             )));
 
             indexer.local_index.add_definition(fully_qualified_name, definition);
@@ -292,10 +307,13 @@ impl Visit<'_> for RubyIndexer {
 
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
             let name_id = NameId::from(&fully_qualified_name);
+            let offset = Offset::from_prism_location(&location);
+            // Uses `location_converter` to evaluate the performance impact of the conversion
+            self.location_converter.offset_to_position(&offset).unwrap();
             let definition = Definition::Constant(Box::new(ConstantDefinition::new(
                 name_id,
                 indexer.uri_id,
-                Offset::from_prism_location(&location),
+                offset,
             )));
 
             indexer.local_index.add_definition(fully_qualified_name, definition);
@@ -313,10 +331,14 @@ impl Visit<'_> for RubyIndexer {
 
                     self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
                         let name_id = NameId::from(&fully_qualified_name);
+                        let offset = Offset::from_prism_location(&location);
+                        // Uses `location_converter` to evaluate the performance impact of the conversion
+                        self.location_converter.offset_to_position(&offset).unwrap();
+
                         let definition = Definition::Constant(Box::new(ConstantDefinition::new(
                             name_id,
                             indexer.uri_id,
-                            Offset::from_prism_location(&location),
+                            offset,
                         )));
 
                         indexer.local_index.add_definition(fully_qualified_name, definition);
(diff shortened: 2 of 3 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -42,6 +42,10 @@ impl RubyIndexer {
         (self.local_index, self.errors)
     }
 
+    pub fn get_local_index(&self) -> &Graph {
```
soutaro: This is a helper function only for the benchmarking purpose.

rust/index/src/indexing.rs
```diff
@@ -73,26 +73,69 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
-                                document.path(),
-                                e
-                            )));
-                        }
+                    for error in errors {
+                        ruby_indexer.add_error(error);
                     }
                 }
 
+                let converter = UTF16SourceLocationConverter::new(&source);
```
soutaro: I wanted to fix the indexer to use the converter to benchmark, but I couldn't do that with the borrow checker... 😭 

Added a loop over the definitions in the index and compute all of the location of the start/end position of them.

rust/index/src/indexing.rs
```diff
@@ -73,26 +73,69 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
+                            Err(e) => {
+                                errors.push(IndexingError::FileReadError(format!(
+                                    "Failed to read {}: {}",
+                                    document.path(),
+                                    e
+                                )));
+                                String::new()
+                            }
+                        }
+                    };
```
amomchilov: Rust lets you limit the mutability of the `errors` vec by introducing a new scope:

```rust
                let (source, errors) = {
                    let mut errors = Vec::new();

                    let source: String =
                        if let Some(source) = &document.source {
                            source.clone()
                        } else {
                            match fs::read_to_string(document.path()) {
                                Ok(source) => {
                                    source
                                }
                                Err(e) => {
                                    errors.push(IndexingError::FileReadError(format!(
                                        "Failed to read {}: {}",
                                        document.path(),
                                        e
                                    )));
                                    String::new()
                                }
                            }
                        };
                    
                    (source, errors)
                };
```

rust/index/src/indexing.rs
amomchilov: I don't know how to fix this either, but I'd love for someone to teach us! CC @jesse-shopify

rust/index/src/indexing.rs
```diff
@@ -73,26 +73,69 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
                     }
                 }
 
+                let converter = UTF16SourceLocationConverter::new(&source);
+                let local_index = ruby_indexer.get_local_index();
+
+                let mut count = 0usize;
+
+                local_index.definitions().iter().for_each(|def| {
+                    let d = def.1;
```
amomchilov: I think you can destructure in the params:

```suggestion
                local_index.definitions().iter().for_each(|_, d| {
```

rust/index/src/indexing.rs
```diff
@@ -73,26 +73,69 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
+                    let start = d.start();
+                    let end = d.end();
+
+                    let start_pos = converter.byte_offset_to_position(start);
+                    let end_pos = converter.byte_offset_to_position(end);
```
amomchilov: Does each of these calls re-walk the string from the start?

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,596 @@
...
+    text_len: u32,
+
+    /// A vector of the line string slice
+    /// 
+    /// The lines end with '\n' except for the last line.
+    /// If the content ends with a newline, there will be an empty line at the end.
+    line_slices: Vec<&'a str>,
+
+    /// The byte offsets of the start of each line
+    line_byte_offsets: Vec<u32>
```
amomchilov: Do we need to store these separately?

Couldn't we compute these by checking the byte distance between the start of the string, and the end of the slices in `line_slices`?

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,596 @@
...
+    #[must_use]
+    pub fn get_line_slice(&self, line: u32) -> Option<(u32, &str)> {
+        let line_index = line as usize;
+        let line_offset = self.line_byte_offsets.get(line_index)?;
+        let slice = self.line_slices.get(line_index)?;
+        Some((*line_offset, *slice))
+    }
+
+    /// Returns the possible line number of the given byte offset.
+    #[must_use]#[allow(clippy::cast_possible_truncation)]
```
amomchilov:
```suggestion
    #[must_use]
    #[allow(clippy::cast_possible_truncation)]
```

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,596 @@
...
+    pub fn get_line_number(&self, byte_offset: u32) -> Option<u32> {
+        if byte_offset > self.text_len {
+            return None;
+        }
+
+        if self.text_len == 0 && byte_offset == 0 {
```
amomchilov: Does the text length matter in this case?

```suggestion
        if byte_offset == 0 {
```

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,596 @@
...
+        }
+
+        let line_index = match self.line_byte_offsets.binary_search(&byte_offset) {
+            Ok(line_index) => line_index,
+            Err(line_index) => line_index.saturating_sub(1),
+        };
+
+        if line_index >= self.line_byte_offsets.len() {
+            return None;
+        }
```
amomchilov: This should be impossible, no? Should this be an assertion instead?

```suggestion
        assert!(line_index >= self.line_byte_offsets.len())
```

rust/index/src/source_location.rs
soutaro: I was assuming we need this for `binary_search`, but `binary_seach_by_key` looks like the best option to go, without having two vectors for offsets and slices.

rust/index/src/indexing.rs
soutaro: It finds the line with `binary_search` and scans the lines to find the column number.
UTF-8 would run faster because it simply returns the byte offset, without scanning the line.

rust/index/src/indexing/ruby_indexer.rs
soutaro: We no longer need this.

rust/index/src/indexing.rs
soutaro: Sending the result of `into_parts` instead of RubyIndexer itself solved my problem. 🎉

rust/index/src/source_location.rs
soutaro: We should clarify the preconditions and panic conditions; not finished yet.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -230,16 +235,19 @@ impl RubyIndexer {
...
-impl Visit<'_> for RubyIndexer {
+impl Visit<'_> for RubyIndexer<'_> {
     fn visit_class_node(&mut self, node: &ruby_prism::ClassNode<'_>) {
         let name = Self::location_to_string(&node.constant_path().location());
-
+        
         self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
             let name_id = NameId::from(&fully_qualified_name);
+            let offset = Offset::from_prism_location(&node.location());
+            self.location_converter.offset_to_position(&offset).unwrap();
```
soutaro, line 246: You can use the `location_converter` from the RubyIndexer to convert the `offset` to line-column pairs.

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,502 @@
...
+
+    #[allow(clippy::cast_possible_truncation)]
+    fn position_with_line_offset(&self, line: u32, line_offset: u32, line_slice: &str) -> Position {
+        let mut column = 0u32;
+
+        for (byte_pos, ch) in line_slice.char_indices() {
+            if byte_pos as u32 == line_offset {
+                break;
+            }
+            assert!((byte_pos as u32) < line_offset, "The line_offset doesn't look like between characters: {line_offset}, {line_slice}");
```
soutaro: It panics if the `byte_offset` is not at the gap between utf-8 codepoints.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -9,23 +9,27 @@ use crate::model::definitions::{
 use crate::model::graph::Graph;
 use crate::model::ids::{NameId, UriId};
 use crate::offset::Offset;
+use crate::source_location::SourceLocationConverter;
 
 use ruby_prism::Visit;
 
+pub type IndexerParts = (Graph, Vec<IndexingError>);
```
jenny-codes, line 16: Nit: I wonder if it would be more clear to use something like `IndexerResult` as we are using this struct to pass back the result. Subsequently maybe the `into_parts` function can also be `to_result` or something in that direction?

rust/index/src/indexing/ruby_indexer.rs
st0012: 2 questions about this:

- I'm a Rust noob so may read this wrong, but it looks like we're not using the return value of `offset_to_position`. Does that mean we're relying on its side effect to `offset`?
- It looks like it always go hand in hand with `let offset = Offset::from_prism_location(&node.location());`. Should we have a helper function to wrap them together?

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        let line_index = match self.lines.binary_search_by_key(&byte_offset, |(offset, _slice, _len)| *offset) {
+            Ok(line_index) => line_index,
+            Err(line_index) => line_index.saturating_sub(1),
+        };
+
+        Some(line_index as u32)
+    }
+
+    fn get_line_slice_from_byte_offset(&self, byte_offset: u32) -> Option<(u32, u32, &str, u32)> {
```
jenny-codes: Suggestion: Maybe we can use a helper struct to make the current `(u32, u32, &str, u32)` more readable

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -453,13 +487,15 @@ impl Visit<'_> for RubyIndexer {
                     self.with_updated_nesting(&name, |indexer, fully_qualified_name| {
                         let writer_name = format!("{fully_qualified_name}=");
                         let name_id = NameId::from(&writer_name);
+                        let offset = Offset::from_prism_location(&location);
+                        self.location_converter.offset_to_position(&offset).unwrap();
```
jenny-codes, line 503: Can you explain what this does? We are calling the offset_to_position function but not using its return value?

soutaro: I'm sorry for the confusion. It's a demo and for benchmarking. 🙇‍♂️ It discards the result, but you would write code using the values.

rust/index/src/indexing/ruby_indexer.rs
soutaro: * It's not using the value because it's a demo and for benchmarking. 🙇‍♂️ You usually use the result of the location conversion.
* Adding a utility function would make sense. I'm assuming it will be done by the user of the converter.

rust/index/src/indexing/ruby_indexer.rs
soutaro: I'm open to renaming the type to something else, but let me share some of my thoughts behind the current names.

I'm assuming that the `into_parts` function name is a common pattern in Rust, as it moves and deconstructs. Therefore, I'm not sure if renaming it to `to_result` makes a lot of sense, and the same goes for the `IndexerResult` type name.

rust/index/src/indexing/ruby_indexer.rs
amomchilov: I'm surprised that didn't give an error, because `fn offset_to_position` is marked with `#[must_use]`.

I would have expected it to require an explicit discard like `_ = ...`

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -9,23 +9,27 @@ use crate::model::definitions::{
...
 ///
 /// It implements the `Visit` trait from `ruby_prism` to visit the AST and create a hash of definitions that must be
 /// merged into the global state later.
-pub struct RubyIndexer {
+pub struct RubyIndexer<'a> {
     uri_id: UriId,
     local_index: Graph,
     nesting_stacks: Vec<Vec<String>>,
     errors: Vec<IndexingError>,
+    location_converter: &'a dyn SourceLocationConverter,
```
amomchilov, line 27: I see why you used `dyn` here (to support any of the 3 implementations for UTF-8/16/32), but it makes all the method calls require dynamic dispatch.

I wonder if instead the RubyIndex should be generic over its `SourceLocationConverter`, so that it can monomorphise and use static dispatch to a concrete `SourceLocationConverter` implementation.

@jesse-shopify WDYT?

In any case, this is just a performance consideration that shouldn't block this Pr. This is good as-is, and we can iterate in a follow-up PR.

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,502 @@
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct Position {
+    pub line: u32,
+    pub column: u32,
+}
+
+pub trait SourceLocationConverter {
+    /// Converts a byte offset to a Position (line and column).
+    /// 
+    /// Note that the ``byte_offset`` is the number based on an UTF-8 encoded source code.
```
amomchilov, line 10: > Note that the ``byte_offset`` is the number based on an UTF-8 encoded source code.

Is this still true, given the 3 different UTF-8/16/32 impls below?

rust/index/src/source_location.rs
amomchilov: Is this more clear?

```suggestion
            assert!((byte_pos as u32) < line_offset, "The line_offset doesn't look like it's on a boundary between characters: {line_offset}, {line_slice}");
```

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,502 @@
...
+
+    #[allow(clippy::cast_possible_truncation)]
+    fn position_with_line_offset(&self, line: u32, line_offset: u32, line_slice: &str) -> Position {
+        let mut column = 0u32;
+
+        for (byte_pos, _ch) in line_slice.char_indices() {
+            if byte_pos as u32 == line_offset {
+                break;
+            }
+            assert!((byte_pos as u32) < line_offset, "The line_offset doesn't look like between characters: {line_offset}, {line_slice}");
```
amomchilov:
```suggestion
            assert!((byte_pos as u32) < line_offset, "The line_offset doesn't look like it's on a boundary between characters: {line_offset}, {line_slice}");
```

rust/index/src/source_location.rs
```diff
@@ -0,0 +1,502 @@
...
+        Some(self.position_with_line_offset(line, byte_offset - offset, slice))
+    }
+
+    /// Converts a pair of byte offsets (start and end) to a pair of Positions.
+    /// 
+    /// If both offsets are valid, returns ``Some((start_position, end_position))``.
+    /// If either offset is invalid, returns ``None``.
+    /// 
+    /// There are currently no optimization for this implementation.
+    /// It simply translates each offset individually.
```
amomchilov, line 25: Could you please open an issue for this?

And to confirm my understanding: currently, each `offset_to_position` call ends up redoing the binary search, but it would be better to accumulate every offset we care about, and do a single walk through the source text to convert them all in bulk. Is that right?

rust/index/src/indexing/ruby_indexer.rs
st0012: > It's not using the value because it's a demo and for benchmarking.

Can you add comments on all the code that serve this purpose?

rust/index/src/indexing.rs
```diff
@@ -72,29 +72,44 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
-                if let Some(source) = &document.source {
-                    ruby_indexer.index(source);
+                let (source, errors) = {
+                    let mut errors = Vec::new();
+
+                    let source: String =
+                        if let Some(source) = &document.source {
+                            source.clone()
+                        } else {
+                            match fs::read_to_string(document.path()) {
```
jesse-shopify: This looks like it could be replaced with `map_err`

rust/index/src/indexing/ruby_indexer.rs
jesse-shopify: This is a good thought but may add some confusion in the future. In Rust it is common to have something like `OperationResult` types. In the context of Rust this is assumed to be either success or failure but never both. In the case of parsing and AST assembly it's common to want a best effort success and a list of warnings/errors which does not fit the `Result` idiom.

rust/index/src/indexing/ruby_indexer.rs
jesse-shopify: This is a non-trivial question. I'd say this is good for now, but definitely something to consider in the future. The guideline that has worked for me is to not introduce Generics until they are really needed. They are needed for something like a container data structure but after that they are better left until the performance implications are better understood. A recent post that covers this:

https://corrode.dev/blog/simple/

rust/index/src/indexing/ruby_indexer.rs
jesse-shopify: In Rust, and any other LLVM based compiler, calls to functions with no side-effects and unused return values will be eliminated in optimization. You can use `std::hint::black_box` for benchmarking unused values, and this is best constrained to prototypes. See: https://doc.rust-lang.org/beta/std/hint/fn.black_box.html

rust/index/src/source_location.rs
jesse-shopify: Would this be an invalid encoding?

rust/index/src/indexing/ruby_indexer.rs
st0012: @jesse-shopify  TIL, thanks

rust/index/src/indexing/ruby_indexer.rs
alexcrocha: That was a great read! Some good gems there

rust/index/src/source_location.rs
soutaro: Yes. This is correct. The byte-offsets we want to convert are calculated by Prism based on the UTF-8 encoded `str`.

The UTF-8/16/32 encodings are just for the meaning of the *column* number. We have one byte-offset, and the number can be converted to three line-column pairs.

(I also got confused on this and wrote a broken implementation.)

rust/index/src/indexing/ruby_indexer.rs
soutaro: The value of `offset_to_position` is used for `unwrap`, which cannot be eliminated by the compiler since it may panic.

Changing the implementation of `offset_to_position`, which may return `None` sometimes, resulted in different output, and I'm assuming it's not eliminated by the compiler.

```rust
        // Give invalid byte offset to return `None` from `offset_to_position` function
        let end = self.byte_offset_to_position(offset.end()+10)?;
```

Output with the modification above. It failed index the files -- looks like it panicked.

```
Indexed 10000 files
Found 0 names
Found 0 definitions
Found 0 URIs
```

/cc @jesse-shopify

rust/index/src/source_location.rs
soutaro: It's when the byte_offset is middle of the code-point in UTF-8.

For example, some Japanese characters like `シ` consist of three bytes. If we pass `byte_pos = 2` for this line, this assertion is triggered.

rust/index/src/source_location.rs
soutaro: Yes, I mean that optimization.
Typically we need both start and end positions of things, so if we can convert them simultaneously, the conversion could be up to twice(?) as fast.

PR #169, opened 2025-09-09
Parse comments during indexing time

jenny-codes:
For https://github.com/Shopify/index/issues/86

I've explored two approaches for comment parsing:

Approach 1 (this one) - Preprocessing: Parse and store comments in the database during indexing

- Trade-offs: +5% peak memory, +3% indexing time (compared to main)
- Benefit: Fast retrieval performance during queries

Approach 2 - On-demand (#169): Parse comments only when requested (e.g., hover requests)

- Trade-offs: No additional memory overhead
- Benefit: Zero indexing cost
- Drawback: 4.2x slower query performance compared to preprocessing

I am leaning towards the preprocessing approach for the following reasons:

- The memory/indexing overhead is minimal (5%/3% respectively)
- Query performance is critical for developer experience - 4.2x slower responses would be noticeable
- Comments are frequently accessed features (hover, documentation, etc.)
- The upfront cost during indexing pays dividends throughout the development workflow

The benchmark data suggests preprocessing offers the better user experience trade-off. Thoughts on this approach?

### Main

Initialization:    0.459s (  2.2%)
Indexing:          9.428s ( 45.4%)
Querying:          0.000s (  0.0%)
Cleanup:          10.889s ( 52.4%)
Total:            20.777s

Maximum Resident Set Size: 7508230144 bytes (7160.40 MB)
Peak Memory Footprint:     7854135296 bytes (7490.28 MB)
Execution Time:            21.07 seconds

### On-demand parsing

Initialization:    0.441s (  0.9%)
Indexing:          9.816s ( 20.0%)
Querying:         27.589s ( 56.3%)
Cleanup:          11.121s ( 22.7%)
Total:            48.967s

Maximum Resident Set Size: 6754041856 bytes (6441.15 MB)
Peak Memory Footprint:     7835867136 bytes (7472.86 MB)
Execution Time:            49.28 seconds

### Pre-process parsing

Initialization:    0.343s (  1.2%)
Indexing:          9.733s ( 34.7%)
Querying:          6.550s ( 23.4%)
Cleanup:          11.384s ( 40.6%)
Total:            28.010s

Maximum Resident Set Size: 7299235840 bytes (6961.09 MB)
Peak Memory Footprint:     8322078848 bytes (7936.55 MB)
Execution Time:            28.26 seconds

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -124,6 +124,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytecount"
+version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "175812e0be2bccb6abe50bb8d566126198344f707e304f45c648fd8f2cc0365e"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -479,6 +485,7 @@ name = "index"
 version = "0.1.0"
 dependencies = [
  "assert_cmd",
+ "bytecount",
  "clap",
  "glob",
  "predicates",
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -11,9 +11,25 @@ use crate::model::ids::{NameId, UriId};
 use crate::offset::Offset;
 use crate::source_location::SourceLocationConverter;
 
-use ruby_prism::Visit;
+use ruby_prism::{ParseResult, Visit};
 
 pub type IndexerParts = (Graph, Vec<IndexingError>);
+
+const MAGIC_AND_RBS_COMMENT_PREFIX: &[&str] = &[
+    "frozen_string_literal:",
+    "typed:",
+    "compiled:",
+    "encoding:",
+    "shareable_constant_value:",
+    "warn_indent:",
+    "rubocop:",
+    "nodoc:",
+    "doc:",
+    "coding:",
+    "warn_past_scope:",
+    "#:",
+    "#|",
+];
 
 /// The indexer for the definitions found in the Ruby source code.
 ///
@@ -25,11 +41,13 @@ pub struct RubyIndexer<'a> {
     nesting_stacks: Vec<Vec<String>>,
     errors: Vec<IndexingError>,
     location_converter: &'a dyn SourceLocationConverter,
+    comments: Vec<CommentGroup>,
+    source: &'a str,
 }
 
 impl<'a> RubyIndexer<'a> {
     #[must_use]
-    pub fn new(uri: String, location_converter: &'a dyn SourceLocationConverter) -> Self {
+    pub fn new(uri: String, location_converter: &'a dyn SourceLocationConverter, source: &'a str) -> Self {
         let mut local_index = Graph::new();
         let uri_id = local_index.add_uri(uri);
 
@@ -39,6 +57,8 @@ impl<'a> RubyIndexer<'a> {
             local_index,
             errors: Vec::new(),
             location_converter,
+            comments: Vec::new(),
+            source,
         }
     }
 
@@ -51,13 +71,69 @@ impl<'a> RubyIndexer<'a> {
         self.errors.push(error);
     }
 
-    pub fn index(&mut self, source: &str) {
-        let result = ruby_prism::parse(source.as_ref());
+    pub fn index(&mut self) {
+        let result = ruby_prism::parse(self.source.as_bytes());
+        self.comments = self.parse_comments_into_groups(&result);
         self.visit(&result.node());
+    }
+
+    fn parse_comments_into_groups(&mut self, result: &ParseResult<'_>) -> Vec<CommentGroup> {
+        let mut iter = result.comments().peekable();
+        let mut groups = Vec::new();
+
+        while let Some(comment) = iter.next() {
+            let mut group = CommentGroup::new();
+            group.add_comment(&comment);
+            while let Some(next_comment) = iter.peek() {
+                if group.accepts(next_comment, self.source) {
+                    let next = iter.next().unwrap();
+                    group.add_comment(&next);
+                } else {
+                    break;
+                }
+            }
+            groups.push(group);
+        }
+        groups
     }
 
     fn location_to_string(location: &ruby_prism::Location) -> String {
         String::from_utf8_lossy(location.as_slice()).to_string()
+    }
+
+    fn find_comments_for(&self, offset: u32) -> Option<String> {
+        let offset_usize = offset as usize;
+        if self.comments.is_empty() {
+            return None;
+        }
+
+        let idx = match self.comments.binary_search_by_key(&offset_usize, |g| g.end_offset) {
+            Ok(_) => {
+                // This should never happen in valid Ruby syntax - a comment cannot end exactly
+                // where a definition begins (there must be at least a newline between them)
+                debug_assert!(false, "Comment ends exactly at definition start - this indicates a bug");
+                return None;
+            }
+            Err(i) if i > 0 => i - 1,
+            Err(_) => return None,
+        };
+
+        let group = &self.comments[idx];
+        let between = &self.source.as_bytes()[group.end_offset..offset_usize];
+        if !between.iter().all(|&b| b.is_ascii_whitespace()) {
+            return None;
+        }
+
+        // We allow at most one blank line between the comment and the definition
+        if bytecount::count(between, b'\n') > 2 {
+            return None;
+        }
+
+        if group.comments.is_empty() {
+            None
+        } else {
+            Some(group.comments.clone())
+        }
     }
 
     fn collect_parameters(node: &ruby_prism::DefNode) -> Vec<Parameter> {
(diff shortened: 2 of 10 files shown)


rust/Cargo.lock
```diff
@@ -123,6 +123,12 @@ dependencies = [
  "serde",
 ]
 
+[[package]]
+name = "bytecount"
```
alexcrocha, line 127: Are we able to get away without introducing a new dependency? Or does `bytecount` help with performance?

jenny-codes: Clippy suggested (warned) to use this instead of natively counting the bytes so I took its word for it :) 

And after taking a deeper look, it looks like it does help with performance (with this algorithm so named `hyperscreamingcount`) so in both ways adding this package seems justifiable

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -230,16 +306,67 @@ impl RubyIndexer {
...
+            comments: String::new(),
+        }
+    }
+
+    // Accepts the next line if it is continuous
+    fn accepts(&self, next: &ruby_prism::Comment, source: &[u8]) -> bool {
+        let current_end_offset = self.end_offset;
+        let next_line_start_offset = next.location().start_offset();
+
+        if current_end_offset >= next_line_start_offset {
```
alexcrocha: How would this happen? Can you provide an example?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -230,16 +306,67 @@ impl RubyIndexer {
...
+            return false;
+        }
+
+        let between = &source[current_end_offset..next_line_start_offset];
+        if !between.iter().all(|&b| b.is_ascii_whitespace()) {
+            return false;
+        }
+
+        bytecount::count(between, b'\n') <= 1
```
alexcrocha, line 340: Can we have a comment explaining what the between calculation is for?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -230,16 +306,67 @@ impl RubyIndexer {
...
+    }
+
+    // For the magic comments, what we want to do is the following:
+    // 1. still move the group end offset to the end of the magic comment
+    // 2. not add the comment to the comments array
+    fn add_comment(&mut self, comment: &ruby_prism::Comment) {
```
alexcrocha, line 346: Thinking more about this, I wonder if we should not be filtering out magic comments and rbs comments here.

The Ruby LSP does not care about those types of comments for documentation on hover, but what if another client wants that information?
In that case, should the responsibility of implementing the filtering logic be with the Ruby LSP? Or, we could provide another API for that specific case.

It would be good to see a benchmark comparison without the pre-filtering work.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -25,6 +41,8 @@ pub struct RubyIndexer<'a> {
     nesting_stacks: Vec<Vec<String>>,
     errors: Vec<IndexingError>,
     location_converter: &'a dyn SourceLocationConverter,
+    comments: Vec<CommentGroup>,
+    source: Vec<u8>,
```
soutaro: I think using `String` for `source` text is more straightforward.
Some operations like `is_ascii_whitespace` can now run on UTF-8 code points, and looks itself safe. (I confirmed if UTF-8 second/third/fourth byte can be `is_ascii_whitespace`. => No, the byte based code is safe.)

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -39,6 +57,8 @@ impl<'a> RubyIndexer<'a> {
             local_index,
             errors: Vec::new(),
             location_converter,
+            comments: Vec::new(),
+            source: Vec::new(),
```
soutaro: I don't want this empty field here—it’s only updated once in the index function, and its content can be initialized in the new function, as we already do with `location_converter`.
The main issue is having mutable and immutable parts in the same struct. [The builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) might be a solution, but I'm not sure if we should rewrite this right now.

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: This shouldn't happen. I have removed it for simplicity

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Good call. Updated

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: As we have discussed offline, I am leaning towards keeping this for now because if nothing else it matches the lsp indexer's behavior. It is an easily reversible decision though so we can revisit it later, unless anyone has other opinions on this

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Updated. Thanks!

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Updated. Now we store the source in the struct during initialization instead of inside the index function

rust/index/src/model/definitions.rs
```diff
@@ -138,15 +154,17 @@ pub struct ClassDefinition {
     name_id: NameId,
     uri_id: UriId,
     offset: Offset,
+    comments: String,
```
Morriar, line 157: The comment should be `Optional<String>`. We shouldn't have to pass an empty string to initialize a definition without a comment.

Also, how do you differentiate an empty comment vs. no comment:

```rb
#
class Foo; end

class Bar; end
```

Morriar: Since we also do the work of figuring the new lines, can we split the string into an array of comment for each line?

So for this:

```rb
# comment1
# comment2
def foo; end
```

The definition of `foo` would have an array of two comments:

```rb
foo.comments => ["comment 1", "comment 2"]
```

Morriar: I also wonder if we should make comment a struct so we can also store the comment location.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -235,20 +311,69 @@ impl<'a> RubyIndexer<'a> {
...
+        bytecount::count(between, b'\n') <= 1
+    }
+
+    // For the magic comments, what we want to do is the following:
+    // 1. still move the group end offset to the end of the magic comment
+    // 2. not add the comment to the comments array
+    fn add_comment(&mut self, comment: &ruby_prism::Comment) {
+        self.end_offset = comment.location().end_offset();
+        let text = String::from_utf8_lossy(comment.location().as_slice()).to_string();
+        if MAGIC_AND_RBS_COMMENT_PREFIX.iter().any(|&prefix| text.contains(prefix)) {
```
Morriar, line 349: We shouldn't filter the comments at this stage. It makes it impossible for the client to access the comments that where ignored.

The indexer should keep the list of all comments, the client is then responsible of filtering what makes sense for its purpose.

rust/index/src/model/graph.rs
```diff
@@ -73,6 +73,22 @@ impl Graph {
         )
     }
 
+    #[must_use]
+    pub fn get_documentation(&self, name: &str) -> Option<String> {
```
vinistock, line 77: This should ideally be returned as part of the `get` API. We don't need a specialized API just for documentation - since we always show a combination of all of the declaration's information on hover.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -51,15 +71,71 @@ impl<'a> RubyIndexer<'a> {
...
+        };
+
+        let group = &self.comments[idx];
+        let between = &self.source.as_bytes()[group.end_offset..offset_usize];
+        if !between.iter().all(|&b| b.is_ascii_whitespace()) {
+            return None;
+        }
+
+        // We allow at most one blank line between the comment and the definition
+        if bytecount::count(between, b'\n') > 2 {
```
vinistock, line 128: What does this dependency do? What's the different between this and simply getting the string's byte length directly?

rust/index/src/model/definitions.rs
jenny-codes: What would be the use case of having locations with the comments?

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Clippy suggested (warned) to use this instead of natively counting the bytes. 

Looks like on top of it being a more solid way to do it, it helps with performance (with this algorithm so named hyperscreamingcount) too

rust/index/src/model/graph.rs
jenny-codes: I see. I will update this.

rust/index/src/model/definitions.rs
jenny-codes: > can we split the string into an array of comment for each line?

I can update the code use an array instead of string for to give the client more control over the comments

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Ok. I added it to match with the lsp indexer's behavior. I can remove this.

rust/index/src/model/graph.rs
jenny-codes: https://github.com/Shopify/index/pull/194

PR #170, opened 2025-09-12
Include phase breakdown in benchmark

jenny-codes:
This allows us to benchmark on a finer level, with a breakdown among
the setting up, indexing, querying, and cleaning up phases

An example output after running `utils/bench`

```
PERFORMANCE BREAKDOWN
Initialization:    0.400s (  1.2%)
Indexing:         10.308s ( 29.7%)
Querying:         10.686s ( 30.8%)
Cleanup:          13.319s ( 38.4%)
Total:            34.713s

Maximum Resident Set Size: 7002750976 bytes (6678.34 MB)
Peak Memory Footprint:     8339200192 bytes (7952.88 MB)
Execution Time:            35.17 seconds

Maximum Resident Set Size: 7255179264 bytes (6919.07 MB)
Peak Memory Footprint:     8295700480 bytes (7911.39 MB)
Execution Time:            29.92 seconds
```

### DIFF
--- a/rust/index/src/bin/bench.rs
+++ b/rust/index/src/bin/bench.rs
@@ -1,0 +1,89 @@
+use std::error::Error;
+use std::time::{Duration, Instant};
+
+use clap::Parser;
+
+use index::{
+    indexing::{self, errors::MultipleErrors},
+    model::graph::Graph,
+};
+
+#[derive(Parser, Debug)]
+#[command(name = "bench", about = "Benchmark the Ruby indexer", version)]
+struct Args {
+    #[arg(value_name = "CORPUS_PATH")]
+    corpus_path: String,
+}
+
+fn time_it<T, F>(f: F) -> (T, Duration)
+where
+    F: FnOnce() -> T,
+{
+    let start = Instant::now();
+    let result = f();
+    let duration = start.elapsed();
+    (result, duration)
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let args = Args::parse();
+    let corpus_path = std::path::Path::new(&args.corpus_path);
+
+    let total_start = Instant::now();
+    let setup_duration;
+    let indexing_duration;
+
+    {
+        let ((mut graph, documents, errors), setup_dur) = time_it(|| {
+            let mut graph = Graph::new();
+            graph
+                .set_configuration(format!("{}/graph.db", corpus_path.display()))
+                .unwrap();
+            let (documents, errors) = indexing::collect_documents(vec![corpus_path.to_string_lossy().to_string()]);
+            (graph, documents, errors)
+        });
+        setup_duration = setup_dur;
+
+        if !errors.is_empty() {
+            return Err(Box::new(MultipleErrors(errors)));
+        }
+
+        let ((), indexing_dur) = time_it(|| {
+            indexing::index_in_parallel(&mut graph, documents).unwrap();
+        });
+        indexing_duration = indexing_dur;
+    }
+
+    let total_duration = total_start.elapsed();
+    let cleanup_duration = total_duration - (setup_duration + indexing_duration);
+
+    println!();
+    println!("PERFORMANCE BREAKDOWN");
+    println!();
+
+    let total_time = total_duration.as_secs_f64();
+    let setup_time = setup_duration.as_secs_f64();
+    let indexing_time = indexing_duration.as_secs_f64();
+    let cleanup_time = cleanup_duration.as_secs_f64();
+
+    println!(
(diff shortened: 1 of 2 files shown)


rust/index/src/bench.rs
alexcrocha, line 1: Let's move this bench file into `src/bin`. 

This way we can also remove the bin config in `Cargo.toml` as Cargo automatically discovers binaries in `src/bin`. See more [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries)

rust/index/src/bench.rs
```diff
@@ -0,0 +1,103 @@
...
+    let setup_duration;
+    let indexing_duration;
+    let querying_duration;
+
+    {
+        let ((mut graph, documents, errors), setup_dur) = time_it(|| {
+            let mut graph = Graph::new();
+            graph
+                .set_configuration(format!("{}/graph.db", corpus_path.display()))
+                .unwrap();
```
alexcrocha: It would be nice to propagate the error instead of panicking here.
```suggestion
            graph.set_configuration(format!("{}/graph.db", corpus_path.display()))?;
```
Although, I see this is inside of a closure so it might be a bit trickier. Not blocking, we can fix forward.

rust/index/src/bench.rs
```diff
@@ -0,0 +1,103 @@
...
+            (graph, documents, errors)
+        });
+        setup_duration = setup_dur;
+
+        if !errors.is_empty() {
+            return Err(Box::new(MultipleErrors(errors)));
+        }
+
+        let ((), indexing_dur) = time_it(|| {
+            indexing::index_in_parallel(&mut graph, documents).unwrap();
```
alexcrocha:
```suggestion
            indexing::index_in_parallel(&mut graph, documents)?;
```
Same as above. Not blocking, we can fix forward.

rust/index/Cargo.toml
```diff
@@ -8,6 +8,10 @@ rust-version = "1.89.0"
 name = "index_cli"
 path = "src/main.rs"
 
+[[bin]]
+name = "bench"
+path = "src/bench.rs"
+
```
alexcrocha:
```suggestion
```
Can be removed once we move `src/bench.rs` to `src/bin/bench.rs` as it becomes redundant.

rust/index/src/bench.rs
jenny-codes: Thanks for the tip! Updated

rust/index/src/bench.rs
jenny-codes: For the purpose of benchmarking I think we want to fail fast when it does not work. Also using unwrap makes the control flow simpler, so I'm leaning towards keeping them as is. What do you say?

rust/index/src/bin/bench.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        let ((), indexing_dur) = time_it(|| {
+            indexing::index_in_parallel(&mut graph, documents).unwrap();
+        });
+        indexing_duration = indexing_dur;
+
+        let ((), querying_dur) = time_it(|| {
+            println!("Found {} names", graph.declarations().len());
+            println!("Found {} definitions", graph.definitions().len());
+            println!("Found {} URIs", graph.documents().len());
```
st0012: Do these functions actually perform "queries" in a way we want to measure? I think they just return the attributes of the graph?
If that's the case, can we use other query API for this measurement. Or if we don't have one now, let's just not including query duration yet and have a comment stating the reason.

jenny-codes: 👍🏻 I have taken this querying phase out in this pr

rust/index/src/bin/bench.rs
```diff
@@ -0,0 +1,89 @@
...
+            graph
+                .set_configuration(format!("{}/graph.db", corpus_path.display()))
+                .unwrap();
+            let (documents, errors) = indexing::collect_documents(vec![corpus_path.to_string_lossy().to_string()]);
+            (graph, documents, errors)
+        });
+        setup_duration = setup_dur;
+
+        if !errors.is_empty() {
+            return Err(Box::new(MultipleErrors(errors)));
```
st0012, line 48: For my own learning: I assume this and `fn main() -> Result<(), Box<dyn Error>>` means that the errors would be displayed by the executable?

jenny-codes: Yes. When main returns an `Err`, Rust's runtime automatically prints the error using `Debug` formatting and exits with a non-zero status code.

Rust book reference: https://doc.rust-lang.org/stable/book/ch09-02-recoverable-errors-with-result.html

PR #173, opened 2025-09-19
Sync with DB state during indexing

jenny-codes:
> Update this is now closed in favor of https://github.com/Shopify/saturn/compare/main.../jennyshih/save-updated-graph

The current indexing flow, when we call the indexing entrypoint, does not interact with the db. We have the api ready but it's not hooked into the indexing flow yet.

Current flow:

```
indexing starts 
-> collect documents from workspace 
-> index 
-> build global graph in memory
```

Updated flow:

```
indexing starts 
-> collect documents from workspace 
-> *calculate diff against db state*
-> *delete stale entries from db* 
-> index
-> build global graph in memory 
-> *save graph to db*
```

To be able to know if the file has changed (and thus needs re-indexing) we will introduce a `content_hash` to the `documents` table in the db. Then during the diff checking stage we will categories them into the following categories

- new: incoming documents don't exist in the db yet
- deleted: db documents that are not present in the incoming documents
- existing: documents that exist in both the incoming and the db, and their content hashes are the same
- updated: documents that exist in both the incoming and the db, and their content hashes are not the same

From here we produce two sets:

- `new_uris`​: It's the union of (new + updated). We pass them on to be index and then save them to both memory and db
- `stale_uris`​: It's the union of (deleted + updated). We clean those up from the db as well as the memory

When we partially delete/create entries we shouldn't need to worry about cascading updates to entries from documents that remain unchanged, because we have not implemented more complex relationships in the graph yet (i.e., inheritance).

### DIFF
--- a/rakelib/index_top_gems.rake
+++ b/rakelib/index_top_gems.rake
@@ -51,6 +51,7 @@ task index_top_gems: :compile_release do
 
             # Index the gem's files and yield errors back to the main Ractor
             graph = Index::Graph.new
+            graph.set_configuration(File.join(@gems_dir, "#{gem}.db"))
             error = graph.index_all(Dir.glob("#{gem_dir}/**/*.rb"))
             next unless error
 
--- a/rust/index-sys/src/index_api.rs
+++ b/rust/index-sys/src/index_api.rs
@@ -56,7 +56,7 @@ pub unsafe extern "C" fn idx_index_all_c(
     all_errors.extend(document_errors);
 
     with_graph(pointer, |graph| {
-        if let Err(errors) = indexing::index_in_parallel(graph, documents) {
+        if let Err(errors) = indexing::index_and_sync(graph, documents) {
             all_errors.extend(errors.0);
         }
 
--- a/rust/index/src/db/schema.sql
+++ b/rust/index/src/db/schema.sql
@@ -4,7 +4,8 @@
 -- Table for storing documents
 CREATE TABLE IF NOT EXISTS documents (
     id INTEGER PRIMARY KEY,  -- Blake3 hash converted to hex
-    uri TEXT NOT NULL UNIQUE
+    uri TEXT NOT NULL UNIQUE,
+    content_hash INTEGER NOT NULL
 );
 
 -- Table for storing code declarations (classes, modules, methods, etc.)
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -11,17 +11,18 @@ use std::sync::{
     Arc, Mutex,
     mpsc::{Receiver, Sender},
 };
-use std::{fs, path::Path};
+use std::{collections::HashMap, collections::HashSet, fs, path::Path};
 use std::{sync::mpsc, thread};
 use url::Url;
+use xxhash_rust::xxh3::xxh3_64;
 pub mod errors;
 pub mod ruby_indexer;
 
-// Represents a document to be indexed. If `source` is `Some(String)`, we're indexing a file state that's not committed
-// to disk yet and should then use the given `source` instead of reading from disk
+#[derive(Clone)]
 pub struct Document {
     uri: Url,
-    source: Option<String>,
+    source: String,
+    content_hash: i64,
 }
 
 impl Document {
@@ -29,9 +30,19 @@ impl Document {
     ///
     /// Creating a new document fails on invalid URIs
     pub fn new(uri: &str, source: Option<String>) -> Result<Self, IndexingError> {
+        let uri = Url::parse(uri).map_err(|e| IndexingError::InvalidUri(e.to_string()))?;
+
+        // Try to get the file content from the source first. If not try to read from the disk
+        let file_content = source.map_or_else(
+            || fs::read_to_string(uri.path()).map_err(|e| IndexingError::FileReadError(e.to_string())),
+            Ok,
+        )?;
+        let content_hash = xxh3_64(file_content.as_bytes()).cast_signed();
+
         Ok(Self {
-            uri: Url::parse(uri).map_err(|e| IndexingError::InvalidUri(format!("Failed to parse URI '{uri}': {e}")))?,
-            source,
+            uri,
+            source: file_content,
+            content_hash,
         })
     }
 
@@ -39,16 +50,26 @@ impl Document {
     ///
     /// Errors if the file path cannot be turned into a URI
     pub fn from_file_path(path: &str) -> Result<Self, IndexingError> {
+        let uri = Url::from_file_path(path)
+            .map_err(|_e| IndexingError::InvalidUri(format!("Couldn't build URI from path '{path}'")))?;
+        let file_content = fs::read_to_string(path).map_err(|e| IndexingError::FileReadError(e.to_string()))?;
+        let content_hash = xxh3_64(file_content.as_bytes()).cast_signed();
+
         Ok(Self {
-            uri: Url::from_file_path(path)
-                .map_err(|_e| IndexingError::InvalidUri(format!("Couldn't build URI from path '{path}'")))?,
-            source: None,
+            uri,
+            source: file_content,
+            content_hash,
         })
     }
 
     #[must_use]
     pub fn path(&self) -> &str {
         self.uri.path()
+    }
+
+    #[must_use]
+    pub fn content_hash(&self) -> i64 {
+        self.content_hash
     }
 }
 
@@ -73,35 +94,9 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
 
         let handle = thread::spawn(move || {
             while let Some(document) = { queue.lock().unwrap().pop() } {
-                let (source, errors) = {
-                    let mut errors = Vec::new();
-
-                    let source: String = if let Some(source) = &document.source {
-                        source.clone()
-                    } else {
-                        fs::read_to_string(document.path()).unwrap_or_else(|e| {
-                            errors.push(IndexingError::FileReadError(format!(
-                                "Failed to read {}: {}",
-                                document.path(),
-                                e
-                            )));
-                            String::new()
-                        })
-                    };
-
-                    (source, errors)
-                };
-
-                let converter = UTF8SourceLocationConverter::new(&source);
-                let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter, &source);
-
-                if errors.is_empty() {
-                    ruby_indexer.index();
-                } else {
-                    for error in errors {
-                        ruby_indexer.add_error(error);
-                    }
-                }
+                let converter = UTF8SourceLocationConverter::new(&document.source);
+                let mut ruby_indexer = RubyIndexer::new(&document, &converter);
+                ruby_indexer.index();
 
                 thread_tx
                     .send(ruby_indexer.into_parts())
@@ -177,11 +172,77 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
     (documents, errors)
 }
 
+/// Indexes the given documents and syncs them with the database.
+///
+/// This function performs a complete indexing workflow:
+/// 1. Syncs the documents with the database to determine what needs to be updated
+/// 2. Indexes the documents that require processing in parallel
+/// 3. Saves the updated graph back to the database
+///
+/// # Errors
+///
+/// Returns `MultipleErrors` if any of the following operations fail:
+/// - Database synchronization operations
+/// - Parallel indexing of documents
+/// - Saving the graph to the database
+pub fn index_and_sync(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
+    let db_documents = graph.db.all_documents_hashes_by_uri()?;
+    let (new_uris, stale_uris) = calculate_diff(&documents, &db_documents);
+    stale_uris.iter().try_for_each(|uri| graph.delete_uri(uri))?;
+    let documents_to_index = documents
+        .into_iter()
+        .filter(|doc| new_uris.contains(doc.uri.as_str()))
+        .collect();
+
+    index_in_parallel(graph, documents_to_index)?;
+    graph.save_to_database()?;
+    Ok(())
+}
+
+/// Calculates which documents need to be indexed and which need to be removed.
+///
+/// Returns `(new_uris, stale_uris)` where:
+/// - `new_uris`: Documents to index (includes new and updated documents)
+/// - `stale_uris`: Documents to delete from graph/DB (includes deleted and updated documents)
+///
+/// Updated documents appear in both sets to ensure clean updates: first the old data
+/// is completely removed, then the document is re-indexed fresh with new data.
+fn calculate_diff(
+    incoming_documents: &[Document],
+    db_documents: &HashMap<String, i64>,
+) -> (HashSet<String>, HashSet<String>) {
+    let incoming_docs_lookup: HashMap<&str, i64> = incoming_documents
+        .iter()
+        .map(|doc| (doc.uri.as_str(), doc.content_hash))
+        .collect();
+
+    let db_uris: HashSet<&str> = db_documents.keys().map(String::as_str).collect();
+    let incoming_document_uris: HashSet<&str> = incoming_docs_lookup.keys().copied().collect();
+
+    let new: HashSet<&str> = incoming_document_uris.difference(&db_uris).copied().collect();
+    let deleted: HashSet<&str> = db_uris.difference(&incoming_document_uris).copied().collect();
+    let updated: HashSet<&str> = db_uris
+        .intersection(&incoming_document_uris)
+        .filter_map(|&uri_str| {
+            let incoming_hash = incoming_docs_lookup.get(uri_str)?;
+            let db_hash = db_documents.get(uri_str)?;
+
+            (incoming_hash != db_hash).then_some(uri_str)
+        })
+        .collect();
+
+    let new_uris = new.union(&updated).map(|&s| s.to_string()).collect();
+    let stale_uris = deleted.union(&updated).map(|&s| s.to_string()).collect();
+
+    (new_uris, stale_uris)
+}
+
 #[cfg(test)]
 mod tests {
     use std::path::PathBuf;
 
     use super::*;
+    use crate::model::ids::UriId;
 
     #[test]
     fn collecting_documents_in_parallel() {
@@ -206,5 +267,70 @@ mod tests {
         let (documents, errors) = collect_documents(paths);
         assert!(documents.is_empty());
         assert!(errors.is_empty());
+    }
+
+    #[test]
+    fn index_and_sync_works() {
+        let mut graph = Graph::new();
+        graph.db.initialize_connection(None).unwrap();
+
+        // Set up initial state
+        let initial_documents = vec![
+            Document::new("file:///existing.rb", Some("module Existing; end".to_string())).unwrap(),
+            Document::new("file:///updated.rb", Some("class OldVersion; end".to_string())).unwrap(),
+            Document::new("file:///to_delete.rb", Some("class ToDelete; end".to_string())).unwrap(),
+        ];
+        index_and_sync(&mut graph, initial_documents).unwrap();
+
+        let initial_db_docs = graph.db.all_documents_hashes_by_uri().unwrap();
+        assert_eq!(initial_db_docs.len(), 3);
+
+        // Test sync with updated document set
+        let incoming_documents = vec![
+            Document::new("file:///new.rb", Some("module New; end".to_string())).unwrap(),
+            Document::new("file:///existing.rb", Some("module Existing; end".to_string())).unwrap(), // Same content
+            Document::new("file:///updated.rb", Some("class NewVersion; end".to_string())).unwrap(), // Updated content
+        ];
+
+        index_and_sync(&mut graph, incoming_documents).unwrap();
+
+        let result_db_docs = graph.db.all_documents_hashes_by_uri().unwrap();
+        assert_eq!(result_db_docs.len(), 3);
+        assert!(result_db_docs.contains_key("file:///new.rb"));
+        assert!(result_db_docs.contains_key("file:///existing.rb"));
+        assert!(result_db_docs.contains_key("file:///updated.rb"));
+
+        assert!(
+            graph
+                .db
+                .load_uri(UriId::from("file:///new.rb"))
+                .unwrap()
+                .iter()
+                .any(|result| result.name == "New")
+        );
+        assert!(
+            graph
+                .db
+                .load_uri(UriId::from("file:///existing.rb"))
+                .unwrap()
+                .iter()
+                .any(|result| result.name == "Existing")
+        );
+        assert!(
+            graph
+                .db
+                .load_uri(UriId::from("file:///updated.rb"))
+                .unwrap()
+                .iter()
+                .any(|result| result.name == "NewVersion")
+        );
+
+        graph.assert_integrity();
+
+        let documents: Vec<_> = graph.documents().values().collect();
+        assert_eq!(documents.len(), 3);
+        assert!(documents.iter().any(|doc| doc.uri() == "file:///new.rb"));
+        assert!(documents.iter().any(|doc| doc.uri() == "file:///existing.rb"));
+        assert!(documents.iter().any(|doc| doc.uri() == "file:///updated.rb"));
     }
 }
--- a/rust/index/src/indexing/errors.rs
+++ b/rust/index/src/indexing/errors.rs
@@ -11,11 +11,24 @@ impl std::fmt::Display for MultipleErrors {
 
 impl Error for MultipleErrors {}
 
+impl From<Box<dyn Error>> for MultipleErrors {
+    fn from(error: Box<dyn Error>) -> Self {
+        MultipleErrors(vec![IndexingError::FailedDbOperation(error.to_string())])
+    }
+}
+
+impl From<IndexingError> for MultipleErrors {
+    fn from(error: IndexingError) -> Self {
+        MultipleErrors(vec![error])
+    }
+}
+
 // Enum representing all types of indexing errors that may happen
 #[derive(Debug)]
 pub enum IndexingError {
     FileReadError(String),
     InvalidUri(String),
+    FailedDbOperation(String),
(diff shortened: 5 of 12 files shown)


rust/index-sys/src/index_api.rs
```diff
@@ -56,9 +59,10 @@ pub unsafe extern "C" fn idx_index_all_c(
     all_errors.extend(document_errors);
 
     with_graph(pointer, |graph| {
-        if let Err(errors) = indexing::index_in_parallel(graph, documents) {
-            all_errors.extend(errors.0);
-        }
+        indexing::sync_with_db(graph, &documents)
+            .and_then(|documents_to_index| indexing::index_in_parallel(graph, documents_to_index))
+            .and_then(|()| graph.save_to_database())
```
amomchilov: Where does this unit come from? Can it be dropped, or replaced with `_`?

```suggestion
            .and_then(|_| graph.save_to_database())
```

rust/index/src/indexing.rs
```diff
@@ -1,54 +1,79 @@
...
+            content_hash,
         })
     }
 
     /// # Errors
     ///
     /// Errors if the file path cannot be turned into a URI
     pub fn from_file_path(path: &str) -> Result<Self, IndexingError> {
+        let uri = Url::from_file_path(path)
+            .map_err(|_e| IndexingError::InvalidUri(format!("Couldn't build URI from path '{path}'")))?;
```
amomchilov, line 54: The human-readable part of this is already express in the `Display` trait:

https://github.com/Shopify/index/blob/7ae18e796e17c44b4180a08dca0d1fe2e2e18434/rust/index/src/indexing/errors.rs#L38

As-is, this error would print as:

> Invalid URI: Couldn't build URI from path 'file:///a/b/c.rb'

```suggestion
            .map_err(|_e| IndexingError::InvalidUri(path))?;
```

amomchilov, line 54: @jesse-shopify Is there any convention around linking errors, to form a backtrace?

rust/index/src/indexing.rs
```diff
@@ -177,6 +175,103 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
...
+/// - Database synchronization operations
+/// - Parallel indexing of documents
+/// - Saving the graph to the database
+pub fn index_and_sync(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
+    sync_with_db(graph, documents)
+        .and_then(|documents_to_index| index_in_parallel(graph, documents_to_index))
+        .and_then(|()| graph.save_to_database().map_err(Into::into))
+}
+
+/// Represents the differences between current documents and database state
```
amomchilov:
```suggestion
/// Represents the differences between current documents on disk, and their cached state in database
```

rust/index/src/indexing.rs
```diff
@@ -207,4 +302,47 @@ mod tests {
...
+        // Check new documents
+        assert_eq!(diff.new.len(), 1);
+        assert!(diff.new.contains("file:///new1.rb"));
+
+        // Check deleted documents
+        assert_eq!(diff.deleted.len(), 1);
+        assert!(diff.deleted.contains("file:///deleted.rb"));
+
+        // No updates since we can't easily test content hash changes without filesystem setup
+        assert_eq!(diff.updated.len(), 0);
```
amomchilov: hmm?

rust/index/src/model/db.rs
```diff
@@ -104,6 +104,30 @@ impl Db {
...
+    /// Returns all URIs and their content hash currently stored in the database
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the database query fails
+    ///
+    /// # Panics
+    ///
+    /// Will panic if the database connection is not initialized before trying to interact with it
+    pub fn all_documents_with_content_hash(&self) -> Result<HashMap<String, u64>, Box<dyn Error>> {
```
amomchilov: Hmm, "with constant hash" sounds like a query criteria, where you specify the content hash to look up.

WDYT of:

```suggestion
    pub fn document_ids_by_uri(&self) -> Result<HashMap<String, u64>, Box<dyn Error>> {
```

Which emphasizes that you're calling this not to get the URIs, but mainly the IDs.

rust/index/src/model/db.rs
```diff
@@ -127,6 +151,34 @@ impl Db {
...
+                let placeholders = "?, ".repeat(uri_strings.len());
+                let placeholders = &placeholders[..placeholders.len() - 2]; // Remove trailing ", "
+                let sql = format!("DELETE FROM documents WHERE uri IN ({placeholders})");
+
+                let mut stmt = tx.prepare_cached(&sql)?;
+                stmt.execute(rusqlite::params_from_iter(uri_strings.iter()))?;
+            } // stmt is dropped here, releasing the borrow on tx
+
+            tx.commit()?;
+            Ok(())
```
amomchilov, line 193: I think `commit()` already returns the correct type ([docs](https://docs.rs/rusqlite/latest/rusqlite/struct.Transaction.html#method.commit)), could this simplify like so?

```suggestion
            tx.commit()
```

rust/index/src/model/db.rs
```diff
@@ -310,4 +366,57 @@ mod tests {
...
+                rusqlite::params!["file:///bar.rb", 456i64],
+            )
+            .unwrap();
+        connection
+            .execute(
+                "INSERT INTO documents (uri, content_hash) VALUES (?1, ?2)",
+                rusqlite::params!["file:///baz.rb", 789i64],
+            )
+            .unwrap();
+        drop(borrow);
```
amomchilov: What does this repeated borrowing/droping of connections do?

rust/index/src/indexing.rs
```diff
@@ -1,54 +1,79 @@
...
 impl Document {
     /// # Errors
     ///
     /// Creating a new document fails on invalid URIs
     pub fn new(uri: &str, source: Option<String>) -> Result<Self, IndexingError> {
+        let uri =
+            Url::parse(uri).map_err(|e| IndexingError::InvalidUri(format!("Failed to parse URI '{uri}': {e}")))?;
+
+        // Try to get the file content from the source first. If not try to read from the disk
+        let file_content = match source {
```
jesse-shopify: This looks like it may be a candidate for `unwrap_or_else`.

rust/index/src/indexing.rs
jesse-shopify: There's 2 popular crates for error handling:

- https://docs.rs/anyhow/latest/anyhow/
- https://docs.rs/thiserror/latest/thiserror/

They are not mutually exclusive so it's common to see either or both. And they encourage some idioms around wrapping errors. It can also be done in plain Rust, but at the point of this discussion probably would be worth more consideration.

rust/index/src/indexing.rs
```diff
@@ -177,6 +175,103 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
...
+/// 3. Saves the updated graph back to the database
+///
+/// # Errors
+///
+/// Returns `MultipleErrors` if any of the following operations fail:
+/// - Database synchronization operations
+/// - Parallel indexing of documents
+/// - Saving the graph to the database
+pub fn index_and_sync(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
+    sync_with_db(graph, documents)
```
jesse-shopify: I like the use of `and_then` here but this might be more readable in this case (subjective):

```
    let documents_to_index = sync_with_db(graph, documents)?;
    index_in_parallel(graph, documents_to_index)?;
    graph.save_to_database().map_err(Into::into)
```

rust/index/src/indexing.rs
amomchilov: Yeah I've used anyhow in a toy project before. Would support using it, but idk the tradeoffs

jesse-shopify: Larger conversation that we should start on the channel.

rust/index/src/model/db.rs
jesse-shopify: They have different error types so you'd have to convert the error type, which `?` does implicitly, and perhaps non-intuitively. Maybe `tx.commit().map_err(Into::into)`. But is that more or less readable?

https://stackoverflow.com/questions/78003329/why-does-resultmap-err-behave-differently-when-using-or-return

rust/index/src/model/db.rs
jesse-shopify: Are these held over from earlier code? They don't look like they are necessary. In my experience calls to `drop` have been rare. If one is testing for resource freeing it's probably better to use a scope. eg.

```rust

  let resource_allocator = Something.new();
  {
    let acquire = resource_allocator.get_resource();
    // DO STUFF
  } // implicit drop of acquire

  // assert that resource_allocator resource give to acquire has been freed

```

rust/index/src/model/db.rs
jenny-codes: Ok but since it's to get the content hashes, I will change it to `document_hashes_by_uri`? Although with the `by` I would kind of expect the function to take in a `uri` as an argument and not returning the whole table. 

Maybe `all_document_hashes` or `all_document_hashes_by_uri`?

rust/index-sys/src/index_api.rs
jenny-codes: This cam from the Ok result of `indexing::index_in_parallel`. We cannot drop it, and using `_` would actually trigger a clippy warning that nudges us to use the unit `()` instead for it is more explicit.

rust/index/src/indexing.rs
jenny-codes: Good point. Updated

rust/index/src/indexing.rs
jenny-codes: lol
tests updated

rust/index/src/indexing.rs
```diff
@@ -177,6 +175,103 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
...
+
+/// Represents the differences between current documents and database state
+#[derive(Debug)]
+pub struct DocumentDiff {
+    /// URIs that are new and need to be inserted into the database
+    pub new: HashSet<String>,
+    /// URIs that no longer exist and need to be deleted from the database
+    pub deleted: HashSet<String>,
+    /// URIs that have been updated and need to be re-synchronized with the database
+    pub updated: HashSet<String>,
```
jenny-codes: Pairing with @alexcrocha –we might have a better result using the uri id struct directly because of the string interning

rust/index/src/indexing.rs
jenny-codes: Updated to use `map_or_else` instead! (Which is a similar method that allows us to return a `Result` from `Option`)

rust/index/src/model/db.rs
jenny-codes: I've cleaned up the tests. You should find them much more reasonable now

rust/index/src/model/db.rs
jenny-codes: Subjectively I like the `?` and then `Ok` but also happy to update to `tx.commit().map_err(Into::into)` if anyone desires so

rust/index/src/model/db.rs
jenny-codes: Updated it to `all_documents_hashes_by_uri`

rust/index/src/indexing.rs
jenny-codes: Updated!

rust/index/src/indexing.rs
jenny-codes: Will investigate this as a followup

rust/index/src/indexing.rs
jenny-codes: Created an issue for followup https://github.com/Shopify/team-ruby-dx/issues/1726

rust/index/src/model/db.rs
amomchilov: Ah yes, that makes sense.

rust/index/src/model/db.rs
amomchilov: my bad, thought the types were the same, but they weren't.

rust/index/src/indexing.rs
jenny-codes: I have refactored this part so we are not storing any intermediate list in a struct any more.

rakelib/index_top_gems.rake
```diff
@@ -41,7 +41,7 @@ task index_top_gems: :compile_release do
 
             # Create the directory to extract the gem's files into
             gem_dir = File.join(@gems_dir, gem)
-            FileUtils.mkdir(gem_dir)
+            FileUtils.mkdir_p(gem_dir)
```
jenny-codes: Note: this is currently needed for the test to path. Investigating why it is needed

rakelib/index_top_gems.rake
```diff
@@ -51,6 +51,7 @@ task index_top_gems: :compile_release do
 
             # Index the gem's files and yield errors back to the main Ractor
             graph = Index::Graph.new
+            graph.set_configuration(File.join(@gems_dir, "#{gem}.db"))
```
jenny-codes, line 54: Note: now that we have added db operations, not setting up configuration before calling the index api will raise an error

rust/index/src/indexing.rs
```diff
@@ -11,45 +11,66 @@ use std::sync::{
...
     ///
     /// Creating a new document fails on invalid URIs
     pub fn new(uri: &str, source: Option<String>) -> Result<Self, IndexingError> {
+        let uri = Url::parse(uri).map_err(|e| IndexingError::InvalidUri(e.to_string()))?;
+
+        // Try to get the file content from the source first. If not try to read from the disk
+        let file_content = source.map_or_else(
+            || fs::read_to_string(uri.path()).map_err(|e| IndexingError::FileReadError(e.to_string())),
+            Ok,
+        )?;
```
jenny-codes, line 39: Note: we move this file source reading out of the `index_in_parallel()` because we need the file content earlier to calculate the content hash. This does mean though the file reading will now be done sequentially and will likely affect the performance.

rust/index/src/model/db.rs
```diff
@@ -189,19 +213,19 @@ impl Db {
     }
 
     /// Performs batch insert of documents (URIs) to the database
-    fn batch_insert_documents(conn: &rusqlite::Connection, graph: &Graph) -> Result<(), Box<dyn Error>> {
-        let mut stmt = conn.prepare_cached("INSERT INTO documents (id, uri) VALUES (?, ?)")?;
+    fn batch_find_or_insert_documents(conn: &rusqlite::Connection, graph: &Graph) -> Result<(), Box<dyn Error>> {
+        let mut stmt = conn.prepare_cached("INSERT OR IGNORE INTO documents (id, uri, content_hash) VALUES (?, ?, ?)")?;
```
jenny-codes, line 230: This change is so that when we try to re-insert the existing documents (the documents whose content hasn't changed) the db would not raise a entry-not-unique error. Those didn't need to be reindexed but are still in the graph. We may be able to handle it more elegantly but I chose the simplest way forward for now.

rust/index/src/indexing.rs
```diff
@@ -11,45 +11,66 @@ use std::sync::{
...
 pub mod ruby_indexer;
 
-// Represents a document to be indexed. If `source` is `Some(String)`, we're indexing a file state that's not committed
-// to disk yet and should then use the given `source` instead of reading from disk
+#[derive(Clone)]
 pub struct Document {
     uri: Url,
-    source: Option<String>,
+    source: String,
+    content_hash: i64,
```
amomchilov, line 25: Looks like SQLite doesn't support an unsigned integer type (without resorting to e.g. bigint). https://sqlite.org/datatype3.html

We can cast to/from an `i64` when we interface with the DB, but for the rest of the application, we should keep the data modal as `u64`. An unsigned number more closely expresses "a bucket of bits" (of a hash value), than "a number" (that you intend to use for arithmetic)

```suggestion
    content_hash: u64,
```


Then we won't need the cast below:

```diff
-let content_hash = xxh3_64(file_content.as_bytes()).cast_signed();
+let content_hash = xxh3_64(file_content.as_bytes());
```

rust/index/src/indexing.rs
```diff
@@ -177,11 +172,71 @@ pub fn collect_documents(paths: Vec<String>) -> (Vec<Document>, Vec<IndexingErro
...
+        .iter()
+        .map(|doc| (doc.uri.as_str(), doc.content_hash))
+        .collect();
+
+    let db_uris: HashSet<&str> = db_documents.keys().map(String::as_str).collect();
+    let incoming_document_uris: HashSet<&str> = incoming_docs_lookup.keys().copied().collect();
+
+    let new: HashSet<&str> = incoming_document_uris.difference(&db_uris).copied().collect();
+    let deleted: HashSet<&str> = db_uris.difference(&incoming_document_uris).copied().collect();
+    let updated: HashSet<&str> = db_uris
```
amomchilov, line 224: These 3 can be calculated in a single pass

Alex TODO: check if the stdlib has something for this

rust/index/src/indexing.rs
jenny-codes: I thought about using u64 for the application and only do the recasting at the db layer but decided on this current approach for debugging convenience–this way we won't be surprised that the raw value from the db is different from the value that the app code uses. And we only have to transform the value once here and not worry about casting back and forth. But maybe it's less of a concern?

amomchilov: Is there any other code that might need to interact with the `content_hash`?

If not, then ignore what I said, `i64` is fine

jenny-codes: So far it is only used for diff checking between documents. I will leave this alone for now then.

rust/index/src/indexing.rs
jenny-codes: Maybe something like
- Iterate on the incoming_uris set, where for each incoming_uri:
  - Check in the db hashmap if the uri exists
    - If exists, remove the entry from the db hashmap, and check if the content hashes are the same 
      - if same content hash, add the uri to `updated`
    - Else, add the uri to `new`
- What remains in the db hashmap would be the `deleted` set

?

rakelib/index_top_gems.rake
vinistock: How about we put the database inside each gem directory?
```suggestion
            graph.set_configuration(File.join(gem_dir, "graph.db"))
```

rust/index/src/db/schema.sql
```diff
@@ -4,7 +4,8 @@
 -- Table for storing documents
 CREATE TABLE IF NOT EXISTS documents (
     id INTEGER PRIMARY KEY,  -- Blake3 hash converted to hex
-    uri TEXT NOT NULL UNIQUE
+    uri TEXT NOT NULL UNIQUE,
+    content_hash INTEGER NOT NULL
```
vinistock, line 8: Given our recent schema optimizations, we may want to put this in a `data` blob - depending on what you find with the definitions table.

We also need to take loading into consideration, since with a separate column are you able to avoid using `SELECT *`. Doesn't necessarily need to be answered in this PR.

rust/index/src/indexing/errors.rs
```diff
@@ -11,18 +11,32 @@ impl std::fmt::Display for MultipleErrors {
 
 impl Error for MultipleErrors {}
 
+impl From<Box<dyn Error>> for MultipleErrors {
+    fn from(error: Box<dyn Error>) -> Self {
+        MultipleErrors(vec![IndexingError::FailedDbOperation(error.to_string())])
+    }
+}
```
vinistock, line 18: Am I reading this correctly that every `Box<dyn Error>` will be turned into a `FailedDbOperation` error?

If that's the case, I think this is a little too broad. We could certainly have a `Box<dyn Error>` that isn't related to the database.

rust/index/src/model/db.rs
```diff
@@ -120,39 +144,51 @@ impl Db {
         self.with_connection(|connection| {
             let tx = connection.transaction()?;
 
-            Self::batch_insert_documents(&tx, graph)?;
-            Self::batch_insert_names(&tx, graph)?;
-            Self::batch_insert_definitions(&tx, graph)?;
+            Self::batch_find_or_insert_documents(&tx, graph)?;
```
vinistock, line 147: I think there's a design aspect that I would disagree with here. Our idea is not to try to save the full graph every time. It's too slow and we could end up in a bad situation if the editor decides to force kill the process during the insert query.

Ideally, we only invoke this method to save the entire graph if there was no database to begin with. From that point on, the LSP will listen to files being saved to disk and commit changes to the database one by one, to avoid having to perform more large dumps.

I would keep this method as is, with the intention to use it to bootstrap the database, and add more methods for synchronization.

rust/index/src/model/document.rs
```diff
@@ -6,14 +6,16 @@ use crate::model::ids::DefinitionId;
 pub struct Document {
     uri: String,
     definition_ids: Vec<DefinitionId>,
+    content_hash: Option<i64>,
```
vinistock, line 9: Why is this an `Option`? What scenario leads to us not being able to compute a content hash?

rust/index/src/model/graph.rs
```diff
@@ -26,7 +20,7 @@ pub struct Graph {
     documents: IdentityHashMap<UriId, Document>,
     // Map of definition nodes
     definitions: IdentityHashMap<DefinitionId, Definition>,
-    db: Db,
+    pub db: Db,
```
vinistock, line 23: We should not leak the Db outside of the graph. Why does this need to be public?

rust/index/src/model/graph.rs
```diff
@@ -136,7 +136,7 @@ impl Graph {
     ///
     /// Any database errors will prevent the data from being loaded
     pub fn load_uri(&mut self, uri: String) -> Result<(), Box<dyn Error>> {
-        let uri_id = self.add_uri(uri);
+        let uri_id = self.add_document(uri, None);
```
vinistock, line 139: This would need to be reworked. `db.load_uri` would return us the data with the existing content hash and then we add that to the graph.

rust/index/src/indexing.rs
```diff
@@ -11,45 +11,66 @@ use std::sync::{
...
 
 impl Document {
     /// # Errors
     ///
     /// Creating a new document fails on invalid URIs
     pub fn new(uri: &str, source: Option<String>) -> Result<Self, IndexingError> {
+        let uri = Url::parse(uri).map_err(|e| IndexingError::InvalidUri(e.to_string()))?;
+
+        // Try to get the file content from the source first. If not try to read from the disk
+        let file_content = source.map_or_else(
```
vinistock, line 36: Ideally, we don't want to read from disk here. Currently, each thread reads the files lazily, so that we don't keep blocked on IO in a single thread for too long.

We'll need to think a bit about how to restructure this, but we can't read all files ahead of time in the main thread without impacting performance.

jenny-codes: If we run the benchmark on huge_corpus, we still run fast with this change: 

### Before (main)

```
PERFORMANCE BREAKDOWN

**Initialization:    0.159s (  1.3%)**
Indexing:          5.201s ( 43.9%)
Cleanup:           6.496s ( 54.8%)
Total:            11.856s

Maximum Resident Set Size: 8845279232 bytes (8435.51 MB)
Peak Memory Footprint:     8709495432 bytes (8306.02 MB)
Execution Time:            12.12 seconds
```

### After (this branch)
```
PERFORMANCE BREAKDOWN

**Initialization:    0.408s (  3.0%)** <-- this is where the sequential file reading happens
Indexing:          6.143s ( 44.9%)
Db operation:      0.000s (  0.0%)
Cleanup:           7.123s ( 52.1%)
Total:            13.673s

Maximum Resident Set Size: 8816394240 bytes (8407.96 MB)
Peak Memory Footprint:     8670403184 bytes (8268.74 MB)
Execution Time:            14.09 seconds
```

This change caused the initialization step to run for 0.4 sec instead of 0.16. Although this indeed did not take into account when one long read will block the process

rust/index/src/model/graph.rs
jenny-codes: Can you explain where the issue with this would be?

rust/index/src/model/graph.rs
jenny-codes: Really just for calling `graph.db.all_documents_hashes_by_uri()` inside indexing.rs. I can create a new api on Graph to keep the db private.

rust/index/src/model/document.rs
jenny-codes: It's for when we're loading the data into memory. We don't need the content hash then as this hash is only used during sync and not querying

rust/index/src/model/db.rs
jenny-codes: Good call. I will give this more thoughts.

rust/index/src/db/schema.sql
jenny-codes: I am brooding on this suggestion of storing the content hash in definitions instead of documents. Doing so would mean to store this value into every definition of  the document. I don't see the benefit of doing so even considering when we want to do a more incremental update (e.g., definition level instead of document level) in the future. Am I missing anything?

vinistock: I think we may need multiple levels of hashes to avoid work at different stages. For example, if we know for sure that the contents of the file are identical, we don't even need to parse the file.

We will probably also want to hash a combination of all definitions found in that document, so that we can figure out if the change was destructive or not, but I'm not sure we can remove the first level - as this requires parsing and indexing the file for comparison.

rust/index/src/model/graph.rs
vinistock: Sorry for the vague comment. If we already have the computed hash for a document in the database, why are we setting it as `None`? To save memory?

jenny-codes: It will add some awkwardness for the load query. If we want to have the value we will either modify the existing load_document query

```sql
SELECT
    documents.content_hash,  // here
    names.id,
    names.name,
    definitions.id,
    definitions.data
FROM documents
JOIN definitions ON documents.id = definitions.document_id
JOIN names ON names.id = definitions.name_id
WHERE documents.id = ?
```

and have this same value be repeated for as many definitions as there are associated with the document, or we use a separated query to populate it, but it would be just for the sake of having the field be non-optional

PR #174, opened 2025-09-24
Use i64 instead of u64 for ID

jenny-codes:
This is to fix the issue where the db Integer field expects an i64
value. We are currently trying to pass in a u64 value (as string) for
which a type mismatch error would be raised when we try to save the data.

So here we cast the u64 hash result into i64 to meet the db constraint.

### DIFF
--- a/rust/index/src/model/id.rs
+++ b/rust/index/src/model/id.rs
@@ -1,5 +1,6 @@
 //! This module contains stable ID representations that compose the `Graph` global representation
 
+use rusqlite::types::{FromSql, FromSqlResult, ToSql, ToSqlOutput, Value, ValueRef};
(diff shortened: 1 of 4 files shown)


rust/index/src/model/id.rs
```diff
@@ -10,13 +11,13 @@ use xxhash_rust::xxh3::xxh3_64;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Id<T> {
-    value: u64,
+    value: i64,
```
Morriar, line 14: I'm confused as to why we're using a signed integer for the ID, do we expect hashes to be negative?

PR #175, opened 2025-09-24
Add save_to_db to benchmark

jenny-codes:
First step to https://github.com/Shopify/team-ruby-dx/issues/1725

We haven't been benchmarking the db operations. This allows us to start
measuring the db performances.

### DIFF
--- a/utils/bench
+++ b/utils/bench
@@ -18,6 +18,12 @@ if [ ! -d "$CORPUS_PATH" ]; then
     $SCRIPT_DIR/generate-corpus "$CORPUS_PATH" -s 1000
 fi
 
+# Temporary measure until we have proper db sync capability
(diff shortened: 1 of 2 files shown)


utils/bench
```diff
@@ -18,6 +18,12 @@ if [ ! -d "$CORPUS_PATH" ]; then
     $SCRIPT_DIR/generate-corpus "$CORPUS_PATH" -s 1000
 fi
 
+# Temporary measure until we have proper db sync capability
```
vinistock, line 21: Depends on what we're benchmarking. We'll probably want to benchmark a cold boot with no database and the synchronization of a partially stale database on different occasions.

jenny-codes: I think after we add the db sync to indexing we can make this into an argument for the command

PR #177, opened 2025-09-26
Delete orphaned names when deleting documents

jenny-codes:
The current method that deletes the document is correctly deleting its
associating definitions, but not the names. We fix it by using an
additional query to delete the orphaned names.

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -131,15 +131,28 @@ impl Db {
     }
 
     /// Deletes all of the data related to the given URI ID. This method relies on the foreign key constraints to
-    /// cascade the deletion to definitions and names that are no longer referenced
+    /// cascade the deletion to definitions, and explicitly deletes orphaned names that are no longer referenced
     ///
     /// # Errors
     ///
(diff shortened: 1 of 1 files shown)


rust/index/src/model/db.rs
```diff
@@ -131,15 +131,28 @@ impl Db {
...
+                let mut stmt = tx.prepare_cached("DELETE FROM documents WHERE id = ?")?;
+                stmt.execute([*uri_id])?;
+            }
+
+            tx.execute(
+                "DELETE FROM names WHERE NOT EXISTS (
+                    SELECT 1 FROM definitions WHERE name_id = names.id
+                )",
+                [],
+            )?;
```
Morriar, line 153: I wondering if this kind of integrity queries would be better using triggers? https://sqlite.org/lang_createtrigger.html

jenny-codes: I chose this over triggers for the sake of explicitness. I feel database triggers add hidden complexity and are harder to debug. But I can be easily persuaded to use it if you think it is the right tool for our use case here

PR #180, opened 2025-10-07
Prevent adding duplicate definitions to documents or declarations

vinistock:
Prevent duplicate definitions from being inserted into the vector of definitions inside documents and declarations - now that we no longer have the guarantees of a hashset.

### DIFF
--- a/rust/index/src/model/declaration.rs
+++ b/rust/index/src/model/declaration.rs
@@ -47,6 +47,28 @@ impl Declaration {
     }
 
     pub fn add_definition(&mut self, definition_id: DefinitionId) {
+        debug_assert!(
+            !self.definition_ids.contains(&definition_id),
+            "Cannot add the same exact definition to a declaration twice. Duplicate definition IDs"
+        );
+
         self.definition_ids.push(definition_id);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[should_panic(expected = "Cannot add the same exact definition to a declaration twice. Duplicate definition IDs")]
+    fn inserting_duplicate_definitions() {
+        let mut decl = Declaration::new("MyDecl".to_string());
+        let def_id = DefinitionId::new(123);
+
+        decl.add_definition(def_id);
+        decl.add_definition(def_id);
+
+        assert_eq!(decl.definitions().len(), 1);
     }
 }
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -267,7 +267,7 @@ impl<'a> RubyIndexer<'a> {
                     ruby_prism::Node::StringNode { .. } => {
                         let string = argument.as_string_node().unwrap();
                         let name = String::from_utf8_lossy(string.unescaped()).to_string();
-                        f(name, node.location());
+                        f(name, argument.location());
                     }
                     _ => {}
                 }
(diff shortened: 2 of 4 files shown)


rust/index/src/model/declaration.rs
```diff
@@ -42,6 +42,26 @@ impl Declaration {
     }
 
     pub fn add_definition(&mut self, definition_id: DefinitionId) {
+        if self.definition_ids.contains(&definition_id) {
+            return;
```
st0012: When we enter this code path, do we consider it a bug? If that's the case, can we use `debug_assesrt!` here to help us debug issues in the future?

vinistock: That's a good point. This may only happen if we have a bug or if an add-on is using some public API in a way we don't expect, but I'm not 100% sure what's the best approach here.

An assertion would panic, right? I'm not sure if that's not overkill. Maybe just alerting the user? It might be a bit soon to tell. What do you think?

st0012: [`debug_assert!`](https://doc.rust-lang.org/std/macro.debug_assert.html) only panic in non-release builds. So it'll help us spotting bugs during development, but not affecting users.

I'd also vote for proactively using `panic!` or `unreachable` in our code when we're not sure about if it's a valid behaviour or not. They will serve as todos for us to think more clear about the behaviour. Each time we hit it, either in random test failures or during development, we'd be forced to bump its priority. And if we think they're too aggressive, it's easy to search and remove them later too.
In the contrary, if we use `return` here, other devs will likely assume having duplicates is an intended behaviour, instead of questioning it. And this can hide deeper problems in our code for longer unnecessarily.

rust/index/src/model/declaration.rs
```diff
@@ -42,6 +42,26 @@ impl Declaration {
     }
 
     pub fn add_definition(&mut self, definition_id: DefinitionId) {
+        if self.definition_ids.contains(&definition_id) {
```
Morriar: Now that we're performing a linear search each time we push a value, are we still faster than using hashsets?

rust/index/src/model/declaration.rs
Morriar: Another option would be to follow what HashSet#insert does and return a boolean:

> Returns whether the value was newly inserted. That is:
> * If the set did not previously contain this value, true is returned.
> * If the set already contained this value, false is returned, and the set is not modified: original value is not replaced, and the value passed as argument is dropped.

rust/index/src/model/declaration.rs
vinistock: This is not a performance, but a memory optimization. Switching back to HashSets makes no difference in performance, but it increases the memory usage by a lot.

```
Vectors

Maximum Resident Set Size: 7147749376 bytes (6816.62 MB)
Peak Memory Footprint:     7047894488 bytes (6721.39 MB)
Execution Time:            15.95 seconds

HashSets

Maximum Resident Set Size: 7990018048 bytes (7619.87 MB)
Peak Memory Footprint:     7912085576 bytes (7545.55 MB)
Execution Time:            15.52 seconds
```

rust/index/src/model/declaration.rs
vinistock: I used `debug_assert!` and it immediately proved useful. It uncovered that we cannot use simply URI and start offset to create definition IDs because attribute accessors produce two definitions from the same exact URI + offset combination.

I fixed that bug by including the name into the mix, with the URI and offset.

PR #181, opened 2025-10-07
Index `attr` calls

### LINKED ISSUE #90: Start indexing attributes
Start indexing attributes. Attributes are similar, but not exactly the same as method definitions.

```ruby
class Foo
  attr_reader :foo    # only reader (def foo; end)
  attr_writer :foo    # only writer (def foo=(foo); end)
  attr_accessor :foo # both
  attr :foo, true     # both
  attr :foo, false    # only reader
  attr :foo           # only reader
end
```

st0012:
Closes #90

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -574,7 +574,86 @@ impl Visit<'_> for RubyIndexer<'_> {
         }
     }
 
+    #[allow(clippy::too_many_lines)]
     fn visit_call_node(&mut self, node: &ruby_prism::CallNode) {
+        fn create_attr_accessor(indexer: &mut RubyIndexer, node: &ruby_prism::CallNode) {
+            RubyIndexer::each_string_or_symbol_arg(node, |name, location| {
+                indexer.with_updated_nesting(&name, |indexer, fully_qualified_name| {
+                    let name_id = NameId::from(&fully_qualified_name);
+                    let writer_name = format!("{fully_qualified_name}=");
+                    let writer_name_id = NameId::from(&writer_name);
+                    let offset = Offset::from_prism_location(&location);
+                    let comments = indexer.find_comments_for(offset.start()).unwrap_or_default();
+                    // Uses `location_converter` to evaluate the performance impact of the conversion
+                    indexer.location_converter.offset_to_position(&offset).unwrap();
+                    indexer.local_index.add_definition(
+                        fully_qualified_name,
+                        Definition::AttrAccessor(Box::new(AttrAccessorDefinition::new(
+                            name_id,
+                            indexer.uri_id,
+                            offset,
+                            comments.clone(),
+                        ))),
+                    );
+
+                    indexer.local_index.add_definition(
+                        writer_name,
+                        Definition::AttrAccessor(Box::new(AttrAccessorDefinition::new(
+                            writer_name_id,
+                            indexer.uri_id,
+                            Offset::from_prism_location(&location),
+                            comments.clone(),
+                        ))),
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -574,7 +574,86 @@ impl Visit<'_> for RubyIndexer<'_> {
         }
     }
 
+    #[allow(clippy::too_many_lines)]
     fn visit_call_node(&mut self, node: &ruby_prism::CallNode) {
+        fn create_attr_accessor(indexer: &mut RubyIndexer, node: &ruby_prism::CallNode) {
```
vinistock, line 579: Can we move these functions definitions outside of the `visit_call_node` function? I didn't even realize Rust allowed for nested definitions, but I'd vote for avoiding them.

st0012: I'm on the opposite side tho. If we only use these functions in `visit_call_node`, keeping them in is like having local Proc/Lambda in a method for in-function abstraction. IMO it's a good way to indicate a function's scope & usage without actually document it.
It's a quite common pattern AFAIK and my experience with them in ZJIT made me appreciate them ([example](https://github.com/ruby/ruby/blob/master/zjit/src/hir.rs#L2264)).

vinistock: I see. I can see the appeal, although in this case the `visit_call_node` method ends up being a bit long. Not a big deal though, we can move forward like this.

Morriar: I'm on the fence. I like that this isolate properly the functions but I don't like that it forces a clippy exception?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -1574,4 +1604,85 @@ mod tests {
...
+    }
+
+    #[test]
+    fn index_attr_with_true_parameter() {
+        let mut context = GraphTest::new();
+
+        context.index_uri("file:///foo.rb", {
+            r#"
+            class Foo
+              attr "foo", true
```
Morriar, line 1667: We may want to add a test case showing that anything else than a literal `TrueNode` will be treated as `false`?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -586,79 +665,30 @@ impl Visit<'_> for RubyIndexer<'_> {
...
-                                comments,
-                            ))),
-                        );
-                    });
-                });
+                if create_writer {
+                    create_attr_accessor(self, node);
+                } else {
+                    create_attr_reader(self, node);
+                }
```
Morriar, line 691: I wonder if we should add a `Attr` enum so we can look at the node and know it's coming from an `attr` call? https://github.com/Shopify/index/blob/main/rust/index/src/model/definitions.rs#L39-L41

rust/index/src/indexing/ruby_indexer.rs
st0012: That's a good point. It turns out if you pass symbol of string, you can define multiple attributes at the same time. I'll update the implementation to account for that.
Values that are not string, symbol, or booleans, will raise runtime error.

rust/index/src/indexing/ruby_indexer.rs
st0012: My understanding is that `attr`'s usage is likely low enough to be ignored. Here's a [quick search on grokt](https://grokt.shopify.io/results?q=%5Cbattr%5Cs%3A+%5Cb+case%3Ayes+lang%3Aruby).

PR #182, opened 2025-10-07
Split empty declaration check

vinistock:
Addresses https://github.com/Shopify/index/pull/156#discussion_r2398959831

I split the empty check separately. The code ends up needing a block of scope because `remove` returns `()` and can't be involved in a conditional.

### DIFF
--- a/rust/index/src/model/graph.rs
+++ b/rust/index/src/model/graph.rs
@@ -212,6 +212,7 @@ impl Graph {
                 if let Some(definition) = self.definitions.remove(def_id)
                     && let Some(declaration) = self.declarations.get_mut(definition.name_id())
                     && declaration.remove_definition(def_id)
+                    && declaration.is_empty()
                 {
                     self.declarations.remove(definition.name_id());
                 }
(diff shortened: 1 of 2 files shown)


rust/index/src/model/graph.rs
```diff
@@ -211,7 +211,10 @@ impl Graph {
             for def_id in document.definitions() {
                 if let Some(definition) = self.definitions.remove(def_id)
                     && let Some(declaration) = self.declarations.get_mut(definition.name_id())
-                    && declaration.remove_definition(def_id)
+                    && {
+                        declaration.remove_definition(def_id);
```
Morriar: Can't we make `remove_definition` return a bool (`true` for removed, `false` for no-op) and avoid this block? If we did not remove anything, do we still need to check if it's empty?

I think we can follow most of the established API patterns when we handle collections: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove

vinistock: Switched to returning a boolean.

PR #186, opened 2025-10-07
Remove foreign keys

jenny-codes:
Part of https://github.com/Shopify/team-ruby-dx/issues/1725

Remove the foreign key constraints so we don't spend unnecessary time
validating entries at inserting. Instead we will enforce data integrity
through a different app function call.

This brings down the db time of large_corpus from ~23 seconds (as of https://app.graphite.dev/github/pr/Shopify/index/184/Remove-unused-indexes) to ~18 seconds

### Before

```
PERFORMANCE BREAKDOWN

Initialization:    0.050s (  0.2%)
Indexing:          0.717s (  2.8%)
*Db operation:     23.873s ( 94.9%)*
Cleanup:           0.527s (  2.1%)
Total:            25.168s

Maximum Resident Set Size: 1050460160 bytes (1001.79 MB)
Peak Memory Footprint:     1048675432 bytes (1000.09 MB)
Execution Time:            25.37 seconds
```

### After 

```
PERFORMANCE BREAKDOWN

Initialization:    0.029s (  0.1%)
Indexing:          0.575s (  3.0%)
*Db operation:     18.128s ( 94.2%)*
Cleanup:           0.503s (  2.6%)
Total:            19.235s

Maximum Resident Set Size: 1054425088 bytes (1005.57 MB)
Peak Memory Footprint:     1052623952 bytes (1003.86 MB)
Execution Time:            19.45 seconds
```

This also brings down the db time for the huge corpus to 6.5 minutes (from 12 minutes)

```
PERFORMANCE BREAKDOWN

Initialization:    0.418s (  0.1%)
Indexing:          5.969s (  1.5%)
Db operation:    393.855s ( 96.9%)
Cleanup:           6.398s (  1.6%)
Total:           406.640s

Maximum Resident Set Size: 7834189824 bytes (7471.26 MB)
Peak Memory Footprint:     8819694120 bytes (8411.11 MB)
Execution Time:            406.95 seconds
```

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -64,7 +64,6 @@ impl Db {
         conn.execute_batch(
             "
             PRAGMA synchronous = NORMAL;
-            PRAGMA foreign_keys = ON;
             PRAGMA locking_mode = EXCLUSIVE;
             ",
         )?;
@@ -140,6 +139,33 @@ impl Db {
         self.with_connection(|connection| {
             let mut stmt = connection.prepare_cached("DELETE FROM documents WHERE id = ?")?;
             stmt.execute([*uri_id])?;
+            Ok(())
+        })
+    }
+
+    /// Delete all the definitions and names that are no longer referenced.
+    ///
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn remove_orphaned_entries(&self) -> Result<(), Box<dyn Error>> {
+        self.with_connection(|connection| {
+            let tx = connection.transaction()?;
+            // First, delete definitions where document_id no longer exists
+            tx.prepare_cached(
+                "DELETE FROM definitions WHERE NOT EXISTS (
+                SELECT 1 FROM documents WHERE id = definitions.document_id
+            )",
+            )?
+            .execute([])?;
+            // Then, delete names that are no longer referenced by any definitions
+            tx.prepare_cached(
+                "DELETE FROM names WHERE NOT EXISTS (
+                SELECT 1 FROM definitions WHERE name_id = names.id
+            )",
+            )?
+            .execute([])?;
+            tx.commit()?;
             Ok(())
         })
     }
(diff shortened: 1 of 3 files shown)


rust/index/src/model/db.rs
```diff
@@ -144,6 +143,33 @@ impl Db {
         })
     }
 
+    /// Delete all the definitions and names that are no longer referenced.
+    ///
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn remove_orphaned_entries(&self) -> Result<(), Box<dyn Error>> {
```
vinistock, line 151: Is there a reason for this to be a separate API rather than logic that happens as part of `delete_data_for_uri`?

If not, then let's please make it all a single API to delete a URI. If we forget to invoke the second part method, we can easily get into a state with corrupted data.

jenny-codes: This came from the discussion on this pr https://github.com/Shopify/index/pull/177 which suggests it might be better to perform the cleanup as a separate step for efficiency.

vinistock: I think this part needs a bit more thought. When we remove the data from memory (from the hashmaps), we know which names, documents and definitions got deleted - so re-checking for orphans is duplicate work.

It might be a better idea to accumulate the IDs that got removed and then pass those to the DB delete method, so that we can run a single query and get all the work done in one go.

That approach would also allow us to move the `name_id` and `document_id` inside the `data` blob, which will simply our schema and potentially make the dump faster.

Can we try this:

1. Explore returning a list of deleted IDs from `Graph#remove_definitions_for_uri`
2. Pass those IDs for deletion to the DB abstraction
3. Use a single query to delete everything in one go

jenny-codes: Still this means every delete operation will incur three db mutations. If we have 10 uris to delete, here are the queries we will incur for each approach:

With the approach to have the cleanup as a separate step:
- 12 delete mutations
  - 10 for uris
  - 1 for orphaned definitions 
  - 1 for orphaned names
- *2 read queries
  - 1 for orphaned names
  - 1 for orphaned definitions

\*for now it is 10*2=20 but once we get to the sync pr that actually uses the delete function we can easily move this out to only be triggered once. 

With the approach to pass in ids to be deleted
- 30 delete mutations
  - 10 for uris
  - 10 for orphaned definitions
  - 10 for orphaned names

But I see the second approach does give us

> That approach would also allow us to move the name_id and document_id inside the data blob, which will simply our schema and potentially make the dump faster.

I will explore this but I propose to have it as an immediate follow-up? As the main purpose of this pr is to make sure the system still behaves correctly after removing the foreign key constraint in the schema.

vinistock: I'm fine with exploring as a follow up, but if you already know the IDs that have to be deleted and put everything in a transaction, isn't it a single DB operation?

The way I'm seeing it in my mind is we collect all name, definition and document IDs that got deleted in a vector and then craft a query like this:

```sql
BEGIN TRANSACTION
DELETE FROM names WHERE id in (...)
DELETE FROM documents WHERE id in (...)
DELETE FROM definitions WHERE id in (...)
END TRANSACTION
```

The logic in the `Graph` method already handles orphaned names and documents.

PR #189, opened 2025-10-08
Accept `attr` with multiple names

st0012:


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -674,12 +674,22 @@ impl Visit<'_> for RubyIndexer<'_> {
                 create_attr_writer(self, node);
             }
             "attr" => {
-                // attr :foo, true  => both reader and writer
-                // attr :foo, false => only reader
-                // attr :foo        => only reader
+                // attr :foo, true        => both reader and writer
+                // attr :foo, false       => only reader
+                // attr :foo              => only reader
+                // attr :foo, "bar", :baz => both reader and writer for foo, bar, and baz
```
vinistock, line 680: Is it possible to also do `attr :foo, :bar, false`?

st0012: Nope, it'd raise `TypeError`

```
irb(main):001* class Foo
irb(main):002*   attr :foo, "bar", false
irb(main):003> end
(irb):2:in 'Module#attr': false is not a symbol nor a string (TypeError)
```

vinistock: Interesting! TIL

PR #190, opened 2025-10-08
Delete orphaned entries by id

jenny-codes:
Address https://github.com/Shopify/index/pull/186#discussion_r2414620040

Instead of using a separate method that deletes all orphaned records, we now pass in the list of name_ids and definition_ids to the db to be deleted alongside documents.

### DIFF
--- a/rust/index/src/model/db.rs
+++ b/rust/index/src/model/db.rs
@@ -1,4 +1,4 @@
-use crate::model::graph::Graph;
+use crate::model::graph::{Graph, RemovedIds};
 use crate::model::{
     definitions::Definition,
     ids::{DefinitionId, NameId, UriId},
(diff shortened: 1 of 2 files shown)


rust/index/src/model/db.rs
```diff
@@ -129,42 +129,35 @@ impl Db {
...
+            // Delete the document
+            tx.prepare_cached("DELETE FROM documents WHERE id = ?")?
+                .execute([*uri_id])?;
+
+            // Delete definitions that were removed
+            if !removed_ids.definition_ids.is_empty() {
+                let mut stmt = tx.prepare_cached("DELETE FROM definitions WHERE id = ?")?;
+                for def_id in &removed_ids.definition_ids {
+                    stmt.execute([**def_id])?;
+                }
```
jenny-codes: Here I've tried to batch delete, but the benchmark result on large_corpus (10k uris, 1.4 mil definitions and names each) are actually slower than single delete. Batch delete takes 75 seconds where single delete takes 66 seconds. 

We need some profiling to understand why, but currently the suspects are:
- Cached vs uncached statement: with batch processing we don't get the benefit of using cached statements because of the varied lengths of the params. Even if we cache the statements for the batches that are the length of the BATCH_SIZE limit, the results are the \*same.
- Vector allocation: with batch statement we need to allocate extra vectors.

\* With BATCH_SIZE=100 the the deletion took 75 seconds. With BATCH_SIZE=10 the deletion took 10 seconds. Neither beats the 66 second single delete approach.

PR #191, opened 2025-10-08
Use macro to reduce duplication in Definition match expressions

st0012:


rust/index/src/model/definitions.rs
```diff
@@ -44,37 +44,32 @@ pub enum Definition {
     ClassVariable(Box<ClassVariableDefinition>),
 }
 
+macro_rules! all_definitions {
```
jenny-codes, line 47: Cool!

rust/index/src/model/definitions.rs
```diff
@@ -44,37 +44,32 @@ pub enum Definition {
...
+            Definition::Class($var) => $expr,
+            Definition::Module($var) => $expr,
+            Definition::Constant($var) => $expr,
+            Definition::GlobalVariable($var) => $expr,
+            Definition::InstanceVariable($var) => $expr,
+            Definition::ClassVariable($var) => $expr,
+            Definition::AttrAccessor($var) => $expr,
+            Definition::AttrReader($var) => $expr,
+            Definition::AttrWriter($var) => $expr,
+            Definition::Method($var) => $expr,
```
Morriar, line 59: I think the match is O(1) and the order doesn't matter so in this case, can we sort the entries alphabetically, it makes it easier to add new entries.

Morriar: Also it looks like all of them can be joined into a single union:

```rust
match self {
  Definition::Class(it)
  | Definition::Module(it)
  | Definition::Constant(it)
  | Definition::GlobalVariable(it)
  | Definition::InstanceVariable(it)
  | Definition::ClassVariable(it)
  | Definition::AttrAccessor(it)
  | Definition::AttrReader(it)
  | Definition::AttrWriter(it)
  | Definition::Method(it) => it.offset.start(),
match self {
```

So another possibility would have been something like this:

```rust
macro_rules! all_definition_variants {
    ($var:ident) => {
        Definition::Class($var)
            | Definition::Module($var)
            | Definition::Constant($var)
            | Definition::GlobalVariable($var)
            | Definition::InstanceVariable($var)
            | Definition::ClassVariable($var)
            | Definition::AttrAccessor($var)
            | Definition::AttrReader($var)
            | Definition::AttrWriter($var)
            | Definition::Method($var)
    };
}

match self {
    all_definition_variants!(it) => it.offset.start(),
}
```

vinistock: I believe it checks each condition in order according to the [book](https://doc.rust-lang.org/book/ch06-02-match.html) (search for "in order").

We may want to pay attention and always have the arms ordered based on how common they are.

Morriar: I think it actually depends on what kind of match you do:

* If it's based on enum types, it's optimized with a jump table [see the assembly](https://godbolt.org/z/bzqbdP1eh).
* If it's based on pattern matching, it's a [glorified if else](https://godbolt.org/z/G6TTGnqnf).

In our case I think the match gets optimized as a jump table and its lookup is constant, so whatever the order you use you still get the same performance.

vinistock: That's good news, although I'm a bit surprised the book doesn't mention that. In fact, the example in the book _is an enum_, which makes it seem that even enums are order dependent.

Well, in this case, better to sort it indeed.

st0012: @alexcrocha Since we talked about this earlier today, do you want to give it a look?

Issue #85, opened 2025-08-12
Start indexing constant references


vinistock:
To provide find references and rename functionality in the LSP, we need to keep track of all references to a constant (in addition to the definitions we already track).

The work to track references is a bit more involved since references can either be unresolved or resolved during the initial indexing pass. For example:

```ruby
module Foo
  module Bar
    class Baz
      def do_something
        # This is an unresolved constant reference. When we first see this `String`, we have no idea
        # if this refers to:
        # Foo::Bar::Baz::String
        # Foo::Bar::String
        # Foo::String
        # some inherited String constant
        # or top level String
        #
        # This reference has to be stored in a dedicated list that we can loop through during resolution
        # so that we can resolve this reference and determine which declaration it is pointing to
        String

        # In contrast, this is a constant reference that is already resolved from the moment we see it
        # Thanks to the fact that it's a top level reference, there is no way that this can resolve to 
        # anything that isn't top level String
        ::String
      end
    end
  end
end
```

We will need to prototype how to store resolved vs unresolved references in the index. Do we keep these in memory? Do we keep only unresolved in memory and, once resolution runs, dump everything to the database and never keep it in memory (references are only needed if the user asks for it, so maybe we don't have to waste memory)?

If we're never keeping it in memory, should this be some temporary structure rather than stored in the global index?

PR #192, opened 2025-10-09
Start tracking unresolved constant references

vinistock:
First step for #85

This PR introduces a semi optimized implementation for tracking unresolved constant references. I say semi optimized because we still have to look into interning non fully qualified names, which will likely save a bit more memory.

However, I did use a stack of Name IDs to save memory in remembering the exact nesting where we found the reference. This is really important to make the incremental architecture work because an indirect change may alter what the constant refers to - and without the exact nesting we discovered it, we cannot properly resolve it. For example:

```ruby
# Initially here, CONST does not exist

module Foo
  class Compact::Style
    CONST
  end
end

# Then someone creates `CONST` inside the Compact class in another file

module Foo
  class Compact
    CONST = 123
  end
end

# CONST **still** doesn't exist here because the compact style means we're not inside
# the `Compact` namespace. The only way to properly resolve the `CONST` reference,
# is if we remember the exact nesting where we found  it.
#
# If we simply store `Foo::Compact::Style`, we could mistakenly resolve it to `Foo::Compact::CONST`,
# which is not correct. This code raises

module Foo
  class Compact::Style
    CONST
  end
end
```

I would really like us to ship 3 important things for the prototype:

1. Constant reference tracking (part of this PR)
2. Basic inheritance tracking (includes, prepends, parent classes)
3. Resolution + ancestor linearization

It's way too much to try to put in a single branch, so I want to ship small pieces and then optimize later. It's really hard to see the exact shape of all pieces before implementing the entire thing.

I did benchmark this on Core (while we haven't added constant references to the corpus) and the performance was not degraded much. Memory usage increased, but we can probably reduce it a bit with string interning later.

```
// Maximum Resident Set Size: 432488448 bytes (412.45 MB)
// Peak Memory Footprint:     403456720 bytes (384.76 MB)
// Execution Time:            3.49 seconds

// Maximum Resident Set Size: 698466304 bytes (666.10 MB)
// Peak Memory Footprint:     673645432 bytes (642.43 MB)
// Execution Time:            3.62 seconds
```

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -8,6 +8,7 @@ use crate::model::definitions::{
 };
 use crate::model::graph::Graph;
 use crate::model::ids::{NameId, UriId};
+use crate::model::references::{UnresolvedConstantRef, UnresolvedReference};
 use crate::offset::Offset;
 use crate::source_location::SourceLocationConverter;
 
@@ -309,6 +310,25 @@ impl<'a> RubyIndexer<'a> {
             current_stack.pop();
         }
     }
+
+    fn index_constant_reference(&mut self, location: &ruby_prism::Location) {
+        let offset = Offset::from_prism_location(location);
+        let name = Self::location_to_string(location);
+        let nesting = self
+            .nesting_stacks
+            .last()
+            .expect("There should always be at least one stack. This is a bug")
+            .clone();
+        let name_id_nesting: Vec<NameId> = nesting.iter().map(NameId::from).collect();
+
+        let reference = UnresolvedReference::Constant(Box::new(UnresolvedConstantRef::new(
+            name,
+            name_id_nesting,
+            self.uri_id,
+            offset,
+        )));
+        self.local_index.add_unresolved_reference(reference);
+    }
 }
 
 struct CommentGroup {
@@ -451,6 +471,14 @@ impl Visit<'_> for RubyIndexer<'_> {
         });
 
         self.visit(&node.value());
+    }
+
+    fn visit_constant_read_node(&mut self, node: &ruby_prism::ConstantReadNode<'_>) {
+        self.index_constant_reference(&node.location());
+    }
+
+    fn visit_constant_path_node(&mut self, node: &ruby_prism::ConstantPathNode<'_>) {
+        self.index_constant_reference(&node.location());
     }
 
     fn visit_multi_write_node(&mut self, node: &ruby_prism::MultiWriteNode) {
--- a/rust/index/src/model/references.rs
+++ b/rust/index/src/model/references.rs
@@ -1,0 +1,60 @@
+use crate::{
+    model::ids::{NameId, UriId},
+    offset::Offset,
+};
+
+#[derive(Debug)]
+pub enum UnresolvedReference {
+    Constant(Box<UnresolvedConstantRef>),
+}
+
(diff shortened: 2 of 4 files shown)


rust/index/src/model/references.rs
```diff
@@ -0,0 +1,60 @@
+use crate::{
+    model::ids::{NameId, UriId},
+    offset::Offset,
+};
+
+#[derive(Debug)]
+pub enum Reference {
```
Morriar: Shouldn't this be an `UnresolvedReference`?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -453,6 +454,46 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+        let reference = Reference::UnresolvedConstant(Box::new(UnresolvedConstantRef::new(
+            name,
+            name_id_nesting,
+            self.uri_id,
+            offset,
+        )));
+        self.local_index.add_unresolved_reference(reference);
+    }
+
+    fn visit_constant_path_node(&mut self, node: &ruby_prism::ConstantPathNode<'_>) {
```
Morriar, line 480: Can't we share the same implementation than `visit_constant_read_node`?

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -453,6 +454,46 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+        let name_id_nesting: Vec<NameId> = nesting.iter().map(NameId::from).collect();
+
+        let reference = Reference::UnresolvedConstant(Box::new(UnresolvedConstantRef::new(
+            name,
+            name_id_nesting,
+            self.uri_id,
+            offset,
+        )));
+        self.local_index.add_unresolved_reference(reference);
+    }
```
Morriar, line 482: Do we also need all the and/or/multi etc. variants?

rust/index/src/model/references.rs
st0012: I think this means a new category called "Reference"?

vinistock: Yes, I was hoping to reuse the same enum for resolved and unresolved - in addition to multiple reference types like instance variables, although it could be better to split the two.

rust/index/src/indexing/ruby_indexer.rs
vinistock: I created a shared method, but it might need some adjustments once we stop using `location.slice` to get the constant path names.

rust/index/src/indexing/ruby_indexer.rs
vinistock: I didn't want to include this as part of this PR. This already allows us to prototype the resolution phase and I didn't want to open the discussion about what's the right way to model the other types just yet.

For example, `A ||= 1` actually defines `A` and doesn't raise any name errors. From a correctness perspective, this is a definition and not a reference given that it's how Ruby understands it. Adding a type checking error to it and saying that `A` is not defined would be a mismatch with the runtime.

However, it's a bit odd and conceptually a reference seems to make sense here as well. Maybe it needs to be both somehow? Anyway, we can discuss those in detail in dedicated PRs.

rust/index/src/model/references.rs
vinistock: I renamed the `enum` to `UnresolvedReference` to split the two in different types. I think this will be more ergonomic, but we can switch back if we find that's not the case.

PR #194, opened 2025-10-09
Remove get_documentation api

jenny-codes:
We won't need this as we will be getting the comments from the get() api
for definitions.

Address https://github.com/Shopify/index/pull/169#discussion_r2414651358

### DIFF
--- a/rust/index/src/model/graph.rs
+++ b/rust/index/src/model/graph.rs
@@ -79,22 +79,6 @@ impl Graph {
         )
     }
 
-    #[must_use]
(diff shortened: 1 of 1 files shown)


rust/index/src/model/graph.rs
```diff
@@ -527,34 +511,4 @@ mod tests {
         assert!(context.graph.documents.is_empty());
         assert!(context.graph.get("Foo").is_none());
     }
-
-    #[test]
-    fn get_documentation() {
```
vinistock, line 595: Can we keep this test, just using the `get` API instead?

PR #195, opened 2025-10-09
Remove comment filtering

jenny-codes:
Instead of trying to trim and filter the comments in the indexer, we
will leave it to the clients to decide what to do with the raw strings

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -16,22 +16,6 @@ use ruby_prism::{ParseResult, Visit};
 
 pub type IndexerParts = (Graph, Vec<IndexingError>);
 
-const MAGIC_AND_RBS_COMMENT_PREFIX: &[&str] = &[
-    "frozen_string_literal:",
-    "typed:",
-    "compiled:",
-    "encoding:",
-    "shareable_constant_value:",
-    "warn_indent:",
-    "rubocop:",
-    "nodoc:",
-    "doc:",
-    "coding:",
-    "warn_past_scope:",
-    "#:",
-    "#|",
-];
-
 /// The indexer for the definitions found in the Ruby source code.
 ///
 /// It implements the `Visit` trait from `ruby_prism` to visit the AST and create a hash of definitions that must be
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -340,21 +324,15 @@ impl CommentGroup {
...
-    // 2. not add the comment to the comments array
     fn add_comment(&mut self, comment: &ruby_prism::Comment) {
         self.end_offset = comment.location().end_offset();
         let text = String::from_utf8_lossy(comment.location().as_slice()).to_string();
-        if MAGIC_AND_RBS_COMMENT_PREFIX.iter().any(|&prefix| text.contains(prefix)) {
-            return;
-        }
 
-        let parsed_comment = text.trim().strip_prefix("# ").unwrap_or(text.trim());
+        let trimmed = text.trim();
```
vinistock: Two things here:

1. Do we need to trim the text here and add a new line break below? What would be the result if we just pushed the text directly?
2. I think we still want to strip the `# ` prefix. RBS annotations can easily be differentiated (since they start with `#:` or `#|` and won't be stripped), but there's no point in keeping the comment character and leading space

jenny-codes: 1. I added the trim there just as a precaution to remove unwanted whitespace characters. If we remove the trimming and the newline characters the multiline comments will all be in one line. There is a test that will fail.

2. Updated

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -360,21 +344,15 @@ impl CommentGroup {
...
-    // 2. not add the comment to the comments array
     fn add_comment(&mut self, comment: &ruby_prism::Comment) {
         self.end_offset = comment.location().end_offset();
         let text = String::from_utf8_lossy(comment.location().as_slice()).to_string();
-        if MAGIC_AND_RBS_COMMENT_PREFIX.iter().any(|&prefix| text.contains(prefix)) {
-            return;
-        }
 
-        let parsed_comment = text.trim().strip_prefix("# ").unwrap_or(text.trim());
+        let trimmed = text.trim().strip_prefix("# ").unwrap_or(text.trim());
```
Morriar, line 351: Honestly, I'd not even strip the `# ` prefix at this point since we're not sure what we want to do with `#:` and `#|` yet.

PR #201, opened 2025-10-09
Add stats flag for time breakdown and query statistics

st0012:
1. Added `--stats` flag - Shows performance breakdown (initialization, indexing, integrity check, querying, cleanup) with times and percentages
2. Query statistics - When `--stats` flag is used, displays documentation coverage, multi-definition counts, and definition type breakdown
    - Since we don't actually have querying API now, I think this can temporarily be used for timing querying phase
4. Removed `bench.rs` - Functionality merged into main CLI

This approach take heavy inspiration from how ZJIT/YJIT uses stats heavily for optimization ([example](https://github.com/Shopify/ruby/issues/714)). Unlike them, we don't have a set of known key metrics to target for, but we can start exploring them already.

The timer is implemented using macro so in the future we can easily add new phases by updating 

```rs
make_timer! {
    setup, "Initialization";
    indexing, "Indexing";
    querying, "Querying";
    new_phase, "New phase";
    integrity_check, "Integrity Check";
}

// and then
time_it!(new_phase, {})
```

Example stats look like this:

```
Query statistics

Total declarations:         13903583
With documentation:         1627824 (11.7%)
Without documentation:      12275759 (88.3%)
Total documentation size:   49326730 bytes
Multi-definition names:     0 (0.0%)

Definition breakdown:
  Method               10000000
  InstanceVariable     1996702
  Constant             1000000
  Class                600527
  Module               299509
  ClassVariable          5859
  GlobalVariable          986

Timing breakdown

Initialization      0.173s (  0.0%)
Indexing           14.218s (  1.4%)
Querying           11.339s (  1.1%)
Database         1016.004s ( 97.5%)
Total:           1041.735s

Indexed 100000 files
Found 13903583 names
Found 13903583 definitions
Found 100000 URIs
```


CLAUDE.md
```diff
@@ -57,10 +57,16 @@ AST to save in the global graph
...
+- `cargo run -- <directory> --stats`: runs the indexer with detailed performance breakdown
+- `cargo run -- <directory> --check-integrity`: runs integrity checks on the index after processing
+- `cargo run -- <directory> --visualize`: generates a DOT visualization of the graph
 - `cargo test`: runs Rust tests
-- `cargo test test_name `: runs a specific tests example
+- `cargo test test_name`: runs a specific tests example
 - `bundle exec rake lint_rust`: lints the Rust code
 - `bundle exec rake format_rust`: auto formats the Rust code
+
+Note: The indexer requires absolute paths or shell-expanded variables (e.g., `$HOME/path`). Tilde (`~`) is not automatically expanded by Rust. The SQLite database file (`graph.db`) will be automatically created in the target directory.
```
vinistock: FWIW, requiring absolute paths is actually a bug 😅.

rust/index/src/main.rs
```diff
@@ -64,3 +88,67 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     Ok(())
 }
+
+#[allow(clippy::cast_precision_loss)]
+fn print_query_statistics(graph: &Graph) {
```
vinistock: Should this live inside `Graph`? It seems like it's mostly about its internal structure, so I'd vote for us to encapsulate it there.

rust/index/src/main.rs
```diff
@@ -22,34 +23,57 @@ struct Args {
...
+                    eprintln!("  - {error}");
+                }
+                std::process::exit(1);
             }
-            std::process::exit(1);
-        }
+        });
+    }
+
+    time_it!(timer, database, { graph.save_to_database() })?;
```
vinistock: With this approach, if we wanted to measure the time to index specific nodes, we'd need to pass the timer object all the way, correct?

Could the timer be some sort of global so that we can avoid having to pass it around everywhere and can more easily time any part of the code?

CLAUDE.md
st0012: Since we still need Claude to know about this to compose the correct command, I've noted that it's a bug there.

rust/index/src/main.rs
st0012: Done 👍

rust/index/src/main.rs
st0012: Done 👍

PR #204, opened 2025-10-15
Resolve top-level constant references

### LINKED ISSUE #197: Resolve top-level constants

st0012:
Closes #197

### DIFF
--- a/rust/index/src/indexing/ruby_indexer.rs
+++ b/rust/index/src/indexing/ruby_indexer.rs
@@ -8,7 +8,7 @@ use crate::model::definitions::{
 };
 use crate::model::graph::Graph;
 use crate::model::ids::{NameId, UriId};
-use crate::model::references::{ResolvedConstantRef, ResolvedReference, UnresolvedConstantRef, UnresolvedReference};
+use crate::model::references::{ConstantRef, ResolvedReference, UnresolvedReference};
 use crate::offset::Offset;
 use crate::source_location::SourceLocationConverter;
 
@@ -323,7 +323,7 @@ impl<'a> RubyIndexer<'a> {
 
         // Check if this is a top-level (absolute) constant reference starting with ::
         if name.starts_with("::") {
-            let reference = ResolvedReference::Constant(Box::new(ResolvedConstantRef::new(
+            let reference = ResolvedReference::Constant(Box::new(ConstantRef::new(
                 name,
                 name_id_nesting,
                 self.uri_id,
@@ -331,7 +331,7 @@ impl<'a> RubyIndexer<'a> {
             )));
             self.local_index.add_resolved_reference(reference);
         } else {
-            let reference = UnresolvedReference::Constant(Box::new(UnresolvedConstantRef::new(
+            let reference = UnresolvedReference::Constant(Box::new(ConstantRef::new(
                 name,
                 name_id_nesting,
                 self.uri_id,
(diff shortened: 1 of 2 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -321,13 +321,24 @@ impl<'a> RubyIndexer<'a> {
...
-        self.local_index.add_unresolved_reference(reference);
+        // Check if this is a top-level (absolute) constant reference starting with ::
+        if name.starts_with("::") {
+            let reference = ResolvedReference::Constant(Box::new(ConstantRef::new(
+                name,
+                name_id_nesting,
+                self.uri_id,
+                offset,
+            )));
+            self.local_index.add_resolved_reference(reference);
```
Morriar, line 332: Question for you, is the reference truly resolved if we don't have a definition for the constant?

Let's take this example:

```rb
A = 42

puts ::A
puts ::B
```

Looking at the implementation we would have 2 resolved constant references: `::A` and `::B`, yet while we can resolve the name, we can't resolve `::B` to an actual constant since it doesn't exist.

Do we need the distinction? Or do we just let the client try to access `::B` and fail?

rust/index/src/model/graph.rs
```diff
@@ -39,6 +41,7 @@ impl Graph {
             definitions: IdentityHashMap::default(),
             documents: IdentityHashMap::default(),
             unresolved_references: Vec::new(),
+            resolved_references: Vec::new(),
```
vinistock, line 44: Instead of keeping an array of resolved references, would it make sense to store the references inside the declaration they belong to? For example, we'd have something like this:


```
Declaration {
  name: "Foo",
  references: Vec[Reference { uri, offsets, nesting }]
}
```

It would make it easier to implement rename, find references or dead code detection. We just grab the name we're searching for and check to see if any references exist.

rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -321,13 +321,24 @@ impl<'a> RubyIndexer<'a> {
...
-        let reference = UnresolvedReference::Constant(Box::new(UnresolvedConstantRef::new(
-            name,
-            name_id_nesting,
-            self.uri_id,
-            offset,
-        )));
-        self.local_index.add_unresolved_reference(reference);
+        // Check if this is a top-level (absolute) constant reference starting with ::
+        if name.starts_with("::") {
+            let reference = ResolvedReference::Constant(Box::new(ConstantRef::new(
```
vinistock, line 326: I think we need some API somewhere to house the entire algorithm. Right now, it's only looking at top level constants, so this is simple enough, but we'll need the rest of the algorithm to live together as well.

Can we move this logic into an API? Maybe something like this:

```rust
// Should this return an optional concrete reference or a union of either
// the resolved reference or the original unresolved one?
pub fn resolve_constant_reference(ref: UnresolvedConstantReference) -> Option<ConstantReference> {

  // check top level (fast path for references prefixed with `::`)
  // check surrounding scopes
  // check inheritance
  // check top level (slow path)
}
```

Connecting with what @Morriar said, this makes it easier for us to differentiate between the different types of resolution failure.

If we have `::Foo`, we already know the fully qualified name without running resolution, but the declaration may not exist. This is slightly different than not knowing exactly what the user is referring to and the distinction would be helpful for better tooltips/error messages (e.g.: we couldn't find where `::Foo` exists vs not sure what constant this is supposed to be).

PR #205, opened 2025-10-15
Switch files collection to use the `ignore` crate

Morriar:
Proof of concept showing how we could reduce the time spent listing files using the [ignore](https://crates.io/crates/ignore). I'm sure we can do even better but this already gets us to a better state.

Comparison on shop/world:

```
Timing breakdown    before              after

Initialization      0.003s (  0.0%)    0.002s (  0.0%)
Listing             5.587s ( 51.5%)    1.755s ( 23.1%)
Indexing            4.970s ( 45.8%)    5.568s ( 73.3%)
Querying            0.293s (  2.7%)    0.272s (  3.6%)
Database            0.000s (  0.0%)    0.000s (  0.0%)
Cleanup             0.000s (  0.0%)    0.000s (  0.0%)
Total:             10.853s             7.597s         
```

Comparison on corpus/huge:

```
Timing breakdown    before              after

Initialization      0.004s (  0.0%)    0.002s (  0.0%)
Listing             0.167s (  0.9%)    0.099s (  0.6%)
Indexing           11.354s ( 64.1%)   10.541s ( 62.3%)
Querying            6.184s ( 34.9%)    6.287s ( 37.1%)
Database            0.000s (  0.0%)    0.000s (  0.0%)
Cleanup             0.000s (  0.0%)    0.000s (  0.0%)
Total:             17.709s            16.928s
```

### DIFF
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -6,7 +6,8 @@ use crate::{
     model::graph::Graph,
     source_location::UTF8SourceLocationConverter,
 };
-use glob::glob;
+use ignore::types::TypesBuilder;
+use ignore::{WalkBuilder, WalkState};
 use std::sync::{
     Arc, Mutex,
     mpsc::{Receiver, Sender},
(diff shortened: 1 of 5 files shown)


rust/index/src/indexing.rs
```diff
@@ -135,46 +136,68 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
...
+        let mut types = TypesBuilder::new();
+        types.add("ruby", "*.rb").unwrap();
+        types.select("ruby");
+        let types = types.build().unwrap();
+
+        builder
+            .threads(thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4)) // use all available threads
+            .types(types) // only index ruby files
+            .hidden(true) // ignore hidden files
+            .git_ignore(true) // ignore gitignore files
```
vinistock, line 172: We may want this one to be `false`. For example, Prism auto generates the `node.rb` file. When working on Prism, you'd still like to get features for it.

Morriar: The `git_ignore` was a good way to get `node_modules` and stuffs out of the way though.

iirc in the current indexer we have a config file `.index.yml` should we bring that back now so we can configure all of this?

vinistock: Yes, ideally configuration. I'm not sure if the answer is a file for the indexer or a configuration API, but something for sure.

PR #206, opened 2025-10-15
Add `clear-db` option to drop the database before saving the graph

st0012:
When we repeatedly run the indexer on the same directory, the second run will fail because of unique constraint violations.

Currently, we need to manually drop the database before running the indexer, which is annoying. I think having an option to clear the database is useful.

### DIFF
--- a/rust/index/src/main.rs
+++ b/rust/index/src/main.rs
@@ -11,6 +11,7 @@ use index::{
 
 #[derive(Parser, Debug)]
 #[command(name = "index_cli", about = "A Ruby code indexer", version)]
+#[allow(clippy::struct_excessive_bools)]
 struct Args {
     #[arg(value_name = "DIR", default_value = ".")]
     dir: String,
@@ -23,6 +24,9 @@ struct Args {
 
     #[arg(long = "visualize")]
     visualize: bool,
+
+    #[arg(long = "clear-db", help = "Clear the database before saving the graph")]
+    clear_db: bool,
 
     #[arg(long = "stats", help = "Show detailed performance statistics")]
     stats: bool,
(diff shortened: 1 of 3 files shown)


rust/index/src/main.rs
```diff
@@ -24,6 +24,9 @@ struct Args {
     #[arg(long = "visualize")]
     visualize: bool,
 
+    #[arg(long = "clear-db")]
```
Morriar: Let's add a description to what it does, maybe `Drop internal database before indexing` or something

rust/index/src/model/graph.rs
```diff
@@ -325,6 +341,15 @@ impl Graph {
...
     }
 
+    /// Clears all data from the database
+    ///
+    /// # Errors
+    ///
+    /// This method will return an error if clearing the database fails.
+    pub fn clear_database(&self) -> Result<(), Box<dyn Error>> {
+        self.db.clear_database()
+    }
```
Morriar, line 351: It looks like `Graph` is acting more and more as a proxy to the DB object and that might be a smell.

rust/index/src/main.rs
st0012: Done 👍

Morriar: Looks like you pushed a description in between 👍

rust/index/src/model/graph.rs
st0012: Another way is to add a flag to `save_to_database`, like `save_to_database(clear_db: true)`. Then we don't need to expose this API. Which do you prefer?

Morriar: No need to change anything in this PR but I think we should review the coupling between `Graph` and `DB`

st0012: Yeah I agree.

PR #209, opened 2025-10-15
Add content_hash to document struct

jenny-codes:
Breaking up https://github.com/Shopify/index/pull/173 into smaller chunks.

Here we add content_hash to the Document model.

We are choosing to use `u16`​ here for the content_hash as the purpose of this field is for staleness check between two versions of a document, which is a different use case from the IDs that need the full 64 bits.

Collision risk comparison:

(how likely we would fail to detect a change in the document content for the number of files we have)

| Hash Size | Values | 100 Files Collision Risk |
| --- | --- | --- |
| u8 | 256 | 32\.4% |
| u16 | 65,536 | 0\.15% |
| u32 | 4\.3B | 0\.000002% |

For 100 file updates the collision risk is 0.15% which feels acceptable.

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -17,6 +17,7 @@ use std::{
 };
 use std::{sync::mpsc, thread};
 use url::Url;
+use xxhash_rust::xxh3::xxh3_64;
 pub mod errors;
 pub mod ruby_indexer;
 
@@ -57,6 +58,13 @@ impl Document {
             .to_file_path()
             .map_err(|_e| IndexingError::InvalidUri(format!("Couldn't convert URI '{}' to file path", self.uri)))
     }
+
+    #[must_use]
+    pub fn calculate_content_hash(source: &[u8]) -> u16 {
+        // Explicitly take only the lower 16 bits of the hash
+        // This is intentional as this is only used for document staleness check
+        (xxh3_64(source) & 0xFFFF) as u16
+    }
 }
 
 /// Indexes the given items, reading the content from disk and populating the given `Graph` instance.
@@ -73,7 +81,8 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
         let (source, errors) = read_document_source(document);
 
         let converter = UTF8SourceLocationConverter::new(&source);
-        let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter, &source);
+        let content_hash = Document::calculate_content_hash(source.as_bytes());
+        let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter, &source, content_hash);
 
         if errors.is_empty() {
             ruby_indexer.index();
--- a/rust/saturn/src/model/document.rs
+++ b/rust/saturn/src/model/document.rs
@@ -6,13 +6,15 @@ use crate::model::ids::DefinitionId;
 pub struct Document {
     uri: String,
     definition_ids: Vec<DefinitionId>,
+    content_hash: u16,
 }
 
 impl Document {
     #[must_use]
-    pub fn new(uri: String) -> Self {
+    pub fn new(uri: String, content_hash: u16) -> Self {
         Self {
             uri,
+            content_hash,
             definition_ids: Vec::new(),
         }
     }
@@ -20,6 +22,11 @@ impl Document {
     #[must_use]
     pub fn uri(&self) -> &str {
         &self.uri
+    }
+
+    #[must_use]
+    pub fn content_hash(&self) -> u16 {
+        self.content_hash
     }
 
     #[must_use]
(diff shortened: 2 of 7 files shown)


rust/index/src/indexing/ruby_indexer.rs
```diff
@@ -50,7 +51,9 @@ impl<'a> RubyIndexer<'a> {
     #[must_use]
     pub fn new(uri: String, location_converter: &'a dyn SourceLocationConverter, source: &'a str) -> Self {
         let mut local_index = Graph::new();
-        let uri_id = local_index.add_uri(uri);
+        #[allow(clippy::cast_possible_truncation)]
+        let content_hash = xxh3_64(source.as_bytes()) as u16;
```
jenny-codes: Here we just take the lower bits of the result. The truncated hash should still have good distribution.

From the xxh3 author: 
> Feel free select the lower N or upper N bits as you wish, their distributions should be suitable.

https://fastcompression.blogspot.com/2019/03/presenting-xxh3.html

vinistock: I think this hash should be computed in `indexing` outside of the `RubyIndexer` so that we can skip even instantiating it in the case where the content has not been modified.

We can then just pass the already computed hash so that we can associate it with the document instance.

rust/index/src/model/document.rs
```diff
@@ -22,6 +26,15 @@ impl Document {
         &self.uri
     }
 
+    /// # Panics
+    ///
+    /// Panics if the content hash was not provided during initialization.
+    #[must_use]
+    pub fn content_hash(&self) -> u16 {
+        self.content_hash
+            .expect("content_hash needs to be passed in during initialization.")
```
vinistock: If the content hash _must_ be passed during initialization (which I believe is correct, we want to be able to compare document contents even before the user saved), why is it an `Option<u16>` and not just `u16`?

rust/index/src/model/graph.rs
```diff
@@ -128,7 +130,8 @@ impl Graph {
     ///
     /// Any database errors will prevent the data from being loaded
     pub fn load_uri(&mut self, uri: String) -> Result<(), Box<dyn Error>> {
-        let uri_id = self.add_uri(uri);
+        // We do not pass in content hash here because we do not need it during querying
```
vinistock: I think this assumption may not necessarily hold. For example, if the user makes a quick edit and `CMD + Z` to undo, we could potentially fold those two updates into a single operation and by comparing the hashes in memory we would realize that the contents are the same and no reindexing is necessary.

I would vote for us to start out without making the content hash optional and if we later conclude it's not needed, we can optimize.

rust/index/src/indexing/ruby_indexer.rs
jenny-codes: Ah yes it is true I will update this.

jenny-codes: Moved to indexing

rust/index/src/model/graph.rs
jenny-codes: Good point. I have updated the field to be non-optional

rust/saturn/src/model/document.rs
```diff
@@ -22,6 +26,14 @@ impl Document {
         &self.uri
     }
 
+    /// # Panics
+    ///
+    /// Panics if the content hash was not provided during initialization.
```
vinistock: This comment can be removed as we no longer panic.

rust/saturn/src/indexing.rs
```diff
@@ -106,7 +107,9 @@ pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<
                 };
 
                 let converter = UTF8SourceLocationConverter::new(&source);
-                let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter, &source);
+                #[allow(clippy::cast_possible_truncation)]
+                let content_hash = xxh3_64(source.as_bytes()) as u16;
```
vinistock: Nit: maybe we can move this to a function to avoid having to disable the clippy thing in multiple places.

rust/saturn/src/indexing.rs
```diff
@@ -17,6 +17,7 @@ use std::{
 };
 use std::{sync::mpsc, thread};
 use url::Url;
+use xxhash_rust::xxh3::xxh3_64;
```
Morriar, line 20: Looks like Sorbet is using `blake2` for file content hashing, should we also give it try and compare performance?
 https://github.com/sorbet/sorbet/blob/f491ad543992fa822f3e864962d881014a921b2c/common/crypto_hashing/crypto_hashing.h#L21-L24

rust/saturn/src/indexing.rs
jenny-codes: Extracted into indexing::Document::calculate_content_hash

rust/saturn/src/indexing.rs
jenny-codes: As I understand it blake2 is a cryptographic hash which is inherently slower than xxh3? On top of that we would be introducing a new dependency to the project if we choose blake2

PR #210, opened 2025-10-15
Skip indexing unchanged files

jenny-codes:

See the full work at https://github.com/Shopify/saturn/compare/main.../jennyshih/save-updated-graph

This is a different implementation than https://github.com/Shopify/index/pull/173. In the previous version, here is the flow:

```
indexing starts 
-> 1. collect documents from workspace
      1-1 read content from file system
-> 2. *calculate diff against db state*
-> 3. *delete stale entries from db* 
-> 4. index individual files
-> 5. build global graph in memory 
-> 6. *save graph to db*
```

which has a few issues

1. To do the staleness check we need to read the document source from the filesystem before going into the parallel indexing (step 4), which can be a bottleneck for us down the road.
2. It attempts to save all entries into the database, including the ones that remain unchanged which is wasteful and a bit awkward.

This new attempt here is as follows:

```
-> 1. collect documents from workspace 
-> 2. index individual files
         2-1 read content from file system
         2-2 calculate diff
         2-3 exit early if same entry exists in db
-> 3. build global graph in memory (merging local index into global)
         3-1 when clearing global index duplicate uri data, clear the db too (not yet implemented)
         3-2 before adding the new data to memory, write to db (not yet implemented)
-> 4. delete the deleted entries from the db (not yet implemeneted) 
```

So for our four kinds of uris:

1. new uris: they will be saved individually during parallel indexing
2. updated uris: their data will be deleted from the db and re-inserted during parallel indexing
3. existing uris: we will not attempt to re-index them nor save them to db again. We will do nothing about them.
4. deleted uris: they will be cleaned up when we finish the parallel indexing of the present uris.

This pr implements the part where it looks up the db document entries and use it to inform the indexing worker to skip a file when it has not been changed.

Prs to come

1. Save the new/updated entries to db after indexing
2. Delete the entries in db that are no longer in the filesystem

### Benchmark

Skipping the unchanged files reduced the db save time to 0 when we run the index twice on the same repo, but does not really affect the indexing time.

#### Before

```
Query statistics

Total declarations:         684127
With documentation:         71130 (10.4%)
Without documentation:      612997 (89.6%)
Total documentation size:   211400 bytes
Multi-definition names:     14691 (2.1%)

Definition breakdown:
  Method               315306
  InstanceVariable     178723
  Module               175442
  Class                100527
  Constant              60735
  AttrReader            31333
  AttrAccessor          11458
  AttrWriter              370
  GlobalVariable           36
  ClassVariable            11

Timing breakdown
  Initialization      0.001s (  0.0%)
  Listing             1.440s ( 11.6%)
  Indexing            2.439s ( 19.7%)
  Querying            0.223s (  1.8%)
  Database            8.276s ( 66.9%)
  Total:             12.380s

Maximum RSS: 657997824 bytes (627.52 MB)

Indexed 85386 files
Found 684127 names
Found 873941 definitions
Found 85386 URIs
```

#### After

```
Query statistics

Total declarations:         1
With documentation:         0 (0.0%)
Without documentation:      1 (100.0%)
Total documentation size:   0 bytes
Multi-definition names:     0 (0.0%)

Definition breakdown:

Timing breakdown
  Initialization      0.002s (  0.0%)
  Listing             3.318s ( 58.9%)
  Indexing            2.300s ( 40.9%) 
  Querying            0.000s (  0.0%)
  Database            0.009s (  0.2%) <- this is for the querying
  Total:              5.630s

Maximum RSS: 65519616 bytes (62.48 MB)

Indexed 0 files
Found 1 names
Found 0 definitions
Found 0 URIs
```

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -1,7 +1,7 @@
 use crate::{
     errors::{Errors, MultipleErrors},
     indexing::ruby_indexer::{IndexerParts, RubyIndexer},
-    model::graph::Graph,
+    model::{graph::Graph, ids::UriId},
     source_location::UTF8SourceLocationConverter,
 };
 use glob::glob;
@@ -33,20 +33,24 @@ pub enum IndexResult {
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
 pub fn index_in_parallel(graph: &mut Graph, file_paths: Vec<String>) -> Result<(), MultipleErrors> {
+    if graph.is_db_initialized() {
+        return index_and_sync(graph, file_paths);
+    }
+
     let index_document = |file_path: &String| -> IndexResult {
         let (source, errors) = read_document_source(file_path);
         if !errors.is_empty() {
             return IndexResult::Errored(errors);
         }
 
+        let uri = match uri_from_file_path(file_path) {
+            Ok(uri) => uri,
+            Err(error) => return IndexResult::Errored(vec![error]),
+        };
+
         let converter = UTF8SourceLocationConverter::new(&source);
         let content_hash = calculate_content_hash(source.as_bytes());
-        let mut ruby_indexer = RubyIndexer::new(
-            uri_from_file_path(file_path).unwrap(),
-            &converter,
-            &source,
-            content_hash,
-        );
+        let mut ruby_indexer = RubyIndexer::new(uri, &converter, &source, content_hash);
         ruby_indexer.index();
         IndexResult::Completed(Box::new(ruby_indexer.into_parts()))
     };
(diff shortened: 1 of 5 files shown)


rakelib/index_top_gems.rake
```diff
@@ -51,6 +51,7 @@ task index_top_gems: :compile_release do
 
             # Index the gem's files and yield errors back to the main Ractor
             graph = Saturn::Graph.new
+            graph.set_configuration(File.join(gem_dir, "graph.db"))
```
Morriar: Do we need the database for this task? Do we expect to find the same files multiple times?

rust/saturn/src/model/db.rs
```diff
@@ -279,6 +284,9 @@ impl Db {
         ))?;
 
         for (declaration_id, declaration) in graph.declarations() {
+            if declaration.name() == "<main>" {
```
Morriar: This string comparison on the name is unfortunate as it runs on each iteration.

Can we extract this declaration as a constant and compare the pointers instead?

test/graph_test.rb
```diff
@@ -9,13 +9,15 @@ class GraphTest < Minitest::Test
   def test_indexing_empty_context
     with_context do |context|
       graph = Saturn::Graph.new
+      graph.set_configuration("graph.db")
```
Morriar: Why do we need the db for all these tests?

rust/saturn/src/indexing.rs
```diff
@@ -79,22 +89,33 @@ impl Document {
...
+        let converter = UTF8SourceLocationConverter::new(&source);
+        let mut ruby_indexer = RubyIndexer::new(uri, &converter, &source, content_hash);
         ruby_indexer.index();
-        ruby_indexer.into_parts()
+        IndexResult::Completed(Box::new(ruby_indexer.into_parts()))
     };
 
+    if !graph.is_db_initialized() {
+        return Err(MultipleErrors(vec![IndexingError::DatabaseNotInitialized]));
+    }
```
Morriar: We shouldn't force the use of the database. It's quite costly for scripts that run only once.

jenny-codes: Would it be better if we fallback to use in-memory if the database path is not set?

Morriar: Even in memory database has a cost we shouldn't have to pay for one off uses.

jenny-codes: What do we intend to use this index_in_parallel function for? The underlying motivation to have the db operations here is the idea that we always want to offload the index data to the cache whenever possible.

If we do not want to enforce a database, I can change it so that we only interact with the db when the db is initialized?

Morriar: > we always want to offload the index data to the cache

Not always, only if a database is specified.

If I run a script only once, for example on CI, there is no point in going through the database.

rust/saturn/src/indexing.rs
```diff
@@ -69,6 +69,12 @@ impl Document {
     }
 }
 
+pub enum IndexResult {
+    Completed(Box<IndexerParts>),
```
jenny-codes, line 21: We want to box this one because the byte differences between the three variants are big.

PR #212, opened 2025-10-16
Fix URI to file path conversion for indexing

### LINKED ISSUE #193: Fix URI to file path conversion for indexing
This code

https://github.com/Shopify/index/blob/312d4c51bcb689d49b2030f1db87f847f75f1972/rust/index/src/indexing.rs#L51

is not correct because it does not unescape certain parts of the URI path and ends up resulting in an invalid file path. We should be using `to_file_path`, which returns a `Result<Path>` and fails if the URI is not a `file://` URI.

alexcrocha:
Resolves #193

### DIFF
--- a/rust/index/src/indexing.rs
+++ b/rust/index/src/indexing.rs
@@ -11,7 +11,10 @@ use std::sync::{
     Arc, Mutex,
     mpsc::{Receiver, Sender},
 };
-use std::{fs, path::Path};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 use std::{sync::mpsc, thread};
 use url::Url;
 pub mod errors;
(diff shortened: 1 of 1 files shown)


rust/index/src/indexing.rs
```diff
@@ -46,9 +49,13 @@ impl Document {
         })
     }
 
-    #[must_use]
-    pub fn path(&self) -> &str {
-        self.uri.path()
+    /// # Errors
+    ///
+    /// Errors if the URI is not a valid file:// URI
```
alexcrocha, line 54: Addressing [missing errors doc](https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc) and [double must use](https://rust-lang.github.io/rust-clippy/master/index.html#double_must_use) linting errors here

PR #215, opened 2025-10-17
Add content_hash to load_uri query

jenny-codes:
This is to ensure we will always have a content hash for the Document
model.

### DIFF
--- a/rust/saturn/src/model/db.rs
+++ b/rust/saturn/src/model/db.rs
@@ -14,7 +14,12 @@ const TABLE_DOCUMENTS: &str = "documents";
 
 const ALL_TABLES: &[&str] = &[TABLE_DEFINITIONS, TABLE_DECLARATIONS, TABLE_DOCUMENTS];
 
-pub struct LoadResult {
+pub struct DocumentData {
+    pub content_hash: u16,
+    pub definitions: Vec<DefinitionData>,
+}
+
+pub struct DefinitionData {
     pub declaration_id: DeclarationId,
     pub name: String,
     pub definition_id: DefinitionId,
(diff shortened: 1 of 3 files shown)


rust/saturn/src/model/db.rs
```diff
@@ -87,28 +92,41 @@ impl Db {
...
-                        definition,
-                    })
-                })?
-                .collect::<Result<Vec<_>, _>>()
-                .map_err(std::convert::Into::into)
+            let tx = connection.transaction()?;
+
+            let content_hash: u16 = tx
+                .prepare_cached("SELECT content_hash FROM documents WHERE id = ?")?
+                .query_row([*uri_id], |row| row.get(0))?;
```
vinistock: Could we load the content_hash in the same `load_uri.sql` query? Or is there a reason for this part being separate?

jenny-codes: The separate query is to prevent the edge case where a document does not contain any definitions, in which case no content_hash would be returned. 

But we can also get around it by trying to get it from the load_query first, and if the document is indeed empty (no rows are returned) we do a separate query to get the content hash. Would that be better?

vinistock: Would a `LEFT JOIN` fix the issue? I think it might be a result of using an implicit `INNER JOIN`, but we can change that.

jenny-codes: ...for some reason LEFT JOIN didn't occur to me. Good call. Updating.

jenny-codes: Updated, if you want to give it a second look

PR #216, opened 2025-10-17
Format comments

jenny-codes:
Add backticks for proper documentation

### DIFF
--- a/rust/index/src/model/id.rs
+++ b/rust/index/src/model/id.rs
@@ -11,8 +11,8 @@ use xxhash_rust::xxh3::xxh3_64;
 
 /// A stable ID representation using i64.
 ///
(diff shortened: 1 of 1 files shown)


rust/index/src/model/id.rs
```diff
@@ -11,8 +11,8 @@ use xxhash_rust::xxh3::xxh3_64;
 
 /// A stable ID representation using i64.
 ///
-/// We use i64 instead of u64 because SQLite doesn't support unsigned integers.
-/// IDs are generated from xxh3_64 hashes (u64) then cast to i64, preserving all bits.
+/// We use i64 instead of u64 because `SQLite` doesn't support unsigned integers.
+/// IDs are generated from ``xxh3_64`` hashes (u64) then cast to i64, preserving all bits.
```
vinistock: Do we need double backticks? Or just single?

PR #217, opened 2025-10-17
Use unqualified NameIDs for tracking constant references

vinistock:
This PR continues the work to unlock constant resolution. This one adds an identity map of unqualified names using the `NameId` from #213 and starts using it for constant references.

With unqualified Name IDs + listing namespace members (next PR), we can actually resolve constant references based on lexical scope - unblocking a good chunk of the resolution phase.

**Note**: there's not much difference in terms of indexing time from this change, but memory was reduced quite a bit. Maybe we can store method parameter names as unqualified name ids too?

<details>
<summary>Benchmark on Core</summary>

```
BEFORE

Initialization      0.001s (  0.0%)
Listing             1.234s ( 34.1%)
Indexing            2.156s ( 59.5%)
Querying            0.230s (  6.4%)
Total:              3.621s

Indexed 87014 files
Found 700824 names
Found 894881 definitions
Found 87014 URIs
Maximum Resident Set Size: 708247552 bytes (675.43 MB)
Peak Memory Footprint:     685114280 bytes (653.37 MB)
Execution Time:            3.86 seconds

AFTER

Initialization      0.001s (  0.0%)
Listing             1.230s ( 33.4%)
Indexing            2.173s ( 59.0%)
Querying            0.277s (  7.5%)
Total:              3.681s

Indexed 87014 files
Found 700824 names
Found 894881 definitions
Found 87014 URIs
Maximum Resident Set Size: 577536000 bytes (550.78 MB)
Peak Memory Footprint:     551961392 bytes (526.39 MB)
Execution Time:            3.90 seconds
```

</details>

### DIFF
--- a/rust/saturn/src/model/references.rs
+++ b/rust/saturn/src/model/references.rs
@@ -1,5 +1,5 @@
 use crate::{
-    model::ids::{DeclarationId, UriId},
+    model::ids::{NameId, UriId},
     offset::Offset,
 };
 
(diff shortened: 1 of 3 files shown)


rust/saturn/src/model/references.rs
```diff
@@ -21,30 +21,30 @@ pub enum UnresolvedReference {
 // considered an unresolved constant
 #[derive(Debug)]
 pub struct UnresolvedConstantRef {
-    name: String,
-    nesting: Vec<DeclarationId>,
+    name_id: NameId,
+    nesting: Vec<NameId>,
```
Morriar, line 28: We could update the comment to explain what each field will contain for a reference

vinistock: Added comments for each field. I also made the document use 3 slashes so that it appears on hover.

rust/saturn/src/model/references.rs
```diff
@@ -8,43 +8,48 @@ pub enum UnresolvedReference {
...
+/// considered an unresolved constant
 #[derive(Debug)]
 pub struct UnresolvedConstantRef {
-    name: String,
-    nesting: Vec<DeclarationId>,
+    /// The unqualified name of the constant
+    name_id: NameId,
+    /// The nesting where we found the constant reference. This is a list of unqualified name IDs, so that we can
+    /// traverse the graph through the member relationships
+    nesting: Vec<NameId>,
```
Morriar, line 28: Should we make it clear that `nesting` doesn't include `name_id`?

So for

```rb
module Foo
  module Bar
    BAZ
  end
end
```

`name_id`: `BAZ`
`nesting`: [`Foo`, `Bar`]

vinistock: Sure https://github.com/Shopify/saturn/pull/224. I added an example to the struct.

PR #218, opened 2025-10-17
Add members mapping to declarations

vinistock:
Another step towards constant reference resolution. This PR adds the `members` mapping to the global declarations, so that we can keep track of what is "owned" by each namespace.

For example

```ruby
module Foo
  class Bar
    def baz; end
  end
end
```

In this case, `Foo` is a member of `Object` (which owns all top level constants), `Bar` is a member of `Foo` and `baz` is a member of `Bar`.

The connection is made with `NameId` (unqualified name) to `DeclarationId` (fully qualified name). This allows us to traverse the graph checking if something is defined at each step - which enables resolution.

**Note**: this is currently not being used. The reason is that we need more changes to the indexer to track a stack of owners. We need to remember which namespace we're currently inside to add new definitions as members. We don't currently track that, so I will follow up with a PR to do so, which allows us to populate the `members` map.

### DIFF
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -1,9 +1,22 @@
-use crate::model::ids::DefinitionId;
+use crate::model::{
+    identity_maps::IdentityHashMap,
+    ids::{DeclarationId, DefinitionId, NameId},
+};
 
+/// A `Declaration` represents the global concept of an entity in Ruby. For example, the class `Foo` may be defined 3
+/// times in different files and the `Foo` declaration is the combination of all of those definitions that contribute to
+/// the same fully qualified name
 #[derive(Debug)]
 pub struct Declaration {
+    /// The fully qualified name of this declaration
     name: String,
+    /// The list of definition IDs that compose this declaration
     definition_ids: Vec<DefinitionId>,
+    /// The entities that are owned by this declaration. For example, constants and methods that are defined inside of
+    /// the namespace. Note that this is a hashmap of unqualified name IDs to declaration IDs. That assists the
+    /// traversal of the graph when trying to resolve constant references or trying to discover which methods exist in a
+    /// class
+    members: IdentityHashMap<NameId, DeclarationId>,
 }
 
 impl Declaration {
@@ -12,6 +25,7 @@ impl Declaration {
         Self {
             name,
             definition_ids: Vec::new(),
+            members: IdentityHashMap::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


rust/saturn/src/model/declaration.rs
Morriar: With both `definitions` and `members`, the `is_empty` method is confusing now. What do you think of either renaming it or deleting it as a follow up?

rust/saturn/src/model/declaration.rs
```diff
@@ -54,6 +68,14 @@ impl Declaration {
 
         self.definition_ids.push(definition_id);
     }
+
+    pub fn add_member(&mut self, name_id: NameId, declaration_id: DeclarationId) {
```
Morriar, line 72: Follow up: Can we also reorder `add_definition` and `remove_definition` to follow the same pattern?

rust/saturn/src/model/declaration.rs
vinistock: Can you explain why you find it confusing, so that we can find a better name?

We still need the method. Also, if all definitions got deleted (which is when this method is used), then there cannot be any members either (every definition got removed by the user). In a sense, everything is empty.

Happy to rename though.

st0012: >  if all definitions got deleted (which is when this method is used), then there cannot be any members either

Is it possible to add a `debug_assert` somewhere to enforce this assumption?
I'd also be helpful to add this as a comment for this method.

Morriar: > Also, if all definitions got deleted (which is when this method is used), then there cannot be any members either

But you can have a declaration with a definition that doesn't have any member right?

```rb
# file.rb#1-23
class Foo; end
```

So it would be not be `empty` because it has one definition yet contain no members and thus be "empty"?

Morriar: Do we actually need the method? Can't the client decide to call either `definitions.empty()` or `members.empty()` depending on what it needs?

vinistock: I renamed to `has_no_definitions` to be explicit. I would still rather keep the method instead of making the fields of `Declaration` public, since we need to check this in the graph to decide whether the entire declaration node should be removed.

vinistock: > s it possible to add a debug_assert somewhere to enforce this assumption?

The assert would be fail because we delete the entire node from the graph in one go, making the `members` hash be erased. And since the definitions for the members also got deleted, they follow a similar pattern.

PR #219, opened 2025-10-17
Prototype: C API for declarations and definitions

Morriar:
See Slack conversation before reviewing: https://shopify.slack.com/archives/C014FP56KSP/p1760742959204659

### DIFF
--- a/ext/index/definition.c
+++ b/ext/index/definition.c
@@ -43,10 +43,92 @@ static VALUE rb_definition_kind(VALUE self) {
     return str;
 }
 
+// DefinitionHandle#name -> String
+static VALUE rb_definition_name(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    const char *name = idx_definition_name(graph, (int64_t)data->id);
+    if (name == NULL) {
+        return Qnil;
+    }
+    VALUE str = rb_utf8_str_new_cstr(name);
+    free_c_string(name);
+    return str;
+}
+
+// DefinitionHandle#uri_path -> String
+static VALUE rb_definition_uri_path(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    const char *uri = idx_definition_uri_path(graph, (int64_t)data->id);
+    if (uri == NULL) {
+        return Qnil;
+    }
+    VALUE str = rb_utf8_str_new_cstr(uri);
+    free_c_string(uri);
+    return str;
+}
+
+// DefinitionHandle#start_location -> Integer
+static VALUE rb_definition_start_location(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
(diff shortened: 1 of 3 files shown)


ext/index/graph.c
```diff
@@ -95,11 +100,47 @@ static VALUE rb_graph_set_configuration(VALUE self, VALUE db_path) {
...
+    Check_Type(name, T_STRING);
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+    int64_t id = idx_graph_resolve_name(graph, StringValueCStr(name));
+
+    if (id == 0) {
+        return Qnil;
+    }
+
+    return rb_funcall(cDeclaration, rb_intern("new"), 2, self, LL2NUM(id));
```
vinistock, line 134:
```suggestion
    VALUE args[] = { self, LL2NUM(id) };
    return rb_class_new_instance(2, args, cDeclaration);
```

ext/index/declaration.c
```diff
@@ -0,0 +1,75 @@
...
+
+static VALUE rb_declaration_alloc(VALUE klass) {
+    HandleData *data = ALLOC(HandleData);
+    data->graph_obj = Qnil;
+    data->id = 0;
+
+    return TypedData_Wrap_Struct(klass, &handle_type, data);
+}
+
+static VALUE rb_declaration_initialize(VALUE self, VALUE graph_obj,
```
vinistock, line 17: One downside of this approach is that the ID concept kind of leaks through the Ruby layer since developers can invoke `Declaration.new`, but they don't have access to the internal IDs in the graph to pass.

Unfortunately, there's no way to pass arguments to the custom allocator to get around this.

ext/index/declaration.c
```diff
@@ -0,0 +1,75 @@
...
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    size_t len = 0;
+    const int64_t *def_ids =
+        idx_declaration_definition_ids(graph, (int64_t)data->id, &len);
+    VALUE arr = rb_ary_new_capa((long)len);
+    for (size_t i = 0; i < len; i++) {
+        VALUE handle = rb_funcall(cDefinition, rb_intern("new"), 2,
+                                  data->graph_obj, LL2NUM(def_ids[i]));
```
vinistock, line 60:
```suggestion
        VALUE args[] = { data->graph_obj, LL2NUM(def_ids[i]) };
        VALUE handle = rb_class_new_instance(2, args, cDefinition);
```

ext/index/graph.c
```diff
@@ -95,11 +100,47 @@ static VALUE rb_graph_set_configuration(VALUE self, VALUE db_path) {
...
+static VALUE rb_graph_declarations(VALUE self) {
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    size_t len = 0;
+    const int64_t *ids = idx_graph_declaration_ids(graph, &len);
+    VALUE arr = rb_ary_new_capa((long)len);
+    for (size_t i = 0; i < len; i++) {
+        VALUE handle =
+            rb_funcall(cDeclaration, rb_intern("new"), 2, self, LL2NUM(ids[i]));
```
vinistock, line 114:
```suggestion
        VALUE args[] = { self, LL2NUM(ids[i]) };
        VALUE handle = rb_class_new_instance(2, args, cDeclaration);
```

ext/index/graph.c
```diff
@@ -95,11 +100,47 @@ static VALUE rb_graph_set_configuration(VALUE self, VALUE db_path) {
     return Qnil;
 }
 
+// Graph#declarations: () -> Array[Declaration]
+// Returns an array of all declarations in the graph
+static VALUE rb_graph_declarations(VALUE self) {
```
vinistock, line 105: In terms of API, I wonder if we should instead provide an iterator to avoid instantiating all objects upfront.

PR #220, opened 2025-10-20
Refactor index_in_parallel

jenny-codes:
This is an attempt to separate the thread worker logic from the actual indexing to make this function easier to reason about

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -69,68 +69,10 @@ impl Document {
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing.rs
```diff
@@ -69,6 +69,69 @@ impl Document {
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
+    let results = with_parallel_workers(documents, index_document);
+
+    let mut all_errors = Vec::new();
+    for (local_graph, errors) in results {
+        graph.update(local_graph);
+        all_errors.extend(errors);
```
vinistock, line 77: This separation here means that we're no longer merging results as they come in from the threads. We first accumulate everything in a vector and then merge all indexers in one go.

Did you benchmark to check if there's any performance impact?

jenny-codes: Good point. That was not the intention. I opened https://github.com/Shopify/index/pull/222 to address this

PR #222, opened 2025-10-20
Preserve parallel merging

jenny-codes:
https://github.com/Shopify/index/pull/220 did a refactoring of the
index_in_parallel function that changed the result merging into
sequential instead of in parallel. This pr makes the changes to preserve
that behavior.

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -69,37 +69,26 @@ impl Document {
 /// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
 /// should not be any code that tries to lock the same mutex multiple times in the same thread
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing.rs
```diff
@@ -69,19 +69,9 @@ impl Document {
...
-        graph.update(local_graph);
-        all_errors.extend(errors);
-    }
-
-    if all_errors.is_empty() {
-        Ok(())
-    } else {
-        Err(MultipleErrors(all_errors))
-    }
+    with_parallel_workers(graph, documents, index_document, |global_graph, local_graph| {
```
vinistock: Is it possible to just use the `graph` reference in the body of the closure? Or does Rust complain about that?

I mean this:

```rust
with_parallel_workers(graph, documents, index_document, |local_graph| {
  graph.update(local_graph);
})
```

jenny-codes: Updated. Thanks!

PR #226, opened 2025-10-20
Make result local graph optional

jenny-codes:
Currently when an error happens during file reading we will initialize a
RubyIndexer object, store the file reading errors, skip the indexing,
and return the empty local graph and the list of errors.

We can simplify it by making the local graph an optional return value,
and only return it when there is no error from the file reading step.
This way we save an RubyIndexer initialization as well as an attempt to
merge an empty local graph back to the global one.

This is also paving the way for the incremental sync which will start
skipping the indexing of existing files whose content have not changed.

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -79,19 +79,14 @@ impl Document {
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
     let index_document = |document: &Document| -> IndexerParts {
         let (source, errors) = read_document_source(document);
+        if !errors.is_empty() {
+            return (None, errors);
+        }
 
         let converter = UTF8SourceLocationConverter::new(&source);
         let content_hash = Document::calculate_content_hash(source.as_bytes());
         let mut ruby_indexer = RubyIndexer::new(document.uri.to_string(), &converter, &source, content_hash);
-
-        if errors.is_empty() {
-            ruby_indexer.index();
-        } else {
-            for error in errors {
-                ruby_indexer.add_error(error);
-            }
-        }
-
+        ruby_indexer.index();
         ruby_indexer.into_parts()
     };
 
@@ -143,10 +138,12 @@ where
 
         let handle = thread::spawn(move || {
             while let Some(document) = { queue.lock().unwrap().pop() } {
-                let result = thread_fn(&document);
-                thread_tx
-                    .send(result)
-                    .expect("Receiver end should not be closed until all threads are done");
+                let (result, errors) = thread_fn(&document);
+                if result.is_some() || !errors.is_empty() {
+                    thread_tx
+                        .send((result, errors))
+                        .expect("Receiver end should not be closed until all threads are done");
+                }
             }
         });
 
@@ -156,8 +153,10 @@ where
     drop(tx);
 
     let mut all_errors = Vec::new();
-    for (local_graph, errors) in rx {
-        result_fn(local_graph);
+    for (result, errors) in rx {
+        if let Some(local_graph) = result {
+            result_fn(local_graph);
+        }
         all_errors.extend(errors);
     }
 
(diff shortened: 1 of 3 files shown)


rust/saturn/src/indexing.rs
```diff
@@ -79,19 +79,14 @@ impl Document {
 pub fn index_in_parallel(graph: &mut Graph, documents: Vec<Document>) -> Result<(), MultipleErrors> {
     let index_document = |document: &Document| -> IndexerParts {
         let (source, errors) = read_document_source(document);
+        if !errors.is_empty() {
+            return (None, errors);
```
vinistock, line 83: Instead of dealing with optional returns, can't we just skip to the next iteration without transmitting anything in the channel?

If the refactor from #220 made it impossible to skip iterations and forces us to transmit empty values due to indirection, then I would vote that we go back to the original implementation. It's longer, but it's straight forward.

rust/saturn/src/indexing.rs
```diff
@@ -143,10 +138,12 @@ where
 
         let handle = thread::spawn(move || {
             while let Some(document) = { queue.lock().unwrap().pop() } {
-                let result = thread_fn(&document);
-                thread_tx
-                    .send(result)
-                    .expect("Receiver end should not be closed until all threads are done");
+                let (result, errors) = thread_fn(&document);
+                if result.is_some() || !errors.is_empty() {
```
st0012, line 142: Is it possible to have no result and also no errors?
If there is, can we have a test case covering it? Asking because if I removed this condition, all tests still pass.

rust/saturn/src/indexing.rs
jenny-codes: This is the same behaviour with the existing code though. When a file reading error happens we will send the errors back to the main thread with an empty graph. The difference is that now we do not add the (unindexed) uri to the graph, which imo is the behavior we want because once we get to the part where we start skipping existing unchanged uris, it can prevent us from skipping the ones that were not properly indexed originally?

jenny-codes: > just skip to the next iteration without transmitting anything in the channel?

That is the goal/new behaviour of this pr. The skipping logic is inside the thread-handling function. Currently nothing is skipped because the content hash comparison is not implemented yet, but the idea is if this function returns an empty graph and no errors, that means we are voluntarily skipping the indexing of this uri and we will not send anything back to the main thread.

rust/saturn/src/indexing.rs
jenny-codes: Currently the scenario does not happen, but the idea is that when we start skipping the indexing of existing unchanged files this condition will be false for the skipped files. I will add a test case for it.

rust/saturn/src/indexing.rs
vinistock: With this new code structure, will we be able to compare document hashes and, if they are the same, skip both indexing or transmitting anything through channels?

Or will this structure require that we transmit something even if no indexing happened and no errors are present?

jenny-codes: We will be able to skip the channel transmitting for unchanged documents because now we only send the data conditionally

```rust
                if result.is_some() || !errors.is_empty() {
                    thread_tx
                        .send((result, errors))
                        .expect("Receiver end should not be closed until all threads are done");
                }
```

rust/saturn/src/indexing.rs
```diff
@@ -79,19 +79,14 @@ impl Document {
...
-
-        if errors.is_empty() {
-            ruby_indexer.index();
-        } else {
-            for error in errors {
-                ruby_indexer.add_error(error);
-            }
-        }
-
+        ruby_indexer.index();
```
Morriar, line 89: This PR made me realize we never surface anything about parse errors: https://github.com/Shopify/saturn/blob/main/rust/saturn/src/indexing/ruby_indexer.rs#L81-L83

PR #231, opened 2025-10-21
Extract `utils.h`

Morriar:
Let's have a place we can share functions across C files

### DIFF
--- a/ext/saturn/utils.h
+++ b/ext/saturn/utils.h
@@ -1,0 +1,32 @@
+#ifndef SATURN_UTILS_H
+#define SATURN_UTILS_H
+
+#include "ruby.h"
+
+// Convert a Ruby array of strings into a double char pointer so that we can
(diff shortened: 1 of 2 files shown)


ext/saturn/utils.h
```diff
@@ -0,0 +1,32 @@
+#ifndef SATURN_UTILS_H
+#define SATURN_UTILS_H
+
+#include "ruby.h"
+
+// Convert a Ruby array of strings into a double char pointer so that we can
+// pass that to Rust. This copies the data so it must be freed
+static char **str_array_to_char(VALUE array, size_t length) {
```
vinistock, line 8: What's the convention in C for when to put implementation in a header file vs the C file? Is there some rule we should follow?

Morriar: I think the best way is to make them non-static and share the implementation through a .c file, see: https://github.com/Shopify/saturn/pull/236

PR #237, opened 2025-10-21
Resolve top level constant references

### LINKED ISSUE #197: Resolve top-level constants

st0012:
Closes #197

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -8,7 +8,7 @@ use crate::model::definitions::{
 };
 use crate::model::graph::Graph;
 use crate::model::ids::{DeclarationId, NameId, UriId};
-use crate::model::references::{ConstantRef, UnresolvedReference};
+use crate::model::references::{ConstantReference, UnresolvedReference};
 use crate::offset::Offset;
 use crate::source_location::SourceLocationConverter;
 
@@ -321,7 +321,7 @@ impl<'a> RubyIndexer<'a> {
             is_top_level = true;
         }
 
-        let name_id = self.local_graph.add_name(name.clone());
+        let name_id = self.local_graph.add_name(name);
         let name_id_nesting = if is_top_level {
             vec![]
         } else {
@@ -333,7 +333,7 @@ impl<'a> RubyIndexer<'a> {
             nesting.iter().map(NameId::from).collect()
         };
 
-        let reference = UnresolvedReference::Constant(Box::new(ConstantRef::new(
+        let reference = UnresolvedReference::Constant(Box::new(ConstantReference::new(
             name_id,
             name_id_nesting,
             self.uri_id,
--- a/rust/saturn/src/model/references.rs
+++ b/rust/saturn/src/model/references.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 
 #[derive(Debug)]
-pub struct ConstantRef {
+pub struct ConstantReference {
     /// The unqualified name of the constant
     name_id: NameId,
     /// The nesting where we found the constant reference. This is a list of unqualified name IDs, so that we can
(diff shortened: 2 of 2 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -172,6 +205,7 @@ impl Graph {
     /// different threads, but not meant to handle updates to the existing global representation
     pub fn extend(&mut self, incomplete_index: Graph) {
         self.definitions.extend(incomplete_index.definitions);
+        self.names.extend(incomplete_index.names);
```
st0012, line 208: I think names should be extended too, so this is technically a bug fix?

rust/saturn/src/model/declaration.rs
```diff
@@ -17,6 +18,8 @@ pub struct Declaration {
     /// traversal of the graph when trying to resolve constant references or trying to discover which methods exist in a
     /// class
     members: IdentityHashMap<NameId, DeclarationId>,
+    /// The list of references that are made to this declaration
+    references: Vec<ResolvedReference>,
```
Morriar, line 22: I guess we need this for the `Show references` feature?

I wonder how this will scale for declarations such as `Object`?

st0012: This will also be used for rename symbol.

> I wonder how this will scale for declarations such as Object?

Can you elaborate?

Morriar: Since all the classes inherit implicitly from `Object`, _if_ we were to make this relationship more explicit the number of references to `Object` would increase a lot?

For explicit relationships, I can imagine some names like `Enumerable`, `ApplicationController`, etc. getting a lot of references to them.

If we store all of them all the time, it may become heavy and this can become a problem for the ancestors used most frequently as we may not be able to offload them to the database.

We can revisit later but maybe a flat list indexed by name would be better?

st0012: I don't think references should take inheritance into consideration? What you described will happen if we were to support something like `Show superclasses`, but resolving references is not for that AFIAK.

Morriar: Now I'm confused, let's say I have this:

```rb
class A; end
module M; end

class B < A
  include M
end
```

If I was to rename `A` or `M`, shouldn't we use the references for that?

st0012: We don't store anything under `Object` even in this case tho. The current approach is to store `A`'s references under `A`'s declaration. So if we were to rename `A`, the steps will probably be:

1. Find `A` declaration
2. Get all of  `A`'s references
3. Perform rename on `A`'s references
4. Re-index associated uris

Morriar: So for `B < A`, we do store a reference from `B` to `A`? Same for `include M`, we store a reference from `B` to `M`?

> I don't think references should take inheritance into consideration?

Your comment seemed to imply we were not?

st0012: The references added in this PR is simply direct constant reference. So in `class B < A`, `B` is not a reference to `A`, while the literal `A` is what we store.
We may include `B` as a reference to `A` in the future if we want to include inheritance concept into "reference". But since we're not even resolving ancestors, I don't think we should or can consider that atm.

Morriar: Had a chat with Stan, we confirmed that `references` will be use for any explicit reference to a declaration.

So considering this example:

```rb
class A::B < A::C
  include A::M
end
```

There are 3 references to the declaration `A`: 

1. As part of the class name `A::B`
2. As part of the superclass name `A::C`
3. As part of the included module name `A::M`

On the other hand we would not store implicit references so this doesn't add a reference from `Foo` to `Object` and asking for the references to `Object` wouldn't list `Foo`.

```rb
class Foo; end
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -313,16 +313,27 @@ impl<'a> RubyIndexer<'a> {
...
-        let name_id_nesting: Vec<NameId> = nesting.iter().map(NameId::from).collect();
+        let mut name = Self::location_to_string(location);
+        let mut is_top_level = false;
+
+        if name.starts_with("::") {
+            name = name.strip_prefix("::").unwrap().to_string();
+            is_top_level = true;
+        }
+
+        let name_id = self.local_graph.add_name(name.clone());
```
Morriar: It doesn't look like we use `name` later, do we need to clone?

rust/saturn/src/model/references.rs
```diff
@@ -33,7 +16,29 @@ pub struct UnresolvedConstantRef {
...
+    /// `[NameId::from("Foo")]`.
+    Constant(Box<ConstantRef>),
+}
+
+#[derive(Debug)]
+pub enum ResolvedReference {
+    Constant(Box<ConstantRef>),
+}
+
+impl ConstantRef {
```
Morriar: If everything else says `Reference` I'd also rename this one to be a `ConstantReference`

rust/saturn/src/model/graph.rs
Morriar: I think you're right

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Good catch. This was a leftover from debugging.

PR #238, opened 2025-10-21
Add query for content hash

jenny-codes:
Add query to get a set of content hashes by uri from db. This will be
used later for staleness check.

### DIFF
--- a/rust/saturn/src/model/db.rs
+++ b/rust/saturn/src/model/db.rs
@@ -6,6 +6,8 @@ use crate::model::{
 use rusqlite::{Connection, params};
 use std::{cell::RefCell, error::Error, fs, path::Path};
 
+use super::identity_maps::IdentityHashMap;
+
 const SCHEMA_VERSION: u16 = 1;
 
 const TABLE_DEFINITIONS: &str = "definitions";
@@ -183,6 +185,19 @@ impl Db {
 
             tx.commit()?;
             Ok(())
+        })
+    }
+
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn get_all_content_hashes(&self) -> Result<IdentityHashMap<UriId, u16>, Box<dyn Error>> {
+        self.with_connection(|connection| {
+            let mut statement = connection.prepare("SELECT id, content_hash FROM documents")?;
+            statement
+                .query_map([], |row| Ok((row.get(0)?, row.get(1)?)))?
+                .collect::<Result<IdentityHashMap<UriId, u16>, _>>()
+                .map_err(Into::into)
         })
     }
 
@@ -427,5 +442,25 @@ mod tests {
             .query_row((), |row| row.get(0))
             .unwrap();
         assert_eq!(name_count, 0,);
+    }
+
+    #[test]
+    fn querying_content_hashes_by_uri() {
+        let mut context = GraphTest::new();
+
+        context.index_uri("file:///first.rb", "module FirstModule; end");
+        context.index_uri("file:///second.rb", "class SecondClass; end");
+
+        let mut db = Db::new();
+        db.initialize_connection(None).unwrap();
+        assert!(db.save_full_graph(&context.graph).is_ok());
+
+        let result = db.get_all_content_hashes().unwrap();
+
+        assert_eq!(result.len(), 2);
+        let first_uri = UriId::from("file:///first.rb");
+        let second_uri = UriId::from("file:///second.rb");
+        assert!(result.contains_key(&first_uri));
+        assert!(result.contains_key(&second_uri));
     }
 }
--- a/rust/saturn/src/model/graph.rs
+++ b/rust/saturn/src/model/graph.rs
@@ -192,6 +192,17 @@ impl Graph {
         removed
     }
 
+    /// Get a list of uris and their content hashes from db.
+    /// This is used for synchronization where we will compare the content hashes
+    /// with the current documents to determine if a document needs updating.
+    ///
+    /// # Errors
+    ///
+    /// Any database errors will prevent the data from being loaded
+    pub fn get_all_cached_content_hashes(&self) -> Result<IdentityHashMap<UriId, u16>, Box<dyn Error>> {
+        self.db.get_all_content_hashes()
+    }
+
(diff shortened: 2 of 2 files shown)


rust/saturn/src/model/db.rs
```diff
@@ -186,6 +192,23 @@ impl Db {
...
+    pub fn content_hashes_by_uri(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
+        self.with_connection(|connection| {
+            let mut statement = connection.prepare("SELECT id, content_hash FROM documents")?;
+            Ok(statement
+                .query_map([], |row| {
+                    let uri = row.get::<_, UriId>(0)?;
+                    let content_hash = row.get::<_, u16>(1)?;
+                    Ok((uri, content_hash))
+                })?
+                .filter_map(|result| result.ok().filter(|(uri, _)| uris.contains(uri)))
```
vinistock: Why do we need to filter the result here? Don't we need to load all available URIs, so that we can detect which ones were added, changed or removed?

rust/saturn/src/indexing.rs
```diff
@@ -325,4 +325,24 @@ mod tests {
             "result_fn should not be called when worker_fn returns (None, vec![])"
         );
     }
+
+    #[test]
+    fn with_parallel_workers_skips_empty_results() {
```
Morriar: Is this test duplicated from the previous one? What is it testing?

rust/saturn/src/model/db.rs
```diff
@@ -186,6 +192,23 @@ impl Db {
         })
     }
 
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn content_hashes_by_uri(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
+        self.with_connection(|connection| {
```
Morriar, line 195: Let's return early if `uris` is empty

rust/saturn/src/model/db.rs
```diff
@@ -186,6 +192,23 @@ impl Db {
...
+            let mut statement = connection.prepare("SELECT id, content_hash FROM documents")?;
```
Morriar, line 196: So if I understand correctly, we're pulling _all_ the documents from the database and filtering in memory?

Can't we filter using SQL with a `WHERE ... IN ...` clause?

rust/saturn/src/model/graph.rs
```diff
@@ -171,6 +172,13 @@ impl Graph {
         removed
     }
 
+    /// # Errors
```
Morriar, line 199: Let's add a documentation explaining what this method returns

rust/saturn/src/model/db.rs
```diff
@@ -186,6 +192,23 @@ impl Db {
         })
     }
 
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn content_hashes_by_uri(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
```
Morriar: The name suggests a N:1 mapping hash:uri

```suggestion
    pub fn content_hashes_for_uris(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
```

rust/saturn/src/model/graph.rs
```diff
@@ -171,6 +172,13 @@ impl Graph {
...
+    ///
+    /// Any database errors will prevent the data from being loaded
+    pub fn get_db_content_hashes(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
```
Morriar: Can't we reuse the same name since we just proxy?

```suggestion
    pub fn content_hashes_for_uris(&self, uris: &HashSet<UriId>) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
```

rust/saturn/src/indexing.rs
jenny-codes: hm this diff showed up after the auto rebasing triggered by the upstack merge this morning. Nvm this I will update.

rust/saturn/src/model/db.rs
jenny-codes: There is no easy way to do the `where ... in ...` because to do so we will need batching and dynamic params construction, which as we have discovered with the `batch_insert_*` is actually making the query slower. 

But to Vini's point we actually want to get all the uris from db so the filtering is not needed anymore.

rust/saturn/src/model/db.rs
jenny-codes: Ah yes. Removed the filtering.

rust/saturn/src/model/graph.rs
jenny-codes: Done

rust/saturn/src/model/graph.rs
jenny-codes: I chose to do `get_all_cached_content_hashes` here to make it clear that this is fetching from the cache and not the memory as they might be different. Let me know if you still wish it to be the same as the db method

rust/saturn/src/model/db.rs
vinistock: Yeah, I think it will be faster to get all (uri, hash) pairs in memory and then reduce the list of what was changed, added and deleted.

rust/saturn/src/model/db.rs
```diff
@@ -186,6 +186,19 @@ impl Db {
         })
     }
 
+    /// # Errors
+    ///
+    /// Errors on any type of database connection or operation failure
+    pub fn get_all_content_hashes(&self) -> Result<HashMap<UriId, u16>, Box<dyn Error>> {
```
vinistock: Let's use an `IdentityHashMap` so that we can get the benefits of skipping double-hashing - given that the keys are already hashed IDs.

```suggestion
    pub fn get_all_content_hashes(&self) -> Result<IdentityHashMap<UriId, u16>, Box<dyn Error>> {
```

PR #242, opened 2025-10-21
Use Scope object to keep track of nesting chain

vinistock:
This PR introduces a `Scope` object. This structure has the responsibility to track scope information that we need to build our graph. I tried to be thorough in documenting it to avoid confusion, but please let me know how we can make it clearer.

Basically, this is how `Scope` keeps track of things:

```ruby
# Values are name_stack | nesting

module Foo # ["Foo"] | { DeclarationId(Foo), parent: None }
  class ::Bar  # ["Bar"] | { DeclarationId(Bar), parent: Scope(Foo) }
  end
  # ["Foo"] | { DeclarationId(Foo), parent: None }


  class Baz::Qux # ["Foo::Baz::Qux"] | { DeclarationId([Foo](Foo::Baz::Qux)), parent: Scope(Foo) }
  end
end
```

All names in the stack are always fully qualified and the `Nesting` is a linked list to represent the chain of lexical scopes we find. There's one major gotcha here to pay attention. Consider this:

```ruby
module Foo
  CONST = 42
  class ::Bar
    CONST
  end
end
```
This code does not crash, it works as intended. The reason is because a top level reference resets the fully qualified name, **but not the lexical scope chain**. This was a lightbulb moment for me during implementation. We need to adjust the names for top level references, but lexical scopes should always be linked directly as we find them in the code.

With this code, we can finally properly resolve constants based on lexical scope. All we have to do is traverse the chain of nestings checking if any of them have the constant defined as part of their `members` hash (**we still need to populate members though**).

<details>
<summary>Benchmarks (minor performance difference, some good memory savings)</summary>

```
BEFORE
Initialization      0.002s (  0.1%)
Listing             1.273s ( 36.5%)
Indexing            2.207s ( 63.4%)
Cleanup             0.001s (  0.0%)
Total:              3.482s
Maximum Resident Set Size: 663404544 bytes (632.67 MB)
Peak Memory Footprint:     635208520 bytes (605.78 MB)
Execution Time:            4.05 seconds
---------
AFTER
Initialization      0.003s (  0.1%)
Listing             1.266s ( 36.9%)
Indexing            2.166s ( 63.1%)
Cleanup             0.001s (  0.0%)
Total:              3.436s
Maximum Resident Set Size: 598786048 bytes (571.04 MB)
Peak Memory Footprint:     574800688 bytes (548.17 MB)
Execution Time:            4.01 seconds
```

</details>

### DIFF
--- a/rust/saturn/src/indexing/scope.rs
+++ b/rust/saturn/src/indexing/scope.rs
@@ -1,0 +1,288 @@
+use crate::model::ids::DeclarationId;
+use std::sync::Arc;
+
+/// The nesting structure is a linked list that represents the chain of lexical scopes we found. This is used to resolve
+/// constant references and always links scopes based on their fully qualified name with whatever scope came before.
+/// Take the following example:
+///
+/// ```ruby
+/// module Foo
+///   CONST = 42
+///
+///   class ::Bar
+///     CONST # resolves to Foo::CONST because we're inside the `Foo` namespace
+///   end
+/// end
+///
+/// class Bar
+///   CONST # NameError because this `Bar` is not lexically inside `Foo`
+/// end
+///
+/// ```
+///
+/// In this example, the top level definition of `::Bar` has the fully qualified name `Bar`, but it is _still_ lexically
+/// connected to `Foo` for purposes of constant resolution. The code in the example does not crash and properly finds
+/// `Foo::CONST`. Therefore, to properly resolve constants, we need to remember the exact chain of lexical scopes we
+/// entered, independentally if the fully qualified name was reset by a top level reference.
+#[derive(Debug)]
+pub struct Nesting {
+    /// The parent scope if any
+    parent: Option<Arc<Nesting>>,
+    /// The declaration ID based on the fully qualified name of this scope
+    declaration_id: DeclarationId,
+}
+
+impl Nesting {
+    #[must_use]
+    pub fn new(parent: Option<Arc<Nesting>>, declaration_id: DeclarationId) -> Self {
+        Self { parent, declaration_id }
+    }
+
+    #[must_use]
+    pub fn declaration_id(&self) -> &DeclarationId {
+        &self.declaration_id
+    }
+
+    #[must_use]
+    pub fn parent(&self) -> &Option<Arc<Nesting>> {
+        &self.parent
+    }
+
+    #[cfg(test)]
+    #[must_use]
+    pub fn ids_as_vec(&self) -> Vec<DeclarationId> {
+        if let Some(parent) = &self.parent {
+            let mut vec = parent.ids_as_vec();
+            vec.push(self.declaration_id);
+            vec
+        } else {
+            vec![self.declaration_id]
+        }
+    }
+}
+
+/// The Scope structure maintains both the `Nesting` linked list of all lexical scopes we encountered and a name stack,
+/// so that we can compute fully qualified names ahead of time for all entries in the graph and their respective
+/// declaration IDs
+#[derive(Default)]
+pub struct Scope {
+    /// A vector of **fully qualified names**. For example:
+    ///
+    /// ```ruby
+    /// module Foo
+    ///   class ::Bar
+    ///   end
+    ///   class Baz::Qux
+    ///   end
+    /// end
+    /// ```
+    ///
+    /// Results in the stack being: `["Foo", "Bar", "Foo::Baz::Qux"]`
+    name_stack: Vec<String>,
+    /// The current chain of lexical scopes
+    nesting: Option<Arc<Nesting>>,
+}
+
+impl Scope {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            name_stack: Vec::new(),
+            nesting: None,
+        }
+    }
+
+    #[must_use]
+    pub fn nesting(&self) -> &Option<Arc<Nesting>> {
+        &self.nesting
+    }
+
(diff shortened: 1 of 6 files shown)


rust/saturn/src/model/references.rs
```diff
@@ -24,9 +27,9 @@ pub enum UnresolvedReference {
 pub struct UnresolvedConstantRef {
     /// The unqualified name of the constant
     name_id: NameId,
-    /// The nesting where we found the constant reference. This is a list of unqualified name IDs, so that we can
-    /// traverse the graph through the member relationships
-    nesting: Vec<NameId>,
+    /// The nesting where we found the constant reference. This is a chain of nesting objects representing the lexical
+    /// scopes surrounding the reference
+    nesting: Option<Arc<Nesting>>,
```
amomchilov, line 15: This could get pricy (both in memory, for the 2 words of Arc overhead, and CPU, for all the reference counting traffic). Would it be worth introducing something like a `NestingID`?

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+/// module Foo
+///   CONST = 42
+///
+///   class ::Bar
+///     CONST
+///   end
+/// end
+/// ```
+///
+/// In this example, the top level definition of `::Bar` has the fully qualified name `Bar`, but it is _still_ lexically
```
Morriar, line 23: Very good example

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+/// connected to `Foo` for purposes of constant resolution. The code in the example does not crash and properly finds
+/// `Foo::CONST`. Therefore, to properly resolve constants, we need to remember the exact chain of lexical scopes we
+/// entered, independentally if the fully qualified name was reset by a top level reference.
+#[derive(Debug)]
+pub struct Nesting {
+    /// The parent scope if any
+    parent: Option<Arc<Nesting>>,
```
Morriar, line 30: Why do we need the Arc here?

Since the visit is depth first don't we always hold a reference to the parent in the visitor and can't let it go when moving to the next node?

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+            nesting: None,
+        }
+    }
+
+    #[must_use]
+    pub fn nesting(&self) -> &Option<Arc<Nesting>> {
+        &self.nesting
+    }
+
+    /// Fully qualified the given name according to the current scope. Takes top level references into account
```
Morriar: Let's add a few examples in the doc

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+            trimmed.to_string()
+        } else if let Some(previous) = self.name_stack.last() {
+            format!("{previous}::{name}")
+        } else {
+            name.to_string()
+        }
+    }
+
+    /// Enters a new namespace (module or class), modifying the scope accordingly
+    pub fn enter(&mut self, name: &str) -> (String, DeclarationId) {
```
Morriar, line 124: Why not returning the `Nesting` directly rather than a pair?

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+        scope.leave();
+        assert_eq!(
+            scope.nesting.as_ref().unwrap().ids_as_vec(),
+            vec![DeclarationId::from("Foo")]
+        );
+        assert_eq!(scope.fully_qualify("CONST"), String::from("Foo::CONST"));
+
+        scope.leave();
+    }
+}
```
Morriar, line 288: Let's add a test combining top level and compact:

```rb
class Foo::Bar
  class ::Bar::Baz
  end
end
```

rust/saturn/src/indexing/scope.rs
```diff
@@ -0,0 +1,238 @@
...
+/// Take the following example:
+///
+/// ```ruby
+/// module Foo
+///   CONST = 42
+///
+///   class ::Bar
+///     CONST
+///   end
+/// end
```
vinistock, line 15: Add an example of `Bar` defined outside and how it does not work due to the lexical scopes.

rust/saturn/src/indexing/scope.rs
vinistock: We don't save the fully qualified string names in `Nesting` to save memory when representing the constant references. When we enter a new namespace, we need the fully qualified name to save it in the graph.

rust/saturn/src/model/references.rs
vinistock: We can definitely explore it, but at this moment it's unclear to me that we will actually need to keep these nesting objects around.

The one reason I can think of is synchronizing changes incrementally by re-resolving constants if some declarations change, but maybe there's a way to do it without keeping the nestings around.

rust/saturn/src/indexing/scope.rs
vinistock: If we find many constant references inside the same nesting, they all point to the same parent, so we need the reference counting. Unfortunately, we can't use `Rc` because the references are collected in threads.

rust/saturn/src/indexing/scope.rs
alexcrocha: Perfect! Thank you 💪

Issue #141, opened 2025-08-27
Revisit deterministic IDs


vinistock:
Our first ID implementation was with `blake3`, which is fast but was using 8 bytes - spending more memory we could afford. Currently, we're using xxhash with i64, which is decently fast, but still uses more memory than we would like.

Let's explore if we can use u32 (which matches Pyre and Sorbet).

PR #243, opened 2025-10-21
Add native implementation for `Declaration`

Morriar:
Extracted from https://github.com/Shopify/saturn/pull/219.

### DIFF
--- a/ext/saturn/declaration.c
+++ b/ext/saturn/declaration.c
@@ -1,13 +1,35 @@
 #include "declaration.h"
+#include "graph.h"
 #include "handle.h"
+#include "rustbindings.h"
 
 VALUE cDeclaration;
+
+// Declaration#name -> String
+static VALUE sr_declaration_name(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+    const char *name = sat_declaration_name(graph, data->id);
+
+    if (name == NULL) {
+        return Qnil;
+    }
+
+    VALUE str = rb_utf8_str_new_cstr(name);
+    free_c_string(name);
+
+    return str;
+}
 
 void initialize_declaration(VALUE mSaturn) {
     cDeclaration = rb_define_class_under(mSaturn, "Declaration", rb_cObject);
 
     rb_define_alloc_func(cDeclaration, sr_handle_alloc);
     rb_define_private_method(cDeclaration, "initialize", sr_handle_initialize, 2);
+    rb_define_method(cDeclaration, "name", sr_declaration_name, 0);
 
     rb_undef_method(CLASS_OF(cDeclaration), "new");
 }
--- a/rust/saturn-sys/src/declaration_api.rs
+++ b/rust/saturn-sys/src/declaration_api.rs
@@ -1,0 +1,30 @@
+//! This file provides the C API for the Graph object
+
+use libc::c_char;
+use std::ffi::CString;
+use std::ptr;
+
+use crate::graph_api::{GraphPointer, with_graph};
+use saturn::model::ids::DeclarationId;
+
+/// Returns the UTF-8 name string for a declaration id.
+/// Caller must free with `free_c_string`.
+///
+/// # Safety
+///
+/// Assumes pointer is valid.
+///
+/// # Panics
+///
+/// This function will panic if the name pointer is invalid.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sat_declaration_name(pointer: GraphPointer, name_id: i64) -> *const c_char {
+    with_graph(pointer, |graph| {
+        let name_id = DeclarationId::new(name_id);
+        if let Some(decl) = graph.declarations().get(&name_id) {
+            CString::new(decl.name()).unwrap().into_raw().cast_const()
+        } else {
+            ptr::null()
+        }
+    })
+}
--- a/rust/saturn-sys/src/lib.rs
+++ b/rust/saturn-sys/src/lib.rs
@@ -1,2 +1,3 @@
+pub mod declaration_api;
 pub mod graph_api;
 pub mod utils;
--- a/test/declaration_test.rb
+++ b/test/declaration_test.rb
@@ -23,6 +23,7 @@ class DeclarationTest < Minitest::Test
 
       declaration = graph.declarations.first
       assert_instance_of(Saturn::Declaration, declaration)
+      assert_equal("A", declaration.name)
     end
   end
 end


ext/saturn/declaration.c
```diff
@@ -0,0 +1,53 @@
...
+static VALUE declaration_initialize(VALUE self, VALUE graph_obj, VALUE id_val) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+    data->graph_obj = graph_obj;
+    data->id = NUM2LL(id_val);
+
+    return self;
+}
+
+// DeclarationHandle#name -> String
```
vinistock: The class name is just `Declaration`, no?
```suggestion
// Declaration#name -> String
```

ext/saturn/declaration.c
```diff
@@ -0,0 +1,53 @@
...
+static VALUE declaration_name(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+    const char *name = sat_declaration_name(graph, (int64_t)data->id);
```
vinistock: We're storing the `id` as `long long`, but here we need to cast it to `int64_t`. Could we store it as `int64_t` directly? Are there any disadvantages in doing so?

ext/saturn/declaration.c
```diff
@@ -0,0 +1,53 @@
...
+    free_c_string(name);
+
+    return str;
+}
+
+void initialize_declaration(VALUE mSaturn) {
+    cDeclaration = rb_define_class_under(mSaturn, "Declaration", rb_cObject);
+
+    rb_define_alloc_func(cDeclaration, declaration_alloc);
+    rb_define_private_method(cDeclaration, "initialize", declaration_initialize,
```
vinistock: The indentation on this one is a bit odd. Do we happen to have the line length configured at 80 characters instead of 120?

ext/saturn/handle.h
```diff
@@ -0,0 +1,30 @@
+#ifndef SATURN_HANDLE_H
+#define SATURN_HANDLE_H
+
+#include "ruby.h"
+
+typedef struct {
+    VALUE graph_obj; // Ruby Graph object to keep it alive
+    long long id;    // Canonical ID (i64) of the indexed item (Declaration,
```
vinistock: Can we add a comment linking to the `id.rs` file and reminding us that this has to be kept in sync?

The work in #141 may end up changing it and there's a chance we could forget.

ext/saturn/handle.h
```diff
@@ -0,0 +1,30 @@
...
+                     // Definition, Document, etc.)
+} HandleData;
+
+static void handle_mark(void *ptr) {
+    if (ptr) {
+        HandleData *data = (HandleData *)ptr;
+        rb_gc_mark(data->graph_obj);
+    }
+}
+static void handle_free(void *ptr) {
```
vinistock, line 18:
```suggestion
}

static void handle_free(void *ptr) {
```

rust/saturn-sys/src/graph_api.rs
```diff
@@ -91,3 +91,91 @@ pub unsafe extern "C" fn sat_graph_set_configuration(pointer: GraphPointer, db_p
...
+    Box::into_raw(Box::new(DeclarationsIter { ids, index: 0 }))
+}
+
+/// Returns the total number of IDs in the iterator snapshot.
+///
+/// # Safety
+///
+/// - `iter` must be a valid pointer previously returned by `sat_graph_declarations_iter_new`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sat_graph_declarations_iter_len(iter: *const DeclarationsIter) -> usize {
```
vinistock, line 133: It seems like this is only invoked once, right after creating the iterator. Could we just return this as part of the iterator instead? Or do you foresee us using this separately?

ext/saturn/declaration.c
Morriar: You're right, let's use `int64_t` everywhere

ext/saturn/declaration.c
Morriar: We can: https://github.com/Shopify/saturn/pull/246

rust/saturn-sys/src/graph_api.rs
Morriar: The returned struct is opaque on the C side so we can't access the fields directly.

ext/saturn/declaration.c
```diff
@@ -0,0 +1,52 @@
+#include "declaration.h"
+#include "graph.h"
+#include "handle.h"
+#include "rustbindings.h"
+
+VALUE cDeclaration;
+
+static VALUE sr_declaration_alloc(VALUE klass) {
```
vinistock: The allocator is not exposed on the Ruby side.
```suggestion
static VALUE declaration_alloc(VALUE klass) {
```

ext/saturn/graph.c
```diff
@@ -64,10 +65,63 @@ static VALUE sr_graph_set_configuration(VALUE self, VALUE db_path) {
...
+    sat_graph_declarations_iter_free(iter);
+
+    return SIZET2NUM(len);
+}
+
+// Graph#declarations: () -> Enumerator[Declaration]
+// Returns an enumerator that yields all declarations lazily
+static VALUE sr_graph_declarations(VALUE self) {
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize_with_size(self, rb_str_new2("declarations"), 0, NULL, graph_declarations_size);
```
vinistock, line 107: How does `free` get invoked on this code path?

ext/saturn/declaration.c
Morriar: So `sr_graph_alloc` shouldn't have the prefix?

This is the function backing the [`allocate`](https://ruby-doc.org/core-2.5.4/Class.html#method-i-allocate) method:

```rb
Saturn::Declaration.allocate # will call `sr_declaration_alloc`
```

Calling `new` is basically calling `allocate` then `initialize`.

Are we sure the `sr_` prefix is still relevant?

ext/saturn/graph.c
Morriar: You mean the free for the iterator?

ext/saturn/declaration.c
vinistock: Oh, I forgot about the `allocate` method. I was under the impression that the allocator was only internal and not exposed in any way.

> Are we sure the sr_ prefix is still relevant?

What do you mean? Why is the prefix not relevant anymore?

ext/saturn/graph.c
vinistock: Yeah, cause we're freeing it in a rescue when a block is given. If it's not, how does it get freed?

Morriar: I'm confused, if a block is given we do not create the iterator, I'm not sure what needs to be freed?

ext/saturn/declaration.c
Morriar: > Why is the prefix not relevant anymore?

There seems to be some confusion about when to use it and when not.

vinistock: What's the confusion? I thought we agreed it was for all functions that get exposed as Ruby methods. I just forgot the allocator backs `allocate`.

ext/saturn/graph.c
vinistock: I guess I'm confused as to how this works. If you don't pass a block and return the enumerator, does Ruby use `self` to figure out how to yield to the block later?

Does it call this method again to start yielding items?

Morriar: Calling `rb_enumeratorize_with_size` will create a Ruby enumerator that will call back into `definitions` with a block only if needed.

So with this:

```rb
enumerator = graph.declarations # create an enumerator, no iterator to free yet

assert_equal(2, enumerator.size) # because we used `rb_enumeratorize_with_size` the enumerator can answer without iterator
assert_equal(2, enumerator.count) # `count` uses an iterator and it will be freed in the ensure
assert_equal(2, enumerator.to_a.size) # `to_a` uses another iterator and it will be freed in the ensure
```

The enumerator itself is, I _think_ handled and freed by Ruby? At least memleak seems to be happy?

vinistock: Yeah, I don't think there's a memory leak. Just curious as to how it works. It must use `self` and the name `declarations` to hook it back to the yield? Anyway, interesting mechanics.

PR #199, opened 2025-10-09
Store comments as optional array

jenny-codes:
Instead of formatting the comments as strings, we now store and present them as an optional arrays of strings. Benchmarks return similar numbers. 

### Benchmark

Running with huge_corpus, commenting out the db operation (it still takes too long) 

#### Before

```
Found 13907063 names
Found 13907063 definitions
Found 99999 URIs

PERFORMANCE BREAKDOWN

Initialization:    0.172s (  1.4%)
Indexing:          5.401s ( 43.9%)
Db operation:      0.000s (  0.0%)
Cleanup:           6.718s ( 54.7%)
Total:            12.291s

----------------------------------------
Maximum Resident Set Size: 8844394496 bytes (8434.67 MB)
Peak Memory Footprint:     8709528176 bytes (8306.05 MB)
Execution Time:            12.62 seconds
```

#### After 

```
Found 13907063 names
Found 13907063 definitions
Found 99999 URIs

PERFORMANCE BREAKDOWN

Initialization:    0.151s (  1.2%)
Indexing:          5.640s ( 43.2%)
Db operation:      0.000s (  0.0%)
Cleanup:           7.279s ( 55.7%)
Total:            13.070s

----------------------------------------
Maximum Resident Set Size: 8915632128 bytes (8502.60 MB)
Peak Memory Footprint:     8778373792 bytes (8371.70 MB)
Execution Time:            13.51 seconds
```

PR #248, opened 2025-10-22
Simplify comments handling

Morriar:
Closes #195 and #199.

Let's remove all the complexity of storing comment and delegate it to the clients as it may greatly vary from one use case to another.

1. Let's store the comments as vectors so we keep the lines separated
2. Remove all filtering
3. Remove processing of `# ` and let the client figure if it wants the prefix or not
4. Stop considering comments that are not directly over a definition

We'll revisit what to do with RBS sigs and `#:`/`#|` after but for now, let's just keep them as comments.

cc. @jenny-codes

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use crate::indexing::errors::IndexingError;
 use crate::indexing::scope::Scope;
+use crate::model::comment::Comment;
 use crate::model::definitions::{
     AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
     ConstantDefinition, Definition, GlobalVariableDefinition, InstanceVariableDefinition, MethodDefinition,
@@ -94,7 +95,7 @@ impl<'a> RubyIndexer<'a> {
         String::from_utf8_lossy(location.as_slice()).to_string()
     }
 
-    fn find_comments_for(&self, offset: u32) -> Option<Vec<String>> {
+    fn find_comments_for(&self, offset: u32) -> Option<Vec<Comment>> {
         let offset_usize = offset as usize;
         if self.comments.is_empty() {
             return None;
@@ -287,7 +288,7 @@ impl<'a> RubyIndexer<'a> {
 
 struct CommentGroup {
     end_offset: usize,
-    comments: Vec<String>,
+    comments: Vec<Comment>,
 }
 
 impl CommentGroup {
@@ -320,7 +321,10 @@ impl CommentGroup {
     fn add_comment(&mut self, comment: &ruby_prism::Comment) {
         self.end_offset = comment.location().end_offset();
         let text = String::from_utf8_lossy(comment.location().as_slice()).to_string();
-        self.comments.push(text.trim().to_string());
+        self.comments.push(Comment::new(
+            Offset::from_prism_location(&comment.location()),
+            text.trim().to_string(),
+        ));
     }
 }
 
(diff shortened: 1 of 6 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -131,7 +115,7 @@ impl<'a> RubyIndexer<'a> {
         }
 
         // We allow at most one blank line between the comment and the definition
```
jenny-codes, line 121: This comment needs to change

st0012, line 121: Should this be updated? Or was it incorrect?

Morriar: Good catch

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -130,8 +114,8 @@ impl<'a> RubyIndexer<'a> {
             return None;
         }
 
-        // We allow at most one blank line between the comment and the definition
-        if bytecount::count(between, b'\n') > 2 {
+        // Comments need to be directly above the definition
```
st0012: This will be a problem for some projects because an "old" style is to have one blank line in between. Examples: [RDoc](https://github.com/ruby/rdoc/blob/master/lib/rdoc/code_object.rb), [minitest](https://github.com/minitest/minitest/blob/master/lib/minitest/test.rb)

Morriar: This is going to be an issue for other. Example with Sorbet: https://sorbet.run/#%23%20typed%3A%20true%0A%0A%23%20Not%20comment%0A%0Aclass%20Foo%3B%20end%0A%0A%23%20Comment%0Aclass%20Bar%3B%20end%0A%0Aputs%20Foo%0Aputs%20Bar.

I think the best we can do at this point is keep the comments but also add the location so the client can decide.

vinistock: What's the issue with still allowing for the extra blank line? Is it the fact that Sorbet doesn't allow for it?

Keep in mind that if we disallow that, no one will be able to see documentation for Minitest methods while writing tests, which is a loss compared to what we have today.

Morriar: It's not a problem if we give a way to other tool to discard the comments they don't want to consider and we can do this by adding the location to the comments (see last commit).

That way a tool behaving like Sorbet (Spoom, Tapioca) can compare the comment location and the definition location to decide if it should be kept or not.

vinistock: Can you explain when Spoom or Tapioca would make a decision based on this extra blank line? So that I can keep in mind, I can't remember what would be impacted by this.

I'm a bit concerned with keeping locations for all comments in terms of memory. I would honestly rather even commit to ignoring the comments with the blank line before storing all locations in memory.

And I know the RBS comments need the location, but again we can save the locations only for those annotation comments and not the others.

Morriar: > Can you explain when Spoom or Tapioca would make a decision based on this extra blank line? So that I can keep in mind, I can't remember what would be impacted by this.

https://github.com/Shopify/spoom/blob/main/test/spoom/deadcode/remover_test.rb#L890-L919 is a good example of why we need the locations.

When removing a dead definition, we also want to remove its comments but not unrelated comments. We can use the locations to know which comment lines to remove or to keep.

> I'm a bit concerned with keeping locations for all comments in terms of memory. I would honestly rather even commit to ignoring the comments with the blank line before storing all locations in memory.

Yes this is 300MB more for `huge` but looking at core it appears to be only 7MB:

```
Maximum Resident Set Size: 614285312 bytes (585.82 MB)
Peak Memory Footprint:     499009984 bytes (475.89 MB)
Execution Time:            5.24 seconds
```

```
Maximum Resident Set Size: 621658112 bytes (592.85 MB)
Peak Memory Footprint:     508856768 bytes (485.28 MB)
Execution Time:            5.03 seconds
```

What we _could_ do is not keep the comments in memory once the database in synchronized? But for now it seems low enough.

vinistock: Oh, if it's just 7mb, then it should be fine.

PR #249, opened 2025-10-22
Populate declaration members

vinistock:
Start populating the members of each declaration. Being able to inspect which members exist in each global declaration allows us to perform constant resolution, but also provide completion based on what's available within the current context.

This is still missing the logic to handle deletions, but with this we're able to implement resolving constants based on lexical scopes.

### Implementation

An important aspect of the implementation that I had to test out is what things are associated to when they are defined in the top level. There are essentially two cases:

1. Constants of any kind (modules, classes and other constants) are associated to `Object` when defined at the top level
2. Anything else (instance variables, methods and so on) get associated to `<main>`. Ruby uses a special instance of the `Object` class to represent its top level. Performing some tests on a script, you can get a better sense for the behaviour

I started adding the `<main>`​ declaration artificially since it doesn't actually exist anywhere, it's created by Ruby [here](https://github.com/ruby/ruby/blob/9ad902e55c610e66114f528f77f7895295a242de/vm.c#L4693). This is equivalent to Sorbet's `<root>`, but I figured why not call it the same as Ruby does.

```ruby
attr_reader :foo # => crash! <main> is an instance of `Object` and therefore attr_* methods don't exist
@@class_var = 123 # => crash! can't use class variables on <main>

def foo; end

foo() # ok! `foo` was defined in the <main> instance

obj = Object.new
obj.foo # crash! `foo` is not defined in the `Object` class
```

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1,6 +1,6 @@
 //! Visit the Ruby AST and create the definitions.
 
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use crate::indexing::errors::IndexingError;
 use crate::indexing::scope::Scope;
@@ -19,6 +19,8 @@ use crate::source_location::SourceLocationConverter;
 use ruby_prism::{ParseResult, Visit};
 
 pub type IndexerParts = (Option<Graph>, Vec<IndexingError>);
+static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Object"));
+static MAIN_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("<main>"));
 
 /// The indexer for the definitions found in the Ruby source code.
 ///
@@ -330,6 +332,7 @@ impl CommentGroup {
 
 impl Visit<'_> for RubyIndexer<'_> {
     fn visit_class_node(&mut self, node: &ruby_prism::ClassNode<'_>) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
         let name = Self::location_to_string(&node.constant_path().location());
         let (fully_qualified_name, declaration_id) = self.scope.enter(&name);
         let offset = Offset::from_prism_location(&node.location());
@@ -346,6 +349,7 @@ impl Visit<'_> for RubyIndexer<'_> {
         )));
 
         self.local_graph.add_definition(fully_qualified_name, definition);
+        self.local_graph.add_member(&previous_nesting_id, declaration_id, &name);
 
         if let Some(body) = node.body() {
             self.visit(&body);
@@ -355,6 +359,7 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_module_node(&mut self, node: &ruby_prism::ModuleNode) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
         let name = Self::location_to_string(&node.constant_path().location());
         let (fully_qualified_name, declaration_id) = self.scope.enter(&name);
         let offset = Offset::from_prism_location(&node.location());
@@ -369,6 +374,7 @@ impl Visit<'_> for RubyIndexer<'_> {
             comments,
         )));
         self.local_graph.add_definition(fully_qualified_name, definition);
+        self.local_graph.add_member(&previous_nesting_id, declaration_id, &name);
 
         if let Some(body) = node.body() {
             self.visit(&body);
@@ -378,6 +384,7 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_constant_write_node(&mut self, node: &ruby_prism::ConstantWriteNode) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
         let name_loc = node.name_loc();
         let name = Self::location_to_string(&name_loc);
         let fully_qualified_name = self.scope.fully_qualify(&name);
@@ -395,10 +402,12 @@ impl Visit<'_> for RubyIndexer<'_> {
         )));
 
         self.local_graph.add_definition(fully_qualified_name, definition);
+        self.local_graph.add_member(&previous_nesting_id, declaration_id, &name);
         self.visit(&node.value());
     }
 
     fn visit_constant_path_write_node(&mut self, node: &ruby_prism::ConstantPathWriteNode) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
         let location = node.target().location();
         let name = Self::location_to_string(&location);
         let fully_qualified_name = self.scope.fully_qualify(&name);
@@ -416,6 +425,7 @@ impl Visit<'_> for RubyIndexer<'_> {
         )));
 
         self.local_graph.add_definition(fully_qualified_name, definition);
+        self.local_graph.add_member(&previous_nesting_id, declaration_id, &name);
         self.visit(&node.value());
     }
 
@@ -431,6 +441,7 @@ impl Visit<'_> for RubyIndexer<'_> {
         for left in node.lefts().iter() {
             match left {
                 ruby_prism::Node::ConstantTargetNode { .. } | ruby_prism::Node::ConstantPathTargetNode { .. } => {
+                    let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
                     let location = left.location();
                     let name = Self::location_to_string(&location);
                     let fully_qualified_name = self.scope.fully_qualify(&name);
@@ -449,6 +460,7 @@ impl Visit<'_> for RubyIndexer<'_> {
                     )));
 
                     self.local_graph.add_definition(fully_qualified_name, definition);
+                    self.local_graph.add_member(&previous_nesting_id, declaration_id, &name);
                 }
                 ruby_prism::Node::GlobalVariableTargetNode { .. } => {
                     let location = left.location();
@@ -466,9 +478,11 @@ impl Visit<'_> for RubyIndexer<'_> {
                         comments,
                     )));
 
+                    self.local_graph.add_member(&MAIN_ID, declaration_id, &name);
                     self.local_graph.add_definition(name, definition);
                 }
                 ruby_prism::Node::InstanceVariableTargetNode { .. } => {
+                    let nesting_id = self.scope.current_nesting_id().unwrap_or(*MAIN_ID);
                     let location = left.location();
                     let name = Self::location_to_string(&location);
                     let fully_qualified_name = self.scope.fully_qualify(&name);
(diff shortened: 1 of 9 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -53,6 +70,8 @@ impl<'a> RubyIndexer<'a> {
             comments: Vec::new(),
             source,
             scope: Scope::new(),
+            object_id: DeclarationId::from("Object"),
+            main_id: DeclarationId::from("<main>"),
```
st0012: Is it worth creating a constant so we don't repeatedly call `DeclarationId::from` for every `RubyIndexer` creation (basically every file?):

```diff
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1,6 +1,6 @@
 //! Visit the Ruby AST and create the definitions.
 
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use crate::indexing::errors::IndexingError;
 use crate::indexing::scope::Scope;
@@ -35,6 +35,9 @@ const MAGIC_AND_RBS_COMMENT_PREFIX: &[&str] = &[
     "#|",
 ];
 
+static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Object"));
+static MAIN_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("<main>"));
+
 /// The indexer for the definitions found in the Ruby source code.
 ///
 /// It implements the `Visit` trait from `ruby_prism` to visit the AST and create a hash of definitions that must be
@@ -70,8 +73,8 @@ impl<'a> RubyIndexer<'a> {
             comments: Vec::new(),
             source,
             scope: Scope::new(),
-            object_id: DeclarationId::from("Object"),
-            main_id: DeclarationId::from("<main>"),
+            object_id: *OBJECT_ID,
+            main_id: *MAIN_ID,
         }
     }
```

I know that `DeclarationId::from` is already heavily optimized, but in this case we simply don't need to call it at all.

From what I read, [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) only locks when initializing the value, but becomes lock free when reading it.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -378,6 +406,7 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_constant_write_node(&mut self, node: &ruby_prism::ConstantWriteNode) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(self.object_id);
```
st0012: In other cases, it's clear that `self.scope.enter` can change the scope's nesting so we need to save the previous nesting id. But in this case, we only call `self.scope.fully_qualify`. Does it also change scope's nesting?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -487,26 +523,30 @@ impl Visit<'_> for RubyIndexer<'_> {
...
-                        comments,
-                    )));
+                    if let Some(nesting_id) = self.scope.current_nesting_id() {
+                        let location = left.location();
+                        let name = Self::location_to_string(&location);
+                        let fully_qualified_name = self.scope.fully_qualify(&name);
+
+                        let declaration_id = DeclarationId::from(&fully_qualified_name);
+                        let offset = Offset::from_prism_location(&location);
+                        let comments = self.find_comments_for(offset.start()).unwrap_or_default();
```
st0012, line 514: Not directly related: I didn't realize how much boilerplate code we have involving these same few lines (with tiny variations). I think we should create a few helpers to encapsulate them.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1369,10 +1390,8 @@ mod tests {
             "
         });
 
-        let definitions = context.graph.get("foo=").unwrap();
-        assert_eq!(definitions.len(), 1);
-        assert_definition_offset!(definitions[0], 13, 16);
-
+        // Invoking `attr_*` methods on the <main> object crashes, so we ignore those
+        assert!(context.graph.get("foo=").is_none());
```
st0012: I think this is the same reason we now have `assert!(context.graph.get("foo").is_none());` in several previous examples? 
How about we remove all the top-level `attr_*` cases from existing tests, and just create one `does_not_index_attr_methods_on_main_object` to cover them.

vinistock: Good idea. It better documents the behaviour too.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Yes, but I've been intentionally holding back on trying to consolidate anything. There are still missing pieces for indexing, missing singleton handling, incorrect name handling in a few places and it's unclear at this moment how much common code there will be.

Once we have the complete indexing with the right structures, we can take a broad look at the indexer implementation and refactor.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: No, it doesn't change it. Do you prefer to inline the variable? The line just gets a little long.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Moved it to a const 👍

rust/saturn/tests/cli.rs
```diff
@@ -19,16 +25,19 @@ fn prints_help() {
 fn dir_argument_variants() {
     let mut zero = Command::cargo_bin("saturn_cli").unwrap();
     zero.assert().success().stderr(predicate::str::is_empty());
+    delete_db();
```
Morriar: We could use a closure for this:

```rs
    with_db(|_db_path| {
        // ...
    }); // auto drop
```

rust/saturn/src/model/integrity.rs
```diff
@@ -110,8 +110,8 @@ mod tests {
     fn test_integrity_check_with_custom_rule() {
         let mut checker = IntegrityChecker::new();
 
-        checker.add_rule("Index must be empty", |index, errors| {
-            if !index.declarations().is_empty() {
+        checker.add_rule("Index must be empty except for <main>", |index, errors| {
```
Morriar, line 113: We could also check the only declaration is actually `<main>` to avoid some weird case:

```rs
graph.add_declaration("foo")
graph.declarations.len() == 2
graph.delete_declaration("<name>")
graph.declarations.len() == 1
```

rust/saturn/src/model/graph.rs
```diff
@@ -102,6 +107,27 @@ impl Graph {
...
 
+    /// Register a member relationship from a declaration to another declaration through its unqualified name id. For example, in
+    ///
+    /// ```ruby
+    /// module Foo
+    ///   class Bar; end
+    /// end
+    /// ```
+    ///
+    /// `Foo::Bar` is a member of `Foo` through `NameId(Bar)`
```
Morriar:
```suggestion
    /// module Foo
    ///   class Bar; end
    ///   def baz; end
    /// end
    /// ```
    ///
    /// `Foo` has two members:
    /// {
    ///   NameId(Bar) => DeclarationId(Bar)
    ///   NameId(baz) => DeclarationId(baz)
    /// }
```

rust/saturn/src/model/graph.rs
```diff
@@ -102,6 +107,27 @@ impl Graph {
...
+        &mut self,
+        declaration_id: &DeclarationId,
+        member_declaration_id: DeclarationId,
+        member_name: &str,
+    ) {
+        if let Some(declaration) = self.declarations.get_mut(declaration_id) {
+            let name_id = NameId::from(member_name);
+            declaration.add_member(name_id, member_declaration_id);
+        }
+    }
```
Morriar, line 136: I wonder if we need another integrity check to make sure the same declaration is not added as a member of 2 different declarations?

rust/saturn/src/model/graph.rs
```diff
@@ -36,8 +36,13 @@ pub struct Graph {
 impl Graph {
     #[must_use]
     pub fn new() -> Self {
+        let mut declarations = IdentityHashMap::default();
+        // Insert the magic top level self <main> object into the graph, so that we can associate global variables or
+        // definitions made at the top level with it
+        declarations.insert(DeclarationId::from("<main>"), Declaration::new(String::from("<main>")));
```
Morriar, line 42: Can we extract this as a constant like Sorbet is doing with `core::names::Object` and friends?

Maybe this could even make it a constant struct:

```rs
struct Main {
   declaration_id: DeclarationId,
   declaration: Declaration
}

const MAIN: Main = Main {
  declaration_id: DeclarationId::from("<main>"), 
  declaration: Declaration::new(String::from("<main>")
};
```

And possibly also extract `<main>` as a constant as well

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -330,6 +332,7 @@ impl CommentGroup {
 
 impl Visit<'_> for RubyIndexer<'_> {
     fn visit_class_node(&mut self, node: &ruby_prism::ClassNode<'_>) {
+        let previous_nesting_id = self.scope.current_nesting_id().unwrap_or(*OBJECT_ID);
```
Morriar, line 335: Could we move the `unwrap_or(*OBJECT_ID)` inside `current_nesting_id()` to avoid missing one?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -466,9 +478,11 @@ impl Visit<'_> for RubyIndexer<'_> {
                         comments,
                     )));
 
+                    self.local_graph.add_member(&MAIN_ID, declaration_id, &name);
                     self.local_graph.add_definition(name, definition);
                 }
                 ruby_prism::Node::InstanceVariableTargetNode { .. } => {
+                    let nesting_id = self.scope.current_nesting_id().unwrap_or(*MAIN_ID);
```
Morriar, line 485: I wonder if the different between `unwrap_or(*OBJECT_ID)` and `unwrap_or(*MAIN_ID)` is not a matter of adding things related to the class or the instance? Maybe something like:

```rs
self.scope.current_nesting_id() # returns `*MAIN_ID`
  .class() # returns `*OBJECT_ID`
```

I wonder if adding the singleton concept will allow us to clean this at some point.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1608,4 +1632,216 @@ mod tests {
...
         assert!(context.graph.get("Foo::123").is_none());
     }
+
+    #[test]
+    fn constant_members() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            module Foo
+              module Bar; end
```
st0012, line 1642: Can we add a test case for this:

```rb
class Foo; end
class Foo::Bar; end
```

`Bar` should also be a `Foo`'s member.

vinistock: I added a test showing that our behaviour is currently incorrect and created https://github.com/Shopify/saturn/issues/253 to address it.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: I'm honestly not sure. It feels like being an instance of the class and the attached/singleton class are different chains for me.

For example,

```ruby
# self is <main>
puts class # => Object
puts class.class # => Class
puts class.singleton_class # => <Class:Object>
```

My gut feeling is that `<main>` is the only instance we would keep in the indexed data.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: We can't because it's not always `unwrap_or(*OBJECT_ID)`, sometimes it's `MAIN_ID`. Rust guarantees that we _have_ to handle the `None` case, so I think we're safe from missing any.

rust/saturn/src/model/graph.rs
vinistock: Yes, but I'll create that in a separate PR.

rust/saturn/src/model/graph.rs
vinistock: Yes, I agree. I'll do it in the next PR so that this one doesn't grow too much.

rust/saturn/tests/cli.rs
```diff
@@ -4,6 +4,17 @@ use regex::Regex;
 use std::fs;
 use std::process::Command;
 
+fn with_db_cleanup<F, T>(f: F) -> T
```
Morriar, line 7: Should we move this under https://github.com/Shopify/saturn/tree/main/test/helpers?

vinistock: I tried, but integration tests such as `cli.rs` don't have access to modules that are conditionally compiled under `cfg(test)`.

Let's see if we'll need it in more places and then we can invest more time in figuring it out.

PR #250, opened 2025-10-22
Implement native extension `Graph#[]`

Morriar:
This is a temporary implementation until we have actual name resolution.

### DIFF
--- a/ext/saturn/graph.c
+++ b/ext/saturn/graph.c
@@ -170,6 +170,28 @@ static VALUE sr_graph_documents(VALUE self) {
     return self;
 }
 
+// Graph#[]: (String fully_qualified_name) -> Declaration
+// Returns a declaration handle for the given ID
+static VALUE sr_graph_aref(VALUE self, VALUE key) {
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    if (TYPE(key) != T_STRING) {
+        rb_raise(rb_eTypeError, "expected String");
+    }
+
+    const int64_t *id_ptr = sat_graph_get_declaration(graph, StringValueCStr(key));
+    if (id_ptr == NULL) {
+        return Qnil;
+    }
+
+    int64_t id = *id_ptr;
+    free_i64(id_ptr);
+    VALUE argv[] = {self, LL2NUM(id)};
+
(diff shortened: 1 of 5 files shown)


ext/saturn/graph.c
```diff
@@ -117,11 +117,34 @@ static VALUE sr_graph_declarations(VALUE self) {
...
+}
+
 void initialize_graph(VALUE mSaturn) {
     cGraph = rb_define_class_under(mSaturn, "Graph", rb_cObject);
 
     rb_define_alloc_func(cGraph, sr_graph_alloc);
     rb_define_method(cGraph, "index_all", sr_graph_index_all, 1);
     rb_define_method(cGraph, "set_configuration", sr_graph_set_configuration, 1);
     rb_define_method(cGraph, "declarations", sr_graph_declarations, 0);
+    rb_define_method(cGraph, "[]", sr_graph_aref, 1);
```
vinistock, line 203: What do you think about already calling this `resolve` and just substituting the underlying implementation once we're ready?

Morriar: I was thinking keeping the `[]` and just plumbing it with resolution?

So `graph["Foo"]` would be the Ruby API to resolve the constant name?

vinistock: How would you pass a compact namespace though? Say you have this:

```ruby
module Foo
  class Bar::Baz
    CONST
  end
end
```

What would the call to `Graph#[]` look like to resolve? In my head, it would be something like `resolve("CONST", ["Foo", "Bar::Baz"])`

Morriar: I think this is the "advanced" way to access a declaration.

For may uses cases, you just care about resolving a fully qualified name into a declaration:

```
graph["Foo::Bar::Baz::CONST"]
graph["::Foo::Bar::Baz::CONST"]
graph["Object::Foo::Bar::Baz::CONST"]
graph["::Object::Foo::Bar::Baz::CONST"]
```

Which would call `sat_resolve` with the right arguments. And we can provide another method (maybe `Graph#resolve`) to handle more complex cases if you want to pass the namespace as you did.

vinistock: I think that's fine. This one can be for directly accessing any declaration through their fully qualified name, whereas resolve will be specific for constants and does extra work.

PR #251, opened 2025-10-23
Implement native extension `Definition`

Morriar:
Add access to definitions through:

* `Declaration#definitions`
* `Document#definitions`

### DIFF
--- a/test/definition_test.rb
+++ b/test/definition_test.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 require "test_helper"
+require "helpers/context"
 
 class DefinitionTest < Minitest::Test
+  include Test::Helpers::WithContext
+
   def test_instantiating_a_definition_from_ruby_fails
     e = assert_raises(NoMethodError) do
       Saturn::Definition.new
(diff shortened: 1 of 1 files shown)


ext/saturn/definition.c
```diff
@@ -0,0 +1,100 @@
...
+    default:
+        return cDefinition;
+    }
+}
+
+void initialize_definition(VALUE mSaturn) {
+    cDefinition = rb_define_class_under(mSaturn, "Definition", rb_cObject);
+    rb_define_alloc_func(cDefinition, sr_handle_alloc);
+    rb_define_private_method(cDefinition, "initialize", sr_handle_initialize, 2);
+    rb_undef_method(CLASS_OF(cDefinition), "new");
```
vinistock: It seems we're using these several times now, even in other places (document and declaration). Can we extract this to a utility function where we just pass the related classes/modules?

ext/saturn/definition.c
```diff
@@ -0,0 +1,100 @@
...
+VALUE cConstantDefinition;
+VALUE cMethodDefinition;
+VALUE cAttrAccessorDefinition;
+VALUE cAttrReaderDefinition;
+VALUE cAttrWriterDefinition;
+VALUE cGlobalVariableDefinition;
+VALUE cInstanceVariableDefinition;
+VALUE cClassVariableDefinition;
+
+VALUE definition_class_for_kind(DefinitionKind kind) {
```
vinistock, line 18: Nit: let's add a comment reminding us to keep this in sync with `definitions.rs`.

ext/saturn/definition.c
```diff
@@ -0,0 +1,100 @@
...
+        return cAttrWriterDefinition;
+    case DefinitionKind_GlobalVariable:
+        return cGlobalVariableDefinition;
+    case DefinitionKind_InstanceVariable:
+        return cInstanceVariableDefinition;
+    case DefinitionKind_ClassVariable:
+        return cClassVariableDefinition;
+    case DefinitionKind_Unknown:
+    default:
+        return cDefinition;
```
vinistock: Reaching this point in the code would indicate a bug, since `Definition` is basically an abstract class that shouldn't actually be instantiated.

Should we print a warning here so that it's easier for us to catch?

rust/saturn-sys/src/definition_api.rs
```diff
@@ -0,0 +1,150 @@
+//! This file provides the C API for Definition accessors
+
+use crate::graph_api::{GraphPointer, with_graph};
+use saturn::model::definitions::Definition;
+use saturn::model::ids::DefinitionId;
+
+/// C-compatible enum representing the kind of a definition.
+#[repr(C)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
```
vinistock: FWIW, there's a way to configure `cbindgen.toml` that allows us to define enums using patterns Rust expects, but outputs it modified for C.

It might be worth checking it out, so that we can define Rust enums regularly and still have them look nice in C too.

ext/saturn/definition.c
Morriar: Actually I realized that this is the wrong way to do it, instead we should make the `new` in `Definition` private.

I'll create a follow up to fix it on `Declaration` and `Document`. I also wonder if a common ancestor for `Handle` objects would make it easier.

ext/saturn/definition.c
Morriar: Good call, I removed this enum. I made it panic on the Rust side and made it raise on the Ruby side just in case

rust/saturn-sys/src/definition_api.rs
Morriar: TIL: `prefix_with_name`

Fixed, thanks for the pointer, it is better indeed.

ext/saturn/definition.c
Morriar: Although it should not compile on the Rust side because of the missing enum branch and then raise on the Ruby side because of the unknown kind, I added a comment.

Once we stabilize the native API we should properly document all of this and possibly add some guard rails to insure sync.

PR #252, opened 2025-10-23
[DOC] Add ruby-behaviors.md to document Ruby behaviors related to indexing

st0012:
I asked Claude to extract Ruby behavior context from the indexer's comments & tests as a separate doc, and then manually corrected/edited some of the points.

This can be useful for AI tools to make the right suggestion as well as for onboarding. We can even contribute this as official Ruby doc in the future.


docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+
+- `Foo::Bar`
+- `ActiveRecord::Base`
+- `Rails::Application::Configuration`
+
+**Top-level references** start with `::`:
+
+- `::Foo` (explicitly top-level)
+- `::Array` (explicitly top-level)
+
```
Morriar, line 55: Should we add a note that qualified names can be rooted: `::ActiveRecord::Base`?

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+1. [Lexical Scoping and Constant Resolution](#lexical-scoping-and-constant-resolution)
+2. [Definitions vs Declarations](#definitions-vs-declarations)
+3. [Method Parameters](#method-parameters)
+4. [Attribute Methods](#attribute-methods)
+5. [Variable Scoping](#variable-scoping)
+6. [Constant References](#constant-references)
+7. [Namespace Qualification](#namespace-qualification)
+
+---
+
```
Morriar: Let's remove all of them. This is styling the CSS could add if we wanted it.

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+The same unqualified name refers to different entities depending on context:
+
+```ruby
+module Foo
+  class Bar
+  end
+end
+
+# Inside Foo, "Bar" means "Foo::Bar"
+# Outside Foo, "Bar" means "Bar" (top-level)
```
Morriar: Let's contextualize:

```ruby
module Foo
  class Bar
  end

  puts Bar # Inside Foo, "Bar" means "Foo::Bar"
end

puts Bar # Outside Foo, "Bar" means "Bar" (top-level)
```

But should we also define a Bar outside to avoid confusion on what we're referring to here?

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Table of Contents
+
+1. [Lexical Scoping and Constant Resolution](#lexical-scoping-and-constant-resolution)
+2. [Definitions vs Declarations](#definitions-vs-declarations)
+3. [Method Parameters](#method-parameters)
+4. [Attribute Methods](#attribute-methods)
+5. [Variable Scoping](#variable-scoping)
+6. [Constant References](#constant-references)
+7. [Namespace Qualification](#namespace-qualification)
```
Morriar: I don't think this list is in the correct order

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+   class Foo
+     attr :foo, true   # Creates both foo and foo= (getter and setter)
+   end
+   ```
+
+4. **Second parameter is a string/symbol:** Creates both reader and writer for ALL arguments
+
+   ```ruby
+   class Foo
+     attr :foo, :bar, :baz
```
Morriar, line 326: Is this correct?

```rb
$ irb
irb:* class Foo
irb:*   attr :foo, :bar, :baz
irb:> end
irb:> Foo.new.foo = 1
(irb):in `<main>': undefined method `foo=' for an instance of Foo (NoMethodError)
```

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+- `Rails::Application::Configuration`
+
+**Top-level references** start with `::`:
+
+- `::Foo` (explicitly top-level)
+- `::Array` (explicitly top-level)
+
+### Singleton Method Naming
+
+Singleton methods (class methods) use the same name as instance methods:
```
Morriar:
```suggestion
Singleton methods (class methods) can use the same name as instance methods:
```

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+This creates:
+
+- `Foo::Bar`
+- `Foo::Bar::Baz::Qux` (nested under `Foo::Bar`)
+- `Quuux` (top-level, but lexically inside `Foo::Bar`)
+
+**Important:** `Foo::Bar::Baz::Qux::Quuux` does **not** exist because `::Quuux` resets to top level.
+
```
Morriar, line 153: Should we also explain constant resolution through inheritance?

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+This code creates:
+
+**3 separate definitions:**
+
+1. Module definition for `Foo`
+2. Class definition for `Foo::Bar` (inside `Foo`)
+3. Class definition for `Foo::Bar` (at top level)
+
+**2 distinct entities (declarations):**
```
Morriar: Since this is a core concept of the indexer's model:

```suggestion
**2 distinct declarations:**
```

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,482 @@
...
+
+1. The module `Foo`
+2. The class `Foo::Bar`
+
+**Key Points:**
+
+- A **definition** is a single definition occurrence of a class, module, method, etc. in a specific file at a specific location
+- A **declaration** represents the conceptual entity that may have multiple definitions
```
Morriar:
```suggestion
- A **declaration** represents the named entity that may have multiple definitions
```

docs/ruby-behaviors.md
st0012: Oh wow I think I misunderstood the behaviour and this is actually a bug too.

docs/ruby-behaviors.md
st0012: Added

docs/ruby-behaviors.md
st0012: https://github.com/Shopify/saturn/pull/259

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,524 @@
...
+
+```ruby
+module Foo
+  # "Bar" => "Foo::Bar"
+  # "::Bar" => "Bar"
+  class Baz
+    # "::Qux" => "Qux"
+    # "Qux" => "Foo::Baz::Qux"
+  end
+end
```
vinistock, line 98: I found this example a bit confusing because we don't have the definitions of `Bar` or `Qux`. Ho do we know it refers to `Foo::Bar` and not `Bar` if we don't have its definition?

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,524 @@
...
+- `*c`: rest positional (captures remaining positional args as an array)
+- `d`: post parameter (required positional after rest)
+- `e:`: required keyword parameter
+- `g: 42`: optional keyword parameter with default value
+- `**i`: rest keyword (captures remaining keyword args as a hash)
+- `&j`: block parameter
+
+### Forward Parameters (Ruby 2.7+)
+
+Ruby 2.7 introduced argument forwarding with `...`:
```
vinistock: I'd omit the version here since we only support 3 and forward anyway.

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,524 @@
...
+```
+
+Both `$foo` and `$bar` are accessible globally. The class definition doesn't affect the scope of `$bar`.
+
+### Instance Variables
+
+Instance variables are scoped to instances of classes. When defined at the class level, they belong to the class itself (as the class is an object):
+
+```ruby
+@foo = 1           # Top-level instance variable
```
```suggestion
@foo = 1           # Top-level instance variable (belongs to <main>)
```

docs/ruby-behaviors.md
```diff
@@ -0,0 +1,524 @@
...
+    CONST  # Resolves to Parent::CONST through inheritance
+  end
+end
+
+Child.new.show  # => "from parent"
+```
+
+**Resolution order with inheritance:**
+
+1. Current class
```
vinistock, line 492: Prepended modules comes first

CLAUDE.md
rafaelfranca, line 1: Can we rename this file to `AGENTS.md`? I don't like we are writing documentation that only work in one single tool.

docs/ruby-behaviors.md
vinistock: Sorry, the order is still flipped. Prepended modules come before the class itself.

It's should be:

- Prepended modules
- The class or module itself
- Included modules
- Parent class

And for each one of those, we need to linearize. When you prepend a module, you don't get just the module, but all of its own prepends and includes too.

Here we're only mentioning that this linearization happens for the parent class, but everything is linearized.

PR #255, opened 2025-10-23
Implement native extension `Definition#location`

Morriar:
This requires a few new things:

1. introduce `Location`
2. Add a Rust API to expose the location of a definition to the Ruby side
3. Implement the C plumbing to retrieve locations

I used it in a few places in test to showcase the features.

One note, for now the Ruby `Location` object contains all the things that are useful to different types of clients: uri, path, start/end offsets, lines and columns. Computing the lines and columns _is expensive_, using the `SourceLocationConverter` we need to read file contents and perform bytes arithmetic.

I think a nicer approach going forward would be to split `Definition#offset` from `Definition#location` so you pay the price only if needed. But I'll wait for the Ruby LSP integration to work on this to have a better understanding of the use cases.

While it's expensive, its impact is almost negligible when benching Spoom against core.

### DIFF
--- a/lib/saturn/location.rb
+++ b/lib/saturn/location.rb
@@ -2,6 +2,8 @@
 
 module Saturn
   class Location
+    include Comparable
+
     #: String
     attr_reader :uri
 
@@ -23,6 +25,15 @@ module Saturn
       raise Saturn::Error, "URI is not a file:// URI: #{@uri}" unless uri.scheme == "file"
 
       uri.path
+    end
+
+    #: (other: BasicObject) -> Integer
+    def <=>(other)
+      return -1 unless other.is_a?(Location)
+
+      a = [@uri, @start_line, @start_column, @end_line, @end_column]
+      b = [other.uri, other.start_line, other.start_column, other.end_line, other.end_column]
+      a <=> b
     end
 
     #: -> String
--- a/test/definition_test.rb
+++ b/test/definition_test.rb
@@ -44,20 +44,22 @@ class DefinitionTest < Minitest::Test
       graph = Saturn::Graph.new
       graph.index_all(context.glob("**/*.rb"))
 
-      defs = graph.declarations.map { |d| d.definitions.to_a }.flatten
-      refute_empty(defs)
+      defs = graph.declarations
+        .map { |d| d.definitions.to_a }
+        .flatten
+        .sort_by(&:location)
 
-      # Ensure at least one of each subtype is present
-      assert(defs.any? { |d| d.is_a?(Saturn::ClassDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::ModuleDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::ConstantDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::MethodDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::AttrAccessorDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::AttrReaderDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::AttrWriterDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::GlobalVariableDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::InstanceVariableDefinition) })
-      assert(defs.any? { |d| d.is_a?(Saturn::ClassVariableDefinition) })
+      assert_instance_of(Saturn::ClassDefinition, defs[0])
+      assert_instance_of(Saturn::ClassVariableDefinition, defs[1])
+      assert_instance_of(Saturn::AttrAccessorDefinition, defs[2]) # for the writer
+      assert_instance_of(Saturn::AttrAccessorDefinition, defs[3]) # for the reader
+      assert_instance_of(Saturn::AttrReaderDefinition, defs[4])
+      assert_instance_of(Saturn::AttrWriterDefinition, defs[5])
+      assert_instance_of(Saturn::ModuleDefinition, defs[6])
+      assert_instance_of(Saturn::ConstantDefinition, defs[7])
+      assert_instance_of(Saturn::MethodDefinition, defs[8])
+      assert_instance_of(Saturn::GlobalVariableDefinition, defs[9])
+      assert_instance_of(Saturn::InstanceVariableDefinition, defs[10])
     end
   end
 


lib/saturn/location.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Saturn
+  class Location
+    include Comparable
+
+    #: String
+    attr_reader :uri, :path
+
+    #: Integer
+    attr_reader :start_offset, :end_offset, :start_line, :end_line, :start_column, :end_column
```
vinistock: Do clients need the offsets if they already have the line/column information?

Also, note for posterity: classes in Ruby are the only entity that require a `name_loc` and a `location` to be stored. This is due to some weirdness in the LSP spec where if you try to jump to definition inside the class' own body, the editor fires a find references instead of a go to definition request (we discovered this in the Ruby LSP thanks to our own code).

```ruby
class Entry
  # The Entry reference here for the parent class of `Class` is inside its own 
  # definition body. The `location` of `Entry` fully covers the `Entry` reference
  #
  # This scenario makes the editor incorrectly trigger a references request 
  # instead of a definition. The only way to fix it is to use the extended 
  # location object, which includes the name location and the selection location
  class Class < Entry
  end
end
```

In the Ruby LSP, we ended up including name location for all entries, which is wasteful and unnecessary. I think the only case where this is possible is with classes and modules (although worth trying to see if recursive methods also trigger it).

lib/saturn/location.rb
```diff
@@ -0,0 +1,37 @@
...
+      @end_line = end_line
+      @start_column = start_column
+      @end_column = end_column
+    end
+
+    #: (other: BasicObject) -> Integer
+    def <=>(other)
+      return -1 unless other.is_a?(Location)
+
+      [uri, start_offset, end_offset] <=> [other.uri, other.start_offset, other.end_offset]
```
vinistock: Nit: I always prefer using instance variables instead of attribute readers. Not a big deal, but my vote goes to ivars. Same for `to_s`, if you decide to switch.

rust/saturn-sys/src/location_api.rs
```diff
@@ -0,0 +1,93 @@
...
+use saturn::source_location::{SourceLocationConverter, UTF8SourceLocationConverter};
+use std::ffi::CString;
+use std::fs;
+use url::Url;
+
+/// C-compatible struct representing a definition location with offsets and line/column positions.
+#[repr(C)]
+pub struct Location {
+    pub uri: *const c_char,
+    pub path: *const c_char,
```
vinistock: What do you think about returning only the URI (which contains more information) and letting the client decide if they want the file path?

We can even provide the method in the Ruby API (i.e.: `definition.file_path`), but I'd vote to do it lazily rather than eagerly. The LSP rarely needs file paths as it only deals with URIs.

rust/saturn-sys/src/location_api.rs
```diff
@@ -0,0 +1,93 @@
...
+///
+/// - If the URI cannot be converted to a file path.
+/// - If the file cannot be read.
+/// - If the offset cannot be converted to a position.
+#[must_use]
+pub(crate) fn create_location_for_uri_and_offset(uri: &str, start: u32, end: u32) -> *mut Location {
+    let path_str = file_path_from_uri(uri).unwrap_or_else(|| panic!("Failed to convert URI to file path: {uri}"));
+
+    let source = fs::read_to_string(&path_str).unwrap_or_else(|_| panic!("Failed to read file at path {path_str}"));
+    let converter = UTF8SourceLocationConverter::new(&source);
```
vinistock, line 42: We can definitely start like this, but it would be nice to have the encoding as an optional parameter that defaults to UTF-8, so that we can switch according to editors later.

Also, one thing I discussed with Soutaro that would be nice to prototype is if we can design an API for add-ons to translate their locations since they do not have (and shouldn't have) access to the documents' internal state.

Maybe the LSP will need some sort of method in its global state where it receives a location object with URI + offsets and returns URI + line/column.

rust/saturn-sys/src/location_api.rs
```diff
@@ -0,0 +1,93 @@
...
+    if ptr.is_null() {
+        return;
+    }
+    unsafe {
+        // Take ownership of the box so we can free inner allocations first
+        let mut boxed = Box::from_raw(ptr);
+
+        if !boxed.uri.is_null() {
+            let _ = CString::from_raw(boxed.uri.cast_mut());
+            boxed.uri = std::ptr::null();
```
vinistock: Why do we need to set the `uri` to null here?

ext/saturn/definition.c
```diff
@@ -42,11 +44,53 @@ VALUE definition_class_for_kind(DefinitionKind kind) {
...
 }
 
+// Helper to build a Ruby Saturn::Location from a C Location pointer.
+// Does not take ownership; caller must free the C Location via sat_definition_location_free.
+static VALUE build_location_value(Location *loc) {
+    if (loc == NULL) {
+        return Qnil;
+    }
+
+    VALUE mSaturn = rb_const_get(rb_cObject, rb_intern("Saturn"));
```
vinistock: Nit: we already receive the `Saturn` module in the `initialize` function, so I would add a `static VALUE mSaturn` on the top of this file, so that we can reuse everywhere without having to `const_get`.

rust/saturn-sys/src/location_api.rs
Morriar: But the LSP is the only to use `uri`? Everything else is `path` based

lib/saturn/location.rb
Morriar: Dropped the offsets for now 👍

rust/saturn-sys/src/location_api.rs
Morriar: Yeah I don't think we need it since we drop `boxed` at the end. Removed 👍

ext/saturn/definition.c
Morriar: Good idea, I also initialized the `cLocation` in the initialization to avoid resolving it over and over again.

rust/saturn-sys/src/location_api.rs
Morriar: Yeah I think we can do better for this but I lack use cases to figure the right API at the moment

rust/saturn-sys/src/location_api.rs
vinistock: Yeah, I was thinking we could save one string by making it on demand - since you can go from URI to file path, but it's not as easy to make the opposite transformation if you have an `untitled:Untitled-1` URI or a `git:///foo.rb` URI.

Do you think there would be any slow down from making this be computed on demand vs eagerly? I'm thinking something like:

```ruby
class Location
  #: () -> String?
  def file_path
    uri = URI(@uri)
    return unless uri.scheme == "file"

    uri.path
  end
end
```

Morriar: The thing is we still need to compute the path on the Rust side so we can open the file to translate offsets to line/col so that means doing it twice.

But I moved it as a method like you suggested 👍

vinistock: Ah, I see your point. Yeah, this API is definitely weird, we're going to have to invest more time figuring out how to make this work or make the conversion eager and just store everything we need from the get go.

PR #259, opened 2025-10-24
Fix incorrect attr indexing

st0012:
```ruby
class Foo
  attr "foo", :bar, :baz
end
```

should only create readers for `foo`, `bar`, and `baz`, not writers.


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -677,19 +677,10 @@ impl Visit<'_> for RubyIndexer<'_> {
                 // attr :foo, true        => both reader and writer
                 // attr :foo, false       => only reader
                 // attr :foo              => only reader
-                // attr :foo, "bar", :baz => both reader and writer for foo, bar, and baz
+                // attr :foo, "bar", :baz => only readers for foo, bar, and baz
```
Morriar, line 680: It would be the perfect place to link to your document once merged

PR #264, opened 2025-10-24
Make document serializable

vinistock:
I noticed that I forgot to add `members` handling for the database and that got me thinking that something in our approach to the db wasn't right. I think the issue clicked for me now.

If we keep queryable fields separate from the serialized structures (even if minimal duplication is required), loading things from the database becomes trivial and requires no code changes when adding more fields to an existing struct.

For example, we can make `Declaration` serializable and _still_ keep `name` as a separate column for query only purposes, which allows the load logic to be extremely simple while still allowing the queries we need.

This PR makes `Document` serializable and adds the `serialize` and `deserialize` convenience methods. `Declaration` is a bit more complicated, so I'll do it in a follow up PR.

### DIFF
--- a/rust/saturn/src/db/schema.sql
+++ b/rust/saturn/src/db/schema.sql
@@ -1,26 +1,26 @@
 -- SQLite schema for the index database
--- Contains tables and indexes for storing code declarations and their relationships
 
--- Table for storing documents
 CREATE TABLE IF NOT EXISTS documents (
+    -- Queryable fields
     id INTEGER PRIMARY KEY,  -- Hashed document ID
-    uri TEXT NOT NULL UNIQUE,
-    content_hash INTEGER NOT NULL
+    content_hash INTEGER NOT NULL,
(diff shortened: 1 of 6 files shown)


rust/saturn/src/db/schema.sql
```diff
@@ -1,26 +1,26 @@
...
 
--- Table for storing documents
 CREATE TABLE IF NOT EXISTS documents (
+    -- Queryable fields
     id INTEGER PRIMARY KEY,  -- Hashed document ID
-    uri TEXT NOT NULL UNIQUE,
-    content_hash INTEGER NOT NULL
+    content_hash INTEGER NOT NULL,
+    -- Serialized struct
+    data BLOB NOT NULL
```
vinistock, line 8: Here we do have the option of loading the entire document and extracting the content_hash, if that's fast enough. For now, I just considered content_hash as a queryable field.

PR #265, opened 2025-10-24
Add initial architecture docs

vinistock:
In pairing sessions, I kept finding it useful to come back to our graph's architecture drawing. Let's add it as an initial basic architecture doc that we can keep expanding.


docs/architecture.md
```diff
@@ -0,0 +1,24 @@
...
+relationship between them as edges. The visualization below demonstrates the conceptual structure of the graph. In
+practice, it is implemented through IDs between the entities - an adjacency list approach.
+
+[Open in Excalidraw](https://excalidraw.com/#json=hQiLSD8nJRVxONhuwtSn4,L78TkfeB4YL1HJTf5L0bvw)
+
+![Graph visualization](images/graph.png)
+
+The concepts mentioned are implemented in the following files:
+
+- `model/declaration.rs`: The global concept of a declaration, which may be composed of multiple definitions. Can also be called a symbol
```
Morriar: Let's avoid pushing multiple names for the same concept. We can let the reader make his own associations depending on their background.

```suggestion
- `model/declaration.rs`: The global concept of a declaration, which may be composed of multiple definitions
```

docs/architecture.md
```diff
@@ -0,0 +1,24 @@
...
+- `model/definitions.rs`: The file specific concept of a definition. This node only holds information that's local to a
```
Morriar: I think it's even more specific than a file/document, consider this:

```rb
# file foo.rb
class Foo; end
class Foo; end
```

We have 2 definitions of `Foo` in the same document.

I wonder if we should import the example from https://github.com/Shopify/saturn/blob/main/rust/saturn/src/model/definitions.rs#L3-L24 here to make things clearer?

CLAUDE.md
```diff
@@ -49,8 +49,8 @@ type checkers, linting and other code analysis features.
 
 ### Key files
 
-- `rust/saturn/src/model/graph.rs`: the Graph representation of the codebase, which is composed of nodes
-representing fully qualified names, definitions, URIs and the relationships between them (edges)
+- `rust/saturn/src/model/graph.rs`: the Graph representation of the codebase. Read more about the architecture of the graph
+in `docs/architecture.md`
```
st0012: Let's move this change to `AGENTS.md` as I've merged #252

docs/architecture.md
```diff
@@ -0,0 +1,24 @@
+# Architecture
+
+Saturn's representation of the Ruby code it analyzes is a graph, where key entities are represented as nodes and the
+relationship between them as edges. The visualization below demonstrates the conceptual structure of the graph. In
+practice, it is implemented through IDs between the entities - an adjacency list approach.
+
+[Open in Excalidraw](https://excalidraw.com/#json=hQiLSD8nJRVxONhuwtSn4,L78TkfeB4YL1HJTf5L0bvw)
```
st0012, line 7: I clicked the link and it says it could replace my own Excalidraw board, which is kinda scary.

Image

Since the only options left is to save the graph, perhaps just having the image below is enough?

docs/architecture.md
vinistock: I added an example involving 2 files and multiple definitions to make it clearer. Also reworded the phrase.

docs/architecture.md
vinistock: Why is it scary? My understanding is that Excalidraw can only ever open a single drawing at a time, so you get this warning whenever you're trying to load a drawing that you saved.

The idea of keeping the link there and not just the image is to make it easy to edit the drawing in the future if necessary. You just click the link, replace your current content with the graph, edit and exported it again.

st0012: But how do I recover the current content I have?
On the free plan I use I only have one big board and I don't see an option to create an account to save the current drawings to it.

Morriar: > But how do I recover the current content I have?

That's the fun part, you don't! And that's my issue with Excalidraw :(

vinistock: You need to save your existing content to a file first (or to a URL), then you load the graph and afterwards you load your previous drawing.

Basically, Excalidraw doesn't implement a way to save multiple drawings because it runs mostly only in the Browser with no server, so the database is your file system 😅.

st0012: 🫡  I was too afraid to find out 😂

docs/architecture.md
```diff
@@ -0,0 +1,36 @@
...
+
+- `model/document.rs`: The representation of a document registered in the system
+
+Connections between the nodes are made using IDs defined in `ids.rs`:
+
+- `DeclarationId`: the hashed version of a fully qualified name (e.g.: `Foo::Bar` or `Foo#my_method`)
+- `DefinitionId`: the hashed version of the URI and byte offset combination where we found the definition in addition to
+its name
+- `NameId`: the hashed version of an unqualified name (e.g.: `Bar` instead of `Foo::Bar`)
+- `UriId`: the hashed version of a string URI
```
Morriar: Putting the example first will help explain the concepts:

~~~suggestion
To understand the concepts in this graph, let's take a simple example:

```ruby
# foo.rb
class Foo
end

# other_foos.rb
class Foo; end
class Foo
```

- `model/document.rs`: The representation of a document registered in the system, generally a file like `foo.rb` or `other_foos.rb` in our example.
- `model/declaration.rs`: The global concept of a name, here `Foo` in our example, which may be composed of multiple definitions.
- `model/definitions.rs`: A single definition that contributes to a declaration. In our example one comes from `foo.rb` and two more from `other_foos.rb`.

Connections between the nodes are made using IDs defined in `ids.rs`:

- `DeclarationId`: the hashed version of a fully qualified name (e.g.: `Foo::Bar` or `Foo#my_method`)
- `DefinitionId`: the hashed version of the URI and byte offset combination where we found the definition in addition to its name
- `NameId`: the hashed version of an unqualified name (e.g.: `Bar` instead of `Foo::Bar`)
- `UriId`: the hashed version of a string URI
~~~

PR #267, opened 2025-10-24
Make declaration serializable

vinistock:
This PR makes `Declaration` serializable, which fixes the fact that we're not remembering `references` or `members` in the database. Through the same principle as in #264, it also simplifies loading things a bit further and I want to follow up with a PR that should make this even more straightforward.

Two points to this PR:

1. We need to enable the `rc` feature from `serde` to be able to serialize `Nesting`
2. I noticed that the serialize/deserialize convenience methods are identical no matter what struct we define them on, so I extracted that to a trait that can only be applied to structs that already derive `Serialize` and `Deserialize`

### DIFF
--- a/rust/saturn/Cargo.toml
+++ b/rust/saturn/Cargo.toml
@@ -17,7 +17,7 @@ url = "2.5.4"
 rusqlite = { version = "0.37.0", features = ["bundled"] }
 xxhash-rust = { version = "0.8.15", features = ["xxh3"] }
 clap = { version = "4.5.16", features = ["derive"] }
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive", "rc"] }
 rmp-serde = "1.3.0"
 glob = "0.3.2"
 bytecount = "0.6.9"
--- a/rust/saturn/src/model/serializable.rs
+++ b/rust/saturn/src/model/serializable.rs
@@ -1,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+pub trait Serializable: Serialize + for<'a> Deserialize<'a> {
+    /// Serializes the entity into the byte vector we store in the database
+    ///
+    /// # Panics
+    ///
+    /// This method will only panic if serialization fails, which should never happen
+    #[must_use]
+    fn serialize(&self) -> Vec<u8> {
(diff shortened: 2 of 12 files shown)


rust/saturn/Cargo.toml
```diff
@@ -17,7 +17,7 @@ url = "2.5.4"
 rusqlite = { version = "0.37.0", features = ["bundled"] }
 xxhash-rust = { version = "0.8.15", features = ["xxh3"] }
 clap = { version = "4.5.16", features = ["derive"] }
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive", "rc"] }
```
st0012, line 20: What's this for?

rust/saturn/src/model/serializable.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub trait Serializable: Serialize + for<'a> Deserialize<'a> {
+    /// Serializes the entity into the byte vector we store in the database
+    ///
+    /// # Panics
+    ///
+    /// This method will only panic if serialization fails, which should never happen
+    #[must_use]
+    fn serialize(&self) -> Vec<u8> {
+        rmp_serde::to_vec(self).expect("Serializing document should always succeed")
```
st0012: Does the `document` here mean `Document` or just mean data in general?

rust/saturn/Cargo.toml
vinistock: Serializing and deserializing `Rc` or `Arc` is not enabled by default. This is because there's no way to reconstruct the original structure of references, so it ends up creating extra copies.

I _think_ that should be fine, but we can re-evaluate if we find that we're spending too much memory after loading references from the database.

rust/saturn/src/model/serializable.rs
vinistock: It should've been data in general. Fixed the comment.

st0012: Hmm I don't seem to see updates around this code. Do you mean comment in other places?

vinistock: Huh, me neither. I think Graphite must have had some blip because my push didn't change the branch. https://github.com/Shopify/saturn/pull/276

PR #268, opened 2025-10-24
Add owner_id relationship to declarations

vinistock:
We're now indexing the `members` information, but we have no way of removing entries from it when someone deletes an existing definition. To be able to work backwards from the definition to the members that must be cleaned, we need the reverse of the `members` relationship.

This PR introduces the `owner_id` to declarations, so that we can walk the graph in the opposite direction as well. I'll follow up with the PR to delete `members` entries when definitions are removed, which should be the last piece in managing the `members` data.

### DIFF
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -22,16 +22,19 @@ pub struct Declaration {
     members: IdentityHashMap<NameId, DeclarationId>,
     /// The list of references that are made to this declaration
     references: Vec<ResolvedReference>,
+    /// The ID of the owner of this declaration
+    owner_id: DeclarationId,
 }
 
 impl Declaration {
     #[must_use]
-    pub fn new(name: String) -> Self {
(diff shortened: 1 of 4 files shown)


rust/saturn/src/model/declaration.rs
```diff
@@ -114,7 +122,7 @@ mod tests {
     #[test]
     #[should_panic(expected = "Cannot add the same exact definition to a declaration twice. Duplicate definition IDs")]
     fn inserting_duplicate_definitions() {
-        let mut decl = Declaration::new("MyDecl".to_string());
+        let mut decl = Declaration::new("MyDecl".to_string(), DeclarationId::from("Object"));
```
Morriar, line 125: Can we put `DeclarationId::from("Object")` in a constant somewhere?

vinistock: Yeah, doing that for `<main>` and `Object` is still on my todo list.

PR #269, opened 2025-10-24
Remove deleted declarations from `members`

vinistock:


rust/saturn/src/model/declaration.rs
```diff
@@ -144,4 +153,13 @@ mod tests {
         assert_eq!(removed, Some(member_decl_id));
         assert_eq!(decl.members.len(), 0);
     }
+
+    #[test]
+    fn unqualified_name_id() {
+        let decl = Declaration::new("Foo::Bar".to_string(), DeclarationId::from("Foo"));
```
Morriar, line 162: Let's add a test with just one name like `Object`

vinistock: Good call

rust/saturn/src/model/declaration.rs
```diff
@@ -111,6 +111,15 @@ impl Declaration {
     pub fn owner_id(&self) -> &DeclarationId {
         &self.owner_id
     }
+
+    // This will change once we fix fully qualified names to not use `::` as separators for everything. Also, we may
+    // want to actually store this in the struct. Currently, it is only used to cleanup a member that got deleted from
+    // the graph, so we're avoiding the extra memory cost by computing it on demand.
+    #[must_use]
+    pub fn unqualified_name_id(&self) -> NameId {
```
Morriar, line 119: Thinking more about it, this API is a bit awkward, we should just make an `unqualified_name` and let the client transform it into an ID if needed.

Morriar: https://github.com/Shopify/saturn/pull/278

PR #270, opened 2025-10-24
Index unresolved method references

Morriar:
This is a big piece of what's missing to support Spoom's deadcode detection.

I basically reproduced what we have in https://github.com/Shopify/spoom/blob/main/lib/spoom/model/references_visitor.rb#L23-L24 and imported the tests from https://github.com/Shopify/spoom/blob/main/test/spoom/model/references_visitor_test.rb#L61.

This change has quite an impact on performance: +14% time on indexing (2s more on core) and +51% more memory consumed (~470MB more). The steep memory increase is explained by the amount of references we're now storing: from 5,340,610 to 16,406,106 references: +207%.

I think we could make it optional depending on what the client needs.

Here's the comparison:

|  | Before   | After    | % change   |
| :------------: | :----: | :-----: | :----: |
| Initialization | 0.00   | 0.00    | 0.00   |
| Listing        | 6.85   | 6.69    | \-2.32 |
| Indexing       | 5.41   | 7.37    | +36.22  |
| Querying       | 0.41   | 0.40    | \-1.97 |
| Cleanup        | 0.00   | 0.00    | 0.00   |
| Total:         | 12.67  | 14.46   | +14.15  |
|     |        |         |        |
| Mem            | 915.33 | 1382.91 | +51.08  |

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -108,6 +108,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         println!("Found {} names", graph.declarations().len());
         println!("Found {} definitions", graph.definitions().len());
         println!("Found {} URIs", graph.documents().len());
+        println!("Found {} references", graph.unresolved_references().len());
     }
 
     Ok(())


rust/saturn/src/model/references.rs
```diff
@@ -19,6 +19,16 @@ pub struct ConstantReference {
     offset: Offset,
 }
 
+#[derive(Debug)]
+pub struct MethodReference {
```
vinistock, line 23: Should we store known receiver types here as well? For example:

```ruby
class Foo
  def bar
    something_on_self
    Baz.something_on_singleton
    "Something on literal".upcase
  end
end
```

In these cases, we don't know if the method is inherited or not, but we do know the receiver type and so we can fully resolve the references during the resolution phase (as opposed to other method calls that require type checking).

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -777,10 +917,45 @@ impl Visit<'_> for RubyIndexer<'_> {
             self.visit(&node.value());
         }
     }
+
+    fn visit_block_argument_node(&mut self, node: &ruby_prism::BlockArgumentNode<'_>) {
+        let expression = node.expression();
+        if let Some(expression) = expression {
+            match expression {
+                ruby_prism::Node::SymbolNode { .. } => {
+                    let symbol = expression.as_symbol_node().unwrap();
```
vinistock, line 930: Is this handling this case `map(&:method_ref)`?

Morriar: Yes it does, there is a test for it in `index_unresolved_method_references` 👍

rust/saturn/src/model/references.rs
Morriar: We may but it is still unclear to me what the field for the receiver should be? Are we storing an `ConstantReference` reference? Or the string name? What about `foo.bar`, should the receiver be another `MethodReference`?

I didn't even add the `nesting` like for `ConstantReference` because I couldn't see yet how to use it.

I _think_ this will actually come later with the type inference because we'll need to resolve the type of the receiver expression.

PR #271, opened 2025-10-28
Resolve constant from lexical scope and add resolution stats

st0012:
Resolution stats from Core

```
Overall reference counts:
  Total resolved:         936883
  Total unresolved:       3134499
  Total references:       4071382
  Resolution ratio:       23.0%
```

Entire stats from Core

```
Resolution statistics

Overall reference counts:
  Total resolved:         936883
  Total unresolved:       3134499
  Total references:       4071382
  Resolution ratio:       23.0%

Query statistics
  Total declarations:         697238
  With documentation:         79357 (11.4%)
  Without documentation:      617881 (88.6%)
  Total documentation size:   224973 bytes
  Multi-definition names:     14931 (2.1%)

Definition breakdown:
  Method               321831
  InstanceVariable     182242
  Module               178383
  Class                102105
  Constant              61709
  AttrReader            31875
  AttrAccessor          11766
  AttrWriter              384
  GlobalVariable           30
  ClassVariable            11

Timing breakdown
  Initialization      0.002s (  0.0%)
  Listing             0.946s (  5.4%)
  Indexing            3.814s ( 21.7%)
  Resolving           0.157s (  0.9%)
  Querying            0.201s (  1.1%)
  Database           10.678s ( 60.7%)
  Cleanup             1.789s ( 10.2%)
  Total:             17.587s

Maximum RSS: 736673792 bytes (702.55 MB)

Indexed 86587 files
Found 697238 names
Found 890336 definitions
Found 86587 URIs
```

We don't actually resolve and store any reference in non-stats mode for now. 
But if we do, the max RSS can grow a bit: `654.95 MB` -> `702.55 MB`.

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -90,6 +90,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     }
 
     if args.stats {
+        time_it!(resolving, {
+            graph.print_resolution_statistics();
+        });
         time_it!(querying, {
             graph.print_query_statistics();
         });
@@ -108,7 +111,6 @@ fn main() -> Result<(), Box<dyn Error>> {
         println!("Found {} names", graph.declarations().len());
         println!("Found {} definitions", graph.definitions().len());
         println!("Found {} URIs", graph.documents().len());
-        println!("Found {} references", graph.unresolved_references().len());
     }
 
     Ok(())
--- a/rust/saturn/src/model/graph.rs
+++ b/rust/saturn/src/model/graph.rs
@@ -8,7 +8,8 @@ use crate::model::document::Document;
 use crate::model::identity_maps::IdentityHashMap;
 use crate::model::ids::{DeclarationId, DefinitionId, NameId, UriId};
 use crate::model::integrity::IntegrityChecker;
-use crate::model::references::UnresolvedReference;
+use crate::model::references::{ResolvedReference, UnresolvedReference};
+use crate::stats;
 
 /// Holds IDs of entities that were removed from the graph
 pub struct RemovedIds {
@@ -187,8 +188,18 @@ impl Graph {
     pub fn resolve_reference(&self, reference: &UnresolvedReference) -> Option<&Declaration> {
         match reference {
             UnresolvedReference::Constant(constant) => {
-                if let Some(_nesting) = constant.nesting() {
-                    // Not implemented yet
+                if let Some(nesting) = constant.nesting() {
+                    let mut current_nesting = Some(nesting.as_ref());
+
+                    while let Some(nesting) = current_nesting {
+                        let declaration_id = nesting.declaration_id();
+                        if let Some(declaration) = self.declarations.get(declaration_id)
+                            && let Some(member) = declaration.members().get(constant.name_id())
+                        {
+                            return Some(self.declarations.get(member))?;
+                        }
+                        current_nesting = nesting.parent().as_ref().map(std::convert::AsRef::as_ref);
+                    }
                     None
                 } else {
                     // Top level reference
@@ -484,24 +495,23 @@ impl Graph {
 
         println!();
         println!("Query statistics");
-        println!();
         let total_declarations = self.declarations().len();
-        println!("Total declarations:         {total_declarations}");
+        println!("  Total declarations:         {total_declarations}");
         println!(
-            "With documentation:         {} ({:.1}%)",
+            "  With documentation:         {} ({:.1}%)",
             declarations_with_docs,
-            (declarations_with_docs as f64 / total_declarations as f64) * 100.0
+            stats::percentage(declarations_with_docs, total_declarations)
         );
         println!(
-            "Without documentation:      {} ({:.1}%)",
+            "  Without documentation:      {} ({:.1}%)",
             total_declarations - declarations_with_docs,
-            ((total_declarations - declarations_with_docs) as f64 / total_declarations as f64) * 100.0
+            stats::percentage(total_declarations - declarations_with_docs, total_declarations)
         );
-        println!("Total documentation size:   {total_doc_size} bytes");
+        println!("  Total documentation size:   {total_doc_size} bytes");
         println!(
-            "Multi-definition names:     {} ({:.1}%)",
+            "  Multi-definition names:     {} ({:.1}%)",
             multi_definition_count,
-            (f64::from(multi_definition_count) / total_declarations as f64) * 100.0
+            stats::percentage(multi_definition_count, total_declarations)
         );
 
         println!();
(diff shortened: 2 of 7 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -757,6 +771,13 @@ mod tests {
             "
         });
 
+        // Foo::CONST should be resolved to the Foo declaration
+        let const_ref = context.graph.unresolved_references.remove(0);
+        assert_eq!(
+            context.graph.resolve_reference(&const_ref).unwrap().name(),
```
Morriar: What do you think of extracting this as either a helper function or a macro?

Morriar: While at it, let's sekect the references by their location. It will make it much easier to understand which reference we're looking at instead of relying on the `unresolved_references` order.

```rs
        assert_eq!(
            pick_and_resolve("foo.rb:12:1-12:4"),
            "Foo"
        );
```

rust/saturn/src/model/graph.rs
```diff
@@ -757,6 +771,13 @@ mod tests {
             "
         });
 
+        // Foo::CONST should be resolved to the Foo declaration
```
Morriar: I guess this refers to the `Foo` in `Foo::CONST`? Are we talking about line 12 in `foo.rb`?

rust/saturn/src/model/graph.rs
```diff
@@ -507,6 +507,53 @@ impl Graph {
...
+            } else {
+                // Keep unresolved references
+                self.unresolved_references.push(reference);
+            }
+        }
+
+        // Count total resolved references across all declarations
+        let total_resolved_references: usize = self.declarations.values().map(|decl| decl.references().len()).sum();
+        let total_unresolved_references = self.unresolved_references.len();
+        let total_references = total_resolved_references + total_unresolved_references;
```
Morriar: Can't we move this higher and just do `let total_references = unresolved.len()`?

rust/saturn/src/model/graph.rs
```diff
@@ -507,6 +507,53 @@ impl Graph {
...
+                    unreachable!(
+                        "Declaration found for reference resolution but not found in graph: {}",
+                        declaration_name
+                    );
+                }
+            } else {
+                // Keep unresolved references
+                self.unresolved_references.push(reference);
+            }
+        }
```
Morriar, line 561: Let's move this to a `Graph#resolve_references` function?

rust/saturn/src/model/graph.rs
```diff
@@ -479,22 +479,21 @@ impl Graph {
...
             total_declarations - declarations_with_docs,
             ((total_declarations - declarations_with_docs) as f64 / total_declarations as f64) * 100.0
         );
-        println!("Total documentation size:   {total_doc_size} bytes");
+        println!("  Total documentation size:   {total_doc_size} bytes");
         println!(
-            "Multi-definition names:     {} ({:.1}%)",
+            "  Multi-definition names:     {} ({:.1}%)",
             multi_definition_count,
             (f64::from(multi_definition_count) / total_declarations as f64) * 100.0
```
Morriar: I'm seeing a bunch of percent computation, some using `from` some using `as`, let's extract this into a function.

rust/saturn/src/model/graph.rs
st0012: Yeah, updated the comment

rust/saturn/src/model/graph.rs
st0012: I've added some macros to simplify things while asserting offset and uri.

rust/saturn/src/model/graph.rs
```diff
@@ -768,38 +829,38 @@ mod tests {
...
-        );
+        context.graph.resolve_references();
 
-        // ::Baz doesn't exist
-        let const_ref = context.graph.unresolved_references.remove(0);
-        assert!(context.graph.resolve_reference(&const_ref).is_none());
+        let foo_declaration = context.graph.declarations.get(&DeclarationId::from("Foo")).unwrap();
+        assert_eq!(foo_declaration.references().len(), 2);
+        assert_constant_ref!(foo_declaration, 0, "Foo", "file:///foo.rb", 359, 362);
+        assert_constant_ref!(foo_declaration, 1, "Foo", "file:///bar.rb", 12, 17);
```
Morriar: Sorry I may not have been clear, the idea was to use the location to extract the reference to resolve, resolve it and compare the resolved constant name matches.

So with:

```rs
assert_constant_resolution!("foo.rb:12:1-12:4", "Foo")
```

We would:

1. Translate the location string `"foo.rb:12:1-12:4"` into a (uri, offset) pair using `SourceLocationConverter#byte_offset_from_position`
2. Select from `graph.unresolved_references` the one that matches the (uri, offset) pair
3. Resolve the reference to a declaration
4. Compare the name of the declaration to `"Foo"`

st0012: I'm not sure if querying with uri & offset will be a good experience for test writer.
If I got the column wrong, for example `"foo.rb:12:1-12:3"`, then the error I'd get is that there's no reference found on that location.
I feel it's more natural to 1) find reference with name `Foo`, and then 2) make sure the uri and offset is correct.

So yeah I agree we can improve the marco further, but I feel the assertion steps should be different.

Morriar: > If I got the column wrong, for example "foo.rb:12:1-12:3", then the error I'd get is that there's no reference found on that location.

Sure, but this is easy to fix, if the macro raise you _know_ you got your location wrong.

With the approach of resolving first, if the macro raises maybe the resolution didn't work, maybe your location is wrong, maybe your name is incorrect?

rust/saturn/src/main.rs
```diff
@@ -84,6 +84,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     time_it!(database, { graph.save_to_database() })?;
 
     if args.stats {
+        time_it!(resolving, {
+            graph.print_resolution_statistics();
+        });
```
vinistock, line 95: Is this timing the resolution or the amount of time that it takes to print the statistics?

st0012: Resolution because we currently only call `graph.resolve_references` in `print_resolution_statistics`

rust/saturn/src/model/graph.rs
st0012: Yeah I'm writing test helpers that can do location/offset conversion for, which is more complicated than I expected. But should be ready today

rust/saturn/src/model/graph.rs
```diff
@@ -769,38 +934,50 @@ mod tests {
...
-        );
+        context.graph.resolve_references();
 
-        // ::Baz doesn't exist
-        let const_ref = context.graph.unresolved_references.remove(0);
-        assert!(context.graph.resolve_reference(&const_ref).is_none());
+        let foo_declaration = context.graph.declarations.get(&DeclarationId::from("Foo")).unwrap();
+        assert_eq!(foo_declaration.references().len(), 2, "Foo should have 2 references");
+        assert_reference_to!(context, "Foo", "file:///foo.rb:11:0-11:3");
+        assert_reference_to!(context, "Foo", "file:///bar.rb:1:2-1:7");
```
st0012: @Morriar the new assertion looks like this

Morriar: It looks like lines & columns are 0 indexed? Should we +1 when creating the location so it's less surprising?

rust/saturn/src/model/graph.rs
```diff
@@ -509,13 +519,165 @@ impl Graph {
...
+    ///
+    /// This macro:
+    /// 1. Parses the location string into `(uri, start_offset, end_offset)`
+    /// 2. Finds the declaration by name
+    /// 3. Finds a reference to that declaration at the given uri and start offset
+    /// 4. Asserts the end offset matches
+    ///
+    /// Location format: "uri:start_line:start_column-end_line:end_column"
+    /// Example: `<file:///foo.rb:3:0-3:5>`
+    macro_rules! assert_reference_to {
```
Morriar: Let's make space to also resolve method references

```suggestion
    macro_rules! assert_constant_reference_to {
```

rust/saturn/src/main.rs
vinistock: Oh, I see.

rust/saturn/src/model/graph.rs
vinistock: Personally, I think starting at zero is more intuitive.

st0012: Given that Prism also uses 0 index for column, I'd prefer aligning with it

Morriar: It doesn't really follow what most tools do.

In VScode, the first line is 1, the first column is 1: 
image

Same on Github, the first line is 1: https://github.com/Shopify/saturn/blob/main/saturn.gemspec#L1.

Same on Slack:
image

PR #275, opened 2025-10-28
Clear database before indexing

jenny-codes:
When clear-db option is set, clear the db before indexing. With the
synchronization work, we will start comparing the filesystem with what's
in the db during indexing. We would want to make sure the indexing does
not opearte on the existing cache before we wipe the cache out.

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -1,11 +1,14 @@
-use std::error::Error;
+use std::{error::Error, fs};
 
 use clap::Parser;
 
 use saturn::{
     indexing::{self, errors::MultipleErrors},
     model::graph::Graph,
-    stats::{memory::MemoryStats, timer::{Timer, time_it}},
+    stats::{
+        memory::MemoryStats,
+        timer::{Timer, time_it},
+    },
     visualization::dot,
 };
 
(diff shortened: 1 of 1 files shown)


rust/saturn/src/main.rs
```diff
@@ -39,9 +42,14 @@ fn main() -> Result<(), Box<dyn Error>> {
         Timer::set_global_timer(Timer::new());
     }
 
+    let db_path = format!("{}/graph.db", &args.dir);
+    if args.clear_db {
+        let _ = fs::remove_file(&db_path);
```
st0012, line 47:
```suggestion
        fs::remove_file(&db_path)?;
```

st0012: @jenny-codes could you apply this change? Or do you feel it's not making the code better?

jenny-codes: Ah sorry I didn't see this comment. I didn't choose to return when errored here because then it will exit early when there is not an existing graph.db:

```
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

st0012: That's fair. Thanks 👍

PR #277, opened 2025-10-29
Add commands to dev.yml

jenny-codes:
Add some commands to dev: 

- `dev test`: Invokes both ruby and rust tests
- `dev lint`
- `dev format`
- `dev check`: Invokes ruby test, rust test, and lint

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -2,3 +2,22 @@ up:
   - ruby
   - bundler
   - rust
+
+commands:
+  test:
+    desc: Run Ruby and Rust test suites via Rake
+    run: |
+      bundle exec rake test
+      bundle exec rake cargo_test
+  lint:
+    desc: Run RuboCop and Rust linting
+    run: bundle exec rake lint
+  format:
+    desc: Format Ruby and Rust sources
+    run: bundle exec rake format
+
+# dev check by default runs each command in parallel across available cores
+check:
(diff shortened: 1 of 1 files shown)


dev.yml
```diff
@@ -2,3 +2,21 @@ up:
...
+  lint:
+    desc: Run RuboCop and Rust linting
+    run: bundle exec rake lint
+  format:
+    desc: Format Ruby and Rust sources
+    run: bundle exec rake format
+
+check:
+  test_ruby: bundle exec rake test
+  test_rust: bundle exec rake cargo_test
```
st0012, line 22: Why do we split them here but not in the `test` command? Is it recommended to have smaller commands when building the `check` command?
(I'm not against it, just curious)

jenny-codes: `dev check` runs each command in parallel across available cores, so this is just leveraging this nice feature

Source: https://github.com/shop/world/blob/d88c23052bed200e0b6398e707fb44b4243f7e9a/areas/tools/dev/lib/dev/commands/check.rb#L188

https://github.com/shop/world/blob/d88c23052bed200e0b6398e707fb44b4243f7e9a/areas/tools/dev/lib/dev/commands/check.rb#L117-L118

st0012: Thanks for the context 👍 Maybe add a comment above `check` to mention this?

PR #280, opened 2025-10-29
Remove unnecessary boxing

Morriar:
Not sure why we needed this cast?

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -53,10 +53,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         graph
     });
 
-    let (documents, errors) = time_it!(listing, {
-        let (documents, errors) = indexing::collect_documents(vec![args.dir]);
-        Ok::<_, Box<dyn Error>>((documents, errors))
-    })?;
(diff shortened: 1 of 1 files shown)


rust/saturn/src/main.rs
```diff
@@ -53,10 +53,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         graph
     });
 
-    let (documents, errors) = time_it!(listing, {
-        let (documents, errors) = indexing::collect_documents(vec![args.dir.clone()]);
-        Ok::<_, Box<dyn Error>>((documents, errors))
-    })?;
+    let (documents, errors) = time_it!(listing, { indexing::collect_documents(vec![args.dir.clone()]) });
```
vinistock: Do we need the clone?
```suggestion
    let (documents, errors) = time_it!(listing, { indexing::collect_documents(vec![args.dir]) });
```

Morriar: We don't but Alex is already removing it in https://github.com/Shopify/saturn/pull/272/files#diff-604b7fda5e4629f2d4c5ba13c962f4c37aa57092a48a8cb9a16595e4a9188befL49 👍

PR #285, opened 2025-10-30
Create IndexResult for individual file indexing result

jenny-codes:
To convey the outcome of individual indexing more clearly

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -20,6 +20,12 @@ pub mod errors;
 pub mod ruby_indexer;
 pub mod scope;
 
+pub enum IndexResult {
+    Completed(Box<IndexerParts>),
+    Skipped,
+    Errored(Vec<IndexingError>),
+}
+
 /// Indexes the given items, reading the content from disk and populating the given `Graph` instance.
 ///
 /// # Errors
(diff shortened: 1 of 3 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -60,7 +60,7 @@ impl<'a> RubyIndexer<'a> {
 
     #[must_use]
     pub fn into_parts(self) -> IndexerParts {
-        (Some(self.local_graph), self.errors)
+        (self.local_graph, self.errors)
```
jenny-codes, line 63: There is another potential refactoring opportunity here to simply return the local_graph and not the errors, as the current code does not actually store any error from the indexing. What do we think?

rust/saturn/src/indexing.rs
```diff
@@ -69,6 +69,12 @@ impl Document {
     }
 }
 
+pub enum IndexResult {
+    Completed(Box<IndexerParts>),
```
jenny-codes, line 24: We want to box this one because the byte differences between the three variants are big.

rust/saturn/src/indexing.rs
```diff
@@ -156,11 +164,18 @@ where
...
+        match result {
+            IndexResult::Completed(parts) => {
+                let (local_graph, errors) = *parts;
+                result_fn(local_graph);
+                all_errors.extend(errors);
+            }
+            IndexResult::Errored(errors) => {
+                all_errors.extend(errors);
+            }
+            IndexResult::Skipped => {}
```
vinistock:
```suggestion
            IndexResult::Skipped => {
              unreacheable!("We should not return information about skipped files")
            }
```

PR #287, opened 2025-10-30
Update rake tasks

jenny-codes:
- `bundle exec rake ruby_test` now houses the Ruby test suite, replacing the old `bundle exec
  rake test` role.
 - `bundle exec rake test` now runs both ruby_test and cargo_test.
 - `bundle exec rake check` fan-outs into linting plus both test suites; `dev check` now
  delegates to this command.

This makes the `dev` tasks a very thin layer of the Rake tasks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -18,8 +18,8 @@ test_config = lambda do |t|
   t.ruby_opts << ["--enable=frozen_string_literal"]
   t.test_files = FileList["test/**/*_test.rb"]
 end
-Rake::TestTask.new(test: :compile, &test_config)
-namespace(:test) { RubyMemcheck::TestTask.new(valgrind: :compile, &test_config) }
+Rake::TestTask.new(ruby_test: :compile, &test_config)
+namespace(:ruby_test) { RubyMemcheck::TestTask.new(valgrind: :compile, &test_config) }
 
 RuboCop::RakeTask.new
 
@@ -43,4 +43,7 @@ task compile_release: :clean do
   Rake::Task[:compile].invoke
 end
 
-task default: [:lint, :cargo_test, :test]
+multitask test: [:cargo_test, :ruby_test]
+multitask check: [:lint, :test]
+
+task default: :check
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,7 +40,8 @@ When necessary, commands can be executed for the Ruby code.
 - `bundle exec rake compile`: compiles both the Rust crate and C extension
 - `bundle exec rake lint`: lints both the Ruby and Rust code
 - `bundle exec rake format`: auto formats both the Ruby and Rust code
-- `bundle exec rake test`: runs all automated Ruby tests
+- `bundle exec rake test`: runs the Ruby and Rust test suites
+- `bundle exec rake ruby_test`: runs all automated Ruby tests
 - `bundle exec ruby -Itest test/specific_test.rb`: runs a specific test file
 
 ## Rust workspace
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The command `bundle exec rake compile` can be used to build both the Rust and C 
 
 ### Running tests
 
-Simply use the command `bundle exec rake test`, and the project will be built and tests located within `test/` will be run.
+Run `bundle exec rake test` to execute both the Ruby and Rust test suites. To run only the Ruby tests, use `bundle exec rake ruby_test`. The combined lint-and-test flow is available through `bundle exec rake check`.
 
 ### Running memory leak and other sanitization checks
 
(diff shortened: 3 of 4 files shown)


Rakefile
```diff
@@ -43,4 +43,7 @@ task compile_release: :clean do
   Rake::Task[:compile].invoke
 end
 
-task default: [:lint, :cargo_test, :test]
+multitask test: [:cargo_test, :ruby_test]
```
Morriar, line 46: I'm not convinced by the use of Rake's `multitask` here, `test` will fail silently if either `cargo_test` or `ruby_test` fails and because the two tasks are ran in parallel, if the first one fails, its output will be hidden at the top of the console.

Here's the output with a compilation error in the C extension:

```
$ rake test
(in /Users/at/src/github.com/Shopify/saturn)

******** Running cargo tests ********
cargo test
cd tmp/arm64-darwin23/saturn/3.4.4
/usr/bin/make
make: *** No rule to make target `../../../../ext/saturn/location.h', needed by `declaration.o'.  Stop.
rake aborted!
Command failed with status (2): [/usr/bin/make]
/Users/at/.gem/ruby/3.4.4/gems/rake-compiler-1.3.0/lib/rake/extensiontask.rb:195:in 'block (2 levels) in Rake::ExtensionTask#define_compile_tasks'
/Users/at/.gem/ruby/3.4.4/gems/rake-compiler-1.3.0/lib/rake/extensiontask.rb:194:in 'block in Rake::ExtensionTask#define_compile_tasks'
/Users/at/.gem/ruby/3.4.4/gems/rake-13.3.0/exe/rake:27:in '<top (required)>'
Tasks: TOP => test => ruby_test => compile => compile:arm64-darwin23 => compile:saturn:arm64-darwin23 => copy:saturn:arm64-darwin23:3.4.4 => tmp/arm64-darwin23/saturn/3.4.4/saturn.bundle
(See full trace by running task with --trace)
$     Finished `test` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running unittests src/lib.rs (target/debug/deps/saturn-4b38bdf3ff84276b)

running 86 tests
test model::db::tests::querying_content_hashes_by_uri ... ok
test model::graph::tests::saving_graph_to_database ... ok
test model::db::tests::removing_orphaned_entries ... ok
test model::graph::tests::unloading_a_document ... ok
test model::graph::tests::loading_a_document ... ok
test model::db::tests::saving_graph_to_the_database ... ok

test result: ok. 86 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

     Running unittests src/main.rs (target/debug/deps/saturn_cli-d7af8fc96a626c05)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/cli.rs (target/debug/deps/cli-cc96a9f7430abc5a)

running 4 tests
test prints_help ... ok
test prints_index_metrics ... ok
test visualize_simple_class ... ok
test dir_argument_variants ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.24s

     Running unittests src/lib.rs (target/debug/deps/saturn_sys-e938a81008a0b8ed)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests saturn

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

It's quite hard to notice that the compilation failed.

jenny-codes: > the tests ran on the previously compiled files.

Is it because the error is in the C extension so the cargo tests are not affected? As in, not because of the use of multitask?

Morriar: Indeed but this error is hidden at the beginning of the output as the terminal scrolls to the end of the cargo test. So unless you scroll back to the beginning to double check, nothing tells you it failed.

Morriar: LLM suggest something like this to capture the failure:

```rb
$test_status = {}

task :cargo_test do
  # Simulate possible failure
  # raise "Cargo test failed!" if rand(2) == 0
  puts "Running cargo_test"
end

task :ruby_test do
  # Simulate possible failure
  # raise "Ruby test failed!" if rand(2) == 0
  puts "Running ruby_test"
end

task :cargo_test_with_status do
  begin
    Rake::Task[:cargo_test].invoke
    $test_status[:cargo_test] = :passed
  rescue Exception => e
    $test_status[:cargo_test] = :failed
    puts "cargo_test failed: #{e.message}"
  end
end

task :ruby_test_with_status do
  begin
    Rake::Task[:ruby_test].invoke
    $test_status[:ruby_test] = :passed
  rescue Exception => e
    $test_status[:ruby_test] = :failed
    puts "ruby_test failed: #{e.message}"
  end
end

multitask test: [:cargo_test_with_status, :ruby_test_with_status] do
  if $test_status.values.include?(:failed)
    puts "Some tests failed!"
  else
    puts "All tests passed!"
  end
end
```

I'll be happy with just a `bin/test` script that runs `rake cargo_test && rake ruby_test` not in parallel

jenny-codes: Maybe there is a way to fix the display to make the errors clear when running both tests together, but we can still invoke the individual `ruby_test` and `cargo_test`. The main purpose here is for the `test` command to include all tests and not just part of the test suites.

jenny-codes: Actually just replacing the `multitask` with the regular `task` achieves this

PR #297, opened 2025-10-31
Implement native extension `Reference`

Morriar:
Make references accessible from Ruby through `Graph#unresolved_references`.

### DIFF
--- a/rust/saturn-sys/src/unresolved_reference_api.rs
+++ b/rust/saturn-sys/src/unresolved_reference_api.rs
@@ -1,0 +1,150 @@
+//! C API for exposing unresolved references through ID handles (like definitions)
+
+use std::ffi::CString;
+
+use crate::graph_api::{GraphPointer, with_graph};
+use crate::location_api::{Location, create_location_for_uri_and_offset};
+use libc::c_char;
+use saturn::model::ids::ReferenceId;
+use saturn::model::references::UnresolvedReference;
+
+/// Kind of unresolved reference for FFI dispatch
+#[repr(C)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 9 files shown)


ext/saturn/location.c
```diff
@@ -0,0 +1,22 @@
...
+VALUE cLocation;
+
+VALUE build_location_value(Location *loc) {
+    if (loc == NULL) {
+        return Qnil;
+    }
+
+    VALUE uri = rb_utf8_str_new_cstr(loc->uri);
+
+    VALUE kwargs = rb_hash_new();
```
vinistock: Since we already know the size
```suggestion
    VALUE kwargs = rb_hash_new_capa(5);
```

rust/saturn-sys/src/unresolved_reference_api.rs
```diff
@@ -0,0 +1,152 @@
...
+/// Assumes pointer is valid.
+///
+/// # Panics
+///
+/// This function will panic if the reference cannot be found.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sat_unresolved_reference_name(
+    pointer: GraphPointer,
+    reference_id: i64,
+) -> *const ::std::os::raw::c_char {
```
vinistock: Does this return a `const *char`? We've been using `libc::c_char` everywhere for that. Is there a difference with this other type?

ext/saturn/location.c
Morriar: Good call 👍

rust/saturn-sys/src/unresolved_reference_api.rs
Morriar: Good point, let's use `libc::c_char` here as well 👍

PR #300, opened 2025-10-31
Do not clone comments strings when not needed

Morriar:
Follow up to https://github.com/Shopify/saturn/pull/288#discussion_r2479612753.

### DIFF
--- a/rust/saturn/src/model/graph.rs
+++ b/rust/saturn/src/model/graph.rs
@@ -640,7 +640,7 @@ impl Graph {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_utils::GraphTest;
+    use crate::{model::comment::Comment, test_utils::GraphTest};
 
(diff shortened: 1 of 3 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -955,20 +955,14 @@ mod tests {
...
         );
 
         let definitions = context.graph.get("CommentedModule").unwrap();
         let def = definitions.first().unwrap();
         assert_eq!(
-            def.comments()
-                .iter()
-                .map(|c| c.string().clone())
-                .collect::<Vec<String>>(),
+            def.comments().iter().map(Comment::string).collect::<Vec<&String>>(),
```
st0012, line 965: I think this can be applied to `assert_definition_comments`, and use it in these 2 tests too?

Morriar: Yes, good call

PR #303, opened 2025-11-04
Split constant and method references into their own maps

vinistock:
It was quite odd to have to deal with any type of reference everywhere. We mostly just added `unreacheable` or something of the sort to appease the type system when we knew something wasn't possible.

That's not super nice, so I propose we split each type of reference into their own map. That's all this PR does, splitting method and constant references into distinct hashmaps.

### Notes

I experimented with creating different ID types for constants and methods. I think it's nice, but the moment we do it, then the question becomes: what happens to the list of resolved references inside declarations?

We could put all types of reference IDs in an enum, but that's a bit dirty. It's not valid to have a `MethodRefId` inside a `ClassDeclaration` for example.

If we want to have distinct IDs for references, I'd suggest we also specialize `Declaration` into an enum for all different types of entities that exist.

This is why I did not include this as part of the PR.

### Performance

No difference when it comes to indexing time, but a bit of a reduction in memory usage. In Core:

```
BEFORE
2218.73 MB
AFTER
1633.59MB
```


rust/saturn/src/model/references.rs
```diff
@@ -18,57 +19,50 @@ pub struct ConstantReference {
...
 }
 
+/// An unresolved or resolved constant reference
 #[derive(Debug)]
-pub enum ResolvedReference {
-    Constant(Box<ConstantReference>),
-    Method(Box<MethodReference>),
+pub enum ConstantReference {
+    Resolved(Box<ConstantRef>),
+    Unresolved(Box<ConstantRef>),
```
Morriar: I would expect both structs to have different fields. For example a ResolvedConstantReference should point to the declaration it resolves to?

rust/saturn/src/model/references.rs
```diff
@@ -126,28 +121,35 @@ impl ConstantReference {
     }
 }
 
+/// An unresolved or resolved method reference
+#[derive(Debug)]
+pub enum MethodReference {
+    Resolved(Box<MethodRef>),
+    Unresolved(Box<MethodRef>),
```
Morriar: Same I'd expect the `ResolveMethodReference` to point to a declaration?

Since we do not have resolution here yet, could we have only one enum:

```rs
pub enum MethodReference {
    Unresolved(Box<UnresolvedMethodReference>),
}
```

rust/saturn/src/model/references.rs
vinistock: We currently don't need the back reference because `Declaration` keeps a list of resolved reference IDs. We can definitely add it later if we need it, but currently there's no use case for it.

rust/saturn/src/model/references.rs
vinistock: Same reason as the other one. The distinction will be useful later to avoid work, but we don't currently need the back references.

rust/saturn/src/model/references.rs
Morriar: How will we implement go to declaration when clicking on a constant?

rust/saturn/src/model/references.rs
Morriar: We currently don't resolve anything about methods, can we remove the enum instead and just keep the `UnresolvedMethodReference`?

rust/saturn/src/model/references.rs
vinistock: I added it to this PR. We will eventually need it, but we don't currently use it for anything.

rust/saturn/src/model/references.rs
vinistock: Sure

Issue #253, opened 2025-10-23
Constant paths are being handled incorrectly


vinistock:
Right now, we're considering a constant path's slice as its name. In addition to creating potentially incorrect names, like in this case:

```ruby
class Foo:: # still part of the slice
  Bar
end
```

There's another incorrect aspect, which is where we are associating things. In the example above, we're defining `Bar` as a member of `Foo` - which we actually may not have discovered yet. Similar to constants like here:

```ruby
class Foo
  A::B = 1
end
```

This defines the constant `B` inside `Foo::A` (which has to be a module or a class).

PR #304, opened 2025-11-04
Start tracking parent classes

vinistock:
This PR starts tracking parent classes by assigning a constant reference ID to the `ClassDefinition` struct.

Our idea is that the definition can remember the ID without actually knowing whether the reference has already been resolved or not. That allows us to save a bit of work from not having to resolve a reference again if it was already resolved before and not changed.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -11,7 +11,7 @@ use crate::model::definitions::{
     ModuleDefinition, Parameter, ParameterStruct,
 };
 use crate::model::graph::Graph;
-use crate::model::ids::{DeclarationId, UriId};
+use crate::model::ids::{DeclarationId, ReferenceId, UriId};
 use crate::model::references::{ConstantReference, MethodRef, UnresolvedConstantRef};
 use crate::offset::Offset;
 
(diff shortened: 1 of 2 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -354,10 +359,6 @@ impl Visit<'_> for RubyIndexer<'_> {
             }
         }
 
-        if let Some(superclass) = node.superclass() {
-            self.visit(&superclass);
-        }
-
```
Morriar, line 361: I think we're loosing the references on constant paths:

```rb
class Foo < Bar::Baz
            ^^^ this doesn't get visited anymore
end
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2277,9 +2293,7 @@ mod tests {
 
         assert_eq!(
             collect_constant_reference_names(&context.graph),
-            vec![
-                "C1", "C2", "C3", "C4", "C4::C5", "C6", "C6::C7", "C8", "C9", "C10", "C10::C11",
-            ]
+            vec!["C1", "C2", "C3", "C4::C5", "C6::C7", "C8", "C9", "C10", "C10::C11",]
```
Morriar: Yeah this is a problem, we're missing `C4`  and `C6` now

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2533,4 +2547,25 @@ mod tests {
 
         assert_eq!(collect_method_reference_names(&context.graph), vec!["m1", "m2"]);
     }
+
+    #[test]
+    fn superclasses_are_indexed_as_constant_ref_ids() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            "
+            class Foo < Bar; end
```
Morriar, line 2583: Let's also add a test for `class Foo < Bar::Baz`

rust/saturn/src/model/definitions.rs
```diff
@@ -118,18 +118,31 @@ pub struct ClassDefinition {
     uri_id: UriId,
     offset: Offset,
     comments: Vec<Comment>,
+    super_class: Option<ReferenceId>,
```
Morriar: Why do you think of something like `superclass_ref` to avoid ambiguity with an actual inheritance tree?

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Let's also add a test where the superclass is not a constant: `class Foo < Bar[1.0]`

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: My idea was to leave it as is until we address #253 since we're not currently linking the references correctly. There's no way to actually resolve properly until we fix that.

Does Spoom need the separation to exist already?

Morriar: I'm not sure I understand why we would not want to have those references indexed anymore?

From what I understand you would instead nest them so `UnresolvedConstantReference(Baz)` points to `UnresolvedConstantReference(Bar)`?

If I run an LSP rename constant action for example, shouldn't they still be listed so we don't have to visit each reference going down their scope?

Morriar: It also makes the logic inside Spoom more complex, I'd keep the indexing as it is.

rust/saturn/src/model/definitions.rs
vinistock: Sure

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1032,16 +1052,31 @@ mod tests {
         assert_eq!(definitions.len(), 1);
         assert_definition_offset!(definitions[0], 0, 50);
         assert_owner!(context.graph, "Foo", "Object");
+        if let Definition::Class(class_def) = &definitions[0] {
```
Morriar, line 1055: Not sure these changes are still relevant to this test

vinistock: I just wanted to verify that classes with no parent have `None` and it felt like overkill to create a new example just to verify that. If you want, I can split.

PR #305, opened 2025-11-04
Track includes and prepends

vinistock:
This PR starts tracking includes and prepends. It only comprehends remembering includes and prepends and not resolving inheritance chain.

### Implementation notes

1. We're currently not handling include and prepend at the top level, which end up in the `Object` class. I think we can solve this by creating a lazy `Definition` for `Object` whenever we find top level includes or prepends, so that we can attach them to a specific definition. It would be a bit weird to put the includes and prepends into any random definition for `Object`, so I think lazily creating a new one will probably be less surprising. Sorbet handles this by always considering the top level like a re-opening of `Object`, but that would force us to use a lot more memory. In a Rails app, people rarely define anything at the top level
2. There are minimal changes to how constant path references are tracked, which are still a symptom of #253

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -7,11 +7,11 @@ use crate::indexing::scope::Scope;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
     AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
-    ConstantDefinition, Definition, GlobalVariableDefinition, InstanceVariableDefinition, MethodDefinition,
+    ConstantDefinition, Definition, GlobalVariableDefinition, InstanceVariableDefinition, MethodDefinition, Mixin,
     ModuleDefinition, Parameter, ParameterStruct,
 };
 use crate::model::graph::Graph;
-use crate::model::ids::{DeclarationId, ReferenceId, UriId};
+use crate::model::ids::{DeclarationId, DefinitionId, ReferenceId, UriId};
 use crate::model::references::{ConstantReference, MethodRef, UnresolvedConstantRef};
 use crate::offset::Offset;
 
@@ -32,6 +32,7 @@ pub struct RubyIndexer<'a> {
     comments: Vec<CommentGroup>,
     source: &'a str,
     scope: Scope,
+    namespace_stack: Vec<DefinitionId>,
 }
 
 impl<'a> RubyIndexer<'a> {
@@ -47,6 +48,7 @@ impl<'a> RubyIndexer<'a> {
             comments: Vec::new(),
             source,
             scope: Scope::new(),
+            namespace_stack: Vec::new(),
         }
     }
 
@@ -306,6 +308,57 @@ impl<'a> RubyIndexer<'a> {
         let reference = MethodRef::new(name_id, self.uri_id, offset);
         self.local_graph.add_method_reference(reference);
     }
+
+    fn handle_mixin(&mut self, node: &ruby_prism::CallNode, prepend: bool) {
+        if let Some(arguments) = node.arguments() {
+            // Collect all arguments as constant references. Ignore anything that isn't a constant
+            let reference_ids = arguments
+                .arguments()
+                .iter()
+                .filter_map(|arg| self.index_constant_reference(&arg))
+                .collect::<Vec<_>>();
+
+            if let Some(owner_id) = self.namespace_stack.last()
+                && let Some(current_owner) = self.local_graph.get_definition_mut(*owner_id)
+            {
+                for id in reference_ids {
+                    match current_owner {
+                        Definition::Class(class_def) => {
+                            if prepend {
+                                class_def.add_mixin(Mixin::Prepend(id));
+                            } else {
+                                class_def.add_mixin(Mixin::Include(id));
+                            }
+                        }
+                        Definition::Module(module_def) => {
+                            if prepend {
+                                module_def.add_mixin(Mixin::Prepend(id));
+                            } else {
+                                module_def.add_mixin(Mixin::Include(id));
+                            }
+                        }
+                        _ => {}
+                    }
+                }
+            }
+        }
+    }
+
+    /// Visits every part of a call node, except for the message itself. Convenient for when we're only interested in
+    /// continuing the traversal
+    fn visit_call_node_parts(&mut self, node: &ruby_prism::CallNode) {
+        if let Some(receiver) = node.receiver() {
+            self.visit(&receiver);
+        }
+
+        if let Some(arguments) = node.arguments() {
+            self.visit_arguments_node(&arguments);
+        }
+
+        if let Some(block) = node.block() {
+            self.visit(&block);
+        }
+    }
 }
 
 struct CommentGroup {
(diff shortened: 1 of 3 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -101,6 +101,11 @@ impl Graph {
         )
     }
 
+    #[must_use]
+    pub(crate) fn get_definition(&mut self, definition_id: DefinitionId) -> Option<&mut Definition> {
+        self.definitions.get_mut(&definition_id)
+    }
+
```
Morriar, line 108: Let's not make the `mut` version the default. Do we really need this method in `Graph`?

rust/saturn/src/model/definitions.rs
```diff
@@ -119,11 +139,12 @@ pub struct ClassDefinition {
     offset: Offset,
     comments: Vec<Comment>,
     superclass_ref: Option<ReferenceId>,
+    mixins: Vec<Mixin>,
 }
 
 impl ClassDefinition {
     #[must_use]
-    pub const fn new(
```
Morriar, line 126: Why do we need to remove the const on this one but not on `ModuleDefinition`?

rust/saturn/src/model/definitions.rs
```diff
@@ -105,6 +107,24 @@ impl Definition {
...
 
+/// Represents either an included or a prepended module. Note that `extend` is modeled as an include on the singleton
+/// class
+#[derive(Debug)]
+pub enum Mixin {
+    Include(ReferenceId),
+    Prepend(ReferenceId),
+}
+
+impl Deref for Mixin {
```
Morriar, line 119: Let's add a comment explaining why we need this

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -284,6 +286,46 @@ impl<'a> RubyIndexer<'a> {
...
+                .filter_map(|arg| match arg {
+                    ruby_prism::Node::ConstantReadNode { .. } | ruby_prism::Node::ConstantPathNode { .. } => {
+                        Some(self.index_constant_reference(&arg.location()))
+                    }
+                    _ => None,
+                })
+                .collect::<Vec<_>>();
+
+            if let Some(owner_id) = self.namespace_stack.last()
+                && let Some(current_owner) = self.local_graph.get_definition(*owner_id)
```
Morriar: Can we extract this into a `current_definition` function?

Morriar: Also, let's return earlier if we don't get any `current_definition`

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -757,6 +806,42 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                        self.visit(&receiver);
+                    }
+
+                    if let Some(arguments) = node.arguments() {
+                        self.visit_arguments_node(&arguments);
+                    }
+
+                    if let Some(block) = node.block() {
+                        self.visit(&block);
+                    }
```
Morriar: What do you think of extracting this into a function so you can reuse it in both branch?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -757,6 +806,42 @@ impl Visit<'_> for RubyIndexer<'_> {
                     }
                 }
             }
+            "include" => {
+                let receiver = node.receiver();
+                if receiver.is_none() || receiver.as_ref().is_some_and(|r| r.as_self_node().is_some()) {
+                    self.handle_mixin(node, false);
```
Morriar, line 840: This means we're also loosing the constant references in these calls right?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2328,8 +2413,8 @@ mod tests {
         assert_eq!(
             collect_constant_reference_names(&context.graph),
             vec![
-                "M1", "M2", "M2::M3", "M4", "M5", "M5::M6", "M7", "M8", "M8::M9", "M10", "M11", "M12", "M13", "M14",
-                "M15", "M16", "M17", "M18", "M18::M19",
+                "M1", "M2::M3", "M4", "M5", "M5::M6", "M7", "M8::M9", "M10", "M11", "M12", "M13", "M14", "M15", "M16",
```
Morriar: Let's keep visiting so we don't change this please

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2619,4 +2704,176 @@ mod tests {
...
+        context.index_uri("file:///foo.rb", {
+            "
+            prepend Bar, Baz, ignored, 123
+            prepend Qux
+            "
+        });
+
+        // This is currently wrong! We should be inserting top level includes into an artificial `Object` definition
+        assert!(context.graph.get("Object").is_none());
+    }
```
Morriar, line 2882: We should also show that wee don't currently have support for singleton classes:

```rb
class Foo
  class << self
    include Bar
  end
end
```

This will push the include in the `Foo` class right?

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Fixed.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Let's wait to make this refactor just a bit. Every include always has an owner, which is either going to be the current namespace or `Object`.

I think we may need to create a lazy definition for `Object` when we find top level include/prepend, but that's an entire PR by itself, so I'd like to leave that separate.

rust/saturn/src/model/definitions.rs
vinistock: That was leftovers from a previous approach.

rust/saturn/src/model/graph.rs
vinistock: We do need it because we want to insert the include/prepend entries to a definition that we already inserted in the graph.

With the work to make indexing non-global, this may change, but at this moment the `members` require the definition to be inserted eagerly in the graph, which then requires us to be able to get a mutable reference to modify them in `visit_call_node`.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: We don't currently have much related to singletons, so instead of putting it in this PR, I created a list of everything we have to revisit for them https://github.com/Shopify/saturn/issues/79#issuecomment-3492767750.

rust/saturn/src/model/graph.rs
Morriar: I meant, do we need this method, can't we just call `self.definitions().get_mut(&definition_id)` instead?

vinistock: You can't because `definitions()` returns an immutable reference. If I try to make that change, Rust complains with `cannot get a mutable reference from a &`.

Morriar: Ah true, what do you think renaming this method `get_mut_definition`?

All the other accessors we provide are always immutable and I'd like to avoid confusion.

vinistock: Already did, but I used `get_definition_mut` to stay consistent with how Rust does it.

PR #314, opened 2025-11-07
Ensure resolution results are cached

vinistock:
This PR moves the resolution phase outside of the `Graph` and ensures that we're caching intermediate resolution results.

It's important for us to move this outside of `Graph` for similar reasons as the indexing phase is separate: it's easier to satisfy the borrow rules if we're working with a reference to a graph than trying to do things internally. This will be particularly helpful for when we introduce parallelism to this phase since the `Graph` cannot put itself inside an `Arc`.

### DIFF
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -1,0 +1,317 @@
+use std::sync::Arc;
+
+use crate::{
+    indexing::nesting_stack::Nesting,
+    model::{
+        graph::Graph,
+        ids::{DeclarationId, NameId, ReferenceId},
+        references::ConstantReference,
+    },
+};
+
+/// Attempts to resolve all constant references stored in the graph
+pub fn resolve_all_constants(graph: &mut Graph) {
+    let filter_map: Vec<ReferenceId> = graph
+        .constant_references()
+        .iter()
+        .filter_map(|(id, reference)| match reference {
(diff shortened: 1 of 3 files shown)


rust/saturn/src/resolution.rs
```diff
@@ -0,0 +1,317 @@
...
+    model::{
+        graph::Graph,
+        ids::{DeclarationId, NameId, ReferenceId},
+        references::ConstantReference,
+    },
+};
+
+/// Attempts to resolve all constant references stored in the graph
+pub fn resolve_all_constants(graph: &mut Graph) {
+    let filter_map: Vec<ReferenceId> = graph
```
vinistock, line 14: I'm not super happy with collecting all IDs that need to be resolved ahead of time, but if we try to loop through `graph.constant_references()` we open an immutable reference to the graph that blocks `resolve_constant` from being called.

rust/saturn/src/resolution.rs
```diff
@@ -0,0 +1,317 @@
...
+            ConstantReference::Resolved(_) => None,
+        })
+        .collect();
+
+    for id in filter_map {
+        resolve_constant(graph, id);
+    }
+}
+
+/// Attempts to resolve a constant reference against the graph. Returns the fully qualified declaration ID that the
```
vinistock, line 28: From this point in the file down, the code is exactly the same as before. I only extracted helper methods to make it easier to digest.

PR #318, opened 2025-11-13
Major rewrite of indexing to focus on a local graph

Morriar:
With this change we introduce the concept of a `LocalGraph`:

* A `LocalGraph` represents the contents of one file only (definitions)
* It doesn't concern itself with global concepts (declarations)
* `LocalGraph` instances can be merged into a more global `Graph` that knows about declarations

On side effect of this change is removing the relationship `Definiion` -> `Declaration`.

### DIFF
--- a/rust/saturn/src/model/definitions.rs
+++ b/rust/saturn/src/model/definitions.rs
@@ -28,7 +28,7 @@ use std::ops::Deref;
 use crate::{
     model::{
         comment::Comment,
-        ids::{DeclarationId, ReferenceId, UriId},
+        ids::{DefinitionId, NameId, ReferenceId, UriId},
     },
     offset::Offset,
 };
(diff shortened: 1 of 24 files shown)


rust/saturn/src/model/definitions.rs
```diff
@@ -135,46 +151,90 @@ impl Deref for Mixin {
 /// ```
 #[derive(Debug)]
 pub struct ClassDefinition {
-    declaration_id: DeclarationId,
+    name_id: NameId,
     uri_id: UriId,
     offset: Offset,
     comments: Vec<Comment>,
+    owner_id: Option<DefinitionId>,
+    members: Vec<DefinitionId>,
```
vinistock, line 159: I wonder if we should call this something else to avoid confusion with the global version of members. We don't have to address it in this PR, but I think it might be worth brainstorming some alternatives so that we can more easily distinguish.

rust/saturn/src/model/definitions.rs
```diff
@@ -135,46 +151,90 @@ impl Deref for Mixin {
...
 
-    pub fn add_mixin(&mut self, mixin: Mixin) {
-        self.mixins.push(mixin);
+    #[must_use]
+    pub fn name_id(&self) -> &NameId {
+        &self.name_id
+    }
+
+    #[must_use]
+    pub fn uri_id(&self) -> &UriId {
```
vinistock, line 197: No need to address in this PR (especially to avoid conflicts), but I think we could create a simple macro to generate readers for all fields that are common to all definition structs.

PR #319, opened 2025-11-13
Add the runtime dependency in the gemspec:

Edouard-chin:
- The "extconf_compile_commands_json" dependency is needed by the gem it self at compilation time.

  If we add it to the Gemfile, then running `gem install saturn` can't work as the dependency from the Gemfile don't get installed b RubyGems.

### DIFF
--- a/ext/saturn/extconf.rb
+++ b/ext/saturn/extconf.rb
@@ -71,6 +71,10 @@ new_makefile.gsub!("$(Q) $(POSTLINK)", <<~MAKEFILE.chomp)
 MAKEFILE
 File.write("Makefile", new_makefile)
 
-require "extconf_compile_commands_json"
-ExtconfCompileCommandsJson.generate!
-ExtconfCompileCommandsJson.symlink!
+begin
+  require "extconf_compile_commands_json"
+
+  ExtconfCompileCommandsJson.generate!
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -10,5 +10,4 @@ gem "rake-compiler"
 gem "minitest"
 gem "rubocop"
 gem "rubocop-shopify"
-gem "extconf_compile_commands_json"
```
Morriar, line 13: Could this be a `development` gem only?

Edouard-chin: I don't think so, we need it for compilation (when a user installs the gem) https://github.com/Shopify/saturn/blob/93c2c6eda1bffb7df8bbee9565952fac2df7566b/ext/saturn/extconf.rb#L74

Morriar: This generates a `compile_commands.json` file that is solely used for LSP/editor purposes (see https://github.com/KJTsanaktsidis/extconf_compile_commands_json). I wonder is there is a way to remove this hard dependency 🤔

vinistock: Yeah, this is only for the clangd language server support. Let's just require this conditionally and avoid making it a runtime dependency.

```ruby
begin
  require "extconf_compile_commands_json" 
rescue LoadError
end
```

PR #320, opened 2025-11-13
Make class and module names constant references

vinistock:
This PR makes classes and modules store a constant reference ID rather than a name. This allows us to delay naming classes and modules to the resolution phase, when we can actually figure out the correct fully qualified name.

For now, I didn't make any modifications to resolution to minimize the amount of changes in one go. So we're still incorrectly resolving more complex cases. All we're doing is building the fully qualified name with each nesting part of the name.

I will follow up with changes to make the naming happen during resolution.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1073,7 +1073,7 @@ mod tests {
 
     macro_rules! assert_name_eq {
         ($context:expr, $expect_name_string:expr, $def:expr) => {{
-            let actual_name = $context.graph().strings().get($def.name_id()).unwrap();
+            let actual_name = $context.graph().strings().get($def.str_id()).unwrap();
 
             assert_eq!($expect_name_string, actual_name);
         }};
--- a/rust/saturn/src/model/definitions.rs
+++ b/rust/saturn/src/model/definitions.rs
@@ -327,7 +327,7 @@ impl ModuleDefinition {
 /// ```
 #[derive(Debug)]
 pub struct ConstantDefinition {
-    name_id: StringId,
+    str_id: StringId,
     uri_id: UriId,
     offset: Offset,
     comments: Vec<Comment>,
@@ -337,14 +337,14 @@ pub struct ConstantDefinition {
 impl ConstantDefinition {
     #[must_use]
     pub const fn new(
-        name_id: StringId,
+        str_id: StringId,
         uri_id: UriId,
         offset: Offset,
         comments: Vec<Comment>,
         owner_id: Option<DefinitionId>,
     ) -> Self {
         Self {
-            name_id,
+            str_id,
             uri_id,
             offset,
             comments,
@@ -354,12 +354,12 @@ impl ConstantDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.name_id))
+        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.str_id))
     }
 
     #[must_use]
-    pub fn name_id(&self) -> &StringId {
-        &self.name_id
+    pub fn str_id(&self) -> &StringId {
+        &self.str_id
     }
 
     #[must_use]
@@ -392,7 +392,7 @@ impl ConstantDefinition {
 /// ```
 #[derive(Debug)]
 pub struct MethodDefinition {
-    name_id: StringId,
+    str_id: StringId,
     uri_id: UriId,
     offset: Offset,
     comments: Vec<Comment>,
@@ -404,7 +404,7 @@ pub struct MethodDefinition {
 impl MethodDefinition {
     #[must_use]
     pub const fn new(
-        name_id: StringId,
+        str_id: StringId,
         uri_id: UriId,
         offset: Offset,
         comments: Vec<Comment>,
@@ -413,7 +413,7 @@ impl MethodDefinition {
         is_singleton: bool,
     ) -> Self {
         Self {
-            name_id,
+            str_id,
             uri_id,
             offset,
             comments,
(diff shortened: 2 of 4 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2176,20 +2278,23 @@ mod tests {
...
             "
         });
 
         assert_constant_references_eq!(
             &context,
-            vec!["C1", "C2", "C3", "C4", "C5", "C6", "C7", "C8", "C9", "C10", "C11"]
+            vec![
+                "C1", "C2", "C3", "C4", "C5", "C6", "C7", "C8", "C9", "C10", "C11", "C12", "C13", "C14", "C15", "C16",
+                "C17", "C18"
+            ]
```
Morriar: This is going to be a problem.

Deadcode detection relies on finding constant references to the same name than definitions. So for `class Foo; end`, the class is seen as alive if we have a constant reference to `Foo`.

With this change, there is no easy way to distinguish a constant reference in the name of a definition vs. anywhere else and all classes would always be alive.

rust/saturn/src/model/definitions.rs
Morriar: It also looks like we're conflating "constant reference" with "constant definition", yet actual `ConstantDefinition` names are not made into an `UnresolvedConstantReference`?

Aren't those the same than the name of a class or a module?

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Maybe one way to avoid this would be to not add there constant references to `graph.constant_references`?

rust/saturn/src/model/name.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// ```ruby
+    /// Foo::Bar::Baz
+    /// # ^ parent scope of Bar::Baz
+    /// #      ^ parent scope of Baz
+    /// ```
+    ///
+    /// `None` indicates that this is a simple constant read or a top level reference
+    parent_scope: Option<NameId>,
+    /// The nesting where we found the constant reference. This is a chain of nesting objects representing the lexical
+    /// scopes surrounding the reference
```
Morriar: I think this comment needs to be updated

rust/saturn/src/model/definitions.rs
```diff
@@ -327,7 +327,7 @@ impl ModuleDefinition {
 /// ```
 #[derive(Debug)]
 pub struct ConstantDefinition {
-    name_id: StringId,
+    str_id: StringId,
```
Morriar, line 330: Why isn't this a `NameId`?

rust/saturn/src/model/definitions.rs
```diff
@@ -127,13 +122,13 @@ impl Definition {
 /// class
 #[derive(Debug)]
 pub enum Mixin {
-    Include(ReferenceId),
-    Prepend(ReferenceId),
+    Include(NameId),
+    Prepend(NameId),
```
Morriar, line 126: Should these be `NameRef`?

rust/saturn/src/model/definitions.rs
```diff
@@ -169,7 +164,7 @@ impl ClassDefinition {
         offset: Offset,
         comments: Vec<Comment>,
         owner_id: Option<DefinitionId>,
-        superclass_ref: Option<ReferenceId>,
+        superclass_ref: Option<NameId>,
```
Morriar, line 167: Should this be a `NameRef`?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1054,6 +1054,23 @@ mod tests {
...
+                let name_string = $context.graph().strings().get(parent_name.str()).unwrap();
+
+                concatenated_name.insert_str(0, &format!("{}::", name_string));
+                name = parent_name;
+            }
+
+            concatenated_name
+        }};
+    }
+
```
Morriar, line 1073: I'm wondering if this shouldn't be a method on the `LocalGraph` and `Grap`?

rust/saturn/src/model/name.rs
vinistock: Updated.

rust/saturn/src/model/definitions.rs
vinistock: Too many changes for the same PR. We still need to handle constant paths correctly in constant definitions that aren't modules and classes, but that can come in a separate change.

rust/saturn/src/model/definitions.rs
vinistock: I don't believe so. We're putting name refs as a node in the graph and we only store IDs in the entities that need to refer to them.

If we were to store a `NameRef` here, it would make it more difficult to loop through all unresolved names because they would be spread around classes, modules, includes, prepends, constant defs and so on.

Our current structure also allows us to avoid resolving the same name twice because two identical name structures produce the same `NameId`, which would be more difficult to achieve if we stored a `NameRef` directly inside other structures.

We would probably end up in the scenario where we need to put them in an `Arc` so that equivalent names point to the same `NameRef` to avoid resolving the same name more than once.

We can revisit if we conclude that the IDs are not the right choice, but I believe this approach results in a graph that's easier to work with.

rust/saturn/src/model/definitions.rs
vinistock: Same as my other comment.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Yes, I think there are multiple places in the codebase where we are reaching inside the graph (either local or global graph) in a weird way.

I would explore either:

1. Adding methods to things that receive an immutable graph reference to do what's necessary
2. Adding graph methods that receive the things and do what we need

Something like this:

```rust
// Option 1
impl ClassDefinition {
  pub string_name(graph: &Graph) -> String {
    // uses the graph to return its name
  }
}

// Option 2
impl Graph {
  pub definition_name(definition: &Definition) -> String {
    // Returns the name
  }
}
```

Personally, I prefer option 1 so that methods are attached to the thing they are related to. Otherwise, the graph will grow to have a million methods for each thing we want to do with some other entity. I'll create an issue to explore this in more detail.

Morriar: The problem with option 1 is you can't have it with both `Graph` and `LocalGraph` or we need to bring in a trait

PR #321, opened 2025-11-13
Setup "EasyCompile" to allow packaging gem with precompiled binaries

Edouard-chin:
👋 Thanks for letting me try our tool in this project.

TL;DR This workflow allows with a single click on the GitHub UI to compile/package/test and publish (optionally) a gem with precompiled binaries.

-----------

This workflow was generated running the `easy_compile ci_template` command from the [EasyCompile tool](https://github.com/shopify-playground/edouard-playground) we are currently developing..

I confirmed that it works on this project https://github.com/Shopify/saturn/actions/runs/19342620213.

----------

### How does it work

When you'd like to cut a new release, you can head to the [GitHub action page](https://github.com/Shopify/saturn/actions), select the workflow and click on "Run workflow".

Once clicked here is what happened:

- We spin up machines on various platforms (MacOS and Linux are configured here). And start the compilation. Since the gem supports multiple ruby versions, we perform the compilation for each ruby minor versions supported (3.2, 3.3, 3.4).
- Once the compilation is done on all platforms, we create a matrix of jobs to test against all ruby versions that we compiled and all platforms.
- If all job passes we package the gem with the precompiled binaries.
- Finally, you have the option to publish the gems on RubyGems by ticking this box (for this to work we need to do a small setup on RubyGems.org and setup trusted publishing).
image

If you have any questions, please ask :D !

### DIFF
--- a/.github/workflows/easy-compile.yaml
+++ b/.github/workflows/easy-compile.yaml
@@ -1,0 +1,82 @@
+name: "Release gems with precompiled binaries"
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: "If the whole build passes on all platforms, release the gems on RubyGems.org"
+        required: false
+        type: boolean
+        default: false
+jobs:
+  compile:
+    timeout-minutes: 20
+    name: "Cross compile the gem on different ruby versions"
+    strategy:
+      matrix:
+        os: ["macos-latest", "shopify-ubuntu-20.04"]
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd"
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+        with:
+          ruby-version: "3.2.9"
+          bundler-cache: true
+      - name: "Run easy compile"
+        uses: "shopify-playground/edouard-playground/.github/actions/easy_compile@main"
+        with:
+          step: "compile"
+  test:
+    timeout-minutes: 20
+    name: "Run the test suite"
+    needs: compile
+    strategy:
+      matrix:
+        os: ["macos-latest", "shopify-ubuntu-20.04"]
+        rubies: ["3.2", "3.3", "3.4"]
+        type: ["cross", "native"]
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd"
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+        with:
+          ruby-version: "${{ matrix.rubies }}"
+          bundler-cache: true
+      - name: "Run easy compile"
+        uses: "shopify-playground/edouard-playground/.github/actions/easy_compile@main"
+        with:
+          step: "test_${{ matrix.type }}"
+  install:
+    timeout-minutes: 5
+    name: "Verify the gem can be installed"
+    needs: test
+    strategy:
+      matrix:
+        os: ["macos-latest", "shopify-ubuntu-20.04"]
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+      - name: "Run easy compile"
+        uses: "shopify-playground/edouard-playground/.github/actions/easy_compile@main"
+        with:
+          step: "install"
+  release:
+    permissions:
+      id-token: write
+      contents: read
+    timeout-minutes: 5
+    if: ${{ inputs.release }}
+    name: "Release all gems with RubyGems"
+    needs: install
+    runs-on: "shopify-ubuntu-latest"
+    steps:
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+      - name: "Run easy compile"
+        uses: "shopify-playground/edouard-playground/.github/actions/easy_compile@main"
+        with:
+          step: "release"
--- a/lib/saturn.rb
+++ b/lib/saturn.rb
@@ -4,7 +4,15 @@ require "bundler"
 require "uri"
 
 require "saturn/version"
-require "saturn/saturn"
+begin
+  # Load the precompiled version of the library
+  ruby_version = /(\d+\.\d+)/.match(RUBY_VERSION)
+  require "saturn/#{ruby_version}/saturn"
+rescue LoadError
+  # It's important to leave for users that can not or don't want to use the gem with precompiled binaries.
+  require "saturn/saturn"
+end
+
 require "saturn/location"
 require "saturn/comment"
 require "saturn/graph"


.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+        required: false
+        type: boolean
+        default: false
+jobs:
+  compile:
+    timeout-minutes: 20
+    name: "Cross compile the gem on different ruby versions"
+    strategy:
+      matrix:
+        os: ["macos-latest", "ubuntu-22.04"]
```
Morriar: Any plans to support Windows in the future? We'll need it at some point.

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v5"
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@v1"
```
Morriar: I think Shopify security policies force us to use specific commits rather than versions: https://github.com/Shopify/spoom/blob/main/.github/workflows/ruby.yml#L22-L23

See https://github.com/Shopify/spoom/pull/731 for example.

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          step: "compile"
+  test:
+    timeout-minutes: 20
+    name: "Run the test suite"
+    needs: compile
+    strategy:
+      matrix:
+        os: ["macos-latest", "ubuntu-22.04"]
+        rubies: ["3.4.7", "3.3.9", "3.2.9"]
```
Morriar: Nit, can we sort them in natural order?

```suggestion
        rubies: ["3.2.9", "3.3.9", "3.4.7"]
```

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          step: "install"
+  release:
+    permissions:
+      id-token: write
+      contents: read
+    timeout-minutes: 5
+    if: ${{ inputs.release }}
+    name: "Release all gems with RubyGems"
+    needs: install
```
Morriar, line 74: So the dependency is `release` -> `install` -> `test` -> `compile`. Does that mean we're going to run some steps 4 times like "Setup Ruby"?

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+      matrix:
+        os: ["macos-latest", "ubuntu-22.04"]
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v5"
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@v1"
+        with:
+          ruby-version: "3.2.9"
```
Morriar, line 24: Why is this one not using `3.4.7` like the others?

.github/workflows/easy-compile.yaml
vinistock: We can also get rid of the patch versions so that we're always using the latest patch.

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
+name: "Package gems with precompiled binaries"
```
vinistock: If this workflow publishes the gem to rubygems, we may want to call it publish or release.

In fact, it would be really nice to have this workflow be triggered when a release is manually created in the GitHub repo.

.github/workflows/easy-compile.yaml
Edouard-chin: Absolutely, it's already supported actually. I removed it from the matrix as it slows down the CI builds and I wasn't sure whether you'd be interested in precompiled binaries for windows. FWIW, we also support macos-intel .

.github/workflows/easy-compile.yaml
Edouard-chin: Yes that's correct. Each jobs runs in a different machine so we have to re-setup ruby. But GitHub action is heavily optimized for this, so is ruby/setup-ruby which only take a couple of seconds to run.

.github/workflows/easy-compile.yaml
Edouard-chin: I'll change the sorting and remove the patch versions. I'll make the change in the generator as well

.github/workflows/easy-compile.yaml
Edouard-chin: Great question, thanks for asking. The tool uses the minimum acceptable ruby version of the gem to run the compilation. The reason is purely for working around a toolchain issue on **Windows for Ruby 3.0 and 3.1**. (The issue affected all Ruby versions and a fix was applied to Ruby versions that are maintained https://bugs.ruby-lang.org/issues/21286)

----

Basically, we could use whatever ruby versions, but if a gem was to support Ruby 3.1 we'd have to compile a binary for it and since windows machine on GitHub comes with GCC 15.1, the compilation would fail (it's incompatible).
We could fix this the issue by downgrading gcc to 14.0, but by using the lowest possible accepted versions, we piggy back on top of `ruby/setup-ruby` which does that automatically for us.

https://github.com/ruby/setup-ruby/blob/d5126b9b3579e429dd52e51e68624dda2e05be25/windows-toolchain-versions.json#L155

.github/workflows/easy-compile.yaml
Edouard-chin: Regarding triggering this workflow when a release is published, I wonder if this is a default we want to add in the generator. The concern I have is that we are going to create a GitHub release but if something happens during the compilation we'd have a release without published gems (or at least until we fix the issues).

I feel like manually triggering the workflow, publishing the gems and then making a GitHub release would make a bit more sense. WDYT ?

vinistock: That's totally fair. In fact, we can automatically create the GitHub release from the workflow only after it succeeded.

We have a much more complicated version of this in the [Ruby LSP](https://github.com/Shopify/ruby-lsp/blob/abd88c8153f1aa4cd43e1e8bf75c80eb96281ec4/.github/workflows/release.yml#L80). For Saturn, we can just pass the argument for GitHub's auto-generated release notes and add a `.github/release.yml` file.

No need to be a part of this PR though! We can do it as a follow up.

Edouard-chin: Ah yeah creating a release from the workflow after the gem gets published make sense. I could see this being a sane default when we generate the workflow, although the description of the release may be a bit tricky to infer. I'll open an issue to keep this in mind. Thanks for the suggestion !

.github/workflows/easy-compile.yaml
Morriar: Yes we just had to disable it temporarily because of Prism: https://github.com/Shopify/saturn/pull/324

.github/workflows/easy-compile.yaml
Morriar: Makes sense, thanks for explaining

lib/saturn.rb
```diff
@@ -4,7 +4,14 @@
 require "uri"
 
 require "saturn/version"
-require "saturn/saturn"
+begin
+  ruby_version = /(\d+\.\d+)/.match(RUBY_VERSION)
+  require "saturn/#{ruby_version}/saturn"
```
Morriar, line 10:
```suggestion
  # Load the precompiled version of the library
  ruby_version = /(\d+\.\d+)/.match(RUBY_VERSION)
  require "saturn/#{ruby_version}/saturn"
```

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+    needs: test
+    strategy:
+      matrix:
+        os: ["macos-latest", "shopify-ubuntu-20.04"]
+    runs-on: "${{ matrix.os }}"
+    steps:
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+        with:
+          ruby-version: "3.4.7"
```
vinistock: Could we remove this so that `setup-ruby` picks up the `.ruby-version` file and we don't have to remember to bump this?

.github/workflows/easy-compile.yaml
```diff
@@ -0,0 +1,86 @@
...
+    timeout-minutes: 5
+    if: ${{ inputs.release }}
+    name: "Release all gems with RubyGems"
+    needs: install
+    runs-on: "shopify-ubuntu-latest"
+    steps:
+      - name: "Setup Ruby"
+        uses: "ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71"
+        with:
+          ruby-version: "3.4.7"
```
vinistock: Same here.

PR #322, opened 2025-11-13
Ensure constant reference resolution results are cached

vinistock:
This PR ensures that intermediate resolution results are cached in the graph (so that we don't try to resolve the same constant multiple times).

Additionally, I moved all of the algorithm to `resolution.rs`. Passing a reference to the graph will make it easier to add parallelism afterwards since the graph can't put itself inside an `Arc`.

### DIFF
--- a/rust/saturn/src/model/graph.rs
+++ b/rust/saturn/src/model/graph.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::Entry;
 use std::sync::LazyLock;
 
 use crate::indexing::local_graph::LocalGraph;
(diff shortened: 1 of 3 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -207,47 +208,27 @@ impl Graph {
...
+    pub fn record_resolved_name(&mut self, name_id: NameId, declaration_id: DeclarationId) {
+        match self.names.entry(name_id) {
+            Entry::Occupied(entry) => match entry.get() {
+                NameRef::Unresolved(_) => {
+                    if let NameRef::Unresolved(unresolved) = entry.remove() {
+                        let resolved_name = NameRef::Resolved(Box::new(ResolvedName::new(*unresolved, declaration_id)));
+                        self.names.insert(name_id, resolved_name);
+                    }
                 }
+                NameRef::Resolved(_) => {}
```
st0012: I think this case is worth recording through a counter. Can we add a comment to do that once we've settled on the implementation?

vinistock: Added a comment 👍

PR #324, opened 2025-11-13
Re-enable Windows CI

### LINKED ISSUE #68: Use cross-compilation to fix Windows build
Rust allows us to cross-compile the project in a Mac machine to the Windows GNU target. Let's use that to fix the Windows build and ensure compatibility from the start.

To cross-compile you'll need to:

1. Install the Windows GNU target with rustup
2. Install the mingw system libraries (with Homebrew or Nix)

vinistock:
Closes #68

It seems that https://github.com/ruby/prism/pull/3594 _did_ fix the Windows build 🎉. This PR re-enables it so that we don't accidentally add Windows incompatible code.

Most of the fixes are related to assuming forward slashes in file paths or not accounting for the carriage return character where it may happen.

### DIFF
--- a/lib/saturn/location.rb
+++ b/lib/saturn/location.rb
@@ -24,7 +24,10 @@ module Saturn
       uri = URI(@uri)
       raise Saturn::Error, "URI is not a file:// URI: #{@uri}" unless uri.scheme == "file"
 
-      uri.path
+      path = uri.path
+      # TODO: This has to go away once we have a proper URI abstraction
+      path.delete_prefix!("/") if Gem.win_platform?
+      path
     end
 
     #: (other: BasicObject) -> Integer
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -166,6 +166,8 @@ pub fn collect_file_paths(paths: Vec<String>) -> (Vec<String>, Vec<Errors>) {
 
 #[cfg(test)]
 mod tests {
+    use std::path::PathBuf;
+
     use super::*;
     use crate::test_utils::Context;
 
@@ -190,18 +192,21 @@ mod tests {
     #[test]
     fn collect_all_documents() {
         let context = Context::new();
-        context.touch("bar/baz.rb");
-        context.touch("bar/qux.rb");
-        context.touch("foo/bar.rb");
+        let baz = Path::new("bar").join("baz.rb");
+        let qux = Path::new("bar").join("qux.rb");
+        let bar = Path::new("foo").join("bar.rb");
+        context.touch(&baz);
+        context.touch(&qux);
+        context.touch(&bar);
 
         let (paths, errors) = collect_document_paths(&context, &["foo", "bar"]);
 
         assert_eq!(
             paths,
             vec![
-                "bar/baz.rb".to_string(),
-                "bar/qux.rb".to_string(),
-                "foo/bar.rb".to_string()
+                baz.to_str().unwrap().to_string(),
+                qux.to_str().unwrap().to_string(),
+                bar.to_str().unwrap().to_string()
             ]
         );
         assert!(errors.is_empty());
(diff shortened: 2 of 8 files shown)


lib/saturn/location.rb
```diff
@@ -24,7 +24,9 @@ def path
       uri = URI(@uri)
       raise Saturn::Error, "URI is not a file:// URI: #{@uri}" unless uri.scheme == "file"
 
-      uri.path
+      path = uri.path
+      path.delete_prefix!("/") if Gem.win_platform?
```
vinistock, line 29: This is terrible, but temporary until we move the Ruby LSP's custom URI implementation into Saturn.

The reason we need this is because Windows paths are prefixed with an artificial forward slash to turn them into valid URIs. For example:

```
C:/projects/foo.rb

# Simply prefixing this with the scheme file:// will result in an invalid URI
# You need an extra forward slash (which has to be accounted for when translating
# back into a file path)

file:///C:/projects/foo.rb
```

test/helpers/context.rb
```diff
@@ -26,6 +26,13 @@ def absolute_path_to(relative_path)
         ::File.join(@absolute_path, relative_path)
       end
 
+      #: (String) -> String
+      def uri_to(relative_path)
+        path = absolute_path_to(relative_path)
+        path.prepend("/") if Gem.win_platform?
+        URI::File.build(path: path).to_s
```
vinistock, line 34: Will also be replaced by our custom URI abstraction once we bring that over.

rust/saturn/src/indexing.rs
```diff
@@ -190,18 +192,21 @@ mod tests {
     #[test]
     fn collect_all_documents() {
         let context = Context::new();
-        context.touch("bar/baz.rb");
-        context.touch("bar/qux.rb");
-        context.touch("foo/bar.rb");
+        let baz = Path::new("bar").join("baz.rb");
+        let qux = Path::new("bar").join("qux.rb");
+        let bar = Path::new("foo").join("bar.rb");
```
Morriar, line 197: Can't we move this logic into `Context.touch`?

test/definition_test.rb
```diff
@@ -104,27 +104,42 @@ def test_definition_comments
         # Multi-line comment
         class Foo
           # Method comment
-          def foo; end
+          def bar; end
```
Morriar, line 107: Why do we need to change this?

test/definition_test.rb
```diff
@@ -104,27 +104,42 @@ def test_definition_comments
...
+      bar_comments = graph.documents.first.definitions.find { |d| d.name == "bar" }.comments
+
+      # On Windows, line endings are \r\n which count as 2 code units in UTF-8, so all ending positions need to add 1
+      if Gem.win_platform?
+        assert_equal(
+          [
+            "# This is a class comment (#{context.absolute_path_to("file1.rb")}:1:1-1:27)",
+            "# Multi-line comment (#{context.absolute_path_to("file1.rb")}:2:1-2:22)",
+          ],
+          foo_comments.map { |c| "#{c.string} (#{c.location})" },
```
Morriar: I wonder if we could move this into a macro:

```
assert_comments_for!(graph, "bar")
```

that would also adjust the location if it's windows so we don't have to duplicate the asserts

test/definition_test.rb
vinistock: I wanted to extract the comments into a variable to make the test a bit more readable, but then both variables would be called `foo_comments` because both definitions are called `foo`.

Unless we're testing something related to indexing getting potentially confused due to similar names for different definitions, I'd vote to always use a different name to avoid any confusion.

test/definition_test.rb
vinistock: I created a private method to make the adjustment.

rust/saturn/src/indexing.rs
vinistock: I tried doing it in [this commit](https://github.com/Shopify/saturn/commit/5819c8036c03a66bae012053e5adc08a5a39ffd5), but it honestly makes the tests odd.

When these are running on Windows we need to assert that the result is using backslashes, which we can easily get if we simply do `path.to_str()`.

If we try to continue writing tests with string literals that use forward slashes, then we need to somehow account for the fact that the result will contain forward slashes in Unix and backslashes on Windows - but then at that point, I feel like just dealing with paths is more straight forward.

test/definition_test.rb
st0012: I think in this particular case we can also use `class_comments` and `method_comments` without changing the test itself. But the current version is fine to me too.

PR #325, opened 2025-11-14
Use line-index crate for offset to position conversions

vinistock:
In #126, I made a bit of a faux pas. I totally forgot to call attention to the fact that Rust Analyzer already publishes a number of crates for LSP related things (such as translating offsets into line/col information) and that we should check what's available.

Indeed, they do publish the `line-index` crate, which does everything we need to translate locations including all of the different encodings.

I propose that we switch to using the crate so that we can avoid having to maintain this translation. The crate supports translating in both directions and it supports handling incremental edits as well.

In the future, we can make each `Document` instance hold their own `LineIndex` instance and we'll be able to easily translate positions in both directions - regardless of whether the document has been committed to disk or not.


rust/saturn-sys/src/location_api.rs
```diff
@@ -38,21 +37,18 @@ fn file_path_from_uri(uri: &str) -> Option<String> {
...
-
     let source = fs::read_to_string(&path_str).unwrap_or_else(|_| panic!("Failed to read file at path {path_str}"));
-    let converter = UTF8SourceLocationConverter::new(&source);
-    let offset = Offset::new(start, end);
-    let (start_line, start_column, end_line, end_column) = match converter.offset_to_position(&offset) {
-        Some((start_pos, end_pos)) => (start_pos.line(), start_pos.column(), end_pos.line(), end_pos.column()),
-        None => (0, 0, 0, 0),
-    };
+
+    let line_index = LineIndex::new(&source);
```
vinistock, line 42: We will need to think about how to structure this API in the future, but for now I'm creating a new line index every time.

rust/saturn/src/test_utils/local_graph_test.rs
```diff
@@ -89,33 +90,23 @@ impl LocalGraphTest {
...
-            None => String::from("INVALID_OFFSET"),
-        }
+        let line_index = LineIndex::new(&self.source);
+        let start = line_index.line_col(offset.start().into());
+        let end = line_index.line_col(offset.end().into());
+        format!("{}:{}-{}:{}", start.line + 1, start.col + 1, end.line + 1, end.col + 1)
     }
 
-    fn parse_location_positions(location: &str) -> (String, Position, Position) {
+    fn parse_location_positions(location: &str) -> (String, LineCol, LineCol) {
```
Morriar, line 109: Could we keep our `Position` abstraction on top of `LineCol` so we can add the features we need to `Position` and also keep the same naming through the Rust and Ruby APIs?

vinistock: Do we currently have any custom functionality we need on top? Or will a type alias do for now?

Morriar: I think a type alias may suffice 👍

PR #328, opened 2025-11-20
Alias `LineCol` to `Position`

vinistock:
This PR just aliases `LineCol` to `Position`.


rust/saturn/src/position.rs
```diff
@@ -0,0 +1,6 @@
+use line_index::LineCol;
```
Morriar, line 1: Let's move this out of `model`

vinistock: And move it where?

Morriar: I think it can leave at the top like `offset`?

PR #329, opened 2025-11-20
[DOC] Document singleton class' behaviour

st0012:


docs/ruby-behaviors.md
```diff
@@ -568,3 +569,79 @@ end
 ```
 
 Constant paths are resolved left-to-right: `Object` is resolved first, then `String` is looked up within `Object`.
+
+## Singleton Classes
+
+Ruby can create a *singleton class* for most objects, but it does so eagerly only for
```
vinistock: Does Ruby create it eagerly? Or is this describing the static analysis process?

docs/ruby-behaviors.md
```diff
@@ -568,3 +569,79 @@ end
...
+- Note that a singleton class's own singleton class is *not* eagerly created.
+
+The singleton class is the receiver for `def self.*`, `class << self`, and it is where modules mixed in via `extend`
+actually insert their methods.
+
+```ruby
+class Foo # defining Foo creates Foo's singleton class immediately
+  class << self # reopens Foo's singleton class
+    A = 1
+    def bar; A end
```
```suggestion
    def bar; A; end
```

docs/ruby-behaviors.md
```diff
@@ -568,3 +569,79 @@ end
...
 Constant paths are resolved left-to-right: `Object` is resolved first, then `String` is looked up within `Object`.
+
+## Singleton Classes
+
+Ruby can create a *singleton class* for most objects, but it does so eagerly only for
+classes and modules.
+
+- When a class `Foo` is defined, Ruby immediately creates `Foo`'s singleton class.
+- Other objects get one lazily (if they can have one at all) once you define singleton behavior (e.g. call `singleton_class` method on it).
+- Immediate values such as integers, symbols, `true`, `false`, and `nil` cannot have singleton classes.
```
vinistock: Only these special built-ins can have no singleton class right? Maybe it's worth mentioning that since any user created class or module will always have a singleton.

docs/ruby-behaviors.md
```diff
@@ -568,3 +569,89 @@ end
...
+
+## Singleton Classes
+
+Ruby can create a *singleton class* for most objects, but it only does so lazily for user-defined objects.
+
+- Defining singleton behavior (`def self.*`, `class << Foo`, `extend`) forces Ruby to create the singleton class if it was not created already.
+- Calling `Foo.singleton_class` also forces creation and returns the singleton class object.
+- If a class or module never uses singleton-specific syntax, Ruby may defer allocating its singleton class indefinitely.
+- Regular (non class/module) objects also get singleton classes lazily the first time singleton behavior is defined for them (e.g. when `obj.singleton_class` is called)
+- Immediate values such as integers, fixnums, symbols, `true`, `false`, and `nil` cannot have singleton classes; these are the only objects for which Ruby will never allocate one.
```
Morriar: I'm confused about this statement:

```rb
true.singleton_class.define_method(:foo) do
  puts "foo"
end

true.foo # puts "foo"
```

st0012: Ah I must misremembered the list 😕 
After checking the source code, the list should be:
- Fixnum
- Bignum
- Float
- Symbol
- "Frozen" String

I'll update. Thanks for verifying

Issue #200, opened 2025-10-09
Design resolution phase


vinistock:
We need to design a performant way to resolve constants. One of the challenges of the resolution phase is that resolving a constant and linearizing ancestors are intertwined.

Here's an example I find illustrative:

```ruby
module Bar
end

module Foo
  CONST = Bar
end

class Something
  # including `Foo` brings `CONST` into the class through inheritance
  include Foo
  # we can now include it
  include CONST
end
```

Ideally, we can perform resolution in parallel to speed things up, but we have these interesting scenarios where resolving a constant can depend on the partial ancestor linearization of a namespace.

Let's design a resolution phase that is both correct and high performance.

PR #331, opened 2025-11-21
Compute all global information during resolution

vinistock:
Step towards #200

This PR starts creating all declarations with their memberships during the resolution phase.

The idea is to sort all constant definitions and references based on the complexity of their name, so that processing units in the beginning of the list unlock work that is further ahead and so on.

I tried to document the code, but here are a few explanations.

### Name complexity

Each name part that appears in the nesting counts as 1 to compute the complexity used to order the list. For example:

```ruby
module Foo # 1
  module Bar # 2
    module Baz # 3
    end
  end

  module Qux::Boop # 3
    module Zip # 4
    end
  end
end
```

I named this "complexity score", but please suggest other names if you have ideas. This is simply a score to sort constants in the most likely order of reducing resolution failures.

### Handling parent scopes in definitions

When we don't have a parent scope in a module/class/constant definition (i.e.: a simple name), there's no need to run a resolution algorithm. We can just take the name as is to create the declaration concept.

Thanks to the ordering based on complexity, we should also guarantee that we will never be in a situation where we find a constant with a parent scope or nesting that hasn't yet been resolved - at least until ancestors are added in the mix.

For constants with parent scopes, we first always resolve the reference to figure out what it points to before naming the definition.

### DIFF
--- a/rust/saturn/src/main.rs
+++ b/rust/saturn/src/main.rs
@@ -50,7 +50,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     time_it!(indexing, { indexing::index_in_parallel(&mut graph, file_paths) })?;
 
-    time_it!(resolving, {
+    time_it!(resolution, {
         resolution::resolve_all(&mut graph);
     });
 
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -128,10 +128,9 @@ mod tests {
         let mut decl = Declaration::new("MyDecl".to_string(), DeclarationId::from("Object"));
         let def_id = DefinitionId::new(123);
 
+        // The second call will panic because we're adding the same exact ID twice
         decl.add_definition(def_id);
         decl.add_definition(def_id);
-
-        assert_eq!(decl.definitions().len(), 1);
     }
 
     #[test]
--- a/rust/saturn/src/model/graph.rs
+++ b/rust/saturn/src/model/graph.rs
@@ -10,7 +10,6 @@ use crate::model::ids::{DeclarationId, DefinitionId, NameId, ReferenceId, String
 use crate::model::name::{NameRef, ResolvedName};
 // use crate::model::integrity::IntegrityChecker;
 use crate::model::references::{ConstantReference, MethodRef};
-use crate::resolution::Unit;
 use crate::stats;
 
 pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Object"));
@@ -41,17 +40,8 @@ pub struct Graph {
 impl Graph {
     #[must_use]
     pub fn new() -> Self {
-        // Insert the magic top level self <main> object into the graph, so that we can associate global variables or
-        // definitions made at the top level with it
-        // let main_id = DeclarationId::from("<main>");
-        // declarations.insert(main_id, Declaration::new(String::from("<main>"), main_id));
-
-        // Insert `Object` head of time into the graph
-        let mut declarations = IdentityHashMap::default();
-        declarations.insert(*OBJECT_ID, Declaration::new(String::from("Object"), *OBJECT_ID));
-
         Self {
-            declarations,
+            declarations: IdentityHashMap::default(),
             definitions: IdentityHashMap::default(),
             definitions_to_declarations: IdentityHashMap::default(),
             documents: IdentityHashMap::default(),
@@ -133,106 +123,6 @@ impl Graph {
         };
 
         *id
-    }
-
-    /// Returns a tuple of r vectors:
-    /// - The first one contains all constants, sorted in order for resolution (less complex constant names first)
-    /// - The second one contains all other definitions, in no particular order
-    ///
-    /// # Panics
-    ///
-    /// This will panic if there's a definition with a name ID that doesn't exist
-    #[must_use]
-    pub fn sorted_resolution_units(&self) -> (Vec<Unit>, Vec<DefinitionId>) {
-        let estimated_length = self.definitions.len() / 2;
-        let mut constants = Vec::with_capacity(estimated_length);
-        let mut others = Vec::with_capacity(estimated_length);
-
-        for (id, definition) in &self.definitions {
-            match definition {
-                Definition::Class(def) => {
-                    constants.push((Unit::Definition(*id), self.names.get(def.name_id()).unwrap()));
-                }
-                Definition::Module(def) => {
-                    constants.push((Unit::Definition(*id), self.names.get(def.name_id()).unwrap()));
-                }
-                Definition::Constant(_) => {
-                    // TODO: This should go in the constants vector, but we didn't change constant definitions to hold a
-                    // name id yet. Fix the `resolving_nested_reference` test in resolution.rs too!!!
-                    others.push(*id);
-                }
-                _ => {
-                    others.push(*id);
-                }
-            }
-        }
-
-        for (id, constant_ref) in &self.constant_references {
-            constants.push((Unit::Reference(*id), self.names.get(constant_ref.name_id()).unwrap()));
-        }
-
-        // Sort namespaces based on their name complexity so that simpler names are always first
-        constants.sort_by(|(_, name_a), (_, name_b)| {
-            let complexity_a = self.name_complexity(name_a);
-            let complexity_b = self.name_complexity(name_b);
-            complexity_a.cmp(&complexity_b)
-        });
-
-        others.shrink_to_fit();
-        (constants.into_iter().map(|(id, _)| id).collect(), others)
-    }
-
-    /// Returns a complexity score for a given name, which is used to sort names for resolution. The complexity is based
-    /// on how many parent scopes are involved in a name's nesting. This is because simple names are always
-    /// straightforward to resolve no matter how deep the nesting is. For example:
-    ///
-    /// ```ruby
-    /// module Foo
-    ///   module Bar
-    ///     class Baz; end
-    ///   end
-    /// end
-    /// ```
-    ///
-    /// These are all simple names because they don't require resolution logic to determine the final name of each step.
-    /// We only have to ensure that they are ordered by nesting level. Names with parent scopes require that their parts
-    /// be resolved to determine what they refer to and so they must be sorted last.
-    ///
-    /// ```ruby
-    /// module Foo
-    ///   module Bar::Baz
-    ///     class Qux; end
-    ///  end
-    /// end
-    /// ```
-    ///
-    /// In this case, we need `Bar` to have already been processed so that we can resolve the `Bar` reference inside of
-    /// the `Foo` nesting, which then unblocks the resolution of `Baz` and finally `Qux`. Notice how `Qux` is a simple
-    /// name, but it's nested under a complex name so we have to sort it last. This is why we consider the number of
-    /// parent scopes in the entire nesting, not just for the name itself
-    #[must_use]
-    fn name_complexity(&self, name: &NameRef) -> u32 {
-        let mut current_name = name;
-        let mut complexity = if current_name.parent_scope().is_some() { 2 } else { 1 };
-
-        while let Some(nesting) = current_name.nesting() {
-            if let Some(parent_id) = current_name.parent_scope() {
-                let mut parent = self.names.get(parent_id).unwrap();
-                let mut current_complexity = if current_name.parent_scope().is_some() { 2 } else { 1 };
-
-                while let Some(parent_scope) = parent.parent_scope() {
-                    current_complexity += 1;
-                    parent = self.names.get(parent_scope).unwrap();
-                }
-
-                complexity += current_complexity;
-            }
-
-            complexity += 1;
-            current_name = self.names.get(nesting).unwrap();
-        }
-
-        complexity
     }
 
     // Returns an immutable reference to the strings map
@@ -768,40 +658,5 @@ mod tests {
 
         let foo = context.graph.declarations.get(&DeclarationId::from("Foo")).unwrap();
         assert!(!foo.members().contains_key(&StringId::from("Bar")));
-    }
-
-    #[test]
-    fn name_complexity() {
-        let mut context = GraphTest::new();
-        context.index_uri("file:///foo.rb", {
-            r"
-            module Foo
-              module Bar
-                module Baz
-                end
-              end
-
-              module Qux::Zip
-                module Zap
-                  class Zop::Boop
-                  end
-                end
-              end
-            end
-            "
-        });
-
-        let mut names = context.graph.names().values().collect::<Vec<_>>();
-        assert_eq!(8, names.len());
-
-        names.sort_by_key(|a| context.graph.name_complexity(a));
-
-        assert_eq!(
-            vec!["Foo", "Qux", "Bar", "Baz", "Zip", "Zap", "Zop", "Boop"],
-            names
-                .iter()
-                .map(|n| context.graph.strings.get(n.str()).unwrap().as_str())
-                .collect::<Vec<_>>()
-        );
     }
 }
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -1,7 +1,10 @@
+use std::collections::VecDeque;
+
 use crate::model::{
     declaration::Declaration,
     definitions::Definition,
     graph::{Graph, OBJECT_ID},
+    identity_maps::IdentityHashMap,
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
     name::{Name, NameRef},
 };
@@ -9,6 +12,24 @@ use crate::model::{
 pub enum Unit {
     Definition(DefinitionId),
     Reference(ReferenceId),
+}
+
+enum Outcome {
+    /// The constant was successfully resolved to the given declaration ID
+    Resolved(DeclarationId),
+    /// We had everything we needed to resolved this constant, but we couldn't find it. This means it's not defined (or
+    /// defined in a way that static analysis won't discover it)
+    Unresolved,
+    /// We couldn't resolve this constant right now because certain dependencies were missing. For example, a constant
+    /// reference involved in computing ancestors (like an include) was found, but wasn't resolved yet. We need to place
+    /// this back in the queue to retry once we have progressed further
+    Retry,
+}
+
+impl Outcome {
+    fn is_resolved_or_retry(&self) -> bool {
+        matches!(self, Outcome::Resolved(_) | Outcome::Retry)
+    }
 }
 
 /// Runs the resolution phase on the graph. The resolution phase is when 4 main pieces of information are computed:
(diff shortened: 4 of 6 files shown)


rust/saturn/src/model/graph.rs
```diff
@@ -657,4 +769,39 @@ mod tests {
...
     }
+
+    #[test]
+    fn name_complexity() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            module Foo
+              module Bar
+                module Baz
```
Morriar: Should we add something like this:

```rb
class Foo
  class ::Bar
    class Baz; end
  end
end
```

rust/saturn/src/model/graph.rs
```diff
@@ -127,6 +138,106 @@ impl Graph {
...
+                complexity += current_complexity;
+            }
+
+            complexity += 1;
+            current_name = self.names.get(nesting).unwrap();
+        }
+
+        complexity
+    }
+
```
Morriar: Can't we simplify it a bit with recursion?

Something like this:

```rb
fn score(graph: &Graph, name_id: NameId) -> usize {
    let name = graph.names().get(&name_id).unwrap();
    let parent_depth = name.parent_scope().map_or(0, |id| score(graph, id) + 1);
    let nesting_depth = name.nesting().map_or(0, |id| score(graph, id) + 1);
    parent_depth + nesting_depth
}
```

rust/saturn/src/resolution.rs
```diff
@@ -269,12 +395,101 @@ mod tests {
...
+        context.resolve();
+
+        let foo = context.graph.declarations().get(&DeclarationId::from("Foo")).unwrap();
+        assert_members(foo, &[]);
+        assert_owner(foo, "Object");
+
+        let bar = context.graph.declarations().get(&DeclarationId::from("Bar")).unwrap();
+        assert_members(bar, &["Baz"]);
+        assert_owner(bar, "Object");
+    }
```
Morriar, line 631: Here too I think we're missing an example with `::Baz`

rust/saturn/src/model/graph.rs
```diff
@@ -127,6 +138,106 @@ impl Graph {
...
 
+    /// Returns a tuple of r vectors:
+    /// - The first one contains all constants, sorted in order for resolution (less complex constant names first)
+    /// - The second one contains all other definitions, in no particular order
+    ///
+    /// # Panics
+    ///
+    /// This will panic if there's a definition with a name ID that doesn't exist
+    #[must_use]
+    pub fn sorted_resolution_units(&self) -> (Vec<Unit>, Vec<DefinitionId>) {
```
Morriar: Let's keep this in the `resolution` module?

rust/saturn/src/model/graph.rs
```diff
@@ -46,16 +47,17 @@ pub struct Graph {
...
-        // let mut declarations = IdentityHashMap::default();
         // Insert the magic top level self <main> object into the graph, so that we can associate global variables or
         // definitions made at the top level with it
         // let main_id = DeclarationId::from("<main>");
         // declarations.insert(main_id, Declaration::new(String::from("<main>"), main_id));
 
-        //declarations.insert(object_id, Declaration::new(String::from("Object"), object_id));
+        // Insert `Object` head of time into the graph
+        let mut declarations = IdentityHashMap::default();
+        declarations.insert(*OBJECT_ID, Declaration::new(String::from("Object"), *OBJECT_ID));
```
Morriar: Why do we need to add it both here and in the resolution pass?

rust/saturn/src/resolution.rs
```diff
@@ -1,28 +1,140 @@
...
+            Definition::Constant(it) => it.str_id(),
+            Definition::Method(it) => it.str_id(),
+            Definition::AttrAccessor(it) => it.str_id(),
+            Definition::AttrReader(it) => it.str_id(),
+            Definition::AttrWriter(it) => it.str_id(),
+            Definition::GlobalVariable(it) => it.str_id(),
+            Definition::InstanceVariable(it) => it.str_id(),
+            Definition::ClassVariable(it) => it.str_id(),
+            _ => panic!("Unexpected definition type. This shouldn't happen"),
+        };
```
Morriar, line 129: We may want to move this to `Definition` to make it easier to keep in sync when adding new kinds of definitions

rust/saturn/src/model/graph.rs
```diff
@@ -127,6 +138,106 @@ impl Graph {
...
+        for (id, constant_ref) in &self.constant_references {
+            constants.push((Unit::Reference(*id), self.names.get(constant_ref.name_id()).unwrap()));
+        }
+
+        // Sort namespaces based on their name complexity so that simpler names are always first
+        constants.sort_by(|(_, name_a), (_, name_b)| {
+            let complexity_a = self.name_complexity(name_a);
+            let complexity_b = self.name_complexity(name_b);
+            complexity_a.cmp(&complexity_b)
+        });
```
st0012: Let's make sure we also prefer definition over references when they have the same name complexity score.

rust/saturn/src/model/declaration.rs
```diff
@@ -91,28 +91,24 @@ impl Declaration {
         &self.members
     }
 
-    pub fn add_member(&mut self, name_id: StringId, declaration_id: DeclarationId) {
-        self.members.insert(name_id, declaration_id);
+    pub fn add_member(&mut self, str_id: StringId, declaration_id: DeclarationId) {
```
st0012: Maybe we extract this into another PR?

rust/saturn/src/model/graph.rs
vinistock: No, that was leftovers from iterations.

rust/saturn/src/model/graph.rs
vinistock: I reflected on this and it's already the case. If we have the same exact name structure, both the definition and the reference will share the same `NameId` and we always process definitions first—so there's no risk for trying to resolve a reference before its definition.

rust/saturn/src/model/graph.rs
```diff
@@ -124,6 +131,104 @@ impl Graph {
...
+    /// module Foo
+    ///   module Bar::Baz
+    ///     class Qux; end
+    ///  end
+    /// end
+    /// ```
+    ///
+    /// In this case, we need `Bar` to have already been processed so that we can resolve the `Bar` reference inside of
+    /// the `Foo` nesting, which then unblocks the resolution of `Baz` and finally `Qux`. Notice how `Qux` is a simple
+    /// name, but it's nested under a complex name so we have to sort it last. This is why we consider the number of
```
st0012: Thanks for the explanation. Can it be added to the architecture doc under the section for resolution? (Can be in a follow up PR)

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,237 @@
...
+    Definition(DefinitionId),
+    Reference(ReferenceId),
+}
+
+enum ResolutionResult {
+    /// The constant was successfully resolved to the given declaration ID
+    Resolved(DeclarationId),
+    /// We had everything we needed to resolved this constant, but we couldn't find it. This means it's not defined (or
+    /// defined in a way that static analysis won't discover it)
+    Unresolved,
```
st0012, line 22: Do we want to introduce some enums as unresolved reasons? It could provide interesting stats.

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,237 @@
...
-    // resolution is added
-    if name.nesting().is_some()
-        && let Some(id) = search_lexical_scopes(graph, name, str_id)
-    {
-        return Some(id);
+    if name.nesting().is_some() {
+        let scope_result = search_lexical_scopes(graph, name, str_id);
+
+        // If we already resolved or need to retry, return early
+        if !scope_result.is_unresolved() {
```
st0012: Should we change the function to `is_resolved` to avoid the double negation given this is the only usage of it?

rust/saturn/src/model/graph.rs
vinistock: Added 👍

rust/saturn/src/model/graph.rs
vinistock: We can indeed!

rust/saturn/src/resolution.rs
vinistock: It's a bit of a weird API to name because we want the `str_id` of any definition except for classes, modules or constants.

Instead of creating an API, I explicitly listed all possible cases so that the compiler will error if we add a new one.

rust/saturn/src/model/graph.rs
vinistock: Yes, totally 👍

rust/saturn/src/resolution.rs
vinistock: We'll have to explore if we are able to determine the reason. It would be interested even for diagnostics, but I'm not sure if we are able to understand why something didn't get resolved.

rust/saturn/src/resolution.rs
vinistock: I renamed it to `is_resolved_or_retry` since in both cases we want to return early.

rust/saturn/src/main.rs
```diff
@@ -51,7 +51,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     time_it!(indexing, { indexing::index_in_parallel(&mut graph, file_paths) })?;
 
     time_it!(resolving, {
-        resolve(&mut graph);
+        resolution::resolve_all(&mut graph);
```
Morriar, line 54: Should we also rename the timer `resolution`? Or rename the phase `resolving`.

rust/saturn/src/model/declaration.rs
```diff
@@ -124,43 +120,43 @@ impl Declaration {
...
+        let def_id = DefinitionId::new(123);
 
-    //     decl.add_definition(def_id);
-    //     decl.add_definition(def_id);
+        decl.add_definition(def_id);
+        decl.add_definition(def_id);
 
-    //     assert_eq!(decl.definitions().len(), 1);
-    // }
+        assert_eq!(decl.definitions().len(), 1);
```
Morriar: Wait, isn't this dead code? We expect the test to panic at line 132 right?

Morriar: Could we do something like this to make it clearer what's panicking?

```
    let result = panic::catch_unwind(|| {
        decl.add_definition(def_id);
    });

    assert!(result.is_err(), "Expected panic on duplicate definition ID");
```

rust/saturn/src/model/graph.rs
```diff
@@ -40,14 +40,11 @@ pub struct Graph {
 impl Graph {
     #[must_use]
     pub fn new() -> Self {
-        // let mut declarations = IdentityHashMap::default();
         // Insert the magic top level self <main> object into the graph, so that we can associate global variables or
         // definitions made at the top level with it
         // let main_id = DeclarationId::from("<main>");
         // declarations.insert(main_id, Declaration::new(String::from("<main>"), main_id));
```
Morriar: Is this commented code still relevant?

rust/saturn/src/model/name.rs
```diff
@@ -53,6 +56,52 @@ impl Name {
...
+            let name_ref = names.get(&id).unwrap();
+            name_ref.depth(names)
+        });
+        let nesting_depth = self.nesting().map_or(0, |id| {
+            let name_ref = names.get(&id).unwrap();
+            name_ref.depth(names)
+        });
+
+        parent_depth + nesting_depth + 1
+    }
```
Morriar: Should we move this to `resolution`?

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,239 @@
...
     name::{Name, NameRef},
 };
 
-/// Attempts to resolve all constant references stored in the graph
+pub enum Unit {
+    Definition(DefinitionId),
+    Reference(ReferenceId),
+}
+
+enum ResolutionResult {
```
Morriar: Should they be named `ResolutionUnit` and `ResolutionResult` or `Unit` and `Result`.

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,239 @@
...
+    graph.clear_declarations();
+    // Ensure that Object exists ahead of time so that we can associate top level declarations with the right membership
+    graph
+        .declarations_mut()
+        .insert(*OBJECT_ID, Declaration::new("Object".to_string(), *OBJECT_ID));
+
+    let (mut constant_queue, other_ids) = sorted_units(graph);
+    // Flag to indicate whether we successfully resolved something in the last resolution iteration. If nothing got
+    // resolved, then we stop because trying again would likely not successfully resolve anything new
+    let mut resolved_something = true;
```
Morriar: Should we invert the condition? It's surprising that we state we already resolved something when we actually didn't perform any work yet.

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,239 @@
...
+    // First, handle all constants, ensuring that we create declarations, members, owners and resolve any name
+    // references that are necessary along the way
+    while resolved_something && let Some(unit_id) = constant_queue.pop_front() {
+        resolved_something = match unit_id {
+            Unit::Definition(id) => {
+                let name_id = match graph.definitions().get(&id).unwrap() {
+                    Definition::Class(class) => *class.name_id(),
+                    Definition::Module(module) => *module.name_id(),
+                    Definition::Constant(constant) => *constant.name_id(),
+                    _ => panic!("Expected module or class definition"),
```
Morriar: or constant apparently

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,239 @@
...
+                };
+
+                match handle_constant_declaration(graph, name_id, id) {
+                    ResolutionResult::Retry => {
+                        // There might be dependencies we haven't figured out yet, so we need to retry
+                        constant_queue.push_back(unit_id);
+                        false
```
Morriar: Not sure if the linter will let you do but with a code that complex I'd find it clearer to do ` resolved_something = false` here.

rust/saturn/src/resolution.rs
```diff
@@ -1,77 +1,239 @@
...
+            Definition::Method(it) => it.str_id(),
+            Definition::AttrAccessor(it) => it.str_id(),
+            Definition::AttrReader(it) => it.str_id(),
+            Definition::AttrWriter(it) => it.str_id(),
+            Definition::GlobalVariable(it) => it.str_id(),
+            Definition::InstanceVariable(it) => it.str_id(),
+            Definition::ClassVariable(it) => it.str_id(),
+            Definition::Class(_) | Definition::Module(_) | Definition::Constant(_) => {
+                panic!("Unexpected definition type. This shouldn't happen")
+            }
```
Morriar, line 128: Would this work:

```suggestion
            Definition::Class(_) | Definition::Module(_) | Definition::Constant(_) => {
                panic!("Unexpected definition type. This shouldn't happen")
            }
            _ => it.str_id()
```

rust/saturn/src/main.rs
vinistock: Yes, done!

rust/saturn/src/model/declaration.rs
vinistock: That fails compilation saying that `Declaration` can't be passed through a `catch_unwind`. I think this test is short enough to keep the panic on top.

rust/saturn/src/model/graph.rs
vinistock: No, this is leftovers from #318. Removed it.

rust/saturn/src/model/name.rs
vinistock: I like the idea of keeping the logic in name and passing the graph, kinda of like having `Definition#string_name` which receives the graph to figure out the right string.

Not a strong opinion, but I do like this structure. @st0012 do you have opinions?

rust/saturn/src/resolution.rs
vinistock: `Result` is a bit problematic because that's a built-in, so we end up needing to remove ambiguity for the compiler. I renamed it to `Outcome`, which I think is short and clear.

rust/saturn/src/resolution.rs
vinistock: Instead of reverting the condition, I renamed the variable to `making_progress`, so that we can avoid doing `while !var` instead of `while var`.

rust/saturn/src/resolution.rs
vinistock: Ah, yeah #336 got shipped before this one and changed this.

rust/saturn/src/resolution.rs
vinistock: Changed it 👍

rust/saturn/src/resolution.rs
vinistock: It doesn't work because if you use a catch all arm, the type of the binding remains as the enumerable and not every variant implements `str_id` (i.e.: `str_id()` doesn't exist in `Definition`, it only exists in the variants that actually own a str_id).

rust/saturn/src/model/name.rs
st0012: Right now it's used as `name.depth(names)`, which I find a bit odd. If moving to `resolution` means we'd do something like `resolution.compute_name_depth(name, names)` instead, then I prefer that.

vinistock: Moved it to resolution 👍

rust/saturn/src/model/declaration.rs
Morriar: Okay, then let's just remove the dead code and add a comment on the line we expect to panic.

rust/saturn/src/resolution.rs
Morriar: That's a bummer :(

rust/saturn/src/resolution.rs
```diff
@@ -82,34 +246,120 @@ fn run_resolution(graph: &Graph, name: &Name) -> Option<DeclarationId> {
...
+
+    for (id, definition) in graph.definitions() {
+        match definition {
+            Definition::Class(def) => {
+                constants.push((Unit::Definition(*id), names.get(def.name_id()).unwrap()));
+            }
+            Definition::Module(def) => {
+                constants.push((Unit::Definition(*id), names.get(def.name_id()).unwrap()));
+            }
+            Definition::Constant(def) => {
```
Morriar, line 342: Would this work?

```suggestion
            Definition::Class(def) |
            Definition::Module(def) |
            Definition::Constant(def) => {
```

rust/saturn/src/resolution.rs
```diff
@@ -82,34 +246,120 @@ fn run_resolution(graph: &Graph, name: &Name) -> Option<DeclarationId> {
...
+}
+
+/// Returns a tuple of 2 vectors:
+/// - The first one contains all constants, sorted in order for resolution (less complex constant names first)
+/// - The second one contains all other definitions, in no particular order
+#[must_use]
+fn sorted_units(graph: &Graph) -> (VecDeque<Unit>, Vec<DefinitionId>) {
+    let estimated_length = graph.definitions().len() / 2;
+    let mut constants = Vec::with_capacity(estimated_length);
+    let mut others = Vec::with_capacity(estimated_length);
```
Morriar, line 331: Is this a meaningful optimization at this point?

Also, last time I checked we had around half as much constant than scopes:

| Kind | Count |
| ---- | ------- |
| classes | 703,376 |
| modules | 504,914 |
| singleton_classes | 147,476 |
| constants | 587,248 |

rust/saturn/src/resolution.rs
vinistock: I thought the same when writing these enums, but unfortunately it doesn't work. Each one of these arms binds `def` to a different type (`ClassDefinition`, `ModuleDefinition` or `ConstantDefinition`).

The compiler doesn't inspect the implementation of the arm to understand that we're only accessing a method that exists in all 3, it simply fails immediately saying that we can't bind `def` to multiple distinct types.

rust/saturn/src/resolution.rs
vinistock: I didn't go super deep into estimating the numbers, but notice that `constants` here holds IDs for all classes, modules and constant references.

We would need to double check the ratio against method defs + instance vars + global vars and so on. It'd probably expect it to be more like 1/3 constants and 2/3 other things.

However, I do think it's better than just starting with `Vec::new`. We know for a fact that we need at least a certain ratio of `definitions` in terms of space, so we're probably saving a bit of the reallocation operations.

I created https://github.com/Shopify/saturn/issues/349 to revisit.

rust/saturn/src/model/declaration.rs
vinistock: Done!

PR #333, opened 2025-11-24
Support indexing singleton class

st0012:
This PR adds singleton class definition indexing for `class << X` expressions:

```rb
class Foo
  def self.bar; end       # method with `receiver` `Foo` (no singleton def created)
  @class_ivar = 1         # `ClassInstanceVariableDefinition` owned by `Foo`

  class << self           # creates `SingletonClassDefinition` `<Foo>`
    def baz; end          # method owned by `<Foo>`

    class << self         # creates `SingletonClassDefinition` `<<Foo>>`
      def qux; end        # method owned by `<<Foo>>`
    end
  end
end

class << Foo              # creates `SingletonClassDefinition` `<Foo>`
  def quux; end           # method owned by `<Foo>`
end

class << some_var         # dynamic expression - no definition created
  def dynamic; end
end
```

Key indexing behaviors:
- Only `class << self` or `class << Constant` creates a `SingletonClassDefinition`
- `def self.method` stores a `receiver` field pointing to the class's `NameId`, without creating a singleton definition
- Instance variables in class/module body (`@ivar`) become `ClassInstanceVariableDefinition`
- Instance variables in method bodies inside `class << self` also become `ClassInstanceVariableDefinition`
- Class variables (`@@var`) bypass singleton blocks. They always belong to the base class/module
- `extend SomeModule` is stored as `Mixin::Extend` on the class/module

Stats from Core:

```
Definition breakdown:
  Method                     321829
  Module                     178383
  Class                      102105
  ClassInstanceVariable       95619
  InstanceVariable            86623
  Constant                    61706
  AttrReader                  31875
  SingletonClass              13557
  AttrAccessor                 5886
  AttrWriter                    384
  GlobalVariable                 30
  ClassVariable                  11

Singleton class definitions (class << X):
    On classes:            11413 (84.2%)
    On modules:             2141 (15.8%)
    Nested:                    3 (0.0%)
  With members:            13526 (99.8%)
```

### DIFF
--- a/rust/saturn-sys/src/definition_api.rs
+++ b/rust/saturn-sys/src/definition_api.rs
@@ -13,20 +13,23 @@ use std::ptr;
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum DefinitionKind {
     Class = 0,
-    Module = 1,
-    Constant = 2,
-    Method = 3,
-    AttrAccessor = 4,
-    AttrReader = 5,
-    AttrWriter = 6,
-    GlobalVariable = 7,
-    InstanceVariable = 8,
-    ClassVariable = 9,
+    SingletonClass = 1,
+    Module = 2,
+    Constant = 3,
+    Method = 4,
+    AttrAccessor = 5,
+    AttrReader = 6,
+    AttrWriter = 7,
+    GlobalVariable = 8,
+    InstanceVariable = 9,
+    ClassInstanceVariable = 10,
+    ClassVariable = 11,
 }
 
 pub(crate) fn map_definition_to_kind(defn: &Definition) -> DefinitionKind {
     match defn {
         Definition::Class(_) => DefinitionKind::Class,
+        Definition::SingletonClass(_) => DefinitionKind::SingletonClass,
         Definition::Module(_) => DefinitionKind::Module,
         Definition::Constant(_) => DefinitionKind::Constant,
         Definition::Method(_) => DefinitionKind::Method,
@@ -35,6 +38,7 @@ pub(crate) fn map_definition_to_kind(defn: &Definition) -> DefinitionKind {
         Definition::AttrWriter(_) => DefinitionKind::AttrWriter,
         Definition::GlobalVariable(_) => DefinitionKind::GlobalVariable,
         Definition::InstanceVariable(_) => DefinitionKind::InstanceVariable,
+        Definition::ClassInstanceVariable(_) => DefinitionKind::ClassInstanceVariable,
         Definition::ClassVariable(_) => DefinitionKind::ClassVariable,
     }
 }
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -5,9 +5,10 @@ use crate::errors::Errors;
 use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
-    AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
-    ConstantDefinition, Definition, GlobalVariableDefinition, InstanceVariableDefinition, MethodDefinition, Mixin,
-    ModuleDefinition, Parameter, ParameterStruct,
+    AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition,
+    ClassInstanceVariableDefinition, ClassVariableDefinition, ConstantDefinition, Definition, GlobalVariableDefinition,
+    InstanceVariableDefinition, MethodDefinition, Mixin, ModuleDefinition, Parameter, ParameterStruct,
+    SingletonClassDefinition,
 };
 use crate::model::document::Document;
 use crate::model::ids::{DefinitionId, NameId, StringId, UriId};
@@ -18,6 +19,13 @@ use crate::offset::Offset;
 use ruby_prism::{ParseResult, Visit};
 
 pub type IndexerParts = (LocalGraph, Vec<Errors>);
+
+#[derive(Clone, Copy)]
+enum MixinType {
+    Include,
+    Prepend,
+    Extend,
+}
 
 /// The indexer for the definitions found in the Ruby source code.
 ///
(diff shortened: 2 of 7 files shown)


rust/saturn-sys/src/definition_api.rs
```diff
@@ -13,20 +13,22 @@ use std::ptr;
...
-    Module = 1,
-    Constant = 2,
-    Method = 3,
-    AttrAccessor = 4,
-    AttrReader = 5,
-    AttrWriter = 6,
-    GlobalVariable = 7,
-    InstanceVariable = 8,
-    ClassVariable = 9,
+    SingletonClass = 1,
```
vinistock, line 16: I think this needs an equivalent change in the C side of the API which switches on all of these kinds.

rust/saturn/src/model/definitions.rs
```diff
@@ -156,6 +160,110 @@ pub struct ClassDefinition {
...
 
+/// A singleton class definition (eigenclass).
+/// Singleton classes are created lazily when singleton-specific syntax is encountered:
+/// - `def self.foo` or `def ClassName.foo`
+/// - `class << self` or `class << ClassName`
+/// - `extend Module`
+/// - Class instance variables (`@foo` in class body)
+#[derive(Debug)]
+pub struct SingletonClassDefinition {
+    name_id: NameId,
```
vinistock, line 186: I wonder if you need to store both `name_id` and `owner`. Since a singleton class is also connected to its attached class, do we need both concepts?

Can a singleton ever be owned by something that isn't the attached class?

I suspect we can get away with just `owner_id`, so that the name is always derived from the attached class.

st0012: I think we need `name_id`, `attached_target`, and `owner`.

In this PR, we'll also introduce the concept of receiver to methods, which should be a `NameId`:

```rb
class Foo
  def self.bar; end # `Foo` as the receiver
  def Bar.baz; end  # `Bar` as the receiver
end

class Foo
  class << self
    # name_id: `NameId("Foo::<Foo>")`
    # attached_target: `NameId("Foo")`
    # owner: `Definition("Foo")`

    class << self
      # name_id: `NameId("Foo::<Foo>::<<Foo>>")`
      # attached_target: `NameId("Foo::<Foo>")`
      # The first singleton needs to have `name_id` so we have something to point to here

      # Here, we also need a name_id to point to
      # receiver: `NameId("Foo::<Foo>::<<Foo>>")`
      def self.bar; end
    end

    class << Bar
      # And this shows that we can't rely on `owner` for this
      # name_id: `NameId("Bar::<Bar>")`
      # attached_target: `NameId("Bar")`
      # owner: `Definition("<Foo>")`
    end
  end
end
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -532,8 +662,10 @@ impl Visit<'_> for RubyIndexer<'_> {
 
             let definition_id = self.local_graph.add_definition(definition);
 
-            if let Some(parent_nesting) = self.parent_nesting() {
-                parent_nesting.add_member(definition_id);
+            if let Some(owner_id) = owner_id
```
vinistock: Why are we making this change?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -562,8 +694,10 @@ impl Visit<'_> for RubyIndexer<'_> {
...
```
vinistock: Same question here, it's not clear why this needs to change.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -576,6 +710,83 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+            // `class << Foo` or `class << Foo::Bar` - use the constant's NameId
+            self.index_constant_reference(&expression, true)
+        } else {
+            // Dynamic expression (e.g., `class << some_var`) - skip creating definition
+            self.visit(&expression);
+            None
+        };
+
+        // Create SingletonClassDefinition if we have an attached_target
+        let block_id = if let Some(attached_target) = attached_target {
```
vinistock, line 765: Nit: I think if we use a match we could avoid checking here again. The code would be more nested, but personally I'd find it a bit more readable. Something like:

```rust

match node.expression() {
  Some(ruby_prism::Node::SelfNode { .. }) => {
    // handle self
  }
  Some(ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. } => {
    // handle constant
  }
  Some(other) => {
    self.visit(other)
  }
  None => {}
}
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -576,6 +710,83 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                self.uri_id,
+                offset,
+                comments,
+                owner_id,
+            )));
+
+            let definition_id = self.local_graph.add_definition(definition);
+
+            if let Some(owner_id) = owner_id
+                && let Some(owner) = self.local_graph.get_definition_mut(owner_id)
```
vinistock, line 801: Same thing here, can't we use `parent_nesting`?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -679,9 +876,24 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                // def self.foo - receiver is the current class/module's NameId
+                ruby_prism::Node::SelfNode { .. } => (self.current_nesting_name_id(), owner_id),
+                // def Foo.bar or def Foo::Bar.baz - receiver is the constant's NameId
+                // In this case, the parent nesting is NOT the owner of the method, so we set owner_id to None
+                ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. } => {
+                    let name_id = self.index_constant_reference(&recv_node, true);
+                    (name_id, None)
+                }
+                // Dynamic receiver (def foo.bar) - skip
+                _ => (None, None),
```
vinistock, line 924: Why do these two cases not have owner IDs? They might still be nested under some definition. Remember that here in indexing, `owner_id` just means the nesting parent - we actually assign ownership during resolution.

rust/saturn/src/indexing/ruby_indexer.rs
st0012: I think they're from a bad rebase 🤦‍♂️

PR #335, opened 2025-11-27
[DOC] Document indexing pipeline and principles

st0012:


docs/architecture.md
```diff
@@ -1,37 +1,108 @@
...
+
+**What Resolution does:**
+
+- Compute fully qualified names for all definitions
+- Create `Declaration` objects that group definitions by fully qualified name
+- Resolve constant references to their target declarations
+- Linearize ancestor chains (method resolution order)
+- Assign semantic membership (which methods/constants belong to which class)
+- Create implicit singleton classes from `def self.method` patterns
+- Process `extend` mixins on singleton classes
```
vinistock: Maybe we can fold this into linearize ancestor chains. This is the same thing really.

docs/architecture.md
```diff
@@ -1,37 +1,107 @@
 # Architecture
 
-Saturn's representation of the Ruby code it analyzes is a graph, where key entities are represented as nodes and the
-relationship between them as edges. The visualization below demonstrates the conceptual structure of the graph. In
-practice, it is implemented through IDs between the entities - an adjacency list approach.
+Saturn indexes Ruby codebases in two distinct stages: **Discovery** and **Resolution**. Understanding this separation is crucial for working with the codebase.
```
Morriar, line 3: I'm not a huge fan of "Discovery" as it can be confused as an incremental process. What we do is more immediate or systematic.

Some other words:
* parsing
* extraction
* inspection
* scanning
* listing

Also whatever we use here, we may want to rename the module name to match.

st0012: IMO it's definitely not parsing. Does it make sense to call it `visiting`? And then we rename the struct to `RubyVisitor`.

cc @vinistock @alexcrocha

PR #336, opened 2025-11-27
Change constant def's name field to be a NameId

st0012:


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -563,12 +608,7 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_constant_path_write_node(&mut self, node: &ruby_prism::ConstantPathWriteNode) {
-        let location = node.target().location();
-        self.add_definition_from_location(&location, |str_id, offset, comments, owner_id, uri_id| {
-            Definition::Constant(Box::new(ConstantDefinition::new(
-                str_id, uri_id, offset, comments, owner_id,
-            )))
-        });
+        self.add_constant_definition(&node.target().as_node());
```
vinistock, line 611: Similarly to `index_constant_reference`, I think the method should accept the entire `ConstantPathWriteNode` here, no?

How is the `NameRef` of the target going to be associated with the `NameRef` of the parent scope without indexing the same reference twice (given that we're visiting the parent below)?

For example

```ruby
class Foo
  # This should result in the following structure, with only 1 constant ref
  #
  # Name(
  #   str: StringId(B),
  #   parent_scope: NameId(A),
  #   nesting: NameId(Foo),
  # )
  #
  # ConstantRef(Name(str: A, parent_scope: None, nesting: NameId(Foo))
  A::B = 1
end
```

I think we're either not connecting the names or indexing the same reference twice. I didn't expect the visit to the parent to still be here.

st0012: Good catch. I've removed the duplicated visit

PR #338, opened 2025-11-28
Treat constant or write as constant definitions

st0012:


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2194,8 +2195,8 @@ mod tests {
         assert_constant_references_eq!(
             &context,
             vec![
-                "C1", "C2", "C3", "C4", "C6", "C7", "C8", "C9", "C10", "C11", "C12", "C13", "C14", "C15", "C16", "C17",
```
Morriar, line 2197: Why do we stop considering `C10` but not `C20`?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -586,7 +586,7 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_constant_or_write_node(&mut self, node: &ruby_prism::ConstantOrWriteNode) {
-        self.index_constant_reference(&node.as_node(), true);
+        self.add_constant_definition(&node.as_node());
```
Morriar, line 589: Should we also add a test showing that we indeed created a definition for this?

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Good catch. Will open a PR for it

PR #339, opened 2025-11-28
Add parent class linearization

vinistock:
Add parent class linearization. I'm more interested in seeing how this fits into the general resolution algorithm at this point, but I think this PR also uncovers the need for specializing declarations.

For example, if we turn `Declaration`​ into an enum, we can always pick the first definition we find as the truth for error scenarios, which makes it easier to emit diagnostics. Imagine we find `class Foo`​ and `module Foo`​. When we're processing the second `Foo`​, it would be easier to recognize the mistake by looking at the `Declaration`​. Same thing for superclass mismatches.

In terms of the implications for the resolution phase, I used a `OnceCell`​for ancestors, which gives us interior mutability (avoiding borrow issues). This makes it easier to work with than `RefCell`​ because we never expect to mutate the ancestor vector. We always erase and set the entire vector in one go. By using `OnceCell`​, we don't pay the runtime price of `RefCell`​.

### DIFF
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -1,3 +1,5 @@
+use std::cell::OnceCell;
+
 use crate::model::{
     identity_maps::IdentityHashMap,
     ids::{DeclarationId, DefinitionId, ReferenceId, StringId},
@@ -21,6 +23,7 @@ pub struct Declaration {
     references: Vec<ReferenceId>,
     /// The ID of the owner of this declaration
     owner_id: DeclarationId,
+    ancestors: OnceCell<Vec<DeclarationId>>,
 }
 
 impl Declaration {
@@ -32,6 +35,7 @@ impl Declaration {
             members: IdentityHashMap::default(),
             references: Vec::new(),
             owner_id,
+            ancestors: OnceCell::new(),
         }
     }
 
@@ -107,6 +111,20 @@ impl Declaration {
     #[must_use]
     pub fn owner_id(&self) -> &DeclarationId {
         &self.owner_id
+    }
+
+    /// # Panics
+    ///
+    /// Will panic if invoked twice without clearing ancestors
+    pub fn set_ancestors(&self, ancestors: Vec<DeclarationId>) {
+        self.ancestors
+            .set(ancestors)
+            .expect("ancestors should only be set once per declaration");
+    }
+
+    #[must_use]
+    pub fn ancestors(&self) -> Option<&[DeclarationId]> {
+        self.ancestors.get().map(Vec::as_slice)
     }
 
     // This will change once we fix fully qualified names to not use `::` as separators for everything. Also, we may
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -1,10 +1,13 @@
-use std::collections::VecDeque;
+use std::{
+    collections::{HashSet, VecDeque},
+    hash::BuildHasher,
+};
 
 use crate::model::{
     declaration::Declaration,
     definitions::Definition,
     graph::{Graph, OBJECT_ID},
-    identity_maps::IdentityHashMap,
+    identity_maps::{IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
     name::{Name, NameRef},
 };
@@ -30,6 +33,18 @@ impl Outcome {
     fn is_resolved_or_retry(&self) -> bool {
         matches!(self, Outcome::Resolved(_) | Outcome::Retry)
     }
+}
+
+pub enum AncestorOutcome {
+    /// The ancestors were fully linearized
+    Resolved(Vec<DeclarationId>),
+    /// Represents a cyclic outcome when linearizing ancestors. For example, if a class inherits from itself (either
+    /// directly or through a chain of parent classes). We still try our best effort to approximate something, but this
+    /// case is an error
+    Cyclic(Vec<DeclarationId>),
+    /// Represents that we didn't have all of the information necessary to complete linearizing the ancestors. We must
+    /// retry in the next iteration of the resolution phase
+    Retry,
 }
 
 /// Runs the resolution phase on the graph. The resolution phase is when 4 main pieces of information are computed:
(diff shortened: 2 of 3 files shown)


rust/saturn/src/resolution.rs
```diff
@@ -629,4 +728,66 @@ mod tests {
         context.resolve();
         assert_eq!(1, context.graph.declarations().len());
     }
+
+    #[test]
+    fn linearizing_super_classes() {
```
Morriar, line 1138: Let's add a test with this:

```rb
class Foo < Bar; end
class Bar < Foo; end
```

Or even this:

```rb
class Foo < Bar; end
class Bar < Baz; end
class Baz < Foo; end
```

rust/saturn/src/resolution.rs
```diff
@@ -132,6 +137,43 @@ pub fn resolve_all(graph: &mut Graph) {
...
+    }
+
+    let mut ancestors = Vec::new();
+    let definitions = declaration
+        .definitions()
+        .iter()
+        .filter_map(|def_id| graph.definitions().get(def_id))
+        .collect::<Vec<_>>();
+
+    ancestors.push(declaration_id);
```
Morriar, line 278:
```suggestion
    let mut ancestors = vec![declaration_id];
    
    let definitions = declaration
        .definitions()
        .iter()
        .filter_map(|def_id| graph.definitions().get(def_id))
        .collect::<Vec<_>>();
```

rust/saturn/src/resolution.rs
```diff
@@ -132,6 +137,43 @@ pub fn resolve_all(graph: &mut Graph) {
...
+
+    // TODO: this check is against `Object` for now to avoid infinite recursion. After RBS indexing, we need to change
+    // this to `BasicObject` since it's the only class that cannot have a parent
+    if declaration_id != *OBJECT_ID {
```
Morriar, line 282: I think we can actually move the `let definitions` in this if to avoid computing it for nothing?

Morriar: Maybe we could also just pass ids to `linearize_parent_class` and let it map it to actual structs?

rust/saturn/src/resolution.rs
```diff
@@ -315,6 +381,39 @@ fn sorted_units(graph: &Graph) -> (VecDeque<Unit>, Vec<DefinitionId>) {
...
+            let name = graph.names().get(&superclass).unwrap();
+
+            match name {
+                NameRef::Resolved(resolved) => {
+                    let parent_id = *resolved.declaration_id();
+
+                    // Between the default `Object` parent class and anything else, we pick the explicit one
+                    if parent_id != *OBJECT_ID {
+                        explicit_parents.push(parent_id);
+                    }
```
Morriar: Why do we need this? Isn't the `superclass_ref` always an explicit parent?

rust/saturn/src/resolution.rs
vinistock: Done!

rust/saturn/src/resolution.rs
vinistock: This would work for now because we only have parent classes, but we would need to change it back once we start handling prepended modules.

What do you think about keeping it as is so that the next person working on prepended modules can just plug it before the push?

rust/saturn/src/resolution.rs
vinistock: This is similar to the other comment. Once we have mixins handling, we'll need to pass the definitions to more methods and so retrieving them here allows us to do it only once.

Then we pass it down to collect includes, prepends, parent classes and so on.

rust/saturn/src/resolution.rs
vinistock: My original idea was that we should accept a super class redefinition from explicit `Object` to something else, but I just realized that's actually an error. Only implicit (no parent class) should be allowed to change the parent.

```ruby
class Foo < Object
end

class Foo < Bar; end # not okay (I thought it would be)

class Baz; end
class Baz < Bar; end # okay
```

I removed the check.

rust/saturn/src/resolution.rs
Morriar: > Done!

I do not see the given examples in the tests? Missing a `git push` maybe?

Just to make it clear, both my examples have a cyclic inheritance.

vinistock: Sorry, I completely missed the point of it being a circular dependency. I added this example and the handling for detecting cycles in the linearization.

Since this scenario is an error (for which we'll want a diagnostic later), I opted for a best effort approach of assuming that the class/module involved in the cycle will at least inherit from `Object` in some point of their inheritance chain.

It's not 100% correct, but it would allow us to provide more completion results if the user is in the middle of fixing their class structure. Let me know what you think!

Morriar: I think Object is okay for now but we may want to introduce some kind of `<TODO>` or `<ERROR>` special declaration at some point.

Now that https://github.com/Shopify/saturn/pull/350 is merged, wdyt of adding a diagnostic for this?

vinistock: Let's have a separate task for adding diagnostics for all possible resolution failures. There are more cases than just this one, which will inflate the scope of this PR, and since I'm out tomorrow and Friday and I won't have time to finish everything.

vinistock: https://github.com/Shopify/saturn/issues/354

rust/saturn/src/resolution.rs
```diff
@@ -969,4 +1108,104 @@ mod tests {
...
+                    *OBJECT_ID
+                ],
+                ancestors,
+                "Incorrect ancestors {}",
+                ancestors
+                    .iter()
+                    .map(|id| context.graph.declarations().get(id).unwrap().name())
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            );
```
Morriar: Let's extract this as a macro:

```rs
assert_linearized_ancestors_eq!("Qux", vec!["Qux", "Baz", "Bar", "Foo"])
```

rust/saturn/src/resolution.rs
```diff
@@ -969,4 +1108,104 @@ mod tests {
...
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Bar < Foo
+              CONST
+            end
+            "
+        });
+        context.resolve();
+        assert_eq!(2, context.graph.declarations().len());
```
Morriar, line 1195: Should we also assert on the behavior of `ancestor_of(Bar)`?

rust/saturn/src/resolution.rs
```diff
@@ -215,6 +230,78 @@ fn get_or_create_singleton_class(graph: &mut Graph, owner_decl_id: DeclarationId
     singleton_decl_id
 }
 
+/// Linearizes the ancestors of a declaration, returning the list of ancestor declaration IDs
+///
+/// # Panics
+///
+/// Can panic if there's inconsistent data in the graph
+#[must_use]
+pub fn ancestors_of(graph: &Graph, declaration_id: DeclarationId) -> AncestorOutcome {
```
Morriar, line 239: Will this be called from `resolve_all` at some point?

rust/saturn/src/model/declaration.rs
```diff
@@ -109,6 +113,20 @@ impl Declaration {
         &self.owner_id
     }
 
+    /// # Panics
+    ///
+    /// Will panic if invoked twice without clearing ancestors
+    pub fn set_ancestors(&self, ancestors: Vec<DeclarationId>) {
```
Morriar, line 119: How come this doesn't require a `&mut self`?

Morriar: Ah it's because it's a `OnceCell` 👍

rust/saturn/src/resolution.rs
vinistock: It's already called when trying to resolve a constant. It first searches lexical scopes and then searches the linearized ancestors.

Morriar: Ah it's called from `run_resolution`, I see 👍

rust/saturn/src/resolution.rs
```diff
@@ -969,4 +1130,91 @@ mod tests {
...
+
+        assert_ancestors_eq!(
+            context,
+            "Qux",
+            vec![
+                DeclarationId::from("Qux"),
+                DeclarationId::from("Baz"),
+                DeclarationId::from("Bar"),
+                DeclarationId::from("Foo"),
+                *OBJECT_ID
```
Morriar: Let's just pass the strings and let the macro create the ids in the same way you passed `Qux`

vinistock: Done!

PR #343, opened 2025-12-02
Introduce local graph diagnostics

Morriar:
Step towards https://github.com/Shopify/saturn/issues/340.

Let's collect local diagnostics from document parsing.

### DIFF
--- a/rust/saturn/src/diagnostic.rs
+++ b/rust/saturn/src/diagnostic.rs
@@ -1,0 +1,57 @@
+use crate::{model::ids::UriId, offset::Offset};
+
+#[derive(Debug)]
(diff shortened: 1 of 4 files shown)


rust/saturn/src/diagnostic.rs
```diff
@@ -0,0 +1,37 @@
+use crate::{model::ids::UriId, offset::Offset};
+
+#[derive(Debug)]
+pub struct Diagnostic {
+    pub uri_id: UriId,
+    pub offset: Offset,
+    pub message: String,
+    pub severity: Severity,
```
vinistock: Can we make these private?

PR #344, opened 2025-12-02
Intern parameter names

### LINKED ISSUE #164: Add string interning for parameter names
We intern the name of all declarations as IDs to avoid keeping duplicated strings stored in memory. For method parameters, we are currently not doing that which is a big part of our memory usage explosion. If you comment out method indexing, the memory usage goes down from ~8.7 GB to ~2 GB.

We will need a way to ensure that we're not keeping duplicate strings for the parameter names in memory. My initial hunch is that we do not need deterministic IDs for method parameters like we did for declaration names. Simple string interning with sequential IDs should likely do it.

Investigate how we can intern parameter names efficiently and implement it.

Note: there's some previous work done in #46. We might just want to reuse what we had previously for parameters.

vinistock:
Closes #164

This PR starts interning parameter names, so that we don't keep duplicate strings around.

```
Benchmark average of 5 runs on Core

Indexing time 3.4094 -> 3.1536
Memory 1689.282 -> 1675.4
```

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -137,45 +137,49 @@ impl<'a> RubyIndexer<'a> {
         }
     }
 
-    fn collect_parameters(node: &ruby_prism::DefNode) -> Vec<Parameter> {
+    fn collect_parameters(&mut self, node: &ruby_prism::DefNode) -> Vec<Parameter> {
         let mut parameters: Vec<Parameter> = Vec::new();
 
         if let Some(parameters_list) = node.parameters() {
             for parameter in &parameters_list.requireds() {
                 let location = parameter.location();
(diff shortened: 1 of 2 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1555,35 +1568,43 @@ mod tests {
             assert_eq!(def.parameters().len(), 8);
 
             assert_parameter!(&def.parameters()[0], RequiredPositional, |param| {
-                assert_eq!(param.name(), "a");
+                let string_name = context.graph().strings().get(param.str()).unwrap();
+                assert_eq!(string_name, "a");
```
Morriar: Let's add a `assert_string_eq!(context, name, "foo")` macro for this

vinistock: Done!

PR #352, opened 2025-12-03
[DOC] Document more singleton class behaviours

st0012:


docs/ruby-behaviors.md
```diff
@@ -648,12 +648,95 @@ class Foo; end
...
+
+Class variables (`@@var`) **bypass singleton classes entirely**.
+They always belong to the base class or module, regardless of where they are defined:
+
+```ruby
+class Foo
+  @@in_class_body = 1
+
+  class << self
+    @@in_singleton_scope = 2
```
Morriar, line 680: The fun part is it's also true with this:

```rb
class Foo; end

class Bar
  class << Foo
    @@foo = 42
  end
end

Foo.class_variables # => []
Foo.singleton_class.class_variables # => []

Bar.singleton_class.class_variables # => [:@@foo]
Bar.class_variables # => [:@@foo]
```

st0012: I don't get it tho 🤔 

```rb
class Bar
  class Foo
    @@foo = 1
  end
end

puts Bar::Foo.class_variables # => []
puts Bar::Foo.singleton_class.class_variables # => []

puts '--------------------------------'

puts Bar.singleton_class.class_variables # => [:@@foo]
puts Bar.class_variables # => [:@@foo]

```

```
$ ruby test.rb
@@foo
@@foo
--------------------------------
```

Morriar: Sorry, there was a typo in my example, the class was supposed to be a singleton:

```rb
class Foo; end

class Bar
  class << Foo
    @@foo = 42
  end
end
```

You could expect the `@@foo` to be defined on `Foo` but it's actually on `Bar`.

PR #353, opened 2025-12-03
Introduce Location concept

Morriar:
Rerun of https://github.com/Shopify/saturn/pull/299.

We split the concept into two entities:

* `Offset`: internal representation 0-based byte index
* `Location`: presentation related 1-based line/column range

`Location ` should only be used to present things to the end user in the terminal, in error messages, etc. All other processing should be based on the `Offset`.

For now, the C API only expose the `Location` of entities but I think we should add a `#offset` method as soon as we start basing the LSP on top of it.

One question remains for me: should `Location` be using Position?

```rs
pub struct Location {
    uri: String,
    start: Position,
    end: Position,
}
```

or keep the current form:

```rs
pub struct Location {
    uri: String,
    line_start: u32,
    line_end: u32,
    column_start: u32,
    column_end: u32,
}
```

Closes https://github.com/Shopify/saturn/pull/299.

### DIFF
--- a/rust/saturn/src/location.rs
+++ b/rust/saturn/src/location.rs
@@ -161,6 +161,17 @@ impl Location {
     pub fn column_end(&self) -> u32 {
         self.column_end
     }
+
+    #[must_use]
+    pub fn to_zero_based(&self) -> Location {
+        Location::new(
+            self.uri.clone(),
+            self.line_start - 1,
+            self.line_end - 1,
+            self.column_start - 1,
+            self.column_end - 1,
+        )
+    }
 }
 
 impl Display for Location {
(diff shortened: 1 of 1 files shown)


rust/saturn-sys/Cargo.toml
```diff
@@ -10,7 +10,6 @@ crate-type = ["cdylib", "staticlib"]
 saturn = { path = "../saturn" }
 libc = "0.2.174"
 url = "2.5.4"
-line-index = "0.1.2"
```
vinistock, line 13: How are we going to do code unit column computation for UTF-8, UTF-16 and UTF-32 without this crate?

Morriar: We already inherit the requirement from `saturn`: https://github.com/Shopify/saturn/blob/main/rust/saturn/Cargo.toml#L22

vinistock: Oh, my bad. I didn't realize this was the other `Cargo.toml`.

PR #355, opened 2025-12-03
Discover & resolve singleton classes

st0012:
This PR adds singleton class definition indexing for `class << X` expressions:

```rb
class Foo
  class << self           # creates `SingletonClassDefinition` `<Foo>`
    def baz; end          # method owned by `<Foo>`

    class << self         # creates `SingletonClassDefinition` `<<Foo>>`
      def qux; end        # method owned by `<<Foo>>`
    end
  end
end

class << Foo              # creates `SingletonClassDefinition` `<Foo>`
  def quux; end           # method owned by `<Foo>`
end

class << some_var         # dynamic expression - no definition created
  def dynamic; end
end
```

Key discovering behaviors:
- Only `class << self` or `class << Constant` creates a `SingletonClassDefinition`
- Class variables (`@@var`) bypass singleton blocks. They always belong to the base class/module
- `extend SomeModule` is stored as `Mixin::Extend` on the class/module

### DIFF
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -66,6 +66,7 @@ pub fn resolve_all(graph: &mut Graph) {
                     Definition::Class(class) => *class.name_id(),
                     Definition::Module(module) => *module.name_id(),
                     Definition::Constant(constant) => *constant.name_id(),
+                    Definition::SingletonClass(singleton) => *singleton.name_id(),
                     _ => panic!("Expected constant definitions"),
                 };
 
@@ -340,6 +341,9 @@ fn sorted_units(graph: &Graph) -> (VecDeque<Unit>, Vec<DefinitionId>) {
                 constants.push((Unit::Definition(*id), names.get(def.name_id()).unwrap()));
             }
             Definition::Constant(def) => {
+                constants.push((Unit::Definition(*id), names.get(def.name_id()).unwrap()));
+            }
+            Definition::SingletonClass(def) => {
                 constants.push((Unit::Definition(*id), names.get(def.name_id()).unwrap()));
             }
             _ => {
@@ -717,5 +721,100 @@ mod tests {
                 .map(|n| context.graph.strings().get(n.str()).unwrap().as_str())
                 .collect::<Vec<_>>()
         );
+    }
+
+    #[test]
+    fn resolution_for_singleton_class() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Foo
+              class << self
+                def bar; end
+                BAZ = 123
+              end
+            end
+            "
+        });
+        context.resolve();
+
+        let foo = context.graph.declarations().get(&DeclarationId::from("Foo")).unwrap();
+        assert_members(foo, &["<Foo>"]);
+        assert_owner(foo, "Object");
+
+        let singleton = context
+            .graph
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>"))
+            .unwrap();
+        assert_members(singleton, &["bar", "BAZ"]);
+        assert_owner(singleton, "Foo");
+    }
+
+    #[test]
+    fn resolution_for_nested_singleton_class() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Foo
+              class << self
+                class << self
+                  def baz; end
+                end
+              end
+            end
+            "
+        });
+        context.resolve();
+
+        let foo = context.graph.declarations().get(&DeclarationId::from("Foo")).unwrap();
+        assert_members(foo, &["<Foo>"]);
+
+        let singleton = context
+            .graph
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>"))
+            .unwrap();
+        assert_members(singleton, &["<<Foo>>"]);
+
+        let nested_singleton = context
+            .graph
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>::<<Foo>>"))
+            .unwrap();
+        assert_members(nested_singleton, &["baz"]);
+        assert_owner(nested_singleton, "Foo::<Foo>");
+    }
+
+    #[test]
+    fn resolution_for_singleton_class_of_external_constant() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Foo; end
+            class Bar
+              class << Foo
+                def baz; end
+              end
+            end
+            "
+        });
+        context.resolve();
+
+        let foo = context.graph.declarations().get(&DeclarationId::from("Foo")).unwrap();
+        assert_members(foo, &["<Foo>"]);
+        assert_owner(foo, "Object");
+
+        let bar = context.graph.declarations().get(&DeclarationId::from("Bar")).unwrap();
+        assert_members(bar, &[]);
+        assert_owner(bar, "Object");
+
+        let singleton = context
+            .graph
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>"))
+            .unwrap();
+        assert_members(singleton, &["baz"]);
+        assert_owner(singleton, "Foo");
     }
 }


rust/saturn/src/model/graph.rs
```diff
@@ -457,8 +461,60 @@ impl Graph {
...
         types.sort_by_key(|(_, count)| std::cmp::Reverse(**count));
         for (kind, count) in types {
-            println!("  {kind:20} {count:6}");
+            println!("  {kind:25} {count:7}");
+        }
+
+        self.print_singleton_statistics();
+    }
+
+    fn print_singleton_statistics(&self) {
```
st0012: I'm fixing this now. If it's much work I'll drop it.

rust/saturn/src/model/definitions.rs
```diff
@@ -108,22 +111,24 @@ impl Definition {
...
 
-/// Represents either an included or a prepended module. Note that `extend` is modeled as an include on the singleton
-/// class
+/// Represents a mixin: include, prepend, or extend.
+/// During resolution, `Extend` mixins are attached to the singleton class.
 #[derive(Debug)]
 pub enum Mixin {
     Include(NameId),
     Prepend(NameId),
+    Extend(NameId),
```
vinistock, line 131: A thought for future work: I understand why we would model an extend here. At the indexing phase, we don't have the singleton definition to attach the extend as what is really is (an include in the singleton).

However, how would we handle the case of a prepend to the singleton class? Can we identify some of these? Maybe we need a 4th concept for a prepend that happens to the singleton class using something like this:

```ruby
module Patch
  def override; end

  Something.singleton_class.prepend(self)
end
```

rust/saturn/src/resolution.rs
```diff
@@ -718,4 +722,82 @@ mod tests {
...
+
+        let foo = context.graph.declarations().get(&DeclarationId::from("Foo")).unwrap();
+        assert_members(foo, &["<Foo>"]);
+        assert_owner(foo, "Object");
+
+        let bar = context.graph.declarations().get(&DeclarationId::from("Bar")).unwrap();
+        assert_members(bar, &[]);
+        assert_owner(bar, "Object");
+
+        let singleton = context.graph.declarations().get(&DeclarationId::from("Foo::<Foo>")).unwrap();
```
vinistock: Very cool that the modelling allows us to handle this case correctly. It's not a super common use case of Ruby, but the fact that the model _simply works_ gives me an extra dose of confidence on the approach.

rust/saturn/src/resolution.rs
```diff
@@ -718,4 +722,82 @@ mod tests {
                 .collect::<Vec<_>>()
         );
     }
+
+    #[test]
+    fn resolution_for_singleton_class() {
```
vinistock, line 727: Let's also add an example with a constant defined inside of the singleton block.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -427,54 +448,41 @@ impl<'a> RubyIndexer<'a> {
...
+            let mixin = match mixin_type {
+                MixinType::Include => Mixin::Include(id),
+                MixinType::Prepend => Mixin::Prepend(id),
+                MixinType::Extend => Mixin::Extend(id),
+            };
+
+            match definition {
+                Definition::Class(class_def) => class_def.add_mixin(mixin),
+                Definition::Module(module_def) => module_def.add_mixin(mixin),
+                Definition::SingletonClass(block_def) => block_def.add_mixin(mixin),
```
vinistock: Should it be `singleton_def`?
```suggestion
                Definition::SingletonClass(singleton_def) => singleton_def.add_mixin(mixin),
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -612,6 +620,83 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                    .strings()
+                    .get(name.str())
+                    .expect("Attached target string should exist");
+                format!("<{target_str}>")
+            };
+
+            let string_id = self.local_graph.intern_string(singleton_class_name);
+            let name_id = self
+                .local_graph
+                .add_name(Name::new(string_id, Some(attached_target), None));
```
vinistock, line 665: This idea of considering the target as a parent scope of the singleton is clever 👏.

However, doesn't that mean that we don't need to store the `attached_target` in the definition? It's `name_id` is already connected to the target through the parent scope. The fact that resolution is already working without even referring to `attached_target` seems like evidence that we don't really need to store this.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -612,6 +620,83 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                owner_id,
+            )));
+
+            let definition_id = self.local_graph.add_definition(definition);
+
+            if let Some(owner_id) = owner_id
+                && let Some(owner) = self.local_graph.get_definition_mut(owner_id)
+            {
+                owner.add_member(definition_id);
+            }
```
vinistock: Can we use the same style we're using in other places here?

```rust
if let Some(parent_nesting) = self.parent_nesting() {
  parent_nesting.add_member(definition_id);
}
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -715,9 +800,24 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                // def self.foo - receiver is the current class/module's NameId
+                ruby_prism::Node::SelfNode { .. } => (true, owner_id),
+                // def Foo.bar or def Foo::Bar.baz - receiver is the constant's NameId
+                // In this case, the parent nesting is NOT the owner of the method, so we set owner_id to None
+                ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. } => {
+                    self.visit(&recv_node);
+                    (true, None)
+                }
+                // Dynamic receiver (def foo.bar) - skip
+                _ => (true, None),
```
vinistock: I think the `owner_id` for these two branches should remain the same no? Keep in mind that `owner_id` in the context of the indexer is actually the ID of the parent nesting - which exists even for dynamic definitions.

st0012: In that case, shouldn't `def Foo.bar` be "owned" by the parent nesting too?
Also, do you mind if I update `owner_id` to something like `lexical_nesting_id` so I don't constantly confuse myself 💀

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Yeah good call. Will remove it.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Yes, please let's rename it. And indeed `def Foo.bar` is "owned" by its lexical scope. We need that connection for resolution and proper ownership attribution later.

rust/saturn/src/resolution.rs
st0012: Yeah it's surprisingly easy to get resolution work with singleton class. Great design 👍

rust/saturn/src/indexing/ruby_indexer.rs
st0012: https://github.com/Shopify/saturn/pull/356

rust/saturn/src/resolution.rs
st0012: Done

rust/saturn/src/model/definitions.rs
st0012: > an include in the singleton

Almost but not exactly. With `extend`, the `included` callback is not called:

```rb
# frozen_string_literal: true

module Foo
  def self.included(base)
    puts "included in #{base}"
  end

  def hello
    puts "hello"
  end
end

class Bar
  singleton_class.include(Foo)
end

class Baz
  extend Foo
end

Bar.hello
Baz.hello

# included in #<Class:Bar>
# hello
# hello
```

> However, how would we handle the case of a prepend to the singleton class?

Since we can also do:

```rb
Foo.singleton_class.include(Bar)
```

Should the question be: Do we create singleton class definition when seeing `CONST.singleton_class`?

vinistock: Maybe not a definition, but it would be nice if we successfully tracked some of these includes and prepends on the singleton classes. Also, TIL that `extend` doesn't run the included callback.

Anyway, something to consider for an experiment.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -738,7 +817,7 @@ impl Visit<'_> for RubyIndexer<'_> {
             comments,
             lexical_nesting_id,
             parameters,
-            is_singleton,
+            node.receiver().is_some(),
```
vinistock, line 820: When you get to handling method receivers, I _think_ they will end up being an `Option<NameId>` where

- `None` means it's a direct definition with no receiver
- `Some(NameId)` means it's a constant that we can resolve and correctly associate the method with

And then we discard dynamic definitions.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -958,18 +1045,43 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+            str_id,
+            uri_id,
+            offset,
+            comments,
+            lexical_nesting_id,
+        )));
+        let definition_id = self.local_graph.add_definition(definition);
+
+        if let Some(lexical_nesting_id) = lexical_nesting_id
+            && let Some(parent_nesting) = self.local_graph.get_definition_mut(lexical_nesting_id)
```
vinistock, line 1080: Can we use this style here?

```rust
if let Some(parent_nesting) = self.parent_nesting() {
  parent_nesting.add_member(definition_id);
}
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -958,18 +1045,43 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+        let uri_id = self.uri_id;
+
+        let definition = Definition::ClassVariable(Box::new(ClassVariableDefinition::new(
+            str_id,
+            uri_id,
+            offset,
+            comments,
+            lexical_nesting_id,
+        )));
+        let definition_id = self.local_graph.add_definition(definition);
```
vinistock, line 1077: Defining class variables directly at the top level is not allowed. Are we still discarding those as expected?

We should also add a diagnostic in the future since Ruby crashes.

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Yes that's how I implemented it in the original PR. Will do the same in the follow up 👍

rust/saturn/src/indexing/ruby_indexer.rs
st0012: This `lexical_nesting_id` is not necessarily the last definition id from the stack. In a few lines above, we need to skip to current lexical nesting's lexical nesting if it's a singleton class:

```rust
        while let Some(current_lexical_nesting_id) = lexical_nesting_id {
            if let Some(definition) = self.local_graph.get_definition_mut(current_lexical_nesting_id)
                && matches!(definition, Definition::SingletonClass(_))
                && let Some(definition_lexical_nesting_id) = definition.lexical_nesting_id()
            {
                lexical_nesting_id = Some(*definition_lexical_nesting_id);
            } else {
                break;
            }
        }
```

So I think here we also need to use the found `lexical_nesting_id` for member addition.

rust/saturn/src/model/definitions.rs
st0012: Added https://github.com/Shopify/saturn/issues/362 for this

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Yeah we still and should index them so we can warn users about them later.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1657,11 +1792,213 @@ mod tests {
...
+        });
+    }
+
+    #[test]
+    fn index_class_variable_in_nested_singleton_class_definition() {
+        let context = index_source({
+            "
+            class Foo
+              class << self
+                class << self
```
Morriar, line 1970: Let's also add a test showing `class << foo` is not indexed?

st0012: Will add it as part of my next step PR 👍

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -958,18 +1045,43 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+        while let Some(current_lexical_nesting_id) = lexical_nesting_id {
+            if let Some(definition) = self.local_graph.get_definition_mut(current_lexical_nesting_id)
+                && matches!(definition, Definition::SingletonClass(_))
+                && let Some(definition_lexical_nesting_id) = definition.lexical_nesting_id()
+            {
+                lexical_nesting_id = Some(*definition_lexical_nesting_id);
+            } else {
+                break;
+            }
+        }
```
Morriar, line 1066: I think at the definition level we should still represent these variables as being inside the singleton then move it to the correct scope at the declaration level.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1657,11 +1792,213 @@ mod tests {
...
+        });
+    }
+
+    #[test]
+    fn index_constant_in_singleton_class_definition() {
+        let context = index_source({
+            "
+            class Foo
+              class << self
+                A = 1
```
Morriar, line 1927: You can also add this:

```rb
class Foo
  class << self
    class Bar; end
  end
end
```

rust/saturn/src/resolution.rs
```diff
@@ -718,4 +722,99 @@ mod tests {
...
+        assert_owner(bar, "Object");
+
+        let singleton = context
+            .graph
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>"))
+            .unwrap();
+        assert_members(singleton, &["baz"]);
+        assert_owner(singleton, "Foo");
+    }
```
Morriar, line 819: Maybe we can also show how this should resolve:

```rb
class Foo; end

class Bar
  class << Foo
    class Baz; end
  end
end
```

st0012: Added in https://github.com/Shopify/saturn/pull/363

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Yeah makes sense. Updated in https://github.com/Shopify/saturn/pull/363

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Added in https://github.com/Shopify/saturn/pull/363 as a resolution test.

PR #357, opened 2025-12-03
Index variations of global variables

### LINKED ISSUE #82: Start indexing global variables
Start indexing global variables, including creating a definition type for it.

You can refer to the [Ruby LSP's test](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/test/global_variable_test.rb) for scenarios we have to cover.

For example:

```ruby
# These should be indexed as definitions
$foo = 1
$foo, $bar = [1, 2]
$foo &= 1
$foo &&= 1
$foo ||= 1
```

Note that, because they are global variables, their fully qualified name is just the name itself (e.g.: `$foo`, `$bar`).

vinistock:
Closes #82

This PR adds indexing for the remaining types of global variables. To my surprise, in all of these variant of global variable nodes, Ruby does not crash if the variable doesn't exist beforehand - including the alias global variable node!

None of these crash:

```ruby
$a &= 1
$a &&= 1
$a ||= 1
alias $a $b
```

Because of that, I modelled all of them as definitions, so that we don't end up adding errors that don't match the runtime (same idea as `A ||= 1`).

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1085,6 +1085,44 @@ impl Visit<'_> for RubyIndexer<'_> {
         self.visit(&node.value());
     }
 
+    fn visit_global_variable_and_write_node(&mut self, node: &ruby_prism::GlobalVariableAndWriteNode<'_>) {
+        self.add_definition_from_location(&node.name_loc(), |str_id, offset, comments, nesting_id, uri_id| {
+            Definition::GlobalVariable(Box::new(GlobalVariableDefinition::new(
+                str_id, uri_id, offset, comments, nesting_id,
+            )))
+        });
+        self.visit(&node.value());
+    }
+
+    fn visit_global_variable_or_write_node(&mut self, node: &ruby_prism::GlobalVariableOrWriteNode<'_>) {
+        self.add_definition_from_location(&node.name_loc(), |str_id, offset, comments, nesting_id, uri_id| {
+            Definition::GlobalVariable(Box::new(GlobalVariableDefinition::new(
+                str_id, uri_id, offset, comments, nesting_id,
+            )))
+        });
+        self.visit(&node.value());
+    }
+
+    fn visit_global_variable_operator_write_node(&mut self, node: &ruby_prism::GlobalVariableOperatorWriteNode<'_>) {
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -941,6 +941,44 @@ impl Visit<'_> for RubyIndexer<'_> {
         self.visit(&node.value());
     }
 
+    fn visit_global_variable_and_write_node(&mut self, node: &ruby_prism::GlobalVariableAndWriteNode<'_>) {
+        self.add_definition_from_location(&node.name_loc(), |str_id, offset, comments, owner_id, uri_id| {
+            Definition::GlobalVariable(Box::new(GlobalVariableDefinition::new(
```
st0012, line 1090: Why do global variables need `lexical_nesting_id` in the first place?

vinistock: For the representation of the global understanding, keeping that information is indeed irrelevant. Global variables always belong to the magic `<main>` object and are accessible anywhere.

Keeping the nesting ID here is based on the principle of storing all of the information of what we see in documents to allow for different consumers. For example, if you wanted to write a linting rule that all global variables should be defined outside of any nesting, then you would need this information.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -941,6 +941,44 @@ impl Visit<'_> for RubyIndexer<'_> {
         self.visit(&node.value());
     }
 
+    fn visit_global_variable_and_write_node(&mut self, node: &ruby_prism::GlobalVariableAndWriteNode<'_>) {
+        self.add_definition_from_location(&node.name_loc(), |str_id, offset, comments, owner_id, uri_id| {
```
st0012: Should we still use `owner_id` here?

rust/saturn/src/indexing/ruby_indexer.rs
st0012: I see, that makes sense.

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Switched to `nesting_id`.

PR #359, opened 2025-12-03
Index instance variable and/op/or assigns

Morriar:
Closes https://github.com/Shopify/saturn/issues/83.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1032,6 +1032,54 @@ impl Visit<'_> for RubyIndexer<'_> {
         self.visit(&node.value());
     }
 
+    fn visit_instance_variable_and_write_node(&mut self, node: &ruby_prism::InstanceVariableAndWriteNode) {
+        self.add_definition_from_location(
+            &node.name_loc(),
+            |str_id, offset, comments, lexical_nesting_id, uri_id| {
+                Definition::InstanceVariable(Box::new(InstanceVariableDefinition::new(
+                    str_id,
+                    uri_id,
+                    offset,
+                    comments,
+                    lexical_nesting_id,
+                )))
+            },
+        );
+        self.visit(&node.value());
+    }
+
+    fn visit_instance_variable_operator_write_node(&mut self, node: &ruby_prism::InstanceVariableOperatorWriteNode) {
+        self.add_definition_from_location(
+            &node.name_loc(),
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2284,23 +2332,77 @@ mod tests {
...
 
-        assert_eq!(context.graph().definitions().len(), 5);
-
         assert_definition_at!(&context, "1:1-1:5", InstanceVariable, |def| {
             assert_name_eq!(&context, "@foo", def);
             assert!(def.lexical_nesting_id().is_none());
         });
 
-        assert_definition_at!(&context, "4:3-4:7", InstanceVariable, |def| {
+        assert_definition_at!(&context, "3:1-6:4", Class, |parent_nesting| {
```
st0012: Can we name `parent_nesting` to something more explicit, like `foo_class_def`? 
I know it's following an existing pattern, but as the block gets longer it's now harder to see what `parent_nesting` means at the last few assertions.
It also helps when there are multiple definition blocks in the same tests.
Can be a follow up change.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -2284,23 +2332,77 @@ mod tests {
...
+            assert_definition_at!(&context, "14:3-14:7", InstanceVariable, |def| {
+                assert_name_eq!(&context, "@bar", def);
+                assert_eq!(parent_nesting.id(), def.lexical_nesting_id().unwrap());
+                assert_eq!(parent_nesting.members()[1], def.id());
+            });
+
+            assert_definition_at!(&context, "15:3-15:7", InstanceVariable, |def| {
+                assert_name_eq!(&context, "@baz", def);
+                assert_eq!(parent_nesting.id(), def.lexical_nesting_id().unwrap());
+                assert_eq!(parent_nesting.members()[2], def.id());
```
st0012: I feel we can have a `assert_members` macro similar to what `resolution.rs` has.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Changed 👍 I'll also update the other PR.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Not for these tests since we also want to make sure the offset of each node is correct.

PR #365, opened 2025-12-04
Use diagnostics instead of errors

Morriar:
Now that both the global and local graphs have diagnostics we can rely on them for indexing issues.

This also greatly simplify how we can expose these errors to the Ruby side.

### DIFF
--- a/ext/saturn/diagnostic.c
+++ b/ext/saturn/diagnostic.c
@@ -1,0 +1,19 @@
+#include "diagnostic.h"
+#include "rustbindings.h"
+
+VALUE cDiagnostic;
+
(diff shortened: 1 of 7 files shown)


rust/saturn-sys/src/location_api.rs
```diff
@@ -37,6 +37,18 @@ fn file_path_from_uri(uri: &str) -> Option<String> {
 /// - If the offset cannot be converted to a position.
 #[must_use]
 pub(crate) fn create_location_for_uri_and_offset(uri: &str, offset: &Offset) -> *mut Location {
+    if !offset.is_valid() {
```
st0012, line 40: Why do we need this change?

PR #368, opened 2025-12-05
Add error codes to diagnostics

Morriar:
In https://github.com/Shopify/saturn/issues/340 we want to follow the LSP specification of diagnostics.

I think one important field we're currently missing is the error code: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnostic.

Error codes help to filter the list of errors by allowing/suppressing specific codes. It is super useful when dealing with Sorbet or trying to filter long error list in the editor. It also helps redirecting to some kind of error page: https://sorbet.org/docs/error-reference#2001.

Before adding more diagnostics in the different phases let's see if we can add such a code to our diagnostics.

Closes https://github.com/Shopify/saturn/pull/365.

### DIFF
--- a/ext/saturn/graph.c
+++ b/ext/saturn/graph.c
@@ -8,6 +8,7 @@
 #include "utils.h"
 
 static VALUE cGraph;
+static VALUE eIndexingError;
 
 // Free function for the custom Graph allocator. We always have to call into Rust to free data allocated by it
 static void graph_free(void *ptr) {
@@ -25,8 +26,8 @@ static VALUE sr_graph_alloc(VALUE klass) {
     return TypedData_Wrap_Struct(klass, &graph_type, graph);
 }
 
-// Graph#index_all: (Array[String] file_paths) -> String?
-// Returns the error messages concatenated as a single string if anything failed during indexing or `nil`
+// Graph#index_all: (Array[String] file_paths) -> nil
+// Raises IndexingError if anything failed during indexing
 static VALUE sr_graph_index_all(VALUE self, VALUE file_paths) {
     check_array_of_strings(file_paths);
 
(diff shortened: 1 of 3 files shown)


rust/saturn/src/diagnostic.rs
```diff
@@ -1,7 +1,57 @@
...
+    (2001, ParseWarning, Severity::Warning);
+
+    // 3000 - Indexing errors
+
+    // 4000 - Resolution errors
+}
+
 #[derive(Debug)]
 pub struct Diagnostic {
+    code: u32,
```
vinistock: u32 is pretty big. How about we use u16?

rust/saturn/src/diagnostic.rs
```diff
@@ -1,7 +1,57 @@
...
+            pub fn code(&self) -> u32 {
+                match self {
+                    $(
+                        Code::$variant => $code,
+                    )*
+                }
+            }
+
+            #[must_use]
+            pub fn severity(&self) -> Severity {
```
vinistock: With this change, it seems that both `Code` and `Diagnostic` have `severity`. Can we unify everything in `Code`?

rust/saturn/src/diagnostic.rs
```diff
@@ -1,7 +1,57 @@
...
+        );* $(;)?
+    ) => {
+        #[derive(Debug)]
+        pub enum Code {
+            $(
+                $variant,
+            )*
+        }
+
+        impl Code {
```
vinistock: We currently have `code` as the name of both this abstraction and the actual code (the number) that it stores.

How about we call this `Kind` so that `code` remains the number? Especially so that we can avoid `code.code()` 😂

Issue #79, opened 2025-08-12
Singleton class indexing


vinistock:
Start indexing singleton class entities. Note that this includes these both cases:

```ruby
class Foo
  # Entering this block creates the singleton class and defines `bar` in it
  class << self
    def bar; end
  end

  # This also defines a method in the singleton class (which means that it needs to exist)
  def self.baz; end
end
```

PR #369, opened 2025-12-05
Discover and resolve method definitions with receiver

st0012:
Another step towards #79

### DIFF
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -108,46 +108,111 @@ pub fn resolve_all(graph: &mut Graph) {
     // Handle other definitions that don't require resolution, but need to have their declarations and membership
     // created
     for id in other_ids {
-        let definition = graph.definitions().get(&id).unwrap();
-        let mut lexical_nesting_id = *definition.lexical_nesting_id();
-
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -831,15 +831,30 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+        let receiver = if let Some(recv_node) = node.receiver() {
+            match recv_node {
+                // def self.foo - receiver is the current class/module's NameId
+                ruby_prism::Node::SelfNode { .. } => self.current_nesting_name_id(),
+                // def Foo.bar or def Foo::Bar.baz - receiver is the constant's NameId
+                ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. } => {
+                    self.index_constant_reference(&recv_node, true)
+                }
+                // Dynamic receiver (def foo.bar) - skip
+                _ => return,
```
Morriar: Shouldn't we visit the receiver in this case?

st0012: Updated 👍

PR #374, opened 2025-12-05
Unwrap singleton `attached_target`

Morriar:
We already return early is the `attached_target` is none, we can unwrap it.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -680,60 +680,49 @@ impl Visit<'_> for RubyIndexer<'_> {
             None
         };
 
-        if attached_target.is_none() {
+        let Some(attached_target) = attached_target else {
             return;
-        }
-
-        // Create SingletonClassDefinition if we have an attached_target
-        let definition_id = if let Some(attached_target) = attached_target {
-            let offset = Offset::from_prism_location(&node.location());
-            let comments = self.find_comments_for(offset.start()).unwrap_or_default();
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -653,58 +653,48 @@ impl Visit<'_> for RubyIndexer<'_> {
...
-                    .get(&attached_target)
-                    .expect("Attached target name should exist");
-                let target_str = self
-                    .local_graph
-                    .strings()
-                    .get(name.str())
-                    .expect("Attached target string should exist");
-                format!("<{target_str}>")
-            };
+        let attached_target = attached_target.unwrap();
```
st0012: I feel using `unwrap` here a bit risky because if the conditions above changed we could panic unexpectedly?

How about we change the condition above to:

```rs
        let Some(attached_target) = attached_target else {
            return;
        };
```

And then we don't need the `unwrap` anymore.

Morriar: In this case, shouldn't we crash hard so we can address the problem rather than silencing it?

st0012: It's not hiding the issue tho. It's just 

```rs
        let Some(attached_target) = attached_target else {
            return;
        };
```

Will be the combination of 

```rs
        if attached_target.is_none() {
            return;
        }
        attached_target.unwrap()
```

Morriar: Here you go 👍

PR #376, opened 2025-12-08
Handle mixins with `self`

Morriar:
So we can represent this:

```rb
module Foo
  include self # will error during resolution because of circular dependency
  prepend self # will error during resolution because of circular dependency
  extend self
end
```

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -463,6 +463,24 @@ impl<'a> RubyIndexer<'a> {
         }
     }
 
+    /// Returns the name ID of this definition
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition is not a nesting definition (class, module, or singleton class)
+    #[must_use]
+    fn parent_nesting_name_id(&mut self) -> Option<NameId> {
+        let parent_nesting = self.parent_nesting()?;
+
+        match parent_nesting {
+            Definition::Class(class) => Some(*class.name_id()),
+            Definition::SingletonClass(singleton_class) => Some(*singleton_class.name_id()),
+            Definition::Module(module) => Some(*module.name_id()),
+            Definition::Constant(constant) => Some(*constant.name_id()),
+            _ => panic!("Cannot get name ID of definition: {parent_nesting:?}"),
+        }
+    }
+
     fn handle_mixin(&mut self, node: &ruby_prism::CallNode, mixin_type: MixinType) {
         let Some(arguments) = node.arguments() else {
             return;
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -458,7 +458,15 @@ impl<'a> RubyIndexer<'a> {
         let reference_ids: Vec<_> = arguments
             .arguments()
             .iter()
-            .filter_map(|arg| self.index_constant_reference(&arg, true))
+            .filter_map(|arg| {
+                if arg.as_self_node().is_some() {
+                    // FIXME: Ideally we would want to save the mixin as `self` but we can only save mixins with a name.
+                    // We'll just use the parent nesting name id for now.
```
vinistock, line 496: Do we need to fix this though? It feels like it fits the rest of the implementation.

We'll detect the circular dependency during resolution.

rust/saturn/src/model/definitions.rs
```diff
@@ -110,6 +110,22 @@ impl Definition {
         }
     }
 
+    /// Returns the name ID of this definition
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition is not a nesting definition (class, module, or singleton class)
+    #[must_use]
+    pub fn name_id(&self) -> &NameId {
```
vinistock: Is this common practice in Rust? It's convenient, but it feels weird to me.

I also dislike having to destructure the enums everywhere, but we're moving the error from compilation to the runtime.

@jesse-shopify do you have any suggestions or opinions on best practices regarding this sort of enum situation?

Would our lives become easier if we had a nested enum for `ConstantDefinition` that could be a `Class`, `SingletonClass`, `Module` or `Constant`, so that we have a common type that shares what we need?

I don't know, just thinking out loud.

Morriar: Good point.

I think we can move it as a private method of the indexer instead. See also https://github.com/Shopify/saturn/pull/380.

> We're moving the error from compilation to the runtime.
> Would our lives become easier if we had a nested enum for ConstantDefinition that could be a Class, SingletonClass, Module or Constant, so that we have a common type that shares what we need?

Not that much in the context of the indexer because we stack `NameId`s to we have to do the `NameId -> Definition` dance and we can't guaranty we won't get something else than what we expect and we'll still need a runtime check.

vinistock: Yeah, that's fair, it's not a situation that will happen. My vote goes with a private method in the `RubyIndexer`.

We may end up with some duplication if more concepts need the same logic, but hopefully it won't happen much.

Another option might be using `pub(crate)` so that consumers to the crate can't call it.

PR #378, opened 2025-12-08
Keep track of descendents when linearizing ancestors

vinistock:
This PR starts keeping track of descendants as we linearize ancestors. Knowing descendants is useful for a few things:

- The LSP can show all descendants of a given type
- Linters can have rules that ban inheriting from a certain type (e.g.: a deprecated module or parent class)
- Collecting statistics about a codebase to understand the most used ancestors (higher number of descendants)

The idea of the implementation is to introduce a new parameter `descendant_id` to `linearize_ancestors`. As we recurse during linearization, we have the opportunity to remember the descendants.

### Caveat

At the moment, the implementation doesn't linearize ancestors that we don't have to linearize. They are only linearized if that is required to resolve a constant reference or if you invoke the API directly.

This is an issue for tracking descendants. You can only figure out all descendants by linearizing every single module and class. Unlike ancestor linearization, we wouldn't know _what do analyze_ ahead of time if we had an API like `descendants_of(declaration: DeclarationId)`.

I believe we have two options:

- Always linearize all ancestors. This may prove necessary in the future anyway. Especially when there's a certain amount of type checking going on. While this may sound wasteful at this point in time, I suspect it's probably the easiest path forward and the inevitable path we'll end up taking
- Introduce a `descendants_of` API that first ensures all ancestors have been linearized and then returns the result. This will be more performant right now, but might make the code more complex

### DIFF
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -1,7 +1,7 @@
-use std::cell::OnceCell;
+use std::cell::{OnceCell, Ref, RefCell};
 
 use crate::model::{
-    identity_maps::IdentityHashMap,
+    identity_maps::{IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, ReferenceId, StringId},
 };
 
@@ -24,6 +24,7 @@ pub struct Declaration {
     /// The ID of the owner of this declaration
     owner_id: DeclarationId,
     ancestors: OnceCell<Vec<DeclarationId>>,
+    descendants: RefCell<IdentityHashSet<DeclarationId>>,
 }
 
 impl Declaration {
@@ -36,6 +37,7 @@ impl Declaration {
             references: Vec::new(),
             owner_id,
             ancestors: OnceCell::new(),
+            descendants: RefCell::new(IdentityHashSet::default()),
         }
     }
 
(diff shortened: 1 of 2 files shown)


rust/saturn/src/model/declaration.rs
```diff
@@ -127,6 +129,14 @@ impl Declaration {
         self.ancestors.get().map(Vec::as_slice)
     }
 
+    pub fn add_descendant(&self, descendant_id: DeclarationId) {
+        self.descendants.borrow_mut().insert(descendant_id);
+    }
```
Morriar, line 134: Can't we use a `mut self` and avoid the `RefCell`? If I understand correctly the `RefCell` will be checked at runtime instead of compile time and we also pay the price of borrow checking?

vinistock: I couldn't find a way to structure the code that doesn't lead to borrow issues. This is why I used internal mutability on both ancestors and descendants.

The issue is that resolving constants requires an immutable borrow of the graph's data, but when we linearize ancestors we want to cache intermediate results to avoid doing the work again.

Especially when we introduce parallelism, I believe internal mutability with a `RwLock` will be inevitable because we want the workers to be able to reuse linearized results from other workers.

If you have ideas on other ways of structuring this, let me know, but I couldn't do it without a mess of borrow failures.

rust/saturn/src/resolution.rs
```diff
@@ -1149,6 +1178,32 @@ mod tests {
...
+
+            class Qux < Bar
+              CONST
+            end
+            "
+        });
+        context.resolve();
+
+        assert_descendants!(context, "Foo", [DeclarationId::from("Bar")]);
+        assert_descendants!(context, "Bar", [DeclarationId::from("Baz"), DeclarationId::from("Qux")]);
```
Morriar: Let's use the same format than `assert_ancestors_eq`:


```suggestion
        assert_descendants!(context, "Bar", ["Baz", "Qux"]);
```

PR #384, opened 2025-12-09
Simplify `indexing` errors handling by using a `Result`

Morriar:
More remnants of the database complexity.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1,7 +1,6 @@
 //! Visit the Ruby AST and create the definitions.
 
 use crate::diagnostic::Severity;
-use crate::errors::Errors;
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing.rs
```diff
@@ -96,9 +79,11 @@ where
...
     let mut all_errors = Vec::new();
     for result in rx {
-        let (local_graph, errors) = result;
-        graph.update(local_graph);
-        all_errors.extend(errors);
+        if let Ok(local_graph) = result {
+            graph.update(local_graph);
+        } else {
+            all_errors.push(result.unwrap_err());
+        }
```
vinistock: Nit: I'd use a match statement so that we can destructure both `Ok` and `Err` without needing to `unwrap_err`.

Morriar: Good point 👍

PR #385, opened 2025-12-10
Index `alias` for methods

Morriar:
First step towards https://github.com/Shopify/saturn/issues/348.

### DIFF
--- a/ext/saturn/definition.c
+++ b/ext/saturn/definition.c
@@ -19,6 +19,7 @@ VALUE cAttrWriterDefinition;
 VALUE cGlobalVariableDefinition;
 VALUE cInstanceVariableDefinition;
 VALUE cClassVariableDefinition;
+VALUE cMethodAliasDefinition;
 
 // Keep this in sync with definition.rs
 VALUE definition_class_for_kind(DefinitionKind kind) {
@@ -45,6 +46,8 @@ VALUE definition_class_for_kind(DefinitionKind kind) {
         return cInstanceVariableDefinition;
     case DefinitionKind_ClassVariable:
         return cClassVariableDefinition;
+    case DefinitionKind_MethodAlias:
+        return cMethodAliasDefinition;
     default:
         rb_raise(rb_eRuntimeError, "Unknown DefinitionKind: %d", kind);
     }
@@ -176,4 +179,8 @@ void initialize_definition(VALUE mod) {
     cClassVariableDefinition = rb_define_class_under(mSaturn, "ClassVariableDefinition", cDefinition);
     rb_define_alloc_func(cClassVariableDefinition, sr_handle_alloc);
     rb_define_method(cClassVariableDefinition, "initialize", sr_handle_initialize, 2);
+
+    cMethodAliasDefinition = rb_define_class_under(mSaturn, "MethodAliasDefinition", cDefinition);
+    rb_define_alloc_func(cMethodAliasDefinition, sr_handle_alloc);
+    rb_define_method(cMethodAliasDefinition, "initialize", sr_handle_initialize, 2);
 }
--- a/ext/saturn/definition.h
+++ b/ext/saturn/definition.h
@@ -16,6 +16,7 @@ extern VALUE cAttrWriterDefinition;
 extern VALUE cGlobalVariableDefinition;
 extern VALUE cInstanceVariableDefinition;
 extern VALUE cClassVariableDefinition;
+extern VALUE cMethodAliasDefinition;
 
 void initialize_definition(VALUE mSaturn);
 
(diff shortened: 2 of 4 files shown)


rust/saturn/src/model/definitions.rs
```diff
@@ -1010,3 +1013,75 @@ impl ClassVariableDefinition {
         &self.lexical_nesting_id
     }
 }
+
+#[derive(Debug)]
+pub struct AliasDefinition {
```
vinistock: Should this be `MethodAliasDefinition`?

I think that will be helpful to differentiate between the different types of aliases.

Especially considering that we will need different fields for a `ConstantAliasDefinition`, which will probably store its target's `NameId` instead of a `StringId`.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1296,8 +1296,24 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_alias_method_node(&mut self, node: &ruby_prism::AliasMethodNode<'_>) {
-        let name = Self::location_to_string(&node.old_name().location());
-        self.index_method_reference(name, &node.old_name().location());
+        let new_name = Self::location_to_string(&node.new_name().location());
+        let old_name = Self::location_to_string(&node.old_name().location());
+        let offset = Offset::from_prism_location(&node.location());
```
vinistock, line 1287: I think the location I would associate with the alias definition would be just the one for the new name, instead of the entire thing.

What do you think?

```suggestion
        let new_name_loc = node.new_name().location();
        let new_name = Self::location_to_string(&new_name_loc);
        let old_name = Self::location_to_string(&node.old_name().location());
        let offset = Offset::from_prism_location(&new_name_loc);
```

Morriar: I'd expect it to behave the same way it does for a class or a method and include the whole declaration.

If we were to introduce a concept akin the `selectionRange` from LSP, then I could see the new_name location being used for it.

rust/saturn/src/model/definitions.rs
Morriar: In https://github.com/Shopify/saturn/pull/386 I was hoping to reuse the same alias structure for globals.

In https://github.com/Shopify/saturn/pull/387 I hesitated reusing the same definition for `alias_method`.

vinistock: I think it might work for globals and methods since both would only store a `StringId`, but I suspect that's not the case for constants. E.g.:

```ruby
module Foo
end

# I think we need both sides to be a NameId since we'll already have
# indexed `Foo` as a reference and need to resolve it anyway
A = Foo
```

We could reuse the same thing for methods and globals, but then what do we name it? `ConstantAlias` and just `Alias`? @st0012 do you have thoughts on this one?

rust/saturn/src/indexing/ruby_indexer.rs
vinistock: Yeah, the question comes down to what is a part of the definition. For a class, the entire body is a part of it. But for an alias, is the old name part of the definition?

For example consider this:

```ruby
class Foo
  attr_reader :bar, :baz, :qux
end
```

If I was jumping to definition on `baz`, I'd want to jump to the position of the `baz` name and not the entire attribute. Similarly, I think it could make sense to make the location include only the new name.

@st0012 what do you think?

st0012: Since we can't alias multiple methods/global variables at the same time (can we?), I think it makes sense to consider the whole line as definition.
IMO `attr_*` is a different case because it creates multiple methods in the same line's declaration, so it makes sense to target specific symbols.

rust/saturn/src/model/definitions.rs
st0012: It'd go with having more specific definition structs for now. It's not a strong opinion but I feel it'd make counting stats and asserting in tests a bit more easier.
It shouldn't be too hard to consolidate them later if we want to.

rust/saturn/src/resolution.rs
```diff
@@ -182,6 +182,10 @@ fn extract_owner_strategy(definition: &Definition) -> (StringId, OwnerResolution
         Definition::AttrWriter(it) => (*it.str_id(), OwnerResolution::LexicalNesting(*it.lexical_nesting_id())),
         Definition::GlobalVariable(it) => (*it.str_id(), OwnerResolution::LexicalNesting(*it.lexical_nesting_id())),
         Definition::InstanceVariable(it) => (*it.str_id(), OwnerResolution::LexicalNesting(*it.lexical_nesting_id())),
+        Definition::AliasMethod(it) => (
+            *it.new_name_str_id(),
+            OwnerResolution::LexicalNesting(*it.lexical_nesting_id()),
+        ),
```
st0012: Do we want to have a test case for resolution too?

Morriar: Added 👍

PR #387, opened 2025-12-10
Index `alias_method` as an `AliasDefinition`

Morriar:
Please review https://github.com/Shopify/saturn/pull/385 and https://github.com/Shopify/saturn/pull/386 first.

We reuse the `AliasDefinition` definition to index `alias_method` calls:

```rb
alias_method :foo, :bar
```

Note that calls not using a symbol literal are ignored for now:

```rb
alias_method ignored, :bar
```

This closes https://github.com/Shopify/saturn/issues/348.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -1294,8 +1294,18 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_alias_method_node(&mut self, node: &ruby_prism::AliasMethodNode<'_>) {
-        let new_name = Self::location_to_string(&node.new_name().location());
-        let old_name = Self::location_to_string(&node.old_name().location());
+        let new_name = if let Some(symbol_node) = node.new_name().as_symbol_node() {
+            Self::location_to_string(&symbol_node.value_loc().unwrap())
+        } else {
+            Self::location_to_string(&node.new_name().location())
+        };
+
+        let old_name = if let Some(symbol_node) = node.old_name().as_symbol_node() {
+            Self::location_to_string(&symbol_node.value_loc().unwrap())
+        } else {
+            Self::location_to_string(&node.old_name().location())
+        };
+
         let offset = Offset::from_prism_location(&node.location());
         let comments = self.find_comments_for(offset.start()).unwrap_or_default();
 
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -3710,6 +3738,47 @@ mod tests {
         });
     }
 
+    #[test]
+    fn index_module_alias_method() {
+        let context = index_source({
+            "
+            alias_method :foo, :bar
```
vinistock: This one actually crashes with undefined method `alias_method` for `<main>`.

Should we ignore it instead?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -3710,6 +3738,47 @@ mod tests {
...
+
+            class Foo
+              alias_method :baz, :qux
```
vinistock, line 3753: For completeness sake, `alias_method` also accepts strings.

We have the `each_string_or_symbol_arg` helper to loop through those.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Let's error in the resolution phase instead so the definition exists.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Good call, I forgot about those, updated and added a test 👍

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -3703,11 +3735,60 @@ mod tests {
...
+            alias_method \"foo_string\", \"bar_string\"
+
+            class Foo
+              alias_method :baz, :qux
+            end
+
+            alias_method :baz, ignored
+            alias_method ignored, :qux
+            alias_method ignored, ignored
+            "
```
st0012, line 3759: Not blocking, but we can avoid escaping double quotes with:

```suggestion
            r#"
            alias_method :foo_symbol, :bar_symbol
            alias_method "foo_string", "bar_string"

            class Foo
              alias_method :baz, :qux
            end

            alias_method :baz, ignored
            alias_method ignored, :qux
            alias_method ignored, ignored
            "#
```

And to avoid escaping `#` (like we have `"#{foo}"`), we can also stack `#` like `r####`.

st0012: https://github.com/Shopify/saturn/pull/396

PR #388, opened 2025-12-10
Specialize declarations

vinistock:
Specialize declarations with an enum, so that we know the type of declaration that got assigned to a fully qualified name.

Notable decisions on this PR for feedback:

1. I used `Method` for all methods and attributes. I believe this is the right choice because you could have the same name defined twice, one as a method and another as an attribute. We still remember the definitions, so it's easy to discover the "origin" of the type, but ultimately they are all methods
2. I divided declarations into two categories: namespaces and simple. The only reason for this is avoiding a ton of boilerplate code. If we need to untangle that later, we can do it
3. I made methods that create declarations accept a closure. This is because if a declaration already exists for a name, we only extend it. The closure lazily creates the declaration, which allows us to save an allocation when creating it is not required
4. I removed the `OwnerStrategy` concept and moved all processing in a big match statement. Most of the related bits are still there, but having everything in the match statement allowed me to avoid borrow issues while still making use of the definition type

### DIFF
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -4,7 +4,10 @@ use std::{
 };
 
 use crate::model::{
-    declaration::Declaration,
+    declaration::{
+        ClassDeclaration, ClassVariableDeclaration, ConstantDeclaration, Declaration, GlobalVariableDeclaration,
+        InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration, SingletonClassDeclaration,
+    },
     definitions::Definition,
     graph::{Graph, OBJECT_ID},
     identity_maps::{IdentityHashMap, IdentityHashSet},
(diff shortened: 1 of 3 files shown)


rust/saturn/src/resolution.rs
```diff
@@ -465,12 +539,21 @@ fn search_lexical_scopes(graph: &Graph, name: &Name, str_id: StringId) -> Outcom
...
     let object = graph.declarations().get(&OBJECT_ID).unwrap();
 
-    match object.members().get(&str_id) {
+    match members(object).get(&str_id) {
         Some(member_id) => Outcome::Resolved(*member_id),
         None => Outcome::Unresolved,
     }
 }
 
+fn members(decl: &Declaration) -> &IdentityHashMap<StringId, DeclarationId> {
```
st0012: nit: I think the name can be `members_of`

PR #390, opened 2025-12-11
Loop through all resolution units to decide when to stop

vinistock:
I uncovered a bug while trying to linearized prepended modules. Our resolution loop structure was incorrect.

We can only decide to stop processing constants if we looped through all items (excluding retries added in the current pass). If after trying to process all items in the queue we made no progress, **then** we stop.

Right now, the code stops immediately at the first retry, which simply won't work (especially considering more complex inheritance cases).

I added a test that breaks without the fix. It's a good example of inheritance causing resolution to cut the boundaries of the sorting partition:

```ruby
module Foo
  module Bar
    class Baz
      CONST = 123
    end
  end
end

class Qux < Foo::Bar::Baz
  # This reference is sorted first in the list of units than Qux's parent class. When we
  # find this, we will try to linearize the ancestors of `Qux`, but we won't have resolved
  # `Foo::Bar::Baz` yet - leading to a retry.
  #
  # Only by going through the list of units again will we succeed in resolving this because
  # it's in the second pass that `Foo::Bar::Baz` is resolved, which allows us to linearize
  # the ancestors of `Qux` and finally resolved `CONST`
  CONST
end
```

### DIFF
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -66,57 +66,64 @@ pub fn resolve_all(graph: &mut Graph) {
         .declarations_mut()
         .insert(*OBJECT_ID, Declaration::new("Object".to_string(), *OBJECT_ID));
 
-    let (mut constant_queue, other_ids) = sorted_units(graph);
-    // Flag to indicate whether we successfully resolved something in the last resolution iteration, thus making
-    // progress. If nothing got resolved, then we stop because trying again would likely not successfully resolve
-    // anything new
-    let mut making_progress = true;
+    let (mut unit_queue, other_ids) = sorted_units(graph);
 
-    // First, handle all constants, ensuring that we create declarations, members, owners and resolve any name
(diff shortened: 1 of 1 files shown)


rust/saturn/src/resolution.rs
```diff
@@ -66,58 +66,63 @@ pub fn resolve_all(graph: &mut Graph) {
...
-                        making_progress = false;
+    let (mut unit_queue, other_ids) = sorted_units(graph);
+
+    loop {
+        // Flag to ensure the end of the resolution loop. We go through all items in the queue based on its current
+        // length. If we made any progress in this pass of the queue, we can continue because we're unlocking more work
+        // to be done
+        let mut made_progress = false;
+
+        for _ in 0..unit_queue.len() {
```
st0012, line 79: Can you add a comment to explicitly state that `unit_queue.len()` marks the boundary of the current work items list and it does NOT change as we push retry items back to the same queue.

Issue #354, opened 2025-12-03
Introduce resolution diagnostics


vinistock:
Now that we have a place to store diagnostics, let's add warnings/errors for all of the possible resolution failures. Things like:

- [ ] Something was a module and then became a class (https://github.com/Shopify/rubydex/pull/474)
- [ ] A class declares two different parents in its definitions (https://github.com/Shopify/rubydex/pull/474)
- [ ] A class or module contains a cyclic dependency (e.g.: a class that inherits from itself) (https://github.com/Shopify/rubydex/pull/474)
- [ ] Ambiguous definitions due to parent scope resolution
- [ ] Tried to include/prepend a class or inherit from a module (https://github.com/Shopify/rubydex/pull/474)
- [ ] Privacy violations for accessing private constants outside of their namespace
- [ ] Class variables definitions at the top level (crashes in the runtime)

As part of this work, I would really love for us to design Rust-like help messages that assist the developer in understand _why_ the diagnostic exists. Maybe something like an expanded description, with examples and information that can be used to better understand what the error is saying.

PR #391, opened 2025-12-12
Complete ancestor linearization

### LINKED ISSUE #196: Implement constant resolution
- [x] #197 
- [x] #198
- [x] Resolve constants in the inheritance chain
- [ ] Resolve aliased constants
- [x] Resolve constants under singleton classes
    - This is closely related how we implement https://github.com/Shopify/index/issues/79 because of cases like
        ```rb
        class Foo
          class << self
            A = 10
        
            def get_a
              A # this works
            end
          end
          
          def self.get_a
            A # this doesn't work
          end
        end
        ```

### LINKED ISSUE #200: Design resolution phase

### LINKED ISSUE #345: Handle prepended module linearization
After #339, we'll have parent class linearization, but we're not yet handling module prepends. Add prepend linearization by collecting all prepends discovered in every definition and recursively linearizing the module's own ancestors.

### LINKED ISSUE #346: Handle included module linearization
Similar to #345, but for included modules. One major caveat of included modules is that they are deduped if they have already been prepended before.

For example:

```ruby
module Foo; end

class A
  include Foo
  prepend Foo
end

A.ancestors # => [Foo, A, Foo, Object, Kernel, BasicObject]

class B
  prepend Foo
  include Foo
end

# Foo doesn't appear twice in this case
B.ancestors # => [Foo, B, Object, Kernel, BasicObject]
```

We should also ensure we have tests verifying edge cases, like including a module that got prepended through another module.

```ruby
module Foo
end

module Bar
  include Foo
end

class A
  prepend Bar

  # Is this one deduped too?
  include Foo
end
```

vinistock:
Closes #345, closes #346, closes #196, closes #200

This PR completes ancestor linearization, handling prepends, includes, extends and singleton class parents.

### Implementation notes

- I put declarations inside a `RwLock`, so that we can lazily create singleton classes without getting into borrow checker issues
- I added a dedicated field to keep track of singleton classes instead of making them members, which is a little more convenient
- I added a bunch of `Ancestor` related structures to assist with keeping track of the information
- Descendants are also tracked as part of the implementation
- I started manually adding the `Class` and `Module` classes while we don't have RBS indexing

### DIFF
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -162,18 +162,13 @@ macro_rules! namespace_declaration {
             pub fn descendants(&self) -> Ref<'_, IdentityHashSet<DeclarationId>> {
                 self.descendants.borrow()
             }
-
-            #[must_use]
-            pub fn kind(&self) -> DeclarationKind {
-                DeclarationKind::$variant
-            }
         }
     };
 }
 
 /// Macro to generate a new struct for simple declarations like variables and methods
 macro_rules! simple_declaration {
-    ($variant:ident, $name:ident) => {
+    ($name:ident) => {
         #[derive(Debug)]
         pub struct $name {
             /// The fully qualified name of this declaration
@@ -198,24 +193,8 @@ macro_rules! simple_declaration {
             }
 
             pub fn extend(&mut self, _other: Declaration) {}
-
-            #[must_use]
-            pub fn kind(&self) -> DeclarationKind {
-                DeclarationKind::$variant
-            }
         }
     };
-}
-
-pub enum DeclarationKind {
-    Class,
-    SingletonClass,
-    Module,
-    Constant,
-    Method,
-    GlobalVariable,
-    InstanceVariable,
-    ClassVariable,
 }
 
 /// A `Declaration` represents the global concept of an entity in Ruby. For example, the class `Foo` may be defined 3
@@ -305,21 +284,16 @@ impl Declaration {
     pub fn unqualified_name(&self) -> String {
         all_declarations!(self, it => it.name.rsplit("::").next().unwrap_or(&it.name).to_string())
     }
-
-    #[must_use]
-    pub fn kind(&self) -> DeclarationKind {
-        all_declarations!(self, it => it.kind())
-    }
 }
 
 namespace_declaration!(Class, ClassDeclaration);
 namespace_declaration!(Module, ModuleDeclaration);
 namespace_declaration!(SingletonClass, SingletonClassDeclaration);
-simple_declaration!(Constant, ConstantDeclaration);
-simple_declaration!(Method, MethodDeclaration);
-simple_declaration!(GlobalVariable, GlobalVariableDeclaration);
-simple_declaration!(InstanceVariable, InstanceVariableDeclaration);
-simple_declaration!(ClassVariable, ClassVariableDeclaration);
+simple_declaration!(ConstantDeclaration);
+simple_declaration!(MethodDeclaration);
+simple_declaration!(GlobalVariableDeclaration);
+simple_declaration!(InstanceVariableDeclaration);
+simple_declaration!(ClassVariableDeclaration);
 
 #[cfg(test)]
 mod tests {
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -7,7 +7,7 @@ use std::{
 use crate::model::{
     declaration::{
         Ancestor, Ancestors, ClassDeclaration, ClassVariableDeclaration, ConstantDeclaration, Declaration,
-        DeclarationKind, GlobalVariableDeclaration, InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration,
+        GlobalVariableDeclaration, InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration,
         SingletonClassDeclaration,
     },
     definitions::{Definition, Mixin},
@@ -325,13 +325,13 @@ fn handle_remaining_definitions(
 
                             // Instance variable in singleton method - owned by the receiver's singleton class
                             let owner_id = get_or_create_singleton_class(graph, receiver_decl_id);
-                            debug_assert!(
-                                matches!(
-                                    graph.declarations().get(&owner_id),
-                                    Some(Declaration::SingletonClass(_))
-                                ),
-                                "Instance variable in singleton method should be owned by a SingletonClass"
-                            );
+                            {
+                                let declarations = graph.declarations().read().unwrap();
+                                debug_assert!(
+                                    matches!(declarations.get(&owner_id), Some(Declaration::SingletonClass(_))),
+                                    "Instance variable in singleton method should be owned by a SingletonClass"
+                                );
+                            }
                             create_declaration(graph, str_id, id, owner_id, |name| {
                                 Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(
                                     name, owner_id,
@@ -345,9 +345,11 @@ fn handle_remaining_definitions(
 
                         // If the method is in a singleton class, the instance variable belongs to the class object
                         // Like `class << Foo; def bar; @bar = 1; end; end`, where `@bar` is owned by `Foo::<Foo>`
-                        if let Some(decl) = graph.declarations().get(&method_owner_id)
+                        let declarations = graph.declarations().read().unwrap();
+                        if let Some(decl) = declarations.get(&method_owner_id)
                             && matches!(decl, Declaration::SingletonClass(_))
                         {
+                            drop(declarations);
                             // Method in singleton class - owner is the singleton class itself
                             create_declaration(graph, str_id, id, method_owner_id, |name| {
                                 Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(
@@ -356,6 +358,7 @@ fn handle_remaining_definitions(
                                 )))
                             });
                         } else {
+                            drop(declarations);
                             // Regular instance method
                             // Create an instance variable declaration for the method's owner
                             create_declaration(graph, str_id, id, method_owner_id, |name| {
@@ -375,13 +378,13 @@ fn handle_remaining_definitions(
                             .copied()
                             .unwrap_or(*OBJECT_ID);
                         let owner_id = get_or_create_singleton_class(graph, nesting_decl_id);
-                        debug_assert!(
-                            matches!(
-                                graph.declarations().get(&owner_id),
-                                Some(Declaration::SingletonClass(_))
-                            ),
-                            "Instance variable in class/module body should be owned by a SingletonClass"
-                        );
+                        {
+                            let declarations = graph.declarations().read().unwrap();
+                            debug_assert!(
+                                matches!(declarations.get(&owner_id), Some(Declaration::SingletonClass(_))),
+                                "Instance variable in class/module body should be owned by a SingletonClass"
+                            );
+                        }
                         create_declaration(graph, str_id, id, owner_id, |name| {
                             Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(name, owner_id)))
                         });
@@ -395,13 +398,13 @@ fn handle_remaining_definitions(
                             .copied()
                             .unwrap_or(*OBJECT_ID);
                         let owner_id = get_or_create_singleton_class(graph, singleton_class_decl_id);
-                        debug_assert!(
-                            matches!(
-                                graph.declarations().get(&owner_id),
-                                Some(Declaration::SingletonClass(_))
-                            ),
-                            "Instance variable in singleton class body should be owned by a SingletonClass"
-                        );
+                        {
+                            let declarations = graph.declarations().read().unwrap();
+                            debug_assert!(
+                                matches!(declarations.get(&owner_id), Some(Declaration::SingletonClass(_))),
+                                "Instance variable in singleton class body should be owned by a SingletonClass"
+                            );
+                        }
                         create_declaration(graph, str_id, id, owner_id, |name| {
                             Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(name, owner_id)))
                         });
@@ -538,7 +541,7 @@ pub fn ancestors_of(graph: &Graph, declaration_id: DeclarationId) -> Ancestors {
 /// Can panic if there's inconsistent data in the graph
 #[must_use]
 fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &mut LinearizationContext) -> Ancestors {
-    let (kind, owner_id) = {
+    {
         let declarations = graph.declarations().read().unwrap();
         let declaration = declarations.get(&declaration_id).unwrap();
 
@@ -583,13 +586,11 @@ fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &m
             context.descendants.remove(&declaration_id);
             return estimated_ancestors;
         }
-
-        (declaration.kind(), *declaration.owner_id())
-    };
+    }
 
     // TODO: this check is against `Object` for now to avoid infinite recursion. After RBS indexing, we need to change
     // this to `BasicObject` since it's the only class that cannot have a parent
-    let parent_ancestors = linearize_parent_ancestors(graph, declaration_id, context, &kind, owner_id);
+    let parent_ancestors = linearize_parent_ancestors(graph, declaration_id, context);
 
     let declarations = graph.declarations().read().unwrap();
     let declaration = declarations.get(&declaration_id).unwrap();
@@ -603,7 +604,7 @@ fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &m
 
     // If we're linearizing a singleton class, add the extends of the attached class to the list of mixins to process
     if let Declaration::SingletonClass(_) = declaration {
-        let attached_decl = declarations.get(&owner_id).unwrap();
+        let attached_decl = declarations.get(declaration.owner_id()).unwrap();
 
         let attached_definitions = attached_decl
             .definitions()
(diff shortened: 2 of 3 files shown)


rust/saturn/src/resolution.rs
```diff
@@ -334,33 +411,51 @@ pub fn ancestors_of(graph: &Graph, declaration_id: DeclarationId) -> AncestorOut
 /// # Panics
 ///
 /// Can panic if there's inconsistent data in the graph
+#[allow(clippy::too_many_lines)]
```
vinistock: I disabled the too many lines here temporarily because it's a bit early to refactor. When we linearize included modules, I suspect that deduplication can only work if we track the order in which we find prepend and include operations.

Extracting anything now will just make the next step more difficult, so the next PR should remove this lint.

rust/saturn/src/model/graph.rs
```diff
@@ -215,6 +215,9 @@ impl Graph {
                 Declaration::Class(it) => it.add_member(member_str_id, member_declaration_id),
                 Declaration::Module(it) => it.add_member(member_str_id, member_declaration_id),
                 Declaration::SingletonClass(it) => it.add_member(member_str_id, member_declaration_id),
+                Declaration::Constant(_) => {
+                    // TODO: temporary hack to avoid crashing on `Struct.new`, `Class.new` and `Module.new`
```
Morriar, line 221: Let's not forget about `Data.define`

rust/saturn/src/model/declaration.rs
```diff
@@ -1,10 +1,78 @@
...
+    pub fn to_partial(self) -> Self {
+        match self {
+            Ancestors::Complete(ancestors) | Ancestors::Cyclic(ancestors) | Ancestors::Partial(ancestors) => {
+                Ancestors::Partial(ancestors)
+            }
+        }
+    }
+
+    #[must_use]
+    pub fn filter_map<F>(self, f: F) -> Self
```
st0012: This doesn't seem to be used. Do we keep it just for clients?

rust/saturn/src/resolution.rs
```diff
@@ -679,13 +1044,52 @@ fn sorted_units(graph: &Graph) -> (VecDeque<Unit>, Vec<DefinitionId>) {
...
+                decl.definitions()
+                    .iter()
+                    .filter_map(|def_id| graph.definitions().get(def_id))
+                    .collect::<Vec<_>>()
+            };
+
+            let (picked_parent, partial) = get_parent_class(graph, &definitions);
+            (get_or_create_singleton_class(graph, picked_parent), partial)
+        }
+    }
```
st0012: Not really a suggestion, but I'm wondering what would be the impact if we hold the lock for a bit longer here and just match against `decl` directly, like:


```suggestion
    let declarations = graph.declarations().read().unwrap();
    let decl = declarations.get(&attached_id).unwrap();

    match decl {
        Declaration::Module(_) => (*MODULE_ID, false),
        Declaration::SingletonClass(_) => {
            // For singleton classes, we keep recursively wrapping parents until we can reach the original attached
            // object
            let owner_id = *decl.owner_id();
            drop(declarations);
            let (inner_parent, partial) = singleton_parent_id(graph, owner_id);
            (get_or_create_singleton_class(graph, inner_parent), partial)
        }
        Declaration::Class(_) => {
            // For classes (the regular case), we need to return the singleton class of its parent
            let definitions: Vec<_> = decl
                .definitions()
                .iter()
                .filter_map(|def_id| graph.definitions().get(def_id))
                .collect();
            drop(declarations);

            let (picked_parent, partial) = get_parent_class(graph, &definitions);
            (get_or_create_singleton_class(graph, picked_parent), partial)
        }
        _ => {
            // Other declaration types (constants, methods, etc.) shouldn't reach here,
            // but default to Object's singleton parent
            drop(declarations);
            (*CLASS_ID, false)
        }
    }
```

rust/saturn/src/model/declaration.rs
```diff
@@ -138,10 +218,26 @@ macro_rules! simple_declaration {
...
+
+            #[must_use]
+            pub fn kind(&self) -> DeclarationKind {
+                DeclarationKind::$variant
+            }
         }
     };
 }
 
+pub enum DeclarationKind {
```
st0012: I don't like the idea of adding a new `Kind` enum just to pass data outside of the lock. I think it's possible to release locks explicitly and just use `Declaration` structs in match cases.

This is what I attempted and it passes all tests:

```diff
diff --git a/rust/saturn/src/model/declaration.rs b/rust/saturn/src/model/declaration.rs
index bf33e29..7d40698 100644
--- a/rust/saturn/src/model/declaration.rs
+++ b/rust/saturn/src/model/declaration.rs
@@ -182,18 +182,13 @@ macro_rules! namespace_declaration {
             pub fn descendants(&self) -> Ref<'_, IdentityHashSet<DeclarationId>> {
                 self.descendants.borrow()
             }
-
-            #[must_use]
-            pub fn kind(&self) -> DeclarationKind {
-                DeclarationKind::$variant
-            }
         }
     };
 }
 
 /// Macro to generate a new struct for simple declarations like variables and methods
 macro_rules! simple_declaration {
-    ($variant:ident, $name:ident) => {
+    ($name:ident) => {
         #[derive(Debug)]
         pub struct $name {
             /// The fully qualified name of this declaration
@@ -218,26 +213,10 @@ macro_rules! simple_declaration {
             }
 
             pub fn extend(&mut self, _other: Declaration) {}
-
-            #[must_use]
-            pub fn kind(&self) -> DeclarationKind {
-                DeclarationKind::$variant
-            }
         }
     };
 }
 
-pub enum DeclarationKind {
-    Class,
-    SingletonClass,
-    Module,
-    Constant,
-    Method,
-    GlobalVariable,
-    InstanceVariable,
-    ClassVariable,
-}
-
 /// A `Declaration` represents the global concept of an entity in Ruby. For example, the class `Foo` may be defined 3
 /// times in different files and the `Foo` declaration is the combination of all of those definitions that contribute to
 /// the same fully qualified name
@@ -325,21 +304,16 @@ impl Declaration {
     pub fn unqualified_name(&self) -> String {
         all_declarations!(self, it => it.name.rsplit("::").next().unwrap_or(&it.name).to_string())
     }
-
-    #[must_use]
-    pub fn kind(&self) -> DeclarationKind {
-        all_declarations!(self, it => it.kind())
-    }
 }
 
 namespace_declaration!(Class, ClassDeclaration);
 namespace_declaration!(Module, ModuleDeclaration);
 namespace_declaration!(SingletonClass, SingletonClassDeclaration);
-simple_declaration!(Constant, ConstantDeclaration);
-simple_declaration!(Method, MethodDeclaration);
-simple_declaration!(GlobalVariable, GlobalVariableDeclaration);
-simple_declaration!(InstanceVariable, InstanceVariableDeclaration);
-simple_declaration!(ClassVariable, ClassVariableDeclaration);
+simple_declaration!(ConstantDeclaration);
+simple_declaration!(MethodDeclaration);
+simple_declaration!(GlobalVariableDeclaration);
+simple_declaration!(InstanceVariableDeclaration);
+simple_declaration!(ClassVariableDeclaration);
 
 #[cfg(test)]
 mod tests {
diff --git a/rust/saturn/src/resolution.rs b/rust/saturn/src/resolution.rs
index d41a28c..d19797a 100644
--- a/rust/saturn/src/resolution.rs
+++ b/rust/saturn/src/resolution.rs
@@ -7,7 +7,7 @@ use std::{
 use crate::model::{
     declaration::{
         Ancestor, Ancestors, ClassDeclaration, ClassVariableDeclaration, ConstantDeclaration, Declaration,
-        DeclarationKind, GlobalVariableDeclaration, InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration,
+        GlobalVariableDeclaration, InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration,
         SingletonClassDeclaration,
     },
     definitions::{Definition, Mixin},
@@ -431,7 +431,7 @@ pub fn ancestors_of(graph: &Graph, declaration_id: DeclarationId) -> Ancestors {
 /// Can panic if there's inconsistent data in the graph
 #[must_use]
 fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &mut LinearizationContext) -> Ancestors {
-    let (kind, owner_id) = {
+    {
         let declarations = graph.declarations().read().unwrap();
         let declaration = declarations.get(&declaration_id).unwrap();
 
@@ -476,13 +476,11 @@ fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &m
             context.descendants.remove(&declaration_id);
             return estimated_ancestors;
         }
-
-        (declaration.kind(), *declaration.owner_id())
-    };
+    }
 
     // TODO: this check is against `Object` for now to avoid infinite recursion. After RBS indexing, we need to change
     // this to `BasicObject` since it's the only class that cannot have a parent
-    let parent_ancestors = linearize_parent_ancestors(graph, declaration_id, context, &kind, owner_id);
+    let parent_ancestors = linearize_parent_ancestors(graph, declaration_id, context);
 
     let declarations = graph.declarations().read().unwrap();
     let declaration = declarations.get(&declaration_id).unwrap();
@@ -496,7 +494,7 @@ fn linearize_ancestors(graph: &Graph, declaration_id: DeclarationId, context: &m
 
     // If we're linearizing a singleton class, add the extends of the attached class to the list of mixins to process
     if let Declaration::SingletonClass(_) = declaration {
-        let attached_decl = declarations.get(&owner_id).unwrap();
+        let attached_decl = declarations.get(declaration.owner_id()).unwrap();
 
         let attached_definitions = attached_decl
             .definitions()
@@ -551,18 +549,23 @@ fn linearize_parent_ancestors(
     graph: &Graph,
     declaration_id: DeclarationId,
     context: &mut LinearizationContext,
-    kind: &DeclarationKind,
-    owner_id: DeclarationId,
 ) -> Option<Vec<Ancestor>> {
-    match kind {
-        DeclarationKind::Class if declaration_id != *OBJECT_ID => {
-            let declarations = graph.declarations().read().unwrap();
-            let declaration = declarations.get(&declaration_id).unwrap();
-            let definitions = declaration
+    if declaration_id == *OBJECT_ID {
+        return None;
+    }
+
+    let declarations = graph.declarations().read().unwrap();
+    let declaration = declarations.get(&declaration_id).unwrap();
+
+    match declaration {
+        Declaration::Class(_) => {
+            let definitions: Vec<_> = declaration
                 .definitions()
                 .iter()
                 .filter_map(|def_id| graph.definitions().get(def_id))
-                .collect::<Vec<_>>();
+                .collect();
+            // Release read lock before calling functions that may acquire write locks
+            drop(declarations);
 
             Some(match linearize_parent_class(graph, &definitions, context) {
                 Ancestors::Complete(ids) => ids,
@@ -576,7 +579,11 @@ fn linearize_parent_ancestors(
                 }
             })
         }
-        DeclarationKind::SingletonClass if declaration_id != *OBJECT_ID => {
+        Declaration::SingletonClass(_) => {
+            let owner_id = *declaration.owner_id();
+            // Release read lock before calling functions that may acquire write locks
+            drop(declarations);
+
             let (singleton_parent_id, partial_singleton) = singleton_parent_id(graph, owner_id);
             if partial_singleton {
                 context.partial = true;
@@ -1056,34 +1063,39 @@ fn singleton_parent_id(graph: &Graph, attached_id: DeclarationId) -> (Declaratio
         return (*CLASS_ID, false);
     }
 
-    let (kind, owner_id) = {
-        let declarations = graph.declarations().read().unwrap();
-        let decl = declarations.get(&attached_id).unwrap();
-        (decl.kind(), *decl.owner_id())
-    };
+    let declarations = graph.declarations().read().unwrap();
+    let decl = declarations.get(&attached_id).unwrap();
 
-    match kind {
-        DeclarationKind::Module => (*MODULE_ID, false),
-        DeclarationKind::SingletonClass => {
+    match decl {
+        Declaration::Module(_) => (*MODULE_ID, false),
+        Declaration::SingletonClass(_) => {
             // For singleton classes, we keep recursively wrapping parents until we can reach the original attached
             // object
+            let owner_id = *decl.owner_id();
+            // Release read lock before calling functions that may acquire write locks
+            drop(declarations);
             let (inner_parent, partial) = singleton_parent_id(graph, owner_id);
             (get_or_create_singleton_class(graph, inner_parent), partial)
         }
-        _ => {
+        Declaration::Class(_) => {
             // For classes (the regular case), we need to return the singleton class of its parent
-            let definitions = {
-                let declarations = graph.declarations().read().unwrap();
-                let decl = declarations.get(&attached_id).unwrap();
-                decl.definitions()
-                    .iter()
-                    .filter_map(|def_id| graph.definitions().get(def_id))
-                    .collect::<Vec<_>>()
-            };
+            let definitions: Vec<_> = decl
+                .definitions()
+                .iter()
+                .filter_map(|def_id| graph.definitions().get(def_id))
+                .collect();
+            // Release read lock before calling functions that may acquire write locks
+            drop(declarations);
 
             let (picked_parent, partial) = get_parent_class(graph, &definitions);
             (get_or_create_singleton_class(graph, picked_parent), partial)
         }
+        _ => {
+            // Other declaration types (constants, methods, etc.) shouldn't reach here,
+            // but default to Object's singleton parent
+            drop(declarations);
+            (*CLASS_ID, false)
+        }
     }
 }
 
```

rust/saturn/src/model/declaration.rs
vinistock: Good catch. No, this is leftovers from several attempts to make it work.

rust/saturn/src/resolution.rs
vinistock: This results in a dead lock because `get_or_create_singleton_class` requires a write lock and you can't acquire one while there are still read locks active.

This is why I structured the code to read and copy what's necessary and then trigger the writes.

rust/saturn/src/model/declaration.rs
vinistock: This is great! Thanks

test/graph_test.rb
```diff
@@ -94,9 +94,9 @@ def test_list_all_declarations_enumerator
 
       enumerator = graph.declarations
 
-      assert_equal(3, enumerator.size)
-      assert_equal(3, enumerator.count)
-      assert_equal(3, enumerator.to_a.size)
+      assert_equal(5, enumerator.size)
```
Morriar, line 98: Let's add a comment explaining what are the 5 expected declarations here.

rust/saturn/src/main.rs
```diff
@@ -86,7 +86,8 @@ fn main() -> Result<(), Box<dyn Error>> {
         println!("{}", dot::generate(&graph));
     } else {
         println!("Indexed {} files", graph.documents().len());
-        println!("Found {} names", graph.declarations().len());
+        let read_lock = graph.declarations().read().unwrap();
+        println!("Found {} names", read_lock.len());
```
Morriar, line 90: Would that work?

```suggestion
        println!("Found {} names", graph.declarations().read().unwrap().len());
```

rust/saturn/src/model/declaration.rs
```diff
@@ -162,18 +162,13 @@ macro_rules! namespace_declaration {
             pub fn descendants(&self) -> Ref<'_, IdentityHashSet<DeclarationId>> {
                 self.descendants.borrow()
             }
-
-            #[must_use]
-            pub fn kind(&self) -> DeclarationKind {
```
Morriar: Without this, do we still have a way to compare that two declarations have the same kind?

For example when trying to produce diagnostics for this:

```rb
module Foo; end
class Foo; end # => error: `Foo` was previously declared as a `module`
```

rust/saturn/src/model/graph.rs
vinistock: It's all listed in the issue #392, I just didn't put it in all comments.

rust/saturn/src/main.rs
vinistock: It complains that the temporary value is dropped while borrowed. You need to bind the `read` lock to a variable.

rust/saturn/src/model/declaration.rs
vinistock: Yes, we should still be able to do achieve it 👍

Morriar: I only found 2 ways to do it:

1. Using `match` but this means you need to handle all the variants:

   ```rs
       matches!((a, b),
        (Declaration::Class(_), Declaration::Class(_))
        | (Declaration::Module(_), Declaration::Module(_))
    )
   ```

2. Using `discriminant` but then you can't match the discriminant to a specific enum variant

   ```rs
   use std::mem::discriminant;

    discriminant(a) == discriminant(b)
   ```

Did you have something else in mind?

vinistock: The idea I originally had in mind was this:

- We process definitions one by one. The first one dictates what declaration gets created. In your `Foo` example, it becomes a module declaration
- Then we process the second definition. At this point, I would check the definition against the existing declaration and add an error if it doesn't match

We wouldn't be comparing two declarations because we skip creating if it already exists to avoid the extra allocation.

Morriar: > add an error if it doesn't match

Yes but how do we do compare that the definition kind matches the declaration kind?

vinistock: We can add a method like this to declaration

```rust
fn matches(&self, definition: &Definition) -> bool {
  matches!(definition, ExpectedType(_))
}
```

which could be generated by the current declaration macros so that we don't write these by hand. Or we could bring back an enum for `kind` that gets shared by declarations and definitions. I think we can pick an approach in #354. I haven't played around with this yet.

Morriar: > I think we can pick an approach in https://github.com/Shopify/saturn/issues/354. I haven't played around with this yet

I'm raising this because I'm working on #354.

We can remove it to avoid too much churn and I'll add it back.

PR #393, opened 2025-12-15
Support indexing class instance variables

st0012:
Support class instance variables with singleton class ownership

Class instance variables are now represented as regular `InstanceVariable` declarations, but with a `SingletonClass` owner. This distinguishes them from regular instance variables:

- **Instance variable**: `@foo` inside regular instance methods → owned by the class (e.g., `Foo::@foo`)
- **Class instance variable**: `@foo` in class body, singleton methods (`def self.foo`, `def Foo.bar`), or methods in singleton class (`class << self`) → owned by the singleton class (e.g., `Foo::<Foo>::@foo`)

Key changes:

- Resolution determines the owner based on lexical nesting:
  - Method with explicit receiver (`def self.foo`, `def Foo.bar`) → owned by receiver's singleton class
  - Method inside singleton class → owned by the singleton class
  - Directly in class/module body → owned by class/module's singleton class
  - Regular instance method → owned by the class itself
- Methods are now pushed onto `definitions_stack` during indexing, allowing resolution to walk up the nesting chain
- Renamed helpers for consistency: `current_nesting_name_id` → `current_owner_name_id`, `add_member_to_current_nesting` → `add_member_to_current_owner`
- Skips creating declarations for top-level ivars (`<main>` object) and methods with unresolved receivers

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -261,18 +261,28 @@ impl<'a> RubyIndexer<'a> {
         parameters
     }
 
-    /// Gets the `NameId` of the current class/module/singleton block.
+    /// Gets the `NameId` of the current owner (class/module/singleton class).
     /// Used to resolve `self` to a concrete `NameId` during indexing.
-    fn current_nesting_name_id(&self) -> Option<NameId> {
-        let parent_nesting_id = self.parent_nesting_id()?;
-        let definition = self.local_graph.definitions().get(parent_nesting_id)?;
-
-        match definition {
-            Definition::Class(class_def) => Some(*class_def.name_id()),
-            Definition::Module(module_def) => Some(*module_def.name_id()),
-            Definition::SingletonClass(singleton_class_def) => Some(*singleton_class_def.name_id()),
-            _ => panic!("current nesting is not a class/module/singleton class"),
+    ///
+    /// Iterates through the definitions stack in reverse to find the first class/module/singleton
+    /// class, skipping methods.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition is not a class, module, or singleton class
+    fn current_owner_name_id(&self) -> Option<NameId> {
+        for &id in self.definitions_stack.iter().rev() {
+            if let Some(definition) = self.local_graph.definitions().get(&id) {
+                match definition {
+                    Definition::Class(class_def) => return Some(*class_def.name_id()),
+                    Definition::Module(module_def) => return Some(*module_def.name_id()),
+                    Definition::SingletonClass(singleton_class_def) => return Some(*singleton_class_def.name_id()),
+                    Definition::Method(_) => {}
+                    _ => panic!("current nesting is not a class/module/singleton class: {definition:?}"),
+                }
+            }
         }
+        None
     }
 
     // Runs the given closure if the given call `node` is invoked directly on `self` for each one of its string or
@@ -377,7 +387,7 @@ impl<'a> RubyIndexer<'a> {
         let nesting = if top_level_ref {
             None
         } else {
-            self.current_nesting_name_id()
+            self.current_owner_name_id()
         };
 
         let string_id = self.local_graph.intern_string(name);
@@ -414,8 +424,29 @@ impl<'a> RubyIndexer<'a> {
         let definition = builder(str_id, offset, comments, lexical_nesting_id, uri_id);
         let definition_id = self.local_graph.add_definition(definition);
 
-        self.add_member_to_current_nesting(definition_id);
+        self.add_member_to_current_owner(definition_id);
 
+        definition_id
+    }
+
+    fn add_instance_variable_definition(&mut self, location: &ruby_prism::Location) -> DefinitionId {
+        let name = Self::location_to_string(location);
+        let str_id = self.local_graph.intern_string(name);
+        let offset = Offset::from_prism_location(location);
+        let comments = self.find_comments_for(offset.start()).unwrap_or_default();
+        let lexical_nesting_id = self.parent_nesting_id().copied();
+        let uri_id = self.uri_id;
+
+        let definition = Definition::InstanceVariable(Box::new(InstanceVariableDefinition::new(
+            str_id,
+            uri_id,
+            offset,
+            comments,
+            lexical_nesting_id,
+        )));
+
+        let definition_id = self.local_graph.add_definition(definition);
+        self.add_member_to_current_owner(definition_id);
         definition_id
     }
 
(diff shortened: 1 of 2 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -447,34 +477,50 @@ impl<'a> RubyIndexer<'a> {
...
-
-        match parent_nesting {
-            Definition::Class(class) => class.add_member(member_id),
-            Definition::SingletonClass(singleton_class) => singleton_class.add_member(member_id),
-            Definition::Module(module) => module.add_member(member_id),
-            _ => panic!("Cannot add a member to a non-nesting definition (current nesting is a {parent_nesting:?})",),
+        // Iterate the stack in reverse to find the first namespace (class, module, singleton class)
+        // Skip methods since they don't own members
+        for &nesting_id in self.definitions_stack.iter().rev() {
+            if let Some(nesting) = self.local_graph.get_definition_mut(nesting_id) {
```
Morriar: The none case should never happen right? Should we just unwrap/panic if this is the case rather than skipping silently?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -447,34 +477,50 @@ impl<'a> RubyIndexer<'a> {
...
-        let parent_nesting = self.parent_nesting()?;
-
-        match parent_nesting {
-            Definition::Class(class) => Some(*class.name_id()),
-            Definition::SingletonClass(singleton_class) => Some(*singleton_class.name_id()),
-            Definition::Module(module) => Some(*module.name_id()),
-            Definition::Constant(constant) => Some(*constant.name_id()),
-            _ => panic!("Cannot get name ID of definition: {parent_nesting:?}"),
+        for &nesting_id in self.definitions_stack.iter().rev() {
+            if let Some(definition) = self.local_graph.definitions().get(&nesting_id) {
```
Morriar: Same here, let's let it panic

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -447,34 +477,50 @@ impl<'a> RubyIndexer<'a> {
...
-        };
-
-        match parent_nesting {
-            Definition::Class(class) => class.add_member(member_id),
-            Definition::SingletonClass(singleton_class) => singleton_class.add_member(member_id),
-            Definition::Module(module) => module.add_member(member_id),
-            _ => panic!("Cannot add a member to a non-nesting definition (current nesting is a {parent_nesting:?})",),
+        // Iterate the stack in reverse to find the first namespace (class, module, singleton class)
+        // Skip methods since they don't own members
+        for &nesting_id in self.definitions_stack.iter().rev() {
```
Morriar: Should we extract this as a `parent_namespace`?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -447,34 +477,50 @@ impl<'a> RubyIndexer<'a> {
     ///
     /// Panics if the definition is not a nesting definition (class, module, or singleton class)
     fn add_member_to_current_nesting(&mut self, member_id: DefinitionId) {
```
Morriar: Or maybe we should rename this as `add_member_to_current_namespace`?

rust/saturn/src/resolution.rs
```diff
@@ -1440,6 +1512,176 @@ mod tests {
...
+                def self.nested
+                  @nested = 4
+                end
+              end
+            end
+            "
+        });
+        context.resolve();
+
+        let class_ivar_foo_decl = context
```
Morriar: Can we extract a macro for these assertions?

rust/saturn/src/indexing/ruby_indexer.rs
st0012: Do you mean just replace `self.definitions_stack.iter().rev()`?

Morriar: No the whole loop so it returns the first definition in the stack that is a namespace

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -436,45 +466,60 @@ impl<'a> RubyIndexer<'a> {
...
 
-    /// Adds a member to the current nesting definition.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the definition is not a nesting definition (class, module, or singleton class)
-    fn add_member_to_current_nesting(&mut self, member_id: DefinitionId) {
-        let Some(parent_nesting) = self.parent_nesting() else {
+    /// Returns the first namespace (class, module, singleton class, or constant) definition ID from the stack.
+    fn current_namespace_id(&self) -> Option<DefinitionId> {
```
Morriar: I'm confused, didn't you already change `current_nesting_name_id` to skip methods? Why do we need both changes?

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -3678,6 +3666,74 @@ mod tests {
...
+    #[test]
+    fn index_instance_variable_inside_methods_stay_instance_variable() {
+        let context = index_source({
+            "
+            class Foo
+              def initialize
+                @bar = 1
+              end
+
+              def self.class_method
```
vinistock, line 3802: Let's add a test for methods with a receiver that isn't `self`.

```ruby
class Foo
  def String.bar
    # This instance variable is associated
    # with the singleton class of String
    #
    # During indexing, we can't know that
    # because we haven't resolved String yet
    #
    # Ivars need to remember the NameRef
    # that represents the self
    @var = 123
  end
end
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -489,8 +534,8 @@ impl<'a> RubyIndexer<'a> {
             .filter_map(|arg| {
                 if arg.as_self_node().is_some() {
                     // FIXME: Ideally we would want to save the mixin as `self` but we can only save mixins with a name.
-                    // We'll just use the parent nesting name id for now.
-                    self.parent_nesting_name_id()
+                    // We'll just use the current namespace name id for now.
+                    self.current_namespace_name_id()
```
vinistock: Is there a conceptual change here for renaming nesting to namespace?

rust/saturn/src/model/declaration.rs
```diff
@@ -155,6 +156,7 @@ pub enum Declaration {
     GlobalVariable(Box<GlobalVariableDeclaration>),
     InstanceVariable(Box<InstanceVariableDeclaration>),
     ClassVariable(Box<ClassVariableDeclaration>),
+    ClassInstanceVariable(Box<ClassInstanceVariableDeclaration>),
```
vinistock: Considering that a class instance variable is just an instance variable owned by the singleton class, do we need this new declaration?

rust/saturn/src/resolution.rs
```diff
@@ -228,11 +229,114 @@ fn handle_remaining_definitions(
...
             }
             Definition::InstanceVariable(var) => {
-                let owner_id = resolve_lexical_owner(graph, *var.lexical_nesting_id());
+                let str_id = *var.str_id();
 
-                create_declaration(graph, *var.str_id(), id, owner_id, |name| {
-                    Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(name, owner_id)))
-                });
+                // Top-level instance variables belong to the `<main>` object, not `Object`.
+                // We can't represent `<main>` yet, so skip creating declarations for these.
```
vinistock, line 235: Maybe worth adding a `TODO` so that we can grep easily. We'll need some `<main>` representation.

rust/saturn/src/resolution.rs
```diff
@@ -228,11 +229,114 @@ fn handle_remaining_definitions(
...
+                                graph.declarations().get(&owner_id),
+                                Some(Declaration::SingletonClass(_))
+                            ),
+                            "Instance variable in class/module body should be owned by a SingletonClass"
+                        );
+                        create_declaration(graph, str_id, id, owner_id, |name| {
+                            Declaration::InstanceVariable(Box::new(InstanceVariableDeclaration::new(name, owner_id)))
+                        });
+                    }
+                    // If in a singleton class body directly, the owner is the singleton class itself
```
vinistock: I think this comment is not right or I'm misunderstanding something. Is it talking about this scenario?

```ruby
class Foo
  class << self
    @var = 123
  end
end
```

Because that `@var` is owned by `<<Foo>>` and not `<Foo>`.

rust/saturn/src/resolution.rs
```diff
@@ -1458,6 +1602,129 @@ mod tests {
...
+                def self.nested
+                  @nested = 4
+                end
+              end
+            end
+            "
+        });
+        context.resolve();
+
+        assert_class_instance_variable(&context, "Foo::<Foo>::@foo");
```
vinistock: Nitpick: can we reuse `assert_instance_variable` and just pass the singleton name as the expected owner?

rust/saturn/src/resolution.rs
st0012: Yep and the implementation is actually wrong. Good catch

rust/saturn/src/resolution.rs
```diff
@@ -228,11 +229,117 @@ fn handle_remaining_definitions(
...
+                    }
+                    // If in a singleton class body directly, the owner is the singleton class's singleton class
+                    // Like `class << Foo; @bar = 1; end`, where `@bar` is owned by `Foo::<Foo>::<<Foo>>`
+                    Definition::SingletonClass(_) => {
+                        let singleton_class_decl_id = graph
+                            .definitions_to_declarations()
+                            .get(&nesting_id)
+                            .copied()
+                            .unwrap_or(*OBJECT_ID);
+                        let owner_id = get_or_create_singleton_class(graph, singleton_class_decl_id);
```
vinistock, line 327: Just to make sure this is handling the algorithm correctly rather than treating a single singleton level as a special case, can we add this test case?

```ruby
class Foo
  class << self
    class << self
      @var = 123
    end
  end
end
```

Which should result in `Foo::<Foo>::<<Foo>>::<<<Foo>>>>::@var`.

st0012: Yep this works:

```rs
  context.index_uri("file:///foo.rb", {
      r"
      class Foo
        class << self
          @bar = 1

          class << self
            @baz = 2
          end
        end
      end
      "
  });
  context.resolve();

  assert_instance_variable(&context, "Foo::<Foo>::<<Foo>>::@bar", "Foo::<Foo>::<<Foo>>");
  assert_instance_variable(
      &context,
      "Foo::<Foo>::<<Foo>>::<<<Foo>>>::@baz",
      "Foo::<Foo>::<<Foo>>::<<<Foo>>>",
  );
```

PR #394, opened 2025-12-16
Introduce `JobQueue` and and use it to list files

Morriar:
The queue is backed by `crossbeam::{channel, deque, utils}` and provides parallelism, communication through channels and job stealing with backoff.

In this PR we migrate the file listing to the `JobQueue`, I'll follow up with indexing and resolution.

Benchmarks:

```
# listing files from core

Branch                       Avg Time (s)  Avg MRSS (MB)
main                         0.842         16.3
ignore-crate                 0.364         23.7
job-queue                    0.306         23.8

# listing files from shop/world/

Branch                       Avg Time (s)  Avg MRSS (MB)
main                         8.776         21.3
ignore-crate                 1.791         33.5
job-queue                    1.517         36.2
```

Closes https://github.com/Shopify/saturn/issues/100
Closes https://github.com/Shopify/saturn/pull/205

### DIFF
--- a/rust/saturn-sys/src/graph_api.rs
+++ b/rust/saturn-sys/src/graph_api.rs
@@ -52,14 +52,20 @@ pub unsafe extern "C" fn sat_index_all(
     count: usize,
 ) -> *const c_char {
     let file_paths: Vec<String> = unsafe { utils::convert_double_pointer_to_vec(file_paths, count).unwrap() };
-    let result = listing::collect_file_paths(file_paths);
+    let (file_paths, errors) = listing::collect_file_paths(file_paths);
 
-    if let Err(errors) = result {
-        return CString::new(errors.to_string()).unwrap().into_raw().cast_const();
+    if !errors.is_empty() {
+        let error_messages = errors
+            .iter()
+            .map(std::string::ToString::to_string)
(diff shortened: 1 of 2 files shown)


rust/saturn/src/job_queue.rs
```diff
@@ -0,0 +1,191 @@
...
+                Steal::Empty => {}                       // No jobs in this peer; try next peer.
+            }
+        }
+
+        // No jobs available from any source.
+        None
+    }
+}
+
+impl Default for JobQueue {
```
vinistock: Nit, let's derive default instead of using a manual implementation.

rust/saturn/src/main.rs
```diff
@@ -40,7 +40,11 @@ fn main() -> Result<(), Box<dyn Error>> {
         Timer::set_global_timer(Timer::new());
     }
 
-    let file_paths = time_it!(listing, { listing::collect_file_paths(vec![args.dir]) })?;
+    let (file_paths, errors) = time_it!(listing, { listing::collect_file_paths(vec![args.dir.clone()]) });
```
vinistock: Why do we need to clone here?

Morriar: We don't 🤦

Issue #392, opened 2025-12-12
Handle `Struct.new`, `Data.define`, `Class.new` and `Module.new`


vinistock:
Putting all of these in the same issue as I suspect there'll be some shared logic. Using these methods return classes or modules and if we set those to a constant, they'll become a namespace where modules can be included, definitions can be created and so on.

In all cases where we write a constant, we need to check if the value is a `new` call node with a receiver of `Struct`, `Class`, `Module` (potentially `OpenStruct` too) and then create a class or module instead of a constant.

**Important caveat**: while invoking these methods does produce new classes and modules, they do not produce a new lexical scope like the `class` or `module` keyword. For example:

```ruby
class Foo
  Bar = Struct.new do
    CONST = 123
  end
end
```

The `CONST` definition here is `Foo::CONST` because `Bar` does not advance the lexical scope. In practical terms, it means we need to insert a `ClassDefinition` or `ModuleDefinition` in the graph, but we should not push anything to the definition stack.

PR #402, opened 2025-12-23
Index Class.new and Module.new

vinistock:
Initial step for #392

This PR starts handling `Class.new` and `Module.new`. The shared characteristic between both is that they create a new possible owner for definitions. That means, you can have methods and instance variables be associated with them.

However, neither of those produces of a new lexical scope, so constant references that occur inside of the block ignore these new classes and modules. Only the `class` and `module` keywords can produce new lexical scopes.

To account for this, I turned the `definitions_stack` into a `nesting_stack` where we can have two types of entries:

- LexicalScope: an owner definition that also needs to be accounted for constant references
- Owner: an owner that doesn't influence constant references

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -4,10 +4,10 @@ use crate::diagnostic::Diagnostics;
 use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
-    AnonymousNamespaceDefinition, AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition,
-    ClassVariableDefinition, ConstantDefinition, Definition, DefinitionFlags, GlobalVariableAliasDefinition,
-    GlobalVariableDefinition, InstanceVariableDefinition, MethodAliasDefinition, MethodDefinition, Mixin,
-    ModuleDefinition, Parameter, ParameterStruct, SingletonClassDefinition,
+    AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -281,15 +280,39 @@ impl<'a> RubyIndexer<'a> {
...
     ///
-    /// Iterates through the definitions stack in reverse to find the first class/module/singleton
-    /// class, skipping methods.
+    /// Iterates through the definitions stack in reverse to find the first class/module/singleton class, skipping
+    /// methods. Ignores `Class.new` and other owners that do not produce lexical scopes
     ///
     /// # Panics
     ///
     /// Panics if the definition is not a class, module, or singleton class
+    fn current_lexical_scope_name_id(&self) -> Option<NameId> {
```
st0012, line 288: Will it be better to name it like `current_lexical_owner_name_id`?
Owner includes class/module/singleton class/anonymous class & module. Lexical owner excludes the last one.

vinistock: I think the nomenclature needs to be a touch different. We now have 3 types of nestings:

- LexicalScope: anything that uses the `class` or `module` keywords
- Owner: a namespace that can own things, but isn't a new lexical scope (e.g.: `Class.new`)
- Method: just to indicate we're inside a method, but it doesn't represent a new lexical scope and can't own anything

I think using "lexical owner" can be confusing because it kind of mixes the first two nesting variants. Personally, I would use `lexical_scope` and `owner` separately without mixing the two.

st0012: That's fair. Can this distinction go into the architecture doc?

Also TIL: nomenclature

vinistock: Yes! I'll add it to #423

PR #404, opened 2025-12-23
Add member method to Declaration for looking up namespace methods

bitwise-aiden:
Adds the `members` method to `Declaration` to expose the declarations that belong to that specific declaration. For traversing the ancestor chain, see this PR: https://github.com/Shopify/saturn/pull/403

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -107,6 +107,31 @@ static VALUE sr_declaration_definitions(VALUE self) {
     return self;
 }
 
+// Declaration#member: (String member) -> Declaration
+// Returns a declaration handle for the given member
+static VALUE sr_declaration_member(VALUE self, VALUE name) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    if (TYPE(name) != T_STRING) {
+        rb_raise(rb_eTypeError, "expected String");
+    }
+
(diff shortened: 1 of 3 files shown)


ext/rubydex/declaration.c
```diff
@@ -107,6 +107,31 @@ static VALUE sr_declaration_definitions(VALUE self) {
     return self;
 }
 
+// Declaration#member: (String member) -> Declaration
```
Morriar, line 110: Shouldn't this return an array?

Consider this:

```rb
class Foo
  def foo; end
  def foo; end
end
```

Or this:

```rb
class Foo
  class Array; end
  def Array(); end
end
```

vinistock: I think that's an existing bug actually. We're storing members in a hash of `string_id => declaration_id`.

Morriar: Oh wait it's me getting confused, it's a declaration it's fine you should only get one.

You get back a declaration that has many definitions.

vinistock: No, I think you're right for the array example. We're not using the type of declaration in either string or declaration IDs, so they are going to be identical and override each other.

```
members: {
  # Both method and class will have the same name
  StringId(Array) => DeclarationId(Foo::Array),
  StringId(Array) => DeclarationId(Foo::Array),
}
```

To solve the declaration ID part, I think the fix is making its fully qualified name `Foo#Array` (which we already knew we had to do).

I'm not sure what to do for the StringId though. We wouldn't want both the class and the method to end up in the same exact slot.

Maybe members needs to be split into constants vs everything else?

PR #409, opened 2026-01-05
Support discovery and resolution of constant aliases

### LINKED ISSUE #347: Index constant aliases
We are currently indexing all constant definitions the same. However, constant aliases are important for resolution because they might point to a namespace. For example:

```ruby
module Foo
  CONST = 123
end

module Bar
  ALIAS_CONST = Foo
end

# This works. We can only resolve this if we track aliases
Bar::ALIAS_CONST::CONST
```

I believe we should create a new definition type for constant aliases, which will store a `NameId` to the target constant, so that we can follow aliases properly.

st0012:
Constant aliases (`ALIAS = Foo`) are now indexed and resolved separately
from regular constants. This enables:

- Resolution through aliases: `ALIAS::CONST` resolves to `Foo::CONST`
- Chained aliases: `ALIAS2 = ALIAS1` where `ALIAS1 = Foo`
- Scoped aliases: `Bar::ALIAS = Foo`
- Circular aliases are detected and left unresolved
- Multiple targets: when an alias has definitions in multiple files
pointing to different targets, all targets are tracked

    ```rb
    # file1.rb
    class Foo; end
    ALIAS ||= Foo
    
    # file2.rb
    class Bar; end
    ALIAS ||= Bar
    
    # file3.rb
    # Both Bar and Foo will be ALIAS' target

    # However, constants defined through ALIAS will only be attached to one target (as its member)
    ALIAS::CONST = 123
    ```

Adds `Graph::alias_targets()` for clients to query alias targets.

Closes #347

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -5207,6 +5207,8 @@ mod tests {
             assert_name_id_to_string_eq!(&context, "B", def);
             assert_name_path_eq!(&context, "Target", def.target_name_id());
         });
+
+        assert_constant_references_eq!(&context, vec!["Target"]);
     }
 
     #[test]
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -211,11 +211,7 @@ impl Graph {
 
         let mut targets = Vec::new();
         for definition_id in declaration.definitions() {
-            let Some(definition) = self.definitions.get(definition_id) else {
-                continue;
-            };
-
-            let Definition::ConstantAlias(alias_def) = definition else {
+            let Some(Definition::ConstantAlias(alias_def)) = self.definitions.get(definition_id) else {
                 continue;
             };
 
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1457,6 +1457,53 @@ mod tests {
     use crate::model::ids::UriId;
     use crate::test_utils::GraphTest;
 
+    macro_rules! assert_constant_alias_target_eq {
+        ($context:expr, $alias_name:expr, $target_name:expr) => {{
+            let decl_id = DeclarationId::from($alias_name);
+            let target = $context
+                .graph()
+                .alias_targets(&decl_id)
+                .and_then(|t| t.first().copied());
+            assert_eq!(
+                target,
+                Some(DeclarationId::from($target_name)),
+                "Expected alias '{}' to have primary target '{}'",
+                $alias_name,
+                $target_name
+            );
+        }};
+    }
+
+    macro_rules! assert_no_constant_alias_target {
+        ($context:expr, $alias_name:expr) => {{
+            let decl_id = DeclarationId::from($alias_name);
+            let targets = $context.graph().alias_targets(&decl_id).unwrap_or_default();
+            assert!(
+                targets.is_empty(),
+                "Expected no alias target for '{}', but found {:?}",
+                $alias_name,
+                targets
+            );
+        }};
+    }
+
+    macro_rules! assert_alias_targets_contain {
+        ($context:expr, $alias_name:expr, $($target_name:expr),+ $(,)?) => {{
+            let decl_id = DeclarationId::from($alias_name);
+            let targets = $context.graph().alias_targets(&decl_id).unwrap_or_default();
+            $(
+                let expected_id = DeclarationId::from($target_name);
+                assert!(
+                    targets.contains(&expected_id),
+                    "Expected alias '{}' to contain target '{}', but targets were {:?}",
+                    $alias_name,
+                    $target_name,
+                    targets
+                );
+            )+
+        }};
+    }
+
     /// Asserts that a declaration has a constant reference at the specified location
     ///
     /// This macro:
@@ -1677,19 +1724,6 @@ mod tests {
         }};
     }
 
-    fn get_alias_targets(context: &GraphTest, alias_name: &str) -> Vec<DeclarationId> {
-        let decl_id = DeclarationId::from(alias_name);
-        context.graph().alias_targets(&decl_id).unwrap_or_default()
-    }
-
-    fn get_constant_alias_target(context: &GraphTest, alias_name: &str) -> Option<DeclarationId> {
-        get_alias_targets(context, alias_name).first().copied()
-    }
-
-    fn has_constant_alias_target(context: &GraphTest, alias_name: &str) -> bool {
-        !get_alias_targets(context, alias_name).is_empty()
-    }
-
     #[test]
     fn resolving_top_level_references() {
         let mut context = GraphTest::new();
@@ -1814,7 +1848,7 @@ mod tests {
         });
         context.resolve();
 
-        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
+        assert_no_diagnostics!(&context);
 
         let foo = context.graph().declarations().get(&DeclarationId::from("Foo")).unwrap();
         assert_members(foo, &["Bar", "Baz"]);
@@ -3691,9 +3725,9 @@ mod tests {
             "
         });
         context.resolve();
+        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
 
-        let foo_decl_id = DeclarationId::from("Foo");
-        assert_eq!(get_constant_alias_target(&context, "ALIAS"), Some(foo_decl_id));
+        assert_constant_alias_target_eq!(context, "ALIAS", "Foo");
         assert_constant_reference_to!(context, "Foo::CONST", "file:///foo.rb:5:7-5:12");
     }
 
@@ -3713,9 +3747,9 @@ mod tests {
             "
         });
         context.resolve();
+        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
 
-        let foo_bar_decl_id = DeclarationId::from("Foo::Bar");
-        assert_eq!(get_constant_alias_target(&context, "ALIAS"), Some(foo_bar_decl_id));
+        assert_constant_alias_target_eq!(context, "ALIAS", "Foo::Bar");
         assert_constant_reference_to!(context, "Foo::Bar::CONST", "file:///foo.rb:7:7-7:12");
     }
 
@@ -3735,9 +3769,9 @@ mod tests {
             "
         });
         context.resolve();
+        assert_no_diagnostics!(&context);
 
-        let foo_decl_id = DeclarationId::from("Foo");
-        assert_eq!(get_constant_alias_target(&context, "Bar::MyFoo"), Some(foo_decl_id));
+        assert_constant_alias_target_eq!(context, "Bar::MyFoo", "Foo");
         assert_constant_reference_to!(context, "Foo::CONST", "file:///foo.rb:6:9-6:14");
     }
 
@@ -3758,6 +3792,7 @@ mod tests {
             "
         });
         context.resolve();
+        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
 
         assert_constant_reference_to!(context, "Foo::CONST", "file:///foo.rb:8:7-8:12");
     }
@@ -3777,12 +3812,10 @@ mod tests {
             "
         });
         context.resolve();
+        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
 
-        let alias1_decl_id = DeclarationId::from("ALIAS1");
-        let foo_decl_id = DeclarationId::from("Foo");
-
-        assert_eq!(get_constant_alias_target(&context, "ALIAS1"), Some(foo_decl_id));
-        assert_eq!(get_constant_alias_target(&context, "ALIAS2"), Some(alias1_decl_id));
+        assert_constant_alias_target_eq!(context, "ALIAS1", "Foo");
+        assert_constant_alias_target_eq!(context, "ALIAS2", "ALIAS1");
         assert_constant_reference_to!(context, "Foo::CONST", "file:///foo.rb:6:8-6:13");
     }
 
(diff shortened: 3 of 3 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -3825,4 +3899,103 @@ mod tests {
...
+            CONST = 123
+            STRING = \"hello\"
+            ARRAY = [1, 2, 3]
+            "
+        });
+
+        assert_no_diagnostics!(&context);
+
+        // These should all be regular constants, not aliases
+        assert_definition_at!(&context, "1:1-1:6", Constant, |_c| {});
```
Morriar: We could make the block optional with something like this:

```rust
macro_rules! assert_definition_at {
    // Arm for the case with a block/closure
    ($context:expr, $range:expr, $kind:ident, $block:expr) => {{
        println!("With block: {}", $range);
    }};
    // Arm for the case WITHOUT the closure/block
    ($context:expr, $range:expr, $kind:ident) => {{
        assert_definition_at!($context, $range, $kind, |_c| {});
    }};
}

fn main() {
    let context = "context";
    assert_definition_at!(&context, "1:1-1:6", Constant, |_c| { println!("Hi"); });
    assert_definition_at!(&context, "1:1-1:6", Constant); // No closure needed
}
```

rust/saturn/src/resolution.rs
```diff
@@ -1494,4 +1570,195 @@ mod tests {
...
+        );
+        assert_constant_reference_to!(context, "Foo::CONST", "file:///foo.rb:6:8-6:13");
+    }
+
+    #[test]
+    fn resolving_constant_alias_to_non_existent_target() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            ALIAS = NonExistent
```
Morriar: Let's add an alias to a non-existant constant through another one:

```rb
ALIAS_1 = NonExistent
ALIAS_2 = ALIAS_1
puts ALIAS_2
```

rust/saturn/src/resolution.rs
```diff
@@ -1494,4 +1570,195 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn resolving_circular_constant_aliases() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            A = B
+            B = A
```
Morriar: Let's also have a circular using 3 constants:

```rb
A = B
B = C
C = A
```

rust/saturn/src/indexing/ruby_indexer.rs
st0012: TIL. Updated

rust/saturn/src/model/graph.rs
```diff
@@ -36,6 +36,8 @@ pub struct Graph {
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     // Map of method references that still need to be resolved
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
+    // Map of constant alias declarations to their target declarations
+    constant_aliases: IdentityHashMap<DeclarationId, DeclarationId>,
```
vinistock: Do we need a new map for this?

In the LSP, the way we modelled this was:

- Constants for which the right hand side is another constant are recorded as a `ConstantAlias`, which has a `target` field with the right hand side name
- Any other constants are just `Constant`

So for example:

```ruby
module Foo
  module Bar; end

  # ConstantAlias {
  #   name_id: NameId(CONST, nesting: Foo)
  #   target_id: NameId(Bar, nesting: Foo)
  # }
  CONST = Bar
end
```

This approach means that constant aliases are just another possible definition/declaration type we have to handle, rather than a completely new map.

st0012: The constant alias definition already stores target. I added `ConstantAlias` declaration to store target to and dropped the map.

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -833,13 +868,27 @@ impl Visit<'_> for RubyIndexer<'_> {
     }
 
     fn visit_constant_or_write_node(&mut self, node: &ruby_prism::ConstantOrWriteNode) {
-        self.add_constant_definition(&node.as_node(), true);
-        self.visit(&node.value());
+        match &node.value() {
+            ruby_prism::Node::ConstantReadNode { .. } | ruby_prism::Node::ConstantPathNode { .. } => {
```
vinistock: Is this handling this case?

```ruby
A = B = C
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -521,6 +521,41 @@ impl<'a> RubyIndexer<'a> {
...
+    ) -> Option<DefinitionId> {
+        let name_id = self.index_constant_reference(name_node, also_add_reference)?;
+        let target_name_id = self.index_constant_reference(target_node, true)?;
+
+        // Get the location for the constant name/path only (not including the value)
+        let location = match name_node {
+            ruby_prism::Node::ConstantWriteNode { .. } => name_node.as_constant_write_node().unwrap().name_loc(),
+            ruby_prism::Node::ConstantOrWriteNode { .. } => name_node.as_constant_or_write_node().unwrap().name_loc(),
+            _ => name_node.location(),
+        };
```
vinistock, line 767: Do we also need to handle constant paths here? If I remember correctly, in a definition like `A::B = 2`, the definition's location is just `B`.

rust/saturn/src/model/declaration.rs
```diff
@@ -240,6 +242,48 @@ simple_declaration!(GlobalVariableDeclaration);
...
+    /// The fully qualified name of this declaration
+    name: String,
+    /// The list of definition IDs that compose this declaration
+    definition_ids: Vec<DefinitionId>,
+    /// The list of references that are made to this declaration
+    references: Vec<ReferenceId>,
+    /// The ID of the owner of this declaration
+    owner_id: DeclarationId,
+    /// The resolved target declaration (None if the target hasn't been resolved yet/can't be resolved)
+    target: Option<DeclarationId>,
```
vinistock: I think we don't need this field. We can probably just search the definitions of this constant alias and get the declaration ID from the definition's `target_name_id`.

That already captures if it's resolved or not, which is also the reason why you ended up needing the `Option`.

rust/saturn/src/resolution.rs
```diff
@@ -557,32 +565,115 @@ where
...
+    // in two different files) - the first definition links the alias, subsequent ones skip.
+    if let Some(Declaration::ConstantAlias(alias)) = graph.declarations().get(&declaration_id)
+        && alias.target().is_some()
+    {
+        return Outcome::Resolved(declaration_id);
+    }
+
+    // Try to resolve the target and set it on the declaration.
+    // Example: `ALIAS = Foo` -> set ALIAS's target to Foo's declaration.
+    match resolve_constant(graph, target_name_id) {
```
vinistock: I'm not sure we need to trigger resolution for the alias' target. In fact, I think this might be doing the same work twice.

The right hand side of the alias gets indexed as a constant reference, so it is already enqueued for resolution.

If we follow my suggestion in https://github.com/Shopify/saturn/pull/409/changes#r2673361629, I think it will simplify this because you won't have to worry about the right hand side at all.

You just process the alias as any other constant and the rest of resolution will handle the rest.

rust/rubydex/src/model/graph.rs
```diff
@@ -184,6 +188,53 @@ impl Graph {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the declaration doesn't exist or is not a constant alias.
+    #[must_use]
+    pub fn alias_targets(&self, declaration_id: &DeclarationId) -> Vec<DeclarationId> {
+        let read_lock = self.declarations.read().unwrap();
+        let declaration = read_lock
+            .get(declaration_id)
+            .expect("Declaration not found");
```
vinistock: Is the idea for this API to be public? If so, I think we need to handle the possibility of passing a declaration that doesn't exist.

Otherwise, let's use `pub(crate)`.

rust/rubydex/src/model/graph.rs
```diff
@@ -184,6 +188,53 @@ impl Graph {
...
+
+        let Declaration::ConstantAlias(_) = declaration else {
+            panic!("Declaration is not a constant alias");
+        };
```
vinistock, line 210: We can probably merge this with the one above and do

```rust
let Some(Declaration::ConstantAlias(declaration)) = ...
```

rust/rubydex/src/resolution.rs
```diff
@@ -880,32 +893,153 @@ where
...
+/// * `graph` - The symbol graph
+/// * `name_id` - The name ID of the constant being declared
+///
+/// # Returns
+/// A vector of declaration IDs representing all namespaces (`Class`, `Module`, or
+/// `SingletonClass`) that should contain this constant as a member.
+///
+/// # Panics
+/// Panics if called before the parent scope or nesting has been resolved.
+fn resolve_all_possible_owners(graph: &Graph, name_id: NameId) -> Vec<DeclarationId> {
```
vinistock: I understand the scenario here, but I'm trying to think if we can actually support this correctly at the moment.

Especially without a type system that supports unions.

For example:

```ruby
module Foo
  FOO_CONST = 1
end

module Bar
  BAR_CONST = 2
end

ALIAS = Foo

# this is okay here
ALIAS::FOO_CONST

# some other file
ALIAS = Bar
# this is no longer okay
ALIAS::FOO_CONST
```

We could make the type of `ALIAS` an union of `<Foo> | <Bar>` (which is basically what you're doing manually), but I'm not 100% sure of the implications.

Sorbet doesn't allow this. We may want to start similarly and do the same thing we did for parent classes: pick the first one and add a diagnostic if there is more than one distinct definition.

@at do you have thoughts on this one?

st0012: IMO supporting multiple targets is better for LSP as I think it's better to show all the possible definitions/references in ambiguous cases.
Also from my experience implementing this, it's much easier to take into account of multiple possible targets and then pick the first one, than changing from picking the first one to supporting multiple targets.

rust/rubydex/src/model/graph.rs
st0012: It's a public API. I've updated it to return an option type instead of panicking here, if that's what you meant?

rust/rubydex/src/model/definitions.rs
```diff
@@ -527,6 +530,88 @@ impl ConstantDefinition {
...
+/// ALIAS = Foo
+/// ```
+#[derive(Debug)]
+pub struct ConstantAliasDefinition {
+    name_id: NameId,
+    target_name_id: NameId,
+    uri_id: UriId,
+    offset: Offset,
+    comments: Vec<Comment>,
+    flags: DefinitionFlags,
```
Morriar, line 547: Not a blocker but I wonder if it could just have a `constant: ConstantDefinition` field instead?

rust/rubydex/src/model/graph.rs
```diff
@@ -190,6 +194,47 @@ impl Graph {
...
+
+        let mut targets = Vec::new();
+        for definition_id in declaration.definitions() {
+            let Some(definition) = self.definitions.get(definition_id) else {
+                continue;
+            };
+
+            let Definition::ConstantAlias(alias_def) = definition else {
+                continue;
+            };
```
Morriar: I _think_ you can merge the statements:

```suggestion
            let Some(Definition::ConstantAlias(alias_def)) = definition else {
                continue;
            };
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5005,4 +5106,145 @@ mod tests {
...
+            assert_name_id_to_string_eq!(&context, "Bar::ALIAS", def);
+        });
+    }
+
+    #[test]
+    fn index_chained_constant_alias() {
+        let context = index_source({
+            "
+            module Target; end
+            A = B = Target
```
Morriar, line 5196: Could you add an assert showing we do not register a constant reference to `B` in `A = B`?

rust/rubydex/src/resolution.rs
```diff
@@ -3548,4 +3676,601 @@ mod tests {
...
+            end
+
+            ALIAS = Foo
+            ALIAS::CONST
+            "
+        });
+        context.resolve();
+
+        let foo_decl_id = DeclarationId::from("Foo");
+        assert_eq!(get_constant_alias_target(&context, "ALIAS"), Some(foo_decl_id));
```
Morriar: Can you add a macro to simplify this:

```rs
assert_constant_alias_target!(context, "ALIAS", "Foo")
```

rust/rubydex/src/resolution.rs
```diff
@@ -3548,4 +3676,601 @@ mod tests {
...
+            r"
+            module Foo
+              CONST = 123
+            end
+
+            ALIAS = Foo
+            ALIAS::CONST
+            "
+        });
+        context.resolve();
```
Morriar, line 3727: Let's add `assert_no_diagnostics!(&context);` to all these new tests

rust/rubydex/src/resolution.rs
```diff
@@ -3548,4 +3676,601 @@ mod tests {
...
+    }
+
+    #[test]
+    fn resolving_circular_constant_aliases() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            A = B
+            B = C
+            C = A
```
Morriar, line 3905: Interestingly, because we retry this will pass without diagnostic while it actually crashed at runtime. At some point we may have to consider that for definitions in the same file, the order matters.

rust/rubydex/src/model/definitions.rs
st0012: But when we index `ALIAS = Foo`, we only have the name of `Foo`, not necessarily the definition of it. So unless we want to perform resolution eagerly during discover/extraction, we won't be able to do so? And that'd break the architectural principle too even if it works?

Morriar: No I meant for `ALIAS`

Morriar:
```rs
pub struct ConstantAliasDefinition {
    alias_constant: ConstantDefinition,
    target_name_id: NameId,
}
```

rust/rubydex/src/resolution.rs
st0012: https://github.com/Shopify/rubydex/issues/479

rust/rubydex/src/model/definitions.rs
st0012: I'll give this a try after merging the PR

PR #410, opened 2026-01-06
Add Definition#deprecated?

### LINKED ISSUE #315: Add support for deprecated annotation
On thing missing from our editor experience is the ability to deprecate something and see it crossed out. Based on some [discussions on the LSP repo](https://github.com/Shopify/ruby-lsp/issues/1773), it seems that RBS standardized this using its free style annotation syntax `%a{deprecated}`. Meanwhile, for Ruby files, the community standard seems to remain the `# @deprecated` comment.

Let's add support for deprecating constants and methods.

thomasmarshall:
Closes #315.

This commit adds support for the `@deprecated` tag in comments to mark definitions as deprecated. This is commonly used in Ruby codebases though not currently officially recognized by RDoc.

This is my first Saturn contribution so I'm still getting up to speed with where functionality should be implemented. I wasn't sure if parsing the comments should happen inline in `Definition` or in `Comment`. For now, I went with `Definition` but open to alternatives.

I added tests for the Ruby part but I couldn't see any tests for the `Definition` model on the Rust side. Are the Ruby tests sufficient or should we add tests for the `is_deprecated` function in Rust?

### DIFF
--- a/rust/rubydex-sys/src/definition_api.rs
+++ b/rust/rubydex-sys/src/definition_api.rs
@@ -312,3 +312,20 @@ where
 
     DefinitionsIter::new(entries)
 }
+
+/// Returns true if the definition is deprecated.
+///
+/// # Safety
+/// - `pointer` must be a valid pointer previously returned by `sat_graph_new`.
+/// - `definition_id` must be a valid definition id.
+///
+/// # Panics
+/// This function will panic if a definition cannot be found.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sat_definition_is_deprecated(pointer: GraphPointer, definition_id: i64) -> bool {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        let defn = graph.definitions().get(&def_id).expect("definition not found");
+        defn.is_deprecated()
+    })
+}
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -5,7 +5,7 @@ use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
     AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
-    ConstantDefinition, Definition, GlobalVariableAliasDefinition, GlobalVariableDefinition,
+    ConstantDefinition, Definition, DefinitionFlags, GlobalVariableAliasDefinition, GlobalVariableDefinition,
     InstanceVariableDefinition, MethodAliasDefinition, MethodDefinition, Mixin, ModuleDefinition, Parameter,
     ParameterStruct, SingletonClassDefinition,
 };
@@ -126,10 +126,10 @@ impl<'a> RubyIndexer<'a> {
         String::from_utf8_lossy(location.as_slice()).to_string()
     }
 
-    fn find_comments_for(&self, offset: u32) -> Option<Vec<Comment>> {
+    fn find_comments_for(&self, offset: u32) -> (Vec<Comment>, DefinitionFlags) {
         let offset_usize = offset as usize;
         if self.comments.is_empty() {
-            return None;
+            return (Vec::new(), DefinitionFlags::empty());
         }
 
         let idx = match self.comments.binary_search_by_key(&offset_usize, |g| g.end_offset) {
@@ -137,28 +137,24 @@ impl<'a> RubyIndexer<'a> {
                 // This should never happen in valid Ruby syntax - a comment cannot end exactly
                 // where a definition begins (there must be at least a newline between them)
                 debug_assert!(false, "Comment ends exactly at definition start - this indicates a bug");
-                return None;
+                return (Vec::new(), DefinitionFlags::empty());
             }
             Err(i) if i > 0 => i - 1,
-            Err(_) => return None,
+            Err(_) => return (Vec::new(), DefinitionFlags::empty()),
         };
 
         let group = &self.comments[idx];
         let between = &self.source.as_bytes()[group.end_offset..offset_usize];
         if !between.iter().all(|&b| b.is_ascii_whitespace()) {
-            return None;
+            return (Vec::new(), DefinitionFlags::empty());
         }
 
         // We allow at most one blank line between the comment and the definition
         if bytecount::count(between, b'\n') > 2 {
-            return None;
+            return (Vec::new(), DefinitionFlags::empty());
         }
 
-        if group.comments.is_empty() {
-            None
-        } else {
-            Some(group.comments.clone())
-        }
+        (group.comments(), group.flags())
     }
 
     fn collect_parameters(&mut self, node: &ruby_prism::DefNode) -> Vec<Parameter> {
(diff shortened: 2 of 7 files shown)


ext/saturn/definition.c
```diff
@@ -126,6 +126,18 @@ static VALUE sr_definition_name(VALUE self) {
     return str;
 }
 
+// Definition#deprecated? -> TrueClass | FalseClass
```
thomasmarshall: Are these comments just for documentation? Would this be clearer as `boolean` instead of `TrueClass | FalseClass`?

Morriar: Yep, just doc, it's kind of a RBS syntax so let's go with `bool`

rust/rubydex/src/model/definitions.rs
```diff
@@ -115,6 +115,17 @@ impl Definition {
             Definition::GlobalVariableAlias(_) => "GlobalVariableAlias",
         }
     }
+
+    #[must_use]
+    pub fn is_deprecated(&self) -> bool {
```
Morriar, line 136: Let's add a test on the Rust side for this.

rust/saturn/src/model/definitions.rs
```diff
@@ -1160,3 +1171,44 @@ impl GlobalVariableAliasDefinition {
...
+        let def = create_method_definition(comments);
+        assert!(!def.is_deprecated());
+    }
+
+    #[test]
+    fn is_deprecated_returns_true_when_deprecated_tag_is_multi_line() {
+        let comments = vec![Comment::new(Offset::new(0, 20), "# Foo \n# @deprecated".to_string())];
+        let def = create_method_definition(comments);
+        assert!(def.is_deprecated());
+    }
```
st0012: Can we add a test case for code like:

```rb
# this method checks if the comment has @deprecated tag
def check_deprecated; end
```

And it does NOT mark `check_deprecated` as deprecated?

rust/saturn/src/model/definitions.rs
```diff
@@ -28,12 +28,26 @@ use std::ops::Deref;
...
         ids::{DefinitionId, NameId, StringId, UriId},
         visibility::Visibility,
     },
     offset::Offset,
 };
 
+/// Check if comments contain the `@deprecated` tag.
+///
+/// The tag must appear at the start of a line (after `#` and whitespace).
+fn comments_contain_deprecated(comments: &[Comment]) -> bool {
```
thomasmarshall: I don't know if this is the right place for this function to live, I'm still getting used to how code is organized in the project.

rust/saturn/src/model/definitions.rs
thomasmarshall: Great, I added a test on the Rust side for this too, we already had one on the Ruby side.

rust/saturn/src/model/definitions.rs
vinistock: We already go through each comment [here](https://github.com/Shopify/saturn/blob/3346630de01b1a11323106a03dc2c835143cf17e/rust/saturn/src/indexing/ruby_indexer.rs#L671).

Let's remember as part of the comment group if we found the deprecated directive. Then we pass the flags whenever we create the definitions in the `RubyIndexer`.

rust/saturn/src/model/flags.rs
```diff
@@ -0,0 +1,15 @@
+use bitflags::bitflags;
+
+bitflags! {
+    #[derive(Debug)]
+    pub struct Flags: u8 {
```
Morriar: Let's call this `DefinitionFlags`

rust/saturn/src/model/definitions.rs
thomasmarshall: OK I pushed a new version that does as you suggest, it attaches a `deprecated` boolean to the comment group and then constructs the flags from the comment group when we create the definitions. Is that what you had in mind?

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -312,3 +312,23 @@ where
...
+/// This function will panic if a definition cannot be found.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sat_definition_is_deprecated(pointer: GraphPointer, definition_id: i64) -> bool {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        if let Some(defn) = graph.definitions().get(&def_id) {
+            defn.is_deprecated()
+        } else {
+            panic!("Definition not found: {def_id:?}");
+        }
```
vinistock:
```suggestion
        let defn = graph.definitions().get(&def_id).expect("definition not found");
        defn.is_deprecated()
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -154,10 +154,17 @@ impl<'a> RubyIndexer<'a> {
...
-        } else {
-            Some(group.comments.clone())
+        Some(group)
+    }
+
+    fn extract_comments(group: Option<&CommentGroup>) -> Vec<Comment> {
+        group.map(|g| g.comments.clone()).unwrap_or_default()
+    }
+
+    fn build_flags(group: Option<&CommentGroup>) -> DefinitionFlags {
```
vinistock: I think these two should be methods in the `CommentsGroup` struct since they only operate on that struct's data.

rust/saturn/src/model/definitions.rs
vinistock: Yes, exactly. I left a couple of comments, but I think it looks good 👍

rust/rubydex/src/indexing/ruby_indexer.rs
thomasmarshall: I added the helpers to reduce some repetition when dealing with the optional `CommentGroup`. If we moved the methods into `CommentGroup` we'd need to repeat the optional handling at each of the call sites. Instead, I chose to make `find_comments_for` return a tuple of `(Vec<Comment>, DefinitionFlags)` so the optional and default handling only happens in one place. What do you think?

vinistock: Oh, I see. I like the tuple approach 👍

test/definition_test.rb
```diff
@@ -134,6 +134,32 @@ def bar; end
...
+
+        class NotDeprecated; end
+
+        # Multi-line comment
+        # @deprecated Use something else
+        def deprecated_method; end
+
+        # Not @deprecated
+        def not_deprecated_method; end
+      RUBY
```
Morriar, line 151: Can we add tests showing how the `@deprecated` behaves with new lines, I think with our current implementation we'll accept one:

~~~rb
  def test_definition_deprecated_newlines
    with_context do |context|
      context.write!("file1.rb", <<~RUBY)
        # @deprecated

        class Deprecated; end

        # @deprecated

        class NotDeprecated; end

        # Multi-line comment
        # @deprecated Use something else

        def deprecated_method; end
      RUBY

      graph = Rubydex::Graph.new
      graph.index_all(context.glob("**/*.rb"))

      assert(graph.documents.first.definitions.find { |d| d.name == "Deprecated" }.deprecated?)
      refute(graph.documents.first.definitions.find { |d| d.name == "NotDeprecated" }.deprecated?)
      assert(graph.documents.first.definitions.find { |d| d.name == "deprecated_method" }.deprecated?)
    end
  end
~~~

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -3206,6 +3243,45 @@ mod tests {
...
+
+        assert_definition_at!(&context, "8:1-8:27", Method, |def| {
+            assert_name_eq!(&context, "deprecated_method", def);
+        });
+        assert!(context.definition_at("8:1-8:27").is_deprecated());
+
+        assert_definition_at!(&context, "11:1-11:31", Method, |def| {
+            assert_name_eq!(&context, "not_deprecated_method", def);
+        });
+        assert!(!context.definition_at("11:1-11:31").is_deprecated());
```
Morriar, line 3282: We shouldn't test again the `definition_at`, let's just focus on the deprecated:


```suggestion
        assert!(context.definition_at("2:1-2:22").is_deprecated());
        assert!(!context.definition_at("4:1-4:25").is_deprecated());
        assert!(context.definition_at("8:1-8:27").is_deprecated());
        assert!(!context.definition_at("11:1-11:31").is_deprecated());
```

PR #414, opened 2026-01-06
Add declaration search

vinistock:
This PR implements a simple version of searching declarations based on their name. This will be used to back the LSP's workspace symbol search.

Right now, it's not super smart. It calculates a match score based on character matches that happen in the right order. It already works okay, but we can make it more sophisticated later.

**Note**

I tried using our job infrastructure for this, but I couldn't make it work. The jobs need an `Arc<&Graph>`, so that they can lock declarations for reading. However, if you try to implement that, you get into trouble because Rust will say that the `&'a Graph` reference does not outlive the Job's `'static` requirement.

On the other hand, we shouldn't need to move the graph (pass it without being a reference) just to query its data. For now, I just used scoped threads.

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -109,6 +109,32 @@ static VALUE sr_graph_declarations(VALUE self) {
     return self;
 }
 
+// Graph#search: () -> Enumerator[Declaration]
+// Returns an enumerator that yields all declarations lazily
+static VALUE sr_graph_search(VALUE self, VALUE query) {
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize(self, rb_str_new2("search"), 1, &query);
+    }
+
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    const char *c_query = StringValueCStr(query);
+
+    void *iter = rdx_graph_declarations_search(graph, c_query);
+
+    if (iter == NULL) {
+        // The only case where the iterator will be NULL instead of a list is if the query cannot be converted to a Rust
+        // string
+        rb_raise(rb_eRuntimeError, "Converting query to Rust string failed");
+    }
+
+    VALUE args = rb_ary_new_from_args(2, self, ULL2NUM((uintptr_t)iter));
+    rb_ensure(graph_declarations_yield, args, graph_declarations_ensure, args);
+
+    return self;
+}
+
 // Body function for rb_ensure in Graph#documents
 static VALUE graph_documents_yield(VALUE args) {
     VALUE self = rb_ary_entry(args, 0);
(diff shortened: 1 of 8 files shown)


ext/rubydex/graph.c
```diff
@@ -109,6 +109,25 @@ static VALUE sr_graph_declarations(VALUE self) {
...
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize(self, rb_str_new2("search"), 1, &query);
+    }
+
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    const char *c_query = StringValueCStr(query);
+
+    void *iter = rdx_graph_declarations_search(graph, c_query);
```
st0012, line 124: Do we need to check if `iter` is null here and raise an exception if the query is invalid?

rust/rubydex/src/query.rs
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+
+                local_results
+            });
+
+            threads.push(handle);
+        }
+
+        threads.into_iter().flat_map(|t| t.join().unwrap()).collect()
```
st0012: If we split the items (`declarations.keys()`) upfront we can avoid avoid repeatedly locking the whole list:

```suggestion
    let ids: Vec<DeclarationId> = declarations.keys().copied().collect();

    let chunk_size = ids.len().div_ceil(num_threads);
    if chunk_size == 0 {
        return Vec::new();
    }

    thread::scope(|s| {
        let handles: Vec<_> = ids
            .chunks(chunk_size)
            .map(|chunk| {
                s.spawn(|| {
                    chunk
                        .iter()
                        .filter(|id| {
                            let declaration = declarations.get(id).unwrap();
                            // When the query is empty, we return everything as per the LSP specification.
                            // Otherwise, we compute the match score and return anything with a score greater than zero
                            query.is_empty() || match_score(query, declaration.name()) > 0
                        })
                        .copied()
                        .collect::<Vec<_>>()
                })
            })
            .collect();

        handles.into_iter().flat_map(|h| h.join().unwrap()).collect()
```

ext/rubydex/graph.c
vinistock: I added a check, although I don't think this scenario can actually happen. We receive the `char*` from the Ruby VM, so I believe converting it to a Rust string will always succeed.

rust/rubydex/src/query.rs
vinistock: I switched to this approach, but I've been thinking if maybe having `rayon` added back in isn't a better option. It provides many convenient parallelism options.

This is also why I mentioned in the PR description that I tried to use our new job infrastructure, so that the parallelism would be consistent with the rest. However, I think we need some modifications in the job infra to make it work.

st0012: I think this particular implementation isn't bad. I'd vote against twisting it to pass graph just to use the job queue, even if it works.

PR #415, opened 2026-01-06
Use JobQueue for file indexing.

Morriar:
Follow up on #394, we use the `JobQueue` for file indexing.

The indexing logic is the same, I just moved it into `Job` implementations.

Performances are unchanged we use the same parallelism, channels and work stealing tricks.

### DIFF
--- a/rust/saturn/src/indexing.rs
+++ b/rust/saturn/src/indexing.rs
@@ -1,98 +1,97 @@
 use crate::{
-    errors::{Errors, MultipleErrors},
+    errors::Errors,
     indexing::{local_graph::LocalGraph, ruby_indexer::RubyIndexer},
+    job_queue::{Job, JobQueue},
     model::graph::Graph,
 };
-use std::fs;
-use std::sync::{
-    Arc, Mutex,
-    mpsc::{Receiver, Sender},
-};
-use std::{sync::mpsc, thread};
+use crossbeam_channel::{Sender, unbounded};
+use std::{fs, path::PathBuf, sync::Arc};
 use url::Url;
 
 pub mod local_graph;
 pub mod ruby_indexer;
 
-type LocalGraphResult = Result<LocalGraph, Errors>;
-
-/// Indexes the given items, reading the content from disk and populating the given `Graph` instance.
-///
-/// # Errors
-///
-/// Returns Ok if indexing succeeded for all given documents or a vector of errors for all failures
-///
-/// # Panics
-/// This function will panic in the event of a thread dead lock, which indicates a bug in our implementation. There
-/// should not be any code that tries to lock the same mutex multiple times in the same thread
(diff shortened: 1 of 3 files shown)


rust/saturn/src/indexing.rs
```diff
@@ -1,98 +1,131 @@
...
+
+    fn send_error(&self, error: Errors) {
+        self.errors_tx
+            .send(error)
+            .expect("errors receiver dropped before run completion");
+    }
+}
+
+impl Job for IndexingRubyFileJob {
+    fn run(&self) {
```
vinistock, line 39: Will these be able to receive parameters too? For the LSP, we'll need to pass the source received from the editor instead of reading from disk.

rust/saturn/src/indexing.rs
```diff
@@ -1,98 +1,131 @@
...
+///
+/// Will panic if the graph cannot be wrapped in an Arc<Mutex<>>
+pub fn index_files(graph: &mut Graph, paths: Vec<PathBuf>) -> Vec<Errors> {
+    let queue = Arc::new(JobQueue::new());
+    let (local_graphs_tx, local_graphs_rx) = unbounded();
+    let (errors_tx, errors_rx) = unbounded();
+
+    // We need to wrap the graph in an Arc<Mutex<>> so it can be shared with the MergeJob.
+    // Since we only have a mutable reference, we take the value out, wrap it, and put it back at the end.
+    let shared_graph = Arc::new(Mutex::new(mem::take(graph)));
```
vinistock: Do we need the merge part to be parallelized? We can probably avoid both the merge job and the arc mutex if we use the main thread to perform the merge as before, no?

Or is there some advantage to parallelizing the merge too?

rust/saturn/src/indexing.rs
Morriar: You pass the parameters through the `new` when you push the job onto the queue.

So we can imagine something like this for URI files:

```rs
pub struct IndexingRubyContentJob {
    uri: URI,
    ruby_content: String,
    local_graph_tx: Sender<LocalGraph>,
    errors_tx: Sender<Errors>,
}

queue.push(Box::new(IndexingRubyContentJob::new(
    uri,
    ruby_content
    local_graphs_tx.clone(),
    errors_tx.clone(),
)));
```

rust/saturn/src/indexing.rs
Morriar: It's not necessary, performances are the same since we don't parallelize the merge, we just made it one single job. The idea was more to stick to one pattern.

I removed the job and kept the merging in the main loop 👍

rust/saturn/src/indexing.rs
vinistock: Oh, right! Nice.

PR #416, opened 2026-01-06
Clean errors definition

Morriar:
Uniformize how we raise errors from listing/indexing.

### DIFF
--- a/rust/saturn/src/listing.rs
+++ b/rust/saturn/src/listing.rs
@@ -39,7 +39,7 @@ impl FileDiscoveryJob {
 
     fn handle_symlink(&self, path: &PathBuf) {
         let Ok(canonicalized) = fs::canonicalize(path) else {
-            self.send_error(Errors::FileReadError(format!(
+            self.send_error(Errors::FileError(format!(
                 "Failed to canonicalize symlink: `{}`",
                 path.display(),
             )));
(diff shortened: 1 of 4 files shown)


rust/saturn/src/listing.rs
```diff
@@ -39,7 +39,7 @@ impl FileDiscoveryJob {
 
     fn handle_symlink(&self, path: &PathBuf) {
         let Ok(canonicalized) = fs::canonicalize(path) else {
-            self.send_error(Errors::FileReadError(format!(
+            self.send_error(Errors::PathError(format!(
```
st0012: When do we use PathError and when to use FileReadError?

Morriar: Would this help?

```rs
errors!(
    // Should be used only if the file can't be opened for reading
    FileReadError;
    // Should be used if any path manipulation fails
    PathError;
);
```

Honestly, I think we could have only one error for all these case. Differentiating is not super useful and kind of arbitrary.

st0012: If we don't need the distinction now, I prefer having a generic one and capture context through the message. @vinistock WDYT?

Morriar: I agree, I was trying to keep the original errors but I think just one more generic is fine.

vinistock: Yeah, I think having a single one makes sense.

Morriar: Perfect, removed the errors to keep only one generic `FileError` 👍

PR #417, opened 2026-01-06
Surface indexing diagnostics

Morriar:
Closes https://github.com/Shopify/saturn/issues/340.

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -488,12 +488,24 @@ impl<'a> RubyIndexer<'a> {
             return;
         };
 
+        let lexical_nesting_id = self.parent_nesting_id().copied();
+
         // Collect all arguments as constant references. Ignore anything that isn't a constant
         let reference_ids: Vec<_> = arguments
             .arguments()
             .iter()
             .filter_map(|arg| {
                 if arg.as_self_node().is_some() {
+                    if lexical_nesting_id.is_none() {
+                        self.local_graph.add_diagnostic(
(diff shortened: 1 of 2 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -482,17 +488,37 @@ impl<'a> RubyIndexer<'a> {
...
                 if arg.as_self_node().is_some() {
+                    if lexical_nesting_id.is_none() {
+                        self.local_graph.add_diagnostic(
+                            Diagnostics::TopLevelMixinSelf,
+                            Offset::from_prism_location(&arg.location()),
+                            "Top level mixin self".to_string(),
+                        );
+
+                        return None;
+                    }
```
vinistock, line 507: Could we consider this a circular mixin error and then already handle this case:

```ruby
module Foo
  include self
end
```

Or is there a difference for when it happens in the top level? Although, we do need to allow `extend self` as that one is valid.

Morriar: At the top level you get an error for anything using `self`:

```rb
include self # => wrong argument type Object (expected Module) (TypeError)
prepend self # => wrong argument type Object (expected Module) (TypeError)
extend self # => wrong argument type Object (expected Module) (TypeError)
```

which is different than circular dependencies:

```rb
module Bar
  include self # => cyclic include detected (ArgumentError)
end
```

So I was aiming at surfacing the first problem during `indexing` because we don't create any definition and handling circular dependencies during `resolution` since we also need to handle a similar case:

```rb
module Bar
  include Bar # => cyclic include detected (ArgumentError)
end
```

PR #419, opened 2026-01-07
Fix class variable indexing and resolution

st0012:
Class variables, unlike instance variables, only follow lexical nesting, not the method receiver/singleton class context.

### DIFF
--- a/docs/ruby-behaviors.md
+++ b/docs/ruby-behaviors.md
@@ -416,6 +416,60 @@ $foo, $bar, $baz = 1, 2, 3
 FOO, BAR::BAZ, ::BAZ = 3, 4, 5
 ```
 
+### Instance Variables vs Class Variables: Receiver vs Lexical Scoping
+
+A critical difference between instance variables (`@`) and class variables (`@@`) is how they determine ownership when defined inside a method with an explicit receiver:
+
+- **Instance variables** follow the **receiver** (`self` at runtime)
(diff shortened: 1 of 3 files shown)


docs/ruby-behaviors.md
```diff
@@ -416,6 +416,60 @@ $foo, $bar, $baz = 1, 2, 3
...
+- **Instance variables** follow the **receiver** (`self` at runtime)
+- **Class variables** follow **lexical scope** (where the code is written)
+
+```ruby
+class Foo; end
+
+class Bar
+  # This defines a singleton method on Foo, but lexically inside Bar
+  def Foo.demo
+    @ivar = "instance var"   # Belongs to Foo (the receiver)
```
```suggestion
    @ivar = "instance var"   # Belongs to <Foo> (the receiver)
```

rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -478,21 +504,29 @@ impl<'a> RubyIndexer<'a> {
         Some(definition_id)
     }
 
+    /// Returns the definition ID of the current owner (class, module, or singleton class),
+    /// skipping methods in the definitions stack.
+    fn current_owner_definition_id(&self) -> Option<DefinitionId> {
```
vinistock: Maybe we can call this `current_nesting_definition_id` as in #402 I'm introducing the concept of things that can be owners, but not lexical scopes.

Issue #91, opened 2025-08-12
Start indexing module_function


vinistock:
There's useful context in https://github.com/Shopify/ruby-lsp/pull/3018. The `module_function` method either declares a section for module functions or accepts a method definition (the returned symbol). When a method gets marked as a module_function, the module gets a singleton version of the method and the original instance method becomes private.

For example:

```ruby
module Foo
  module_function

  def bar; end
end

Foo.bar # the public module function bar 👍 

class Something
  include Foo
end

Something.new.bar # => error, can't invoke private method bar
```

Invoking `module_function` in a class raises an error, so we can just ignore it.

PR #420, opened 2026-01-07
Track visibility inside singleton classes.

Morriar:
One more step towards #91.

This time we handle visibility inside `class << self`.

Also fix a bug where we were always pushing a new visibility instead of replacing the last one:

```rb
class Foo
  public # push a new one to the stack
  private # replace the current visibility instead of pushing
end
```

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -709,15 +709,13 @@ impl Visit<'_> for RubyIndexer<'_> {
 
             self.add_member_to_current_owner(definition_id);
 
-            self.definitions_stack.push(definition_id);
-            self.visibility_stack.push(Visibility::Public);
-
             if let Some(body) = node.body() {
+                self.definitions_stack.push(definition_id);
+                self.visibility_stack.push(Visibility::Public);
                 self.visit(&body);
+                self.visibility_stack.pop();
+                self.definitions_stack.pop();
             }
-
-            self.visibility_stack.pop();
-            self.definitions_stack.pop();
         }
     }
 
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -1122,11 +1120,29 @@ impl Visit<'_> for RubyIndexer<'_> {
                     return;
                 }
 
-                self.visibility_stack.push(Visibility::from_string(message.as_str()));
-
                 if let Some(arguments) = node.arguments() {
```
alexcrocha, line 1125: Consider extracting `Visibility::from_string(message.as_str())` to a variable before the `if let` to avoid repeating the conversion in both branches and improve readability a bit:
```suggestion
                let visibility = Visibility::from_string(message.as_str());
                
                if let Some(arguments) = node.arguments() {
```

Also, a personal preference, `match node.arguments() {` would work really well here.

rust/saturn/src/indexing/ruby_indexer.rs
alexcrocha, line 2460: We need to fix this test as well. It's a coincidence that the code example was masking this bug and allowing the test to pass. Removing the last `protected` should do it.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: I pulled the `Visibility::from_string(message.as_str())` out of the branches, but I kept the `if`.

rust/saturn/src/indexing/ruby_indexer.rs
Morriar: Good catch, I changed the last `protected` into a `public` to better show the behavior.

PR #421, opened 2026-01-08
Add nesting enum for indexer

vinistock:
This PR extracts just the new `Nesting` enum portion of #402, to make it a bit easier to handle the existing conflicts with main.

Basically, the `definitions_stack` becomes a `nesting_stack`, where we can store the IDs with a bit of differentiation and perform the right behaviour:

- LexicalScope: anywhere the `class` or `module` keywords get used, which produces a new lexical scope that binds constants and class variables
- Owner: definitions that can own things, but without producing a new lexical scope (like the ones created by `Class.new`)
- Method: a method. Just used for instance variables, so that we know we're inside of a method

### DIFF
--- a/rust/saturn/src/indexing/ruby_indexer.rs
+++ b/rust/saturn/src/indexing/ruby_indexer.rs
@@ -25,6 +25,26 @@ enum MixinType {
     Extend,
 }
 
+enum Nesting {
+    /// Nesting stack entries that produce a new lexical scope to which constant references must be attached to (i.e.:
+    /// the class and module keywords). All lexical scopes are also owner, but the opposite is not true
+    LexicalScope(DefinitionId),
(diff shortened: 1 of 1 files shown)


rust/saturn/src/indexing/ruby_indexer.rs
```diff
@@ -25,6 +25,26 @@ enum MixinType {
     Extend,
 }
 
+enum Nesting {
+    /// Nesting stack entries that produce a new lexical scope to which constant references must be attached to (i.e.:
+    /// the class and module keywords). All lexical scopes are also owner, but the opposite is not true
+    LexicalScope(DefinitionId),
+    /// An owner entry that will be associated with all members encountered, but will not produce a new lexical scope
+    /// (e.g.: Module.new or Class.new)
+    #[allow(dead_code)]
```
vinistock, line 34: This is dead just until the work on the other PR is complete.

PR #424, opened 2026-01-08
Put `ClassDefinition` impl next to struct

vinistock:
Another extraction to simplify #402

The impl block for `ClassDefinition` was the only one that was not next to its struct. This PR just moves it next to it.

### DIFF
--- a/rust/saturn/src/model/definitions.rs
+++ b/rust/saturn/src/model/definitions.rs
@@ -156,6 +156,82 @@ pub struct ClassDefinition {
     mixins: Vec<Mixin>,
 }
 
+impl ClassDefinition {
+    #[must_use]
+    pub const fn new(
+        name_id: NameId,
+        uri_id: UriId,
+        offset: Offset,
+        comments: Vec<Comment>,
+        lexical_nesting_id: Option<DefinitionId>,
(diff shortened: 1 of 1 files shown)


rust/saturn/src/model/definitions.rs
```diff
@@ -308,25 +313,20 @@ impl ClassDefinition {
...
-
     #[must_use]
     pub fn mixins(&self) -> &[Mixin] {
         &self.mixins
     }
 
+    pub fn add_member(&mut self, member_id: DefinitionId) {
+        self.members.push(member_id);
+    }
+
```
Morriar, line 329: Can we preserve the order so `add_member` comes right after `members`?

st0012: Oops I didn't know auto-merge was enabled

vinistock: `add_member` does come after `members`. The diff on GitHub is weird, all I did was move the entire impl block as is.

Morriar: It's in SingletonClass: https://github.com/Shopify/saturn/pull/434

PR #425, opened 2026-01-08
refactor declaration.rs to change conventions

jesse-shopify:
- remove macros
- remove individual classes as they currently do not have data or behavior that differs from their shared components - when/if they do they can be re-added
- use enum Declaration to separate types
- implement helper methods to get at useful structs on enum so match statements don't proliferate. can use:
    - `as_simple`
    - `as_simple_mut`
    - `as_namespace`
    - `as_namespace_mut`

### DIFF
--- a/rust/rubydex/Cargo.toml
+++ b/rust/rubydex/Cargo.toml
@@ -15,6 +15,7 @@ crate-type = ["rlib"]
 test_utils = ["dep:tempfile"]
 
 [dependencies]
+parking_lot = "0.12"
 ruby-prism = "1.6.0"
 url = "2.5.4"
 xxhash-rust = { version = "0.8.15", features = ["xxh3"] }
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -109,6 +109,13 @@ impl SimpleDeclaration {
         self.references.push(id);
     }
 
+    pub fn remove_reference(&mut self, reference_id: &ReferenceId) {
+        if let Some(pos) = self.references.iter().position(|id| id == reference_id) {
+            self.references.swap_remove(pos);
+            self.references.shrink_to_fit();
+        }
+    }
+
     pub fn remove_definition(&mut self, definition_id: &DefinitionId) -> bool {
         if let Some(pos) = self.definition_ids.iter().position(|id| id == definition_id) {
             self.definition_ids.swap_remove(pos);
@@ -351,6 +358,10 @@ impl Declaration {
 
     pub fn add_reference(&mut self, id: ReferenceId) {
         self.as_simple_mut().add_reference(id);
+    }
+
+    pub fn remove_reference(&mut self, reference_id: &ReferenceId) {
+        self.as_simple_mut().remove_reference(reference_id);
     }
 
     // Deletes a definition from this declaration
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -1,4 +1,4 @@
-use std::sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard};
+use parking_lot::{Mutex, MutexGuard, RwLock, RwLockReadGuard};
 
 use crate::model::{
     identity_maps::{IdentityHashMap, IdentityHashSet},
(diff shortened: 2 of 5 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -213,14 +213,11 @@ impl Graph {
...
         if let Some(declaration) = write_lock.get_mut(owner_id) {
-            match declaration {
-                Declaration::Class(it) => it.add_member(member_str_id, member_declaration_id),
-                Declaration::Module(it) => it.add_member(member_str_id, member_declaration_id),
-                Declaration::SingletonClass(it) => it.add_member(member_str_id, member_declaration_id),
-                Declaration::Constant(_) => {
-                    // TODO: temporary hack to avoid crashing on `Struct.new`, `Class.new` and `Module.new`
-                }
-                _ => panic!("Tried to add member to a declaration that isn't a namespace"),
+            if let Some(namespace) = declaration.as_namespace_mut() {
```
jesse-shopify, line 216: cleaner, I think

rust/rubydex/src/resolution.rs
```diff
@@ -1067,12 +1055,9 @@ fn search_top_level(graph: &Graph, str_id: StringId) -> Outcome {
...
 fn members_of(decl: &Declaration) -> &IdentityHashMap<StringId, DeclarationId> {
-    match decl {
-        Declaration::Class(it) => it.members(),
-        Declaration::SingletonClass(it) => it.members(),
-        Declaration::Module(it) => it.members(),
-        _ => panic!("Tried to get members for a declaration that isn't a namespace"),
-    }
+    decl.as_namespace()
+        .expect("Tried to get members for a declaration that isn't a namespace")
+        .members()
```
jesse-shopify, line 1077: cleaner

rust/rubydex/src/model/declaration.rs
```diff
@@ -53,256 +53,317 @@ impl<'a> IntoIterator for &'a Ancestors {
...
-            pub fn descendants(&self) -> Ref<'_, IdentityHashSet<DeclarationId>> {
-                self.descendants.borrow()
-            }
-        }
-    };
+/// Struct for namespace-like declarations such as classes and modules
+#[derive(Debug)]
+pub struct NamespaceDeclaration {
+    /// The base simple declaration fields
+    simple: SimpleDeclaration,
```
Morriar, line 154: I tried this design earlier in the project and having this extra field for all definitions resulted in a lot more memory usage.

Could you compare `main` against your branch with `--stats` on a few corpus sizes?

You can produce test corpuses with https://github.com/Shopify/saturn/blob/main/utils/generate-corpus#L11.

rust/rubydex/Cargo.toml
```diff
@@ -15,6 +15,7 @@ crate-type = ["rlib"]
 test_utils = ["dep:tempfile"]
 
 [dependencies]
+parking_lot = "0.12"
```
jesse-shopify, line 18: Added due to better lock API

rust/rubydex/src/model/declaration.rs
jesse-shopify: The data is below. There was no change in memory or runtime performance.

Morriar: In terms of API, I'm not a fan of having to go through `simple` depending on the field I want to access.

On the other hand, I do like the simplification around handling namespace related fields and avoiding the matches/panic. Could we get the same result with nested enums instead?

jesse-shopify: The thing about Rust is that there are many ways to get to the same place and the only different is ergonomics. Nested enums could work. Another option to improve ergonomics might be Using `Deref` / `DerefMut` traits. Another option would be to have a SimpleDeclaration that has an optional, boxed NamespaceDeclaration, though I suppose that would add an extra 8 bytes to SimpleDeclaration but then Declaration might not need to box which could be a performance and memory win.

Morriar: I tried nested enums in https://github.com/Shopify/rubydex/pull/470, it's not horrible.

jesse-shopify:
Closing as stale.

PR #426, opened 2026-01-08
Start testing diagnostics during resolution

Morriar:
Step towards https://github.com/Shopify/saturn/issues/354, we make sure we don't already have diagnostics for the resolution tests.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1306,6 +1306,7 @@ fn set_singleton_class_id(declaration: &mut Declaration, id: DeclarationId) {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::diagnostic::Rule;
     use crate::model::ids::UriId;
     use crate::test_utils::GraphTest;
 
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1610,22 +1610,27 @@ mod tests {
         }};
     }
 
+    fn format_diagnostics(context: &LocalGraphTest) -> Vec<String> {
```
vinistock, line 1892: How about we implement either `Display`, `Debug` or pretty print for `Diagnostic`?

That way, we get the printing by default and it reduces some of the duplication between indexing and resolution tests.

rust/rubydex/src/resolution.rs
```diff
@@ -1462,29 +1462,71 @@ mod tests {
...
     fn resolving_top_level_references() {
         let mut context = GraphTest::new();
         context.index_uri("file:///bar.rb", {
             r"
             class Bar; end
 
-            ::Bar
-            Bar
+            puts ::Bar
+            puts Bar
```
vinistock: Are the puts all to avoid the `Constant in void context` warnings from Prism?

If so, what do you think of automatically filtering based on code, so that we only assert against resolution level diagnostics?

Right now, adding the `puts` is enough, but as we get more and more diagnostics, writing code that generates zero of them might pollute examples with a lot of code that's unrelated to what's being tested.

st0012: Yeah perhaps the macro can take an optional exceptions list `assert_no_diagnostics!(&context, <exceptions>);`?

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: I don't think we can use either because we need to pass the `source` to be able to translate the byte offset into a line/column

rust/rubydex/src/resolution.rs
Morriar: I like it, updated 👍

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: Ah, right. How about a regular method to reduce the duplication then?

```rust
fn formatted(&self, source: &str) -> String {
  format!(
    "{}: {} ({})",
    self.severity.as_str(),
    self.message,
    self.offset.to_display_range(source)
  )
}
```

PR #435, opened 2026-01-09
Add resolution diagnostic for declaration kind mismatch

Morriar:
One more step towards https://github.com/Shopify/saturn/issues/354.

This required a way to assert on the kind of the declaration. I added a DeclarationKind enum than can be used to both `build` a new declaration (which removes the need for closures in the resolution) and know what kind an existing declaration  is.

Another subtlety is to which definition location to attach the error. Because declarations are created in a specific order that may not match their order in the code, we need to add some logic so the error is displayed on the last introduced one rather than the first one.

This PR is easier to review commit by commit.

### DIFF
--- a/rust/rubydex/src/diagnostic.rs
+++ b/rust/rubydex/src/diagnostic.rs
@@ -20,11 +20,6 @@ impl Diagnostic {
     }
 
     #[must_use]
-    pub fn make(rule: Rule, uri_id: UriId, offset: Offset, message: String) -> Self {
-        Self::new(rule, uri_id, offset, message)
-    }
-
-    #[must_use]
     pub fn rule(&self) -> &Rule {
         &self.rule
     }
--- a/rust/rubydex/src/indexing/local_graph.rs
+++ b/rust/rubydex/src/indexing/local_graph.rs
@@ -131,7 +131,7 @@ impl LocalGraph {
     }
 
     pub fn add_diagnostic(&mut self, rule: Rule, offset: Offset, message: String) {
-        let diagnostic = Diagnostic::make(rule, self.uri_id, offset, message);
+        let diagnostic = Diagnostic::new(rule, self.uri_id, offset, message);
         self.diagnostics.push(diagnostic);
     }
 


rust/rubydex/src/model/graph.rs
```diff
@@ -161,11 +219,18 @@ impl Graph {
...
     }
 
+    // Diagnostics
+
     #[must_use]
     pub fn diagnostics(&self) -> &Vec<Diagnostic> {
         &self.diagnostics
     }
 
+    pub fn add_diagnostic(&mut self, diagnostics: Diagnostics, uri_id: UriId, offset: Offset, message: String) {
```
st0012: This is not used?

rust/rubydex/src/diagnostic.rs
```diff
@@ -111,4 +111,5 @@ diagnostics! {
     (3004, TopLevelMixinSelf, Severity::Warning);
 
     // 4000 - Resolution errors
+    (4001, KindRedefinition, Severity::Error);
```
st0012: Does this mean, if a library of my project has

```rb
# a fixture file bundled in the gem
# not required by my project
class User; end
```

And in my project, I define

```rb
module User
  class Something; end
end
```

I'll then get an error always show up in Ruby LSP?

Morriar: Yes, which should be the same with Sorbet until you ignore the `fixtures` directory.

rust/rubydex/src/model/graph.rs
Morriar: Introduced too early, I can remove it from here 👍

rust/rubydex/src/diagnostic.rs
st0012: But unlike Sorbet, rubydex under the usage of Ruby LSP will index dependencies, regardless whether they'd be required or not (which is different then gem rbi Tapioca generates through runtime). So in this case the chance of having this error will be much higher. And when that happens, users will need to exclude specific gem file/folder, which is harder to configure.
I think we should be more conservative on showing this as errors. IMO warning would be better.

vinistock: My 2 cents: I think this should be an error. We only index gems' require paths. If a gem includes a fixture as part of its require path with code that conflicts with the real gem's implementation, I would argue that's a mistake in the gem.

Re-defining a class/module type crashes in the runtime, so I would stick with error.

st0012: Oh I didn't know we only index the gem's require path. Where is this implemented?
If that's the case, then I'm good with this.

vinistock: This exact line https://github.com/Shopify/rubydex/blob/c8c27f8230906a7663bd802c22cda138bc03bfe2/lib/rubydex/graph.rb#L34

st0012: What about cases like the [recent RDoc patch](https://github.com/ruby/rdoc/blob/master/lib/rdoc/markup/heading.rb#L8-L30)? If we introduce `Struct` as a new type of definition, wouldn't that mean RDoc would always trigger this error?

vinistock: There are two parts to the answer here. For `Struct.new`, it returns a class. We just need special handling for it, which I'm working on #392.

The second aspect I realized is that we actually need to be able to promote declarations since we can't control what kind of meta-programming a gem or project might be using. For example:

```ruby
# Create a class via meta-programming. Rubydex has no idea and thinks this is
# a constant
Foo = meta_programming_class do
  # ....
end

# Then re-open the class and perform operations that are only valid in a namespace
# context, like including a module
class Foo
  include Bar
end
```

At the moment, we would be crashing in this case, so I have a branch where I'm exploring automatic promotion of declarations. Basically, if we find something that was a constant and then gets re-opened as a class, the most correct behaviour might be to promote it.

However, we can't allow redefinition from a class to a module. That's always an error.

Morriar: @st0012 raised interesting questions and I think we can do better, please take a look at https://github.com/Shopify/rubydex/pull/450.

Morriar: I rebased this work using the rules from #450.

rust/rubydex/src/model/graph.rs
```diff
@@ -73,14 +73,71 @@ impl Graph {
...
+                    let old_definition = self.definitions.get(old_definition_id).unwrap();
+                    let new_definition = self.definitions.get(&definition_id).unwrap();
+
+                    let mut definitions = [
+                        (old_definition, old_kind.as_str()),
+                        (new_definition, declaration_kind.as_str()),
+                    ];
+
+                    // We want to show the diagnostic for the most recent definition, so we sort by uri and offset
+                    definitions.sort_by_key(|(d, _)| {
```
vinistock, line 108: Considering we can have definitions across files that we cannot know the load order, does sorting really make a difference?

I'd say we can pick one to add the diagnostic and then we use [related information](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticRelatedInformation) to associate the diagnostic with all other definitions.

rust/rubydex/src/resolution.rs
```diff
@@ -140,24 +136,16 @@ fn handle_definition_unit(
 ) {
     let outcome = match graph.definitions().get(&id).unwrap() {
         Definition::Class(class) => {
-            handle_constant_declaration(graph, *class.name_id(), id, false, |name, owner_id| {
-                Declaration::Class(Box::new(ClassDeclaration::new(name, owner_id)))
-            })
+            handle_constant_declaration(graph, *class.name_id(), &DeclarationKind::Class, id, false)
```
vinistock, line 139: This refactor will conflict with your other refactor idea that requires enqueueing an ancestor linearization unit only when the closure to produce a new declaration gets invoked.

It might be worth splitting this refactor and considering it as part of the other one.

PR #439, opened 2026-01-12
Regenerate the EasyCompile workflow

Edouard-chin:
Now that the tool got renamed from "EasyCompile" to "Cibuildgem" and its now released, we have to regenerate the template (I ran `cibuildgem ci_template` to regenerate it).

### DIFF
--- a/.github/workflows/cibuildgem.yaml
+++ b/.github/workflows/cibuildgem.yaml
@@ -1,4 +1,4 @@
-name: "Release gems with precompiled binaries"
+name: "Package and release gems with precompiled binaries"
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/cibuildgem.yaml
```diff
@@ -23,8 +23,8 @@ jobs:
         with:
           ruby-version: "3.2.9"
           bundler-cache: true
-      - name: "Run easy compile"
-        uses: "shopify-playground/edouard-playground/.github/actions/easy_compile@main"
+      - name: "Run cibuildgem"
```
vinistock, line 26: Should the Ruby version above be a matrix too? Or is the idea to compile agains the oldest supported version?

Edouard-chin: We don't need a matrix. We just need a runtime ruby (we pick the oldest version that is supported by the gem), then the tool will take care of downloading other ruby versions that the gem supports and compile against them.

Issue #330, opened 2025-11-20
Support consuming document changes incrementally


vinistock:
We're getting close to the design of the resolution phase working properly and now a new problem arises: we are only able to take all of the information we indexed and transform it in one go, but we have no way of accounting for incremental changes.

To support usage in LSP mode (and in the future, incremental mode through the database), we need to be able to consume changes incrementally.

As part of this issue, let's understand how we can implement this.

### Notes

The unit of indexing or consuming changes is always the same: a document. The incremental aspect comes from understanding what kind of change happened and matching that to what needs to happen in the subsequent parts of the analysis.

Some examples:

- If someone just adds a bunch of blank lines in a file, we need to index it again (to update the offsets), but there's no need to discard declaration level information or run resolution again
- If someone adds new declarations that don't impact constant and ancestor resolution (like a new method), maybe we can skip most of the resolution phase
- If someone deletes a constant or changes inheritance, we need to try to compute the minimum amount of work that has to happen to account for it

My understanding is that most type checkers use hashed values for this. For example, if we hash all definitions that we got in a local graph from indexing a single document and that doesn't match the previous state, then there are new/modified definitions (maybe we need to remember the definition hashes in `Document`?). Maybe we need an ancestors hash?

vinistock:
Partially complete.

PR #440, opened 2026-01-12
Remove references when document changes

thomasmarshall:
This PR is part of #330. It ensures we remove method and constant references whenever a document changes. In the future, we'd like to make invalidation more granular, but for the first version we will just invalidate/remove everything that may be associated with the document and then reindex and perform resolution.

As part of this change, I added tracking of method/constant references in `Document` so we know which ones need to be removed from the graph. I also changed `Declaration` to store `references` as a set so we can more efficiently remove the ones belonging to the document that was changed. However, I'm not sure if this is the right approach.

The `remove_definitions_for_uri` function is updated to remove all method references from the graph that were associated with the doc (to be reindexed later). As far as I understand, method references do not have resolved names, only . Is that because the resolution work is still in progress for methods?

It then does the same for constant references, except because these are resolved, we also need to remove any dangling references from declarations by looking them up via the resolved name.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -296,16 +296,24 @@ impl Graph {
             return;
         };
 
+        let mut write_lock = self.declarations.write().unwrap();
+
+        // TODO: Remove method references from method declarations once method inference is implemented
         for ref_id in document.method_references() {
             self.method_references.remove(ref_id);
         }
+
         for ref_id in document.constant_references() {
-            self.constant_references.remove(ref_id);
+            if let Some(constant_ref) = self.constant_references.remove(ref_id)
+                && let Some(NameRef::Resolved(resolved)) = self.names.get(constant_ref.name_id())
+                && let Some(declaration) = write_lock.get_mut(resolved.declaration_id())
+            {
+                declaration.remove_reference(ref_id);
+            }
         }
 
         // Vector of (owner_declaration_id, member_name_id) to delete after processing all definitions
         let mut members_to_delete: Vec<(DeclarationId, StringId)> = Vec::new();
-        let mut write_lock = self.declarations.write().unwrap();
 
         for def_id in document.definitions() {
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -296,9 +296,24 @@ impl Graph {
             return;
         };
 
+        let mut write_lock = self.declarations.write().unwrap();
+
+        // TODO: Remove method references from method declarations once method resolution is implemented
```
thomasmarshall: I don't know if this is an accurate TODO or if I misunderstand if/how method references are linked to method declarations.

vinistock: Method references are implemented partially at the moment. We never link them to any declarations.

The context behind this is the following: we added this to be able to support Spoom's dead code detection, which currently doesn't take resolution data like inheritance into account.

Method references are significantly more complex than constant references because you need to type check the receiver to figure out what method you are actually referring to - which won't be in place for a while.

For now, just removing them from the list like you did is enough. The text would be slightly different though because resolution cannot figure out what a method refers to.


```suggestion
        // TODO: Remove method references from method declarations once method inference is implemented
```

thomasmarshall: Thanks for the explanation, I've updated the comment.

rust/rubydex/src/model/document.rs
```diff
@@ -52,4 +74,17 @@ mod tests {
...
+        let mut document = Document::new("file:///foo.rb".to_string());
+        let method_ref = ReferenceId::new(1);
+        let constant_ref = ReferenceId::new(2);
+
+        document.add_method_reference(method_ref);
+        document.add_constant_reference(constant_ref);
+
+        assert_eq!(document.method_references(), &[method_ref]);
+        assert_eq!(document.constant_references(), &[constant_ref]);
+    }
```
Morriar, line 89: Not sold on the value of this test.

PR #441, opened 2026-01-12
Remove diagnostics when documents change

thomasmarshall:
This PR is part of #330. When a document changes or is removed, we want to clear/invalidate any parts of the graph that were created by or associated to that document. Eventually we will implement a more granular approach that clears/invalidates as little as possible, but for now we'll blow away everything to do with the document.

This ensures the diagnostics related to a document are removed when it is changed or deleted. It does so by moving the diagnostics from the top-level of the graph to be stored in the `Document` itself. @vinistock suggested this approach, as there is currently nothing else in the graph that connects to diagnostics and they are always (currently?) associated to a single document.

### DIFF
--- a/rust/rubydex/src/model/document.rs
+++ b/rust/rubydex/src/model/document.rs
@@ -1,4 +1,4 @@
-use crate::model::ids::{DefinitionId, ReferenceId};
+use crate::model::ids::{DefinitionId, DiagnosticId, ReferenceId};
 
 // Represents a document currently loaded into memory. Identified by its unique URI, it holds the edges to all
 // definitions and references discovered in it
@@ -8,6 +8,7 @@ pub struct Document {
     definition_ids: Vec<DefinitionId>,
     method_reference_ids: Vec<ReferenceId>,
     constant_reference_ids: Vec<ReferenceId>,
+    diagnostic_ids: Vec<DiagnosticId>,
 }
 
 impl Document {
@@ -18,6 +19,7 @@ impl Document {
             definition_ids: Vec::new(),
             method_reference_ids: Vec::new(),
             constant_reference_ids: Vec::new(),
+            diagnostic_ids: Vec::new(),
         }
     }
 
@@ -56,6 +58,15 @@ impl Document {
 
     pub fn add_constant_reference(&mut self, reference_id: ReferenceId) {
         self.constant_reference_ids.push(reference_id);
+    }
+
+    #[must_use]
+    pub fn diagnostic_ids(&self) -> &[DiagnosticId] {
+        &self.diagnostic_ids
+    }
+
+    pub fn add_diagnostic_id(&mut self, diagnostic_id: DiagnosticId) {
+        self.diagnostic_ids.push(diagnostic_id);
     }
 }
 
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -7,7 +7,7 @@ use crate::model::declaration::Declaration;
 use crate::model::definitions::Definition;
 use crate::model::document::Document;
 use crate::model::identity_maps::IdentityHashMap;
-use crate::model::ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId, UriId};
+use crate::model::ids::{DeclarationId, DefinitionId, DiagnosticId, NameId, ReferenceId, StringId, UriId};
 use crate::model::name::{NameRef, ResolvedName};
 // use crate::model::integrity::IntegrityChecker;
 use crate::model::references::{ConstantReference, MethodRef};
@@ -38,8 +38,7 @@ pub struct Graph {
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     // Map of method references that still need to be resolved
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
-
-    diagnostics: Vec<Diagnostic>,
+    diagnostics: IdentityHashMap<DiagnosticId, Diagnostic>,
 }
 
 impl Graph {
@@ -54,7 +53,7 @@ impl Graph {
             names: IdentityHashMap::default(),
             constant_references: IdentityHashMap::default(),
             method_references: IdentityHashMap::default(),
-            diagnostics: Vec::new(),
+            diagnostics: IdentityHashMap::default(),
         }
     }
 
@@ -162,8 +161,8 @@ impl Graph {
     }
 
     #[must_use]
-    pub fn diagnostics(&self) -> &Vec<Diagnostic> {
-        &self.diagnostics
+    pub fn diagnostics(&self) -> Vec<&Diagnostic> {
+        self.diagnostics.values().collect()
     }
 
     /// # Panics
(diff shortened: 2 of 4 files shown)


rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -126,13 +123,13 @@ impl LocalGraph {
...
-    pub fn diagnostics(&self) -> &Vec<Diagnostic> {
-        &self.diagnostics
+    pub fn diagnostics(&self) -> &[Diagnostic] {
+        self.document.diagnostics()
     }
 
     pub fn add_diagnostic(&mut self, diagnostics: Diagnostics, offset: Offset, message: String) {
         let diagnostic = Diagnostic::make(diagnostics, self.uri_id, offset, message);
-        self.diagnostics.push(diagnostic);
+        self.document.add_diagnostic(diagnostic);
```
Morriar: Maybe we can delete these methods to make it clearer at this stage the diagnostics are in the `document`?

rust/rubydex/src/model/document.rs
```diff
@@ -87,4 +99,25 @@ mod tests {
...
+            Offset::new(0, 10),
+            "test error".to_string(),
+        );
+
+        document.add_diagnostic(diagnostic);
+
+        assert_eq!(document.diagnostics().len(), 1);
+        assert_eq!(document.diagnostics()[0].message(), "test error");
+        assert!(matches!(document.diagnostics()[0].severity(), Severity::Error));
+    }
```
Morriar: I'm not sold on the value of this test?

rust/rubydex/src/model/graph.rs
```diff
@@ -38,8 +38,6 @@ pub struct Graph {
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     // Map of method references that still need to be resolved
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
-
-    diagnostics: Vec<Diagnostic>,
```
Morriar, line 42: This is going to be a problem.

Indexing diagnostics are indeed related to a specific document but resolution may not. They are more "global".

Take this example:

```rb
# file1.rb
class Foo < Bar; end
```

```rb
# file2.rb
class Bar < Foo; end
```

We have a circular dependency diagnostic and both documents participate to the error.

Let's say we put the error in `file1.rb`, then you update `file2.rb` to change the code to 

```rb
# file2.rb
class Bar; end
```

You'll have to figure out you need to discard the diagnostic in `file1.rb`.

vinistock: That's a good point. Maybe we need diagnostics to be a top level node in the graph so that multiple other nodes can have edges to it?

For example, maybe indexing diagnostics get associated with documents and resolution diagnostics with declarations.

thomasmarshall: Thanks both. I pushed an alternative approach that keeps diagnostics at the top level, but indexed by `DiagnosticId` for lookup purposes.

I used URI/offset/code/message to generate the ID. However, this is not guaranteed to be unique by Prism which we use for parse error diagnostics. In [some cases](https://github.com/ruby/prism/pull/3827) Prism has previously attached identical looking errors at the same location (even though they stem from different issues in the source code). Perhaps that would be considered a Prism bug and we don't care about attaching multiple instances of seemingly identical errors?

> maybe indexing diagnostics get associated with documents and resolution diagnostics with declarations

Do you think they would be linked to the document as well, or just the declaration? Will there always be a suitable declaration for resolution issues?

Another option would be to have a `DiagnosticGroupId` which links together diagnostics for problems that span multiple documents. We could have an `IdentityHashMap<DiagnosticGroupID, Vec[DiagnosticId]>` index in the graph and each `Diagnostic` could store an optional `group_id`. In the cyclic references example there would be a separate `Diagnostic` in both `file1.rb` and `file2.rb` both with the same `group_id`.

rust/rubydex/src/model/document.rs
thomasmarshall: Yeah that's fair, I wasn't sure either. Removed now :+1:

rust/rubydex/src/model/graph.rs
Morriar: What if we attached the resolution diagnostics to the declarations instead of the graph?

```rs
declaration.add_diagnostic(...)
```

That way when we remove a definition we also reset the declaration diagnostics before running the resolution again?

vinistock: ☝️ that's what I was thinking too

> For example, maybe indexing diagnostics get associated with documents and resolution diagnostics with declarations.

I believe if we make diagnostics a hashmap and allow documents and declarations to associate with them, we can account for indexing and resolution cases.

We can probably make this PR just about indexing ones though and then follow up with the resolution part.

rust/rubydex/src/model/graph.rs
```diff
@@ -162,8 +161,8 @@ impl Graph {
     }
 
     #[must_use]
-    pub fn diagnostics(&self) -> &Vec<Diagnostic> {
-        &self.diagnostics
+    pub fn diagnostics(&self) -> Vec<&Diagnostic> {
```
vinistock, line 164: We can probably change this one to return the `IdentityHashMap`, so that it matches the readers we have for the other hashmaps too.

rust/rubydex/src/model/graph.rs
```diff
@@ -266,16 +265,20 @@ impl Graph {
     /// Merges everything in `other` into this Graph. This method is meant to merge all graph representations from
     /// different threads, but not meant to handle updates to the existing global representation
     pub fn extend(&mut self, local_graph: LocalGraph) {
-        let (uri_id, document, definitions, strings, names, constant_references, method_references, diagnostics) =
+        let (uri_id, mut document, definitions, strings, names, constant_references, method_references, diagnostics) =
             local_graph.into_parts();
 
+        for diagnostic in diagnostics {
```
vinistock, line 271: I think we should register the diagnostics inside of the documents during indexing using the `LocalGraph`. That way, this method doesn't change.

rust/rubydex/src/model/graph.rs
thomasmarshall: I have a branch [here](https://github.com/Shopify/rubydex/compare/remove-diagnostics-alt?w=1) that tries that approach, let me know if I've misunderstood:

- Parsing and indexing errors will be attached to `Document`
- Resolution errors will be attached to `Declaration`
- Parsing and indexing errors will be cleared automatically when the document is removed
- Resolution errors will be cleared from any declaration that the document contributes to via its definitions

Does that seem better? In that approach diagnostics are removed from the top-level of the graph and there is no ID lookup. What do you both think?

Morriar: > Parsing and indexing errors will be attached to Document

Yes.

> Resolution errors will be attached to Declaration

For this one I don't know if it will be enough.

Let's consider this code:

```rb
# file1.rb
puts Foo # error: constant `Foo` doesn't exists
```

This is a resolution error (we need to have tried to resolve `Foo` to know it doesn't exist). But we don't really have a Declaration to attach the error? Or maybe it would be in `Object`?

Then if we later add the constant, the diagnostic should disappear:

```rb
# file2.rb

Foo = 42
```

thomasmarshall: @vinistock and I just discussed that maybe we could attach diagnostics to any of documents, declarations, definitions, and constant references, depending on the type of issue. I am also going to take a look at how ty and Pyre to understand how they do it too.

thomasmarshall: - Pyre attaches diagnostics to modules (files) and it has dependency tracking to determine which modules are impacted by changes in other modules.
- ty attaches diagnostics to different scopes (function bodies, class bodies, modules). I don't quite understand the architecture but it uses [Salsa](https://github.com/salsa-rs/salsa) for incremental computation so diagnostics are invalidated too by that mechanism.

vinistock: IMO, it would make sense to have diagnostics attached to all possible nodes:

- Documents: parse errors
- Definitions: indexing errors like dynamic superclass
- Declarations: resolution errors like superclass mismatch
- Constant references: indexing errors like dynamic mixin and resolution errors like trying to include a class

thomasmarshall: I just opened a draft PR [here](https://github.com/Shopify/rubydex/pull/484) with that alternative approach.

PR #443, opened 2026-01-12
Clear ancestor chains when documents change

thomasmarshall:
Part of #330.

This PR ensures we clear any affected ancestor chains for descendants of declarations linked to definitions removed then a document is removed or changes. For example:

```rb
class Foo
end
```

```rb
class Bar < Foo
end
```

If we modify the first file to delete `Foo` we should invalidate the ancestor chain for the `Bar` declaration, as the `Foo` declaration will no longer exist. Even if we don't delete `Foo`, it might change, for example:

```rb
class Foo
  include Baz
end
```

Here we modify the first file to include `Baz` in `Foo`, so we'll need to relinearize the ancestors for the `Bar` declaration.

Conversely, we need to remove any declaration for which we have cleared its ancestors, from the `descendants` _of_ each of its ancestors. For example:

```rb
class Foo
  include Bar
end
```

```rb
class Foo
end
```

```rb
module Bar
end
```

If the first document changes, we remove the `Foo` definition but not the global `Foo` declaration. We clear the ancestors for `Foo` but we also need to remove `Foo` from the `descendants` of `Bar` because we don't know yet whether or not `Foo` does include `Bar`.

In the future we can be smarter and more granular about which changes should clear/invalidate things from the graph, or require resolution or linearization, but for now we're invalidating everything associated to the document.

### DIFF
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -1,4 +1,4 @@
-use std::sync::{Mutex, MutexGuard, RwLock};
+use std::sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard};
 
 use crate::model::{
     identity_maps::{IdentityHashMap, IdentityHashSet},
@@ -150,6 +150,10 @@ macro_rules! namespace_declaration {
                 *ancestors_lock = ancestors;
             }
 
+            fn ancestors(&self) -> RwLockReadGuard<'_, Ancestors> {
+                self.ancestors.read().unwrap()
+            }
+
             #[must_use]
             pub fn clone_ancestors(&self) -> Ancestors {
                 self.ancestors.read().unwrap().clone()
@@ -165,6 +169,10 @@ macro_rules! namespace_declaration {
 
             pub fn add_descendant(&self, descendant_id: DeclarationId) {
                 self.descendants.lock().unwrap().insert(descendant_id);
+            }
+
+            fn remove_descendant(&self, descendant_id: &DeclarationId) {
+                self.descendants.lock().unwrap().remove(descendant_id);
             }
 
             pub fn descendants(&self) -> MutexGuard<'_, IdentityHashSet<DeclarationId>> {
@@ -311,6 +319,57 @@ impl Declaration {
     #[must_use]
     pub fn unqualified_name(&self) -> String {
         all_declarations!(self, it => it.name.rsplit("::").next().unwrap_or(&it.name).to_string())
+    }
+
+    pub fn remove_descendant(declaration: &Declaration, descendant_id: &DeclarationId) {
+        match declaration {
+            Declaration::Class(c) => c.remove_descendant(descendant_id),
+            Declaration::Module(m) => m.remove_descendant(descendant_id),
+            Declaration::SingletonClass(s) => s.remove_descendant(descendant_id),
+            _ => {}
+        }
+    }
+
+    pub fn for_each_ancestor<F>(declaration: &Declaration, mut f: F)
+    where
+        F: FnMut(&Ancestor),
+    {
+        match declaration {
+            Declaration::Class(c) => c.ancestors().iter().for_each(&mut f),
+            Declaration::Module(m) => m.ancestors().iter().for_each(&mut f),
+            Declaration::SingletonClass(s) => s.ancestors().iter().for_each(&mut f),
+            _ => {}
+        }
+    }
+
+    pub fn for_each_descendant<F>(declaration: &Declaration, mut f: F)
+    where
+        F: FnMut(&DeclarationId),
+    {
+        match declaration {
+            Declaration::Class(c) => c.descendants().iter().for_each(&mut f),
+            Declaration::Module(m) => m.descendants().iter().for_each(&mut f),
+            Declaration::SingletonClass(s) => s.descendants().iter().for_each(&mut f),
+            _ => {}
+        }
+    }
+
+    pub fn clear_ancestors(declaration: &Declaration) {
+        match declaration {
+            Declaration::Class(c) => c.set_ancestors(Ancestors::Partial(vec![])),
+            Declaration::Module(m) => m.set_ancestors(Ancestors::Partial(vec![])),
+            Declaration::SingletonClass(s) => s.set_ancestors(Ancestors::Partial(vec![])),
+            _ => {}
+        }
+    }
+
+    pub fn clear_descendants(declaration: &Declaration) {
+        match declaration {
+            Declaration::Class(c) => c.descendants().clear(),
+            Declaration::Module(m) => m.descendants().clear(),
+            Declaration::SingletonClass(s) => s.descendants().clear(),
+            _ => {}
+        }
     }
 }
 
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -3,10 +3,10 @@ use std::sync::LazyLock;
 
 use crate::diagnostic::Diagnostic;
 use crate::indexing::local_graph::LocalGraph;
-use crate::model::declaration::Declaration;
+use crate::model::declaration::{Ancestor, Declaration};
 use crate::model::definitions::Definition;
 use crate::model::document::Document;
-use crate::model::identity_maps::IdentityHashMap;
+use crate::model::identity_maps::{IdentityHashMap, IdentityHashSet};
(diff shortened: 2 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -17,6 +17,76 @@ pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::
 pub static MODULE_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Module"));
 pub static CLASS_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Class"));
 
+fn remove_descendant(decl: &Declaration, descendant_id: DeclarationId) {
```
thomasmarshall: I'm not sure really where this should live. Any thoughts?

rust/rubydex/src/model/graph.rs
```diff
@@ -17,6 +17,76 @@ pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::
...
+    match decl {
+        Declaration::Class(c) => c.remove_descendant(&descendant_id),
+        Declaration::Module(m) => m.remove_descendant(&descendant_id),
+        Declaration::SingletonClass(s) => s.remove_descendant(&descendant_id),
+        _ => panic!("Tried to remove a descendant from a declaration that isn't a namespace"),
+    }
+}
+
+fn invalidate_ancestor_chains(
```
thomasmarshall: I had to pull this out of `remove_definitions_for_uri` to satisfy clippy.

rust/rubydex/src/model/graph.rs
vinistock: Yeah, it's a similar issue as what we see quite a bit in resolution.

What do you think about making all of these convenience things as associated functions in the `Declaration` struct? That way, it becomes centralized.

Example:

```rust
impl Declaration {
  fn members_of(declaration: &Declaration) -> &IdentityHashMap<...> {

    // do the work here
  }
}

// Usages

Declaration::members_of(&declaration)
```

rust/rubydex/src/model/graph.rs
vinistock: I think splitting is fine, but if we change the implementation slightly to accept `&self`, then we can put this inside of the `Graph`.

What do you think?

rust/rubydex/src/model/graph.rs
```diff
@@ -17,6 +17,76 @@ pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::
...
+            Declaration::Class(c) => {
+                for ancestor in c.ancestors().iter() {
+                    if let Ancestor::Complete(ancestor_id) = ancestor
+                        && let Some(ancestor_decl) = declarations.get(ancestor_id)
+                    {
+                        remove_descendant(ancestor_decl, declaration_id);
+                    }
+                }
+                c.set_ancestors(Ancestors::Partial(vec![]));
+                for descendant_id in c.descendants().iter() {
```
vinistock: I think we need to clear the descendants here as well.

Otherwise, when we linearize the ancestors for the descendants that got impacted, we may have leftovers from before the change.

rust/rubydex/src/model/graph.rs
```diff
@@ -616,6 +697,49 @@ mod tests {
...
+        }
+
+        context.index_uri("file:///a.rb", "");
+
+        let read_lock = context.graph().declarations.read().unwrap();
+
+        let Declaration::Class(foo) = read_lock.get(&DeclarationId::from("Foo")).unwrap() else {
+            panic!("Expected Foo to be a class");
+        };
+        assert!(matches!(*foo.ancestors(), Ancestors::Partial(ref a) if a.is_empty()));
```
vinistock: Let's also assert that the descendant. They get re-populated once resolution runs again.

rust/rubydex/src/model/graph.rs
```diff
@@ -17,6 +17,76 @@ pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::
...
+    while let Some(declaration_id) = queue.pop() {
+        if !visited.insert(declaration_id) {
+            continue;
+        }
+
+        let Some(decl) = declarations.get(&declaration_id) else {
+            continue;
+        };
+
+        match decl {
```
vinistock: I'm trying to think if there's a way to reduce the duplication on the different branches, but I'm honestly not finding many great ideas.

Maybe we can have a method that moves the ancestor data out of `Declaration` for processing? Something like this:

```rust
let (ancestors, descendants) = match decl {
  Declaration::Class(c) => c.into_ancestor_data(),
  Declaration::Module(m) => m.into_ancestor_data(),
  Declaration::SingletonClass(sc) => sc.into_ancestor_data(),
  _ => panic!("wrong!"),
}

// process ancestors and descendants the same
```

rust/rubydex/src/model/graph.rs
thomasmarshall: I like that! I wasn't sure if it would be acceptable for us to have namespace related functionality directly on `Declaration` since there are non-namespace declarations too. But if you think it's acceptable then I like it.

rust/rubydex/src/model/graph.rs
thomasmarshall: I moved it to an associated function on `Graph` rather than a method, because otherwise (I believe) we would have to either release the write lock and take a new one inside the method, or pass it through. Neither seemed quite right, but perhaps I've misunderstood.

rust/rubydex/src/model/graph.rs
thomasmarshall: OK I see, I added another `clear_descendants` convenience method to `Declaration` which we now call from `invalidate_ancestor_chains`.

rust/rubydex/src/model/graph.rs
thomasmarshall: I added some more convenience methods: `for_each_ancestor` and `for_each_descendant`. Do you think that's taking it a little far? If so, I will try your suggestion.

rust/rubydex/src/model/graph.rs
thomasmarshall: Oh wait, you said associated functions and I added them as methods. Let me take another look. Any reason you suggested associated functions instead of methods?

vinistock: Oh, just because they don't take `self` and they are defined on the enum instead of the variants. So you wouldn't do `declaration.members_of`, you'd do `Declaration::members_of(&declaration)`.

rust/rubydex/src/model/declaration.rs
```diff
@@ -298,6 +306,57 @@ impl Declaration {
...
+    pub fn clear_ancestors(&self) {
+        match self {
+            Declaration::Class(c) => c.set_ancestors(Ancestors::Partial(vec![])),
+            Declaration::Module(m) => m.set_ancestors(Ancestors::Partial(vec![])),
+            Declaration::SingletonClass(s) => s.set_ancestors(Ancestors::Partial(vec![])),
+            _ => {}
+        }
+    }
+
+    pub fn clear_descendants(&self) {
```
vinistock: This pattern works well for methods that return void, but it gets weird if there's an actual return. You end up needing to pass closures everywhere because you don't have anything to return on the catchall match arm.

Personally, I think an associated function would be a bit cleaner. I don't know if others have a different opinion.

```rust
Declaration::clear_descendants(&declaration)
```

rust/rubydex/src/model/graph.rs
```diff
@@ -348,6 +359,39 @@ impl Graph {
         }
     }
 
+    fn invalidate_ancestor_chains(
+        declarations: &IdentityHashMap<DeclarationId, Declaration>,
```
vinistock: What do you think about accepting `self` here and then acquiring a new lock?

I don't think the overhead of re-acquiring is significant and that allows this method to be self contained.

rust/rubydex/src/model/declaration.rs
```diff
@@ -150,6 +150,10 @@ macro_rules! namespace_declaration {
                 *ancestors_lock = ancestors;
             }
 
+            pub fn ancestors(&self) -> RwLockReadGuard<'_, Ancestors> {
```
vinistock: I think we can make this and `remove_descendant` both private since they are only used in the same file.

thomasmarshall: I've updated now :+1:

PR #444, opened 2026-01-12
Update ruby behavior doc

st0012:
These are the behaviours we added support in the last few weeks.

### DIFF
--- a/docs/ruby-behaviors.md
+++ b/docs/ruby-behaviors.md
@@ -8,11 +8,14 @@ This document describes various Ruby language behaviors, compiled from observati
 2. [Lexical Scoping and Constant Resolution](#lexical-scoping-and-constant-resolution)
 3. [Definitions vs Declarations](#definitions-vs-declarations)
 4. [Method Parameters](#method-parameters)
-5. [Attribute Methods](#attribute-methods)
-6. [Variable Scoping](#variable-scoping)
-7. [Constant References](#constant-references)
-8. [Constant Aliases](#constant-aliases)
-9. [Singleton Classes](#singleton-classes)
+5. [Method Visibility](#method-visibility)
+6. [Method Aliasing](#method-aliasing)
+7. [Attribute Methods](#attribute-methods)
+8. [Variable Scoping](#variable-scoping)
+9. [Mixins](#mixins)
+10. [Constant References](#constant-references)
+11. [Constant Aliases](#constant-aliases)
+12. [Singleton Classes](#singleton-classes)
 
 ## Namespace Qualification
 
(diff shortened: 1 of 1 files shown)


docs/ruby-behaviors.md
```diff
@@ -248,6 +251,166 @@ Results in:
...
+    def m2; end    # public (resets again)
+  end
+
+  def m3; end      # private (still in Foo's private scope)
+end
+```
+
+### Visibility Does NOT Apply to `def self.method`
+
+Visibility modifiers have **no effect** on singleton methods defined with `def self.method`:
```
vinistock, line 316: It might be worth mentioning that for singleton methods you have to use `private_singleton_method`, although I don't think we added support for it yet.

docs/ruby-behaviors.md
```diff
@@ -248,6 +251,166 @@ Results in:
...
+  alias_method :baz, :bar           # symbol syntax
+  alias_method "qux", "bar"         # string syntax
+end
+```
+
+Unlike `alias`, `alias_method` requires symbols or strings, not bare method names.
+
+### Top-Level Aliases
+
+Aliases can be defined at the top level:
```
vinistock: Only with the `alias` keyword. `alias_method` crashes at the top level

docs/ruby-behaviors.md
```diff
@@ -471,6 +634,201 @@ Foo.instance_variables  # => [:@ivar2]
...
+- `extend` calls the `extended` hook
+- `include` in singleton class calls the `included` hook
+
+### Multiple Mixins
+
+Multiple modules can be mixed in a single call:
+
+```ruby
+class Foo
+  include Bar, Baz   # Baz is included first, then Bar
```
vinistock, line 725: This caught a bug! https://github.com/Shopify/rubydex/pull/445

docs/ruby-behaviors.md
```diff
@@ -471,6 +634,201 @@ Foo.instance_variables  # => [:@ivar2]
...
+  include B
+end
+
+Child.ancestors  # => [Child, Parent, B, A, Object, ...]
+# B and A appear once (child's include is deduplicated against parent)
+```
+
+### Cyclic Mixins
+
+A module can include or prepend itself (though it's unusual):
```
vinistock: This crashes in the runtime. The only reason we must explicitly handle this is because Rubydex itself would get into an infinite recursion loop.

docs/ruby-behaviors.md
st0012: I think you mean `private_class_method`?

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
 Bar.class_variables     # => [:@@cvar2]
 ```
 
+## Mixins
+
+Ruby provides three ways to mix modules into classes or other modules: `include`, `prepend`, and `extend`.
+
+### Include
+
+`include` adds module methods as instance methods, inserted **after** the class in the ancestor chain:
```
Morriar, line 645: From the document it looks like it's more than just methods:

> Ruby’s default implementation is to add the constants, methods, and module variables of this module to mod

Morriar: > adds module methods as instance methods

Though it seems that functionally it's just adding an ancestor in the chain and the methods get "added" as a side effect?

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
...
+  include B
+  include C
+end
+
+Foo.ancestors  # => [Foo, C, B, A]
+# A appears only once despite being in both B and C
+```
+
+### Duplicate Mixins Across Parent and Child
+
```
Morriar, line 773:
```suggestion
```

let's remove the section header, both examples should go together

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
...
+When both parent and child include the same module, the module appears **twice** in the ancestor chain:
```
Morriar, line 774: You say "twice" here but the example shows only once

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
...
+
+### Top-Level Mixins
+
+Mixins at the top level affect `Object`:
+
+```ruby
+include Foo  # Makes Foo's methods available everywhere
+```
+
+**Note:** `include self`, `prepend self`, or `extend self` at the top level is invalid and will produce a warning.
```
Morriar, line 803:
```suggestion
**Note:** `include self`, `prepend self`, or `extend self` at the top level is invalid and will produce an error.
```

```
> RUBY_VERSION
=> "4.0.0"
> include self # => wrong argument type Object (expected Module) (TypeError)
> prepend self # => undefined method 'prepend' for main (NoMethodError)
> extend self # => wrong argument type Object (expected Module) (TypeError)
```

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
...
+
+### Extend Self Pattern
```
Morriar, line 805: Let's move this in the `Extend` section?

docs/ruby-behaviors.md
```diff
@@ -471,6 +636,188 @@ Foo.instance_variables  # => [:@ivar2]
...
+  extend self
+
+  def bar
+    "bar"
+  end
+end
+
+Foo.bar  # => "bar"
+```
+
```
Morriar, line 820: Should we add a note about cyclic includes/prepends and how you can't do `include self`/`prepend self`?

PR #446, opened 2026-01-13
Add module_function support to indexer

### LINKED ISSUE #91: Start indexing module_function

alexcrocha:
This PR adds support for indexing `module_function` declarations and closes #91.

 When `module_function` is active, method definitions create two entries:
  - A public singleton method on the module 
  - A private instance method

We model this by creating two `Definition` objects at the same source location, differentiated by their receiver.

For `attr_*` methods, `module_function` acts like `private` without creating singleton methods.

## Changes
  - New `ModuleFunction` visibility variant
  - `MethodDefinition::id()` includes receiver so the two definitions don't collide
  - Test helper `all_definition_at()` for locations with multiple definitions

### DIFF
--- a/rust/rubydex/src/test_utils/local_graph_test.rs
+++ b/rust/rubydex/src/test_utils/local_graph_test.rs
@@ -43,7 +43,7 @@ impl LocalGraphTest {
     ///
     /// Panics if a definition cannot be found at the given location.
     #[must_use]
-    pub fn definition_at<'a>(&'a self, location: &str) -> &'a Definition {
+    pub fn all_definitions_at<'a>(&'a self, location: &str) -> Vec<&'a Definition> {
         let (uri, offset) = self.parse_location(&format!("{}:{}", self.uri(), location));
         let uri_id = UriId::from(&uri);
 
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/test_utils/local_graph_test.rs
```diff
@@ -43,7 +43,7 @@ impl LocalGraphTest {
     ///
     /// Panics if a definition cannot be found at the given location.
     #[must_use]
-    pub fn definition_at<'a>(&'a self, location: &str) -> &'a Definition {
+    pub fn all_definition_at<'a>(&'a self, location: &str) -> Vec<&'a Definition> {
```
Morriar:
```suggestion
    pub fn all_definitions_at<'a>(&'a self, location: &str) -> Vec<&'a Definition> {
```

alexcrocha: Thanks!

Issue #447, opened 2026-01-13
Add API to get entry based on nesting


vinistock:
For the LSP, we need to be able to get declarations based on the current nesting for hover and go to definition. For example, instance and class variables.

We need to pass the nesting, so that we can build a `NameRef` in the `rubydex-sys` API and pass that to the underlying implementation. That way, the LSP doesn't have to implement constant resolution.

```ruby
module Foo
  class Bar::Baz
    def Something.other_thing
      @ivar # <<<< go to definition here
    end
  end
end
```

In this example, the LSP should fully delegate the core intelligence to Rubydex, but it might need to construct the `NameRef` structure that represents the `Something` constant within its nesting + the fact that `@ivar` will be connected to the singleton of the resolved declaration.

Issue #412, opened 2026-01-06
Add API to get method based on receiver type


vinistock:
For the LSP, we need an API where we can pass a method name and a receiver fully qualified name, and it then finds the method and returns (if it exists of course).

The API needs an option to return only inherited methods too, which is what we use to provide go to definition on `super`. For example:

```ruby
class Foo < Bar
  def baz
    # To provide go to definition here, we need to search for Foo#baz, but we don't want to find
    # this exact `baz` definition. We want to find the first one in the inheritance chain _after_ Foo
    # itself
    super
  end
end
```

Issue #59, opened 2025-08-01
Add API for retrieving completion candidates


vinistock:
We need an API to provide completion candidates based on the current context and node. In the Ruby LSP's indexer, we have separate APIs for each type of completion, which turned out to not be a great implementation.

Let's try to implement a single API that can return all applicable completion candidates based on where it is being triggered. For example:

```ruby
b # <- if I type `b` here, we have several types of applicable candidates
# It could be a method, local variable, keyword (like `begin`) and so on

foo.b # <- in this case, the applicable completions narrow a lot. Only method candidates are possible
# since you cannot refer to anything else with the dot syntax

Foo:: # <- in this one, both constants and singleton methods defined in `<Class:Foo>` are available
```

Start simple, with only one type of completion and grow the implementation in parts:

- [ ] Keywords (probably the easiest one as this is a static array of names that requires little type checking)
- [ ] Global variables
- [ ] Constants
- [ ] Instance variables (requires inferring type of `self`)
- [ ] Class variable (requires inferring type of `self`)
- [ ] Local variables (requires implementing scope tracking)
- [ ] Methods (requires inferring type of receiver)

- [ ] Add more specific errors for completion.

PR #449, opened 2026-01-13
Expose resolve constant as a Ruby API

vinistock:
Add a Ruby API to be able to resolve a constant given a name ref. This will be necessary for #447, #412 and #59 since we don't want clients of the API to have to perform the resolution of the nesting constants themselves, but instead rely on Rubydex.

The idea is basically to match what the core implementation does. Clients of the Ruby API have to construct a `NameRef` structure, which we translate from Ruby to Rust to perform resolution.

**Caveat**: for live document modifications like in the LSP, this API expects that the client will already have invoked our future API to handle modifications (which already resolves all references).

In this case, the API will retrieve the result directly. The `NameRef` structure is simply important to represent the data correctly and generate the expected ID on the Rust side.


rust/rubydex/src/resolution.rs
```diff
@@ -926,7 +926,11 @@ fn name_owner_id(graph: &mut Graph, name_id: NameId) -> Outcome {
 /// Attempts to resolve a constant reference against the graph. Returns the fully qualified declaration ID that the
 /// reference is related to or `None`. This method mutates the graph to remember which constants have already been
 /// resolved
-fn resolve_constant(graph: &mut Graph, name_id: NameId) -> Outcome {
+///
+/// # Panics
+///
+/// Only panics if there's inconsistent data in the graph
+pub fn resolve_constant(graph: &mut Graph, name_id: NameId) -> Outcome {
```
Morriar: I'm not sure it makes sense to return `retry`, can't we expose a `resolve_constant(graph: &mut Graph, name_id: NameId) -> Option<DeclarationId>` instead?

lib/rubydex/name_ref.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Rubydex
+  # A reference to a constant name, which includes its nesting and parent scope
+  #
+  # Example
+  # ```ruby
+  # # NameRef("Foo", nesting: nil, parent_scope: nil)
```
Morriar: Do we really need this abstraction at the Ruby level? Can we do `graph.resolve_constant("Foo", nesting: nil, parent_scope: nil)` instead?

vinistock: How would you request the resolution of `Baz` in that example?

rust/rubydex/src/resolution.rs
vinistock: This method specifically has to return retries for the resolution phase itself, but we can create a proxy method that returns `Option<Declaration>`.

Morriar: > we can create a proxy method that returns `Option<Declaration>`.

Yes 👍

lib/rubydex/name_ref.rb
Morriar: Can't you do `graph.resolve_constant("Baz", nesting: "Foo::Bar")`?

vinistock: I added a commit that changes the API to accept a constant name and nesting array of strings. I initially thought we might get into trouble without a proper representation of names on the Ruby side, but maybe it's enough.

It's much simpler, so let's start like this.

rust/rubydex/src/resolution.rs
```diff
@@ -130,6 +130,15 @@ pub fn resolve_all(graph: &mut Graph) {
...
 
+/// Resolves a single constant against the graph. This method is not meant to be used by the resolution phase, but by
+/// the Ruby API
+pub fn resolve_single_constant(graph: &mut Graph, name_id: NameId) -> Option<DeclarationId> {
+    match resolve_constant(graph, name_id) {
+        Outcome::Resolved(id, _) => Some(id),
+        Outcome::Unresolved(_) | Outcome::Retry => None,
+    }
+}
+
```
Morriar: Let's expose the nice name and keep the not-nice one internal:

```suggestion
pub fn resolve_constant(graph: &mut Graph, name_id: NameId) -> Option<DeclarationId> {
    match resolve_constant_internal(graph, name_id) {
        Outcome::Resolved(id, _) => Some(id),
        Outcome::Unresolved(_) | Outcome::Retry => None,
    }
}

```

test/graph_test.rb
```diff
@@ -166,6 +166,45 @@ def test_graph_search
...
+        end
+      RUBY
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      const = graph.resolve_constant("CONST", ["Foo", "Bar::Baz"])
+      assert_equal("Foo::CONST", const.name)
+    end
```
Morriar, line 244: Can you add a test showing what happens for unresolved constants?

```rb
      const = graph.resolve_constant("NOT_FOUND", ["Foo", "Bar::Baz"])
      assert_nil(const)
```

rust/rubydex-sys/src/name_api.rs
```diff
@@ -0,0 +1,72 @@
...
+/// # Panics
+///
+/// Should not panic because `const_name` will always be turned into a name
+pub fn nesting_stack_to_name_id(graph: &mut Graph, const_name: &str, nesting: Vec<String>) -> NameId {
+    let mut current_nesting = None;
+    let mut current_name = None;
+
+    for entry in nesting {
+        for part in entry.split("::").map(String::from) {
+            let str_id = graph.intern_string(part);
```
Morriar: Why do we need to intern the strings?

If I understand correctly, the client can do something like this:

```rb
graph = Rubydex::Graph.new

1_000_000.times do |i|
  graph.resolve_constant("Class#{i}")
end
```

and we'll intern 1,000,000 strings?

vinistock: I initially thought we needed the strings registered to properly perform resolution, but we indeed don't need it. We do need to ensure names are registered though.

rust/rubydex-sys/src/name_api.rs
```diff
@@ -0,0 +1,76 @@
...
+    name::Name,
+};
+
+/// Takes a constant name and a nesting stack (e.g.: `["Foo", "Bar::Baz", "Qux"]`) and transforms it into a `NameId`,
+/// registering each required part in the graph
+///
+/// # Panics
+///
+/// Should not panic because `const_name` will always be turned into a name
+pub fn nesting_stack_to_name_id(graph: &mut Graph, const_name: &str, nesting: Vec<String>) -> NameId {
```
Morriar: If we rebase on https://github.com/Shopify/rubydex/pull/458, could we increment the reference when creating the name, then decrement after the resolution so the name gets removed?

vinistock: Great idea! That way we don't get extra names. I made the change.

rust/rubydex-sys/src/graph_api.rs
```diff
@@ -60,6 +60,37 @@ pub unsafe extern "C" fn rdx_graph_declarations_search(
...
+
+        let resolved_id = match resolution::resolve_constant(graph, name_id) {
+            Some(id) => Box::into_raw(Box::new(*id)).cast_const(),
+            None => ptr::null(),
+        };
+
+        for name_id in names_to_untrack {
+            graph.untrack_name(name_id);
+        }
+        resolved_id
```
Morriar, line 94: I wonder if there is a higher level function we could add to `Graph` here so we could:

* keep `untrack_name` private
* add a proper test in `Graph` to make sure we untrack the names properly
* benefit from "simplified" `resolve_constant` function on the Rust side

vinistock: It feels like crossing the boundaries of the FFI to me a little bit. The graph operates on names and their IDs, but we'd be exposing a public method where we allow interaction with strings and vectors - which are only there to be the glue between both sides.

We can definitely add a test in the sys crate to verify everything works as expected, but I'd vote for keeping FFI related concerns outside of `rubydex` and exclusively in `rubydex-sys`.

Morriar: fair 👍

test/graph_test.rb
```diff
@@ -221,6 +221,50 @@ def test_graph_set_encoding_with_invalid_value
...
+  end
```
st0012, line 245: Now that constant alias support has been merged, can we add a test case with aliased constant after rebasing?

PR #450, opened 2026-01-13
Use rules instead of diagnostics code and severity

Morriar:
This PR changes the way we produce diagnostics to follow more closely what `ty` is doing for python: https://docs.astral.sh/ty/rules/.

The idea would be for rubydex to produce diagnostics and let the client define if this is an error or a warning. See https://docs.astral.sh/ty/rules/#rule-levels.

Also learning from our experience with Sorbet, error codes are generally confusing. You see the number and it doesn't mean much so you have to refer to the documentation or the code. Using names to represent diagnostics avoids this. It also makes configuration easier to understand.

### DIFF
--- a/lib/rubydex/diagnostic.rb
+++ b/lib/rubydex/diagnostic.rb
@@ -2,19 +2,19 @@
 
 module Rubydex
   class Diagnostic
+    #: Symbol
+    attr_reader :rule
+
     #: String
     attr_reader :message
-
-    #: Integer
-    attr_reader :code
 
     #: Location
     attr_reader :location
 
-    #: (message: String, code: Integer, location: Location) -> void
-    def initialize(message:, code:, location:)
+    #: (rule: Symbol, message: String, location: Location) -> void
+    def initialize(rule:, message:, location:)
+      @rule = rule
       @message = message
-      @code = code
       @location = location
     end
   end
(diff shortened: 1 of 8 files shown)


lib/rubydex/diagnostic.rb
```diff
@@ -2,23 +2,22 @@
...
   class Diagnostic
-    #: String
-    attr_reader :message
-
     #: Symbol
-    attr_reader :severity
+    attr_reader :rule
 
     #: Integer
     attr_reader :code
```
st0012:
```suggestion
```

rust/rubydex/src/diagnostic.rs
```diff
@@ -45,70 +43,58 @@ impl Diagnostic {
...
-    Warning,
 }
 
-impl Severity {
-    #[must_use]
-    pub fn as_str(&self) -> &str {
-        match self {
-            Severity::Error => "Error",
-            Severity::Warning => "Warning",
+fn camel_to_snake(s: &str) -> String {
```
vinistock, line 48: Not critical, but what do you all think about just using the camelcase names directly?

It simplifies the code, but I also personally think it looks nicer:

```
MyRule: something is not right
vs
my_rule: something is not right
```

Morriar: I went with lowercase and dashes to make it easier to map the error to an option or config, as with `ty`:

```
ty check \
  --warn unused-ignore-comment \
  --ignore redundant-cast \
  --error possibly-missing-attribute \
```

```toml
[tool.ty.rules]
unused-ignore-comment = "warn"
redundant-cast = "ignore"
possibly-missing-attribute = "error"
```

From the LSP we can remap the rule string to either a code or a camelcase ID.

PR #451, opened 2026-01-13
Improve tests for @deprecated tag

thomasmarshall:
This PR includes a couple of test improvements suggested in #410:

- Remove unnecessary definition assertions
- Make explicit the behavior when there are newlines

The latter was already the existing behavior based on how comment groups are attached to definitions and matches the behavior of RDoc and YARD.

### DIFF
--- a/test/definition_test.rb
+++ b/test/definition_test.rb
@@ -146,6 +146,10 @@ class DefinitionTest < Minitest::Test
         # @deprecated Use something else
         def deprecated_method; end
 
+        # @deprecated
+        # more comment
+        def also_deprecated_method; end
+
         # Not @deprecated
         def not_deprecated_method; end
       RUBY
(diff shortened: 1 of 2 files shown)


test/definition_test.rb
```diff
@@ -160,6 +160,36 @@ def not_deprecated_method; end
...
+        # @deprecated
+
+        class DeprecatedWithBlank; end
+
+        # @deprecated Use something else
+
+        class DeprecatedWithMessage; end
+
+        # Multi-line comment
+        # @deprecated
```
Morriar, line 180: Just to be sure:

```suggestion
        # @deprecated
        # more comment
```

thomasmarshall: I've added that to the test now (both with and without the empty line in between). The result is that the definition in both cases is marked as deprecated (matching YARD behavior).

PR #454, opened 2026-01-14
Move LineIndex into Document

vinistock:
The LSP needs to be able to get correct line/column locations for documents that haven't been saved yet and only exist in memory.

Additionally, we only need to re-compute the `LineIndex` structure if the file gets modified and not every time we need a location.

This PR moves the `LineIndex` inside of documents, so that every time we index a file, the structure gets created and can be reused.

This will also account for unsaved files, because in that case we'll just change the `source` we're indexing (it won't be read from disk).

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -61,7 +61,7 @@ impl<'a> RubyIndexer<'a> {
     #[must_use]
     pub fn new(uri: String, source: &'a str) -> Self {
         let uri_id = UriId::from(&uri);
-        let local_graph = LocalGraph::new(uri_id, Document::new(uri));
+        let local_graph = LocalGraph::new(uri_id, Document::new(uri, source));
 
         Self {
             uri_id,
@@ -1893,7 +1893,10 @@ mod tests {
         let mut diagnostics = context.graph().diagnostics().iter().collect::<Vec<_>>();
 
         diagnostics.sort_by_key(|d| d.offset());
-        diagnostics.iter().map(|d| d.formatted(context.source())).collect()
+        diagnostics
+            .iter()
+            .map(|d| d.formatted(context.graph().document()))
+            .collect()
     }
 
     macro_rules! assert_diagnostics_eq {
--- a/rust/rubydex/src/model/document.rs
+++ b/rust/rubydex/src/model/document.rs
@@ -1,3 +1,5 @@
+use line_index::LineIndex;
+
 use crate::model::ids::{DefinitionId, ReferenceId};
 
 // Represents a document currently loaded into memory. Identified by its unique URI, it holds the edges to all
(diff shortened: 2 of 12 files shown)


rust/rubydex/src/model/document.rs
```diff
@@ -1,20 +1,24 @@
+use line_index::LineIndex;
+
 use crate::model::ids::{DefinitionId, ReferenceId};
 
 // Represents a document currently loaded into memory. Identified by its unique URI, it holds the edges to all
 // definitions and references discovered in it
 #[derive(Debug)]
 pub struct Document {
     uri: String,
+    line_index: LineIndex,
```
Morriar, line 10: What's the memory impact of this? If it's high, what if we just stored the source directly in the document? It would simplify _a lot_ other places in the code where we need to pass around the source.

Then once we add the DB logic, we can figure a way to unload it?

vinistock: I benchmarked comparing main and this branch and it makes no difference in memory usage. LineIndex is quite efficient because it extracts the indices of line breaks and which characters are wide, so the memory cost isn't significant.

Also, the other places in the code that pass the source can now just use the line index, no?

Morriar: > I benchmarked comparing main and this branch and it makes no difference in memory usage

Can you save the benchmark results in this PR description so we keep a trace please?

> the other places in the code that pass the source can now just use the line index

In most cases yes. I was looking for a way to better separate the phases where `indexer` wouldn't have to read/parse the file but we can revisit later 👍

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1893,7 +1893,10 @@ mod tests {
         let mut diagnostics = context.graph().diagnostics().iter().collect::<Vec<_>>();
 
         diagnostics.sort_by_key(|d| d.offset());
-        diagnostics.iter().map(|d| d.formatted(context.source())).collect()
+        diagnostics
+            .iter()
+            .map(|d| d.formatted(context.graph().document()))
```
Morriar, line 1898: Does that mean we can remove https://github.com/Shopify/rubydex/blob/ca42d8b28af0016ba470ee7f356d49781e8dd661/rust/rubydex/src/test_utils/local_graph_test.rs#L40-L43?

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -135,11 +136,10 @@ impl GraphTest {
         );
     }
 
-    fn line_index_for(&self, uri: &str) -> LineIndex {
-        let source = self
-            .get_source(uri)
```
Morriar, line 140: Does that mean we can remove https://github.com/Shopify/rubydex/blob/ca42d8b28af0016ba470ee7f356d49781e8dd661/rust/rubydex/src/test_utils/graph_test.rs#L57-L61?

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: https://github.com/Shopify/rubydex/pull/469

rust/rubydex/src/test_utils/graph_test.rs
vinistock: https://github.com/Shopify/rubydex/pull/469

PR #455, opened 2026-01-14
Support LSP location encodings

vinistock:
This PR adds support for configuring the graph's encoding for the LSP, so that returned locations and match what's expected by editors according to the spec.

The idea is to have our own small encoding enum because `line_index::WideEncoding` actually doesn't have a value for `UTF-8`. We then widen the encoding when necessary to get the right columns when multibyte characters are involved.

### DIFF
--- a/rust/rubydex-sys/src/location_api.rs
+++ b/rust/rubydex-sys/src/location_api.rs
@@ -2,6 +2,7 @@
 
 use libc::c_char;
 use rubydex::model::document::Document;
+use rubydex::model::graph::Graph;
 use rubydex::offset::Offset;
 use std::ffi::CString;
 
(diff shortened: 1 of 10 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +41,7 @@ pub struct Graph {
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
 
     diagnostics: Vec<Diagnostic>,
+    encoding: Encoding,
```
Morriar: Is this related to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities?

```suggestion
    position_encoding: Encoding,
```

test/graph_test.rb
```diff
@@ -166,6 +166,47 @@ def test_graph_search
     end
   end
 
+  def test_graph_set_encoding
+    with_context do |context|
+      context.write!("foo.rb", <<~RUBY)
+        class Foo
+          def initialize
+            @叫聲 = "喵"
```
Morriar: Would something like `@叫聲😍x = "喵"` work so you have different positions in each encoding?

rust/rubydex/src/model/graph.rs
Morriar: Can we also add a comment explaining it only impact how we compute location, not how we read the content of the files?

rust/rubydex-sys/src/location_api.rs
```diff
@@ -25,17 +26,30 @@ pub struct Location {
 /// - If the file cannot be read.
 /// - If the offset cannot be converted to a position.
 #[must_use]
-pub(crate) fn create_location_for_uri_and_offset(document: &Document, offset: &Offset) -> *mut Location {
+pub(crate) fn create_location_for_uri_and_offset(graph: &Graph, document: &Document, offset: &Offset) -> *mut Location {
```
Morriar, line 29: If we had a higher level concept of location like with https://github.com/Shopify/saturn/pull/353, and since we put the encoding in the graph, this could be moved in the graph as well:

```rs
let loc = graph.location_for(&document, &offset)
```

or maybe

```rs
let loc = graph.location_for(&definition)
```

rust/rubydex/src/model/graph.rs
vinistock: Yes, this is the position encodings that dictate what location values have to be returned to the editor.

rust/rubydex-sys/src/location_api.rs
vinistock: Yeah, I think we can discuss how we want our first-class citizen `Location` to look like and push some changes afterwards.

URIs are already stored in documents and all entities are able to trace back to the document where they were discovered, so maybe `Location` can just store line/column information. So you'd do something like:

```ruby
# Returns proper URI object so that clients can inspect any part of the URI
# they need
definition.uri # => URI("file:///foo.rb")

# Returns byte offsets (either a range or dedicated object)
definition.offsets # => (5, 10)

# Returns the line/column information
definition.location #  => #<Location @start_line=1, @start_column=1 ...>
```

test/graph_test.rb
```diff
@@ -166,6 +166,47 @@ def test_graph_search
...
+
+      # UTF-8: code units => number of bytes
+      loc = foo.location
+      assert_equal(3, loc.start_line)
+      assert_equal(5, loc.start_column)
+      assert_equal(3, loc.end_line)
+      assert_equal(17, loc.end_column)
+
+      # UTF-16: code units => 1 for 1,2 byte characters, 2 for 3,4 byte characters
+      graph.set_encoding("utf16")
```
Morriar, line 193: Let's add a few tests where we pass garbage to `set_encoding`

PR #458, opened 2026-01-14
Track names for cleanup

thomasmarshall:
Part of #330.

This PR implements ref count tracking for names so we can clear them (if they are unused) from the graph when a document is changed or removed. Whenever a name is used we increment its counter and then during cleanup we decrement the counter. Once the counter reaches zero we know nothing else is using the name and it can be deleted.

Originally I implemented this as a `names_count` mapping on the graph, but @vinistock suggested we could add a count field directly to the `Name` struct—I think this works nicely.

@Morriar suggested I also try the simplest approach, which would be to essentially rebuilt the full `names` hash. Here is the code I used for that:

```rust
pub fn rebuild_names(&mut self) {
    let mut used_names: std::collections::HashSet<NameId> = std::collections::HashSet::new();

    for def in self.definitions.values() {
        if let Some(name_id) = def.name_id() {
            used_names.insert(*name_id);
        }
    }

    for reference in self.constant_references.values() {
        used_names.insert(*reference.name_id());
    }

    self.names.retain(|id, _| used_names.contains(id));
}
```

It loops through all definitions and references finding which ones to retain. This is slightly different than blowing the hash away and recreating from scratch, but I think it should be similar. See below for benchmarks.

<details>
  <summary>Benchmarks</summary>

  Medium corpus:

```
=== Ref-counting approach ===
Time: 63.917µs
Names: 21973 -> 21962 (11 removed)
```

```
=== Rebuild approach ===
Time: 1.624375ms
Names: 21973 -> 21962 (11 removed)
```

Huge corpus:

```
=== Ref-counting approach ===
Time: 450.791µs
Names: 2203531 -> 2203522 (9 removed)
```

```
=== Rebuild approach ===
Time: 259.197583ms
Names: 2203531 -> 2203522 (9 removed)
```
</details>

Since the ref counting approach isn't too complex and substantially faster, I would like to go with this. We could do something similar for `strings` too.

The diff is a little simpler to review with [whitespace changes ignored](https://github.com/Shopify/rubydex/pull/458/files?w=1).

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -191,6 +191,14 @@ impl Graph {
         &self.names
     }
 
+    fn untrack_name(&mut self, name_id: NameId) {
+        if let Some(name_ref) = self.names.get_mut(&name_id)
+            && !name_ref.decrement_ref_count()
+        {
+            self.names.remove(&name_id);
+        }
+    }
+
     /// Register a member relationship from a declaration to another declaration through its unqualified name id. For example, in
     ///
     /// ```ruby
@@ -315,11 +323,13 @@ impl Graph {
         }
 
         for ref_id in document.constant_references() {
-            if let Some(constant_ref) = self.constant_references.remove(ref_id)
-                && let Some(NameRef::Resolved(resolved)) = self.names.get(constant_ref.name_id())
-                && let Some(declaration) = self.declarations.get_mut(resolved.declaration_id())
-            {
-                declaration.remove_reference(ref_id);
+            if let Some(constant_ref) = self.constant_references.remove(ref_id) {
+                if let Some(NameRef::Resolved(resolved)) = self.names.get(constant_ref.name_id())
+                    && let Some(declaration) = self.declarations.get_mut(resolved.declaration_id())
+                {
+                    declaration.remove_reference(ref_id);
+                }
+                self.untrack_name(*constant_ref.name_id());
             }
         }
 
@@ -329,17 +339,22 @@ impl Graph {
         let mut declarations_to_invalidate_ancestor_chains: Vec<DeclarationId> = Vec::new();
 
         for def_id in document.definitions() {
-            if let Some(_definition) = self.definitions.remove(def_id)
-                && let Some(declaration_id) = self.definitions_to_declarations.remove(def_id)
-                && let Some(declaration) = self.declarations.get_mut(&declaration_id)
-                && declaration.remove_definition(def_id)
-            {
-                declarations_to_invalidate_ancestor_chains.push(declaration_id);
+            if let Some(definition) = self.definitions.remove(def_id) {
+                if let Some(declaration_id) = self.definitions_to_declarations.remove(def_id)
+                    && let Some(declaration) = self.declarations.get_mut(&declaration_id)
+                    && declaration.remove_definition(def_id)
+                {
+                    declarations_to_invalidate_ancestor_chains.push(declaration_id);
 
-                if declaration.has_no_definitions() {
-                    let unqualified_str_id = StringId::from(&declaration.unqualified_name());
-                    members_to_delete.push((*declaration.owner_id(), unqualified_str_id));
-                    declarations_to_delete.push(declaration_id);
+                    if declaration.has_no_definitions() {
+                        let unqualified_str_id = StringId::from(&declaration.unqualified_name());
+                        members_to_delete.push((*declaration.owner_id(), unqualified_str_id));
+                        declarations_to_delete.push(declaration_id);
+                    }
+                }
+
+                if let Some(name_id) = definition.name_id() {
+                    self.untrack_name(*name_id);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -191,6 +191,15 @@ impl Graph {
         &self.names
     }
 
+    fn untrack_name(names: &mut IdentityHashMap<NameId, NameRef>, name_id: NameId) {
+        if let Some(name_ref) = names.get_mut(&name_id) {
+            name_ref.decrement_ref_count();
+            if name_ref.ref_count() == 0 {
```
Morriar: Should we expose `has_ref` or something like this?

rust/rubydex/src/model/graph.rs
```diff
@@ -276,7 +285,13 @@ impl Graph {
         self.documents.insert(uri_id, document);
         self.definitions.extend(definitions);
         self.strings.extend(strings);
-        self.names.extend(names);
+        for (name_id, name_ref) in names {
+            if let Some(existing) = self.names.get_mut(&name_id) {
+                existing.add_ref_count(name_ref.ref_count());
```
Morriar: Maybe we need a `Name#merge`?

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -92,7 +92,11 @@ impl LocalGraph {
 
     pub fn add_name(&mut self, name: Name) -> NameId {
         let name_id = name.id();
-        self.names.insert(name_id, NameRef::Unresolved(Box::new(name)));
+        if let Some(existing) = self.names.get_mut(&name_id) {
+            existing.add_ref_count(1);
```
Morriar:
```suggestion
            existing.increment_ref_count();
```

rust/rubydex/src/model/graph.rs
thomasmarshall: Yes, that's nice.

rust/rubydex/src/indexing/local_graph.rs
thomasmarshall: I added `merge` as suggested above, which removes the need for either `add_ref_count` or `increment_ref_count`.

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -92,7 +92,12 @@ impl LocalGraph {
 
     pub fn add_name(&mut self, name: Name) -> NameId {
         let name_id = name.id();
-        self.names.insert(name_id, NameRef::Unresolved(Box::new(name)));
+        let name_ref = NameRef::Unresolved(Box::new(name));
+        if let Some(existing) = self.names.get_mut(&name_id) {
+            existing.merge(&name_ref);
+        } else {
+            self.names.insert(name_id, name_ref);
+        }
```
vinistock: Nit: I'd use this style here:

```rust
match self.names().entry(name_id) {
  Entry::Occupied(entry) => {}
  Entry::Vacant(entry) => {}
}
```

Also, I would change the API slightly to avoid the need to allocate the enum variant + boxing the name.

Additionally, we add these names one by one. There's never a scenario inside of this method where the ref_count of `name` isn't equal to 1.

Could we instead do something simpler like:

```rust
match self.names().entry(name_id) {
  Entry::Occupied(entry) => {
    entry.get().increment_ref_count(1)
  }
  Entry::Vacant(entry) => {
    // ...
  }
}
```

rust/rubydex/src/model/name.rs
```diff
@@ -120,6 +122,36 @@ impl NameRef {
...
+    }
+
+    pub fn merge(&mut self, other: &NameRef) {
+        match self {
+            NameRef::Unresolved(name) => name.ref_count += other.ref_count(),
+            NameRef::Resolved(resolved_name) => resolved_name.name.ref_count += other.ref_count(),
+        }
+    }
+
+    pub fn decrement_ref_count(&mut self) {
```
vinistock: We could make this method return a boolean, so that we structure the other code like:

```rust
if decrement_ref_count() {
  // remove name
}
```

rust/rubydex/src/model/graph.rs
```diff
@@ -276,7 +285,13 @@ impl Graph {
         self.documents.insert(uri_id, document);
         self.definitions.extend(definitions);
         self.strings.extend(strings);
-        self.names.extend(names);
+        for (name_id, name_ref) in names {
+            if let Some(existing) = self.names.get_mut(&name_id) {
+                existing.merge(&name_ref);
```
vinistock: Personally, I would do something like:

```rust
existing.increment(name_ref.ref_count())
```

rust/rubydex/src/indexing/local_graph.rs
thomasmarshall: > avoid the need to allocate the enum variant + boxing the name

Ah yeah, I switched to that because I thought it was nice to use the same `merge` method in both places, but didn't think about the unnecessary allocation. I have switched to this style now, removing `merge` again and implementing `increment_ref_count(n)`.

rust/rubydex/src/model/name.rs
thomasmarshall: Nice, yes.

rust/rubydex/src/model/graph.rs
thomasmarshall: @Morriar suggested `merge` (I previously had `add_ref_count`). I guess I like having symmetry in the two places we do this, so if `increment_ref_count(n)` works in both, then let's do that.

vinistock: I would personally favour `increment_ref_count` over `merge` because you can't merge two distinct names into one. You can only merge two identical names and all it does is increasing the ref count.

To me, `merge` communicates combining information more strongly.

PR #459, opened 2026-01-14
Add uri/offset combination to sort

Morriar:
Make sure classes are always processed in the same order when they have the same `name_depth`.

This makes the declaration / linearization computation more deterministic.


rust/rubydex/src/resolution.rs
```diff
@@ -1117,23 +1117,37 @@ fn name_depth(name: &NameRef, names: &IdentityHashMap<NameId, NameRef>) -> u32 {
 #[must_use]
 fn sorted_units(graph: &Graph) -> (VecDeque<Unit>, Vec<DefinitionId>) {
     let estimated_length = graph.definitions().len() / 2;
-    let mut constants = Vec::with_capacity(estimated_length);
+    let mut definitions = Vec::with_capacity(estimated_length);
```
vinistock: This should actually be `units`. It includes constant references too.

Morriar: Renamed 👍

PR #461, opened 2026-01-14
Do not mutate the graph to access declarations ancestors

Morriar:
Before this PR we were lazily computing the ancestors only for top level declarations. All the nested declarations where computed eagerly. With this change we perform all declaration computation eagerly.

With this, we can serve the ancestors of a declaration after the resolution phase and without a mutable graph.

Performance is unchanged:

```
# before

Timing breakdown
  Listing             0.825s (  2.7%)
  Indexing            7.728s ( 25.4%)
  Resolution         21.420s ( 70.4%)
  Querying            0.438s (  1.4%)
  Total:             30.411s

# after

Timing breakdown
  Listing             0.542s (  1.7%)
  Indexing            7.182s ( 23.2%)
  Resolution         22.913s ( 73.9%)
  Querying            0.350s (  1.1%)
  Total:             30.987s
```

### DIFF
--- a/rust/rubydex/src/test_utils/graph_test.rs
+++ b/rust/rubydex/src/test_utils/graph_test.rs
@@ -30,11 +30,6 @@ impl GraphTest {
     }
 
     #[must_use]
-    pub fn graph_mut(&mut self) -> &mut Graph {
-        &mut self.graph
-    }
-
-    #[must_use]
     fn index_source(uri: &str, source: &str) -> LocalGraph {
         let mut indexer = RubyIndexer::new(uri.to_string(), source);
         indexer.index();


rust/rubydex/src/model/graph.rs
```diff
@@ -361,6 +361,23 @@ impl Graph {
...
+    #[must_use]
+    pub fn ancestors_of(&self, declaration_id: DeclarationId) -> Ancestors {
+        let declaration = self.declarations.get(&declaration_id).unwrap();
+        match declaration {
+            Declaration::Class(class) => class.clone_ancestors(),
+            Declaration::Module(module) => module.clone_ancestors(),
+            Declaration::SingletonClass(singleton) => singleton.clone_ancestors(),
+            Declaration::Constant(_) => Ancestors::Complete(vec![]),
+            _ => panic!("Tried to get ancestors for a declaration that isn't a namespace or a constant"),
+        }
```
vinistock: How about we put this as an associated function to `Declaration` (similar to what we discussed [here](https://github.com/Shopify/rubydex/pull/443#discussion_r2686316937))?

That way we can do `Declaration::ancestors_of(&decl)`.

Morriar: That works too 👍

PR #462, opened 2026-01-14
More graph functions for resolution

Morriar:
In the same vein than #461, move a few useful accessors from the resolution to the graph.

### DIFF
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -388,6 +388,10 @@ impl Namespace {
         all_namespaces!(self, it => it.has_complete_ancestors())
     }
 
+    pub fn add_descendant(&self, descendant_id: DeclarationId) {
+        all_namespaces!(self, it => it.add_descendant(descendant_id));
+    }
+
     pub fn remove_descendant(&self, descendant_id: &DeclarationId) {
         all_namespaces!(self, it => it.remove_descendant(descendant_id));
     }
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -655,7 +655,13 @@ impl<'a> Resolver<'a> {
             // Automatically track descendants as we recurse. This has to happen before checking the cache since we may have
             // already linearized the parent's ancestors, but it's the first time we're discovering the descendant
             for descendant in &context.descendants {
-                self.add_descendant(declaration_id, *descendant);
+                self.graph
+                    .declarations_mut()
+                    .get_mut(&declaration_id)
+                    .unwrap()
+                    .as_namespace_mut()
+                    .unwrap()
+                    .add_descendant(*descendant);
             }
         }
 
@@ -876,7 +882,13 @@ impl<'a> Resolver<'a> {
             for ancestor in cached {
                 if let Ancestor::Complete(ancestor_id) = ancestor {
                     for descendant in descendants.iter() {
-                        self.add_descendant(*ancestor_id, *descendant);
+                        self.graph
+                            .declarations_mut()
+                            .get_mut(ancestor_id)
+                            .unwrap()
+                            .as_namespace_mut()
+                            .unwrap()
+                            .add_descendant(*descendant);
                     }
                 }
             }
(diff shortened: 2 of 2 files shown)


rust/rubydex/src/model/declaration.rs
```diff
@@ -385,6 +385,45 @@ impl Declaration {
             _ => {}
         }
     }
+
+    /// # Panics
+    ///
+    /// Panics if the declaration is not a namespace
+    #[must_use]
+    pub fn members(&self) -> &IdentityHashMap<StringId, DeclarationId> {
```
vinistock: Not trying to be pedantic, but I was actually thinking of an associated function instead of a method.

It feels a bit weird to do `declaration.members()` and have it panic. If you do `Declaration::members(&decl)` though, it feels more expected because it's plausible that you passed incorrect arguments.

What do you think?

```suggestion
    pub fn members(declaration: &Declaration) -> &IdentityHashMap<StringId, DeclarationId> {
```

Morriar: Not a fan of the `Declaration::members(decl)` idea.

I think a more Rust way would be to have a nested enum for `Namespace` and add the function on it, on the call site, this would look like this:

```rs
decl.as_namespace().members()
```

We can choose to make `as_namespace()` return an optional or panic if it can be down casted.

Considering https://github.com/Shopify/rubydex/pull/463, is ready to merge and the amount of conflicts, I'd like to merge this one and we can change things as a follow up?

rust/rubydex/src/resolution.rs
```diff
@@ -612,7 +612,7 @@ impl<'a> Resolver<'a> {
 
             // Return the cached ancestors if we already computed them. If they are partial ancestors, ignore the cache to try
             // again
-            if Self::has_complete_ancestors(declaration) {
+            if Self::has_complete_ancestors(declaration.as_namespace().unwrap()) {
```
```suggestion
            if declaration.as_namespace().unwrap().has_complete_ancestors() {
```

rust/rubydex/src/resolution.rs
```diff
@@ -629,7 +629,7 @@ impl<'a> Resolver<'a> {
                 } else {
                     Ancestors::Cyclic(vec![])
                 };
-                Self::set_ancestors(declaration, estimated_ancestors.clone());
+                Self::set_ancestors(declaration.as_namespace().unwrap(), estimated_ancestors.clone());
```
```suggestion
               declaration.as_namespace().unwrap().set_ancestors(estimated_ancestors.clone());
```

rust/rubydex/src/resolution.rs
```diff
@@ -692,7 +692,15 @@ impl<'a> Resolver<'a> {
...
-        Self::set_ancestors(self.graph.declarations().get(&declaration_id).unwrap(), result.clone());
+        Self::set_ancestors(
+            self.graph
+                .declarations()
+                .get(&declaration_id)
+                .unwrap()
+                .as_namespace()
+                .unwrap(),
+            result.clone(),
+        );
```
```suggestion
            self.graph
                .declarations()
                .get(&declaration_id)
                .unwrap()
                .as_namespace()
                .unwrap()
                .set_ancestors(result.clone());
```

rust/rubydex/src/resolution.rs
```diff
@@ -1388,21 +1416,19 @@ impl<'a> Resolver<'a> {
         }
     }
 
-    fn has_complete_ancestors(declaration: &Declaration) -> bool {
-        match declaration {
-            Declaration::Namespace(Namespace::Class(class)) => class.has_complete_ancestors(),
-            Declaration::Namespace(Namespace::Module(module)) => module.has_complete_ancestors(),
-            Declaration::Namespace(Namespace::SingletonClass(singleton)) => singleton.has_complete_ancestors(),
-            _ => panic!("Tried to check complete ancestors for a declaration that isn't a namespace"),
+    fn has_complete_ancestors(namespace: &Namespace) -> bool {
```
vinistock: I'd remove all of these now that we can much more conveniently just access it directly on the namespace.

rust/rubydex/src/resolution.rs
Morriar: Good point 👍

PR #464, opened 2026-01-15
Split declaration member types

vinistock:
We had a bug with two facets for this example:

```ruby
module Kernel
  class Array; end

  def Array(*elements); end
end
```

1. The first facet of the bug is that all members (both constants and others) are stored in the same `members` hash. The keys are the `StringId` for the interned name, which are identical. This meant that members could override other members and we missed the information
2. The second part of the bug is that the declarations for the array constant and method were incorrectly sharing the same fully qualified name. We depend on `DeclarationId` being both deterministic and unique, so this caused us to also incorrectly override declarations

This PR does two things to fully fix the bug:

1. Splits constant members into their own storage. There's never a situation where we don't know what kind of member we are looking for. Even if that was so, we could search both `constant_members` and `members` hashes. This avoids the conflicting name issue while preserving our ability to find things based on unqualified names (like asking is `baz` defined in `Foo`)
2. I switched all non-constant declarations to use the `#` name separator. This is something we already knew was missing. It both looks nicer and eliminates the lack of uniqueness since now the two array declarations are identified by:

- Foo::Array (constant)
- Foo#Array (method)

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1655,8 +1655,7 @@ mod tests {
 
         for member in members {
             assert!(
-                constant_members.contains_key(&StringId::from(*member))
-                    || other_members.contains_key(&StringId::from(*member)),
+                actual_members.contains_key(&StringId::from(*member)),
                 "Expected member '{member}' not found"
             );
         }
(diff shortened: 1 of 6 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -3491,4 +3471,40 @@ mod tests {
...
+              class Array; end
+            end
+            "
+        });
+        context.resolve();
+
+        let foo = context.graph().declarations().get(&DeclarationId::from("Foo")).unwrap();
+
+        // Both entries exist as unique members
+        assert!(Declaration::members_of(foo).contains_key(&StringId::from("Array")));
```
Morriar: You can rebase on https://github.com/Shopify/rubydex/pull/462 to use the new helpers:

```rs
assert_members_eq!(context, "Foo", vec!["Array"]);
```

PR #465, opened 2026-01-15
Track strings for cleanup

thomasmarshall:
Part of #330.

Similar to how we did ref count tracking for names in #458, this PR adds ref count tracking for strings so they can be cleaned up when a document is changed or removed. This is a little more involved since we needed to add a wrapper struct to store the `ref_count` and update all the places that previously expected `String` to work with `StringRef`. There are also more places we add strings than where we add names.

I tried to keep the diff as small and clean as possible, but I had to change a bunch of places to call `as_str()` which also impacted other things like formatting. The alternative approach would be to implement various traits on `StringRef` so we can use it the same way we did `String`, such as `Eq`, `Ord`, `PartialEq` etc. I'm not sure which is preferable or which is more idiomatic.

The basic mechanism is this: when a string is added to the graph for the first time, the `ref_count` is 1. If we encounter the same string elsewhere, we increment the `ref_count` by one. When we merge a local graph into the main graph we increment the `ref_count` by the number of times that string appears in the local graph. Finally, when we remove a document we loop traverse through all the definitions and references in that document, decrementing the `ref_count` for any strings we find, and then remove the strings from the graph if they are unused.

### DIFF
--- a/rust/rubydex/src/indexing/local_graph.rs
+++ b/rust/rubydex/src/indexing/local_graph.rs
@@ -7,13 +7,14 @@ use crate::model::identity_maps::IdentityHashMap;
 use crate::model::ids::{DefinitionId, NameId, ReferenceId, StringId, UriId};
 use crate::model::name::{Name, NameRef};
 use crate::model::references::{ConstantReference, MethodRef};
+use crate::model::string_ref::StringRef;
 use crate::offset::Offset;
 
 type LocalGraphParts = (
     UriId,
     Document,
     IdentityHashMap<DefinitionId, Definition>,
-    IdentityHashMap<StringId, String>,
+    IdentityHashMap<StringId, StringRef>,
     IdentityHashMap<NameId, NameRef>,
     IdentityHashMap<ReferenceId, ConstantReference>,
     IdentityHashMap<ReferenceId, MethodRef>,
@@ -25,7 +26,7 @@ pub struct LocalGraph {
     uri_id: UriId,
     document: Document,
     definitions: IdentityHashMap<DefinitionId, Definition>,
-    strings: IdentityHashMap<StringId, String>,
+    strings: IdentityHashMap<StringId, StringRef>,
     names: IdentityHashMap<NameId, NameRef>,
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
@@ -80,13 +81,20 @@ impl LocalGraph {
     // Strings
 
     #[must_use]
-    pub fn strings(&self) -> &IdentityHashMap<StringId, String> {
+    pub fn strings(&self) -> &IdentityHashMap<StringId, StringRef> {
         &self.strings
     }
 
     pub fn intern_string(&mut self, string: String) -> StringId {
         let string_id = StringId::from(&string);
-        self.strings.insert(string_id, string);
+        match self.strings.entry(string_id) {
+            Entry::Occupied(mut entry) => {
+                entry.get_mut().increment_ref_count(1);
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(StringRef::new(string));
+            }
+        }
         string_id
     }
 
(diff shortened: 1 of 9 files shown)


rust/rubydex/src/model/string_ref.rs
```diff
@@ -0,0 +1,38 @@
...
+        self.ref_count
+    }
+
+    pub fn increment_ref_count(&mut self, count: usize) {
+        self.ref_count += count;
+    }
+
+    #[must_use]
+    pub fn decrement_ref_count(&mut self) -> bool {
+        self.ref_count -= 1;
```
st0012, line 33: Let's add `debug_assert!(self.ref_count > 0)` here to make sure we don't accidentally call this but not remove the reference?

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -80,13 +81,20 @@ impl LocalGraph {
...
     #[must_use]
-    pub fn strings(&self) -> &IdentityHashMap<StringId, String> {
+    pub fn strings(&self) -> &IdentityHashMap<StringId, StringRef> {
         &self.strings
     }
 
     pub fn intern_string(&mut self, string: String) -> StringId {
         let string_id = StringId::from(&string);
-        self.strings.insert(string_id, string);
+        match self.strings.entry(string_id) {
```
st0012, line 90: TIL this pattern

rust/rubydex/src/model/definitions.rs
```diff
@@ -146,6 +146,28 @@ impl Definition {
         }
     }
 
+    #[must_use]
+    pub fn str_ids(&self) -> Vec<StringId> {
```
vinistock: Not a super strong opinion, but I'd move this to a helper method in the graph and avoid the vector allocations.

rust/rubydex/src/model/string_ref.rs
```diff
@@ -0,0 +1,38 @@
...
+        self.ref_count > 0
+    }
+}
+
+// Allows calling String methods like .as_str() on StringRef
```
vinistock: This is true, but not exactly what `Deref` does. Implementing `Deref` means defining what `*string_ref` does and the Rust compiler can automatically figure out when to deref something in certain situations, which then makes methods that you can invoke on the underlying string available.

rust/rubydex/src/model/string_ref.rs
```diff
@@ -0,0 +1,38 @@
+use std::ops::Deref;
+
+#[derive(Debug)]
+pub struct StringRef {
```
vinistock, line 10: Let's add some doc comments explaining why we ended up adding this struct.

rust/rubydex/src/model/graph.rs
```diff
@@ -179,8 +180,9 @@ impl Graph {
     /// Registers a name in the graph unless already registered. In regular indexing, this only happens in the local
     /// graph. This method is only used to back the `Graph#resolve_constant` Ruby API because every name must be
     /// registered in the graph to properly resolve
-    pub fn add_name(&mut self, name: Name) -> NameId {
+    pub fn add_name(&mut self, name: Name, str_value: String) -> NameId {
```
vinistock: This makes the API a bit weird because you already need to intern the string to create the `Name` object.

We originally avoided interning strings in `name_api` so that the memory wouldn't grow indefinitely, but now that we can untrack strings, we don't have to worry about it.

Let's not do this here and instead require that all users of this API intern their strings first (i.e.: let's intern the strings in `name_api`).

thomasmarshall: Ah yes sorry, you're right. I'll remove.

rust/rubydex/src/model/string_ref.rs
```diff
@@ -0,0 +1,44 @@
+use std::ops::Deref;
+
+/// A reference-counted string used in the graph.
+///
+/// This struct wraps a `String` with a reference count to track how many times
+/// the string is used across the graph. When a document is removed, we decrement
+/// the reference count for each string it uses, and remove the string from the
+/// graph when its count reaches zero.
+#[derive(Debug)]
+pub struct StringRef {
```
Morriar, line 10: I'm wondering if we could generalize the concept between String and Name:

```rs
pub struct RefCount<T> {
    value: T,
    ref_count: usize,
}

impl<T> Deref for RefCount<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.value
    }
}
```

thomasmarshall: I'll take a look, thanks for the suggestion.

vinistock: I'd say, feel free to ship and open a PR with the refactor, so that we minimize the changes.

thomasmarshall: Yes, I will do that instead :+1:

PR #467, opened 2026-01-15
Add documentation to clarify untrack_name

thomasmarshall:
This PR adds a note to explain that names in `parent_scope` or `nesting` are tracked and untracked separately.

For example:

```rb
class Foo::Bar::Baz
end
```

This will create and `Foo` and `Bar` names with count `1` via the constant references and the `Baz` name with count `1` via the definition.

```rb
module A
  B
end
```

This will create the `A` name with count `1` via a definition and the `B` name with count `1` via a constant reference.

In both cases, when the document is cleared, we loop through each constant reference and definition and decrement the associated names once. Since each name is accessed via its own constant reference or definition, it's not necessary to increment or decrement the counts for `parent_scope` or `nesting` names.

It also adds a new `untrack_name_recursive` that will help with other use-cases, such as temporarily adding a name to the graph as proposed in #449.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -254,11 +254,36 @@ impl Graph {
         &self.names
     }
 
+    /// Decrements the ref count for a name and removes it if the count reaches zero.
+    ///
+    /// This does not recursively untrack `parent_scope` or `nesting` names.
     pub fn untrack_name(&mut self, name_id: NameId) {
         if let Some(name_ref) = self.names.get_mut(&name_id)
             && !name_ref.decrement_ref_count()
         {
             self.names.remove(&name_id);
+        }
+    }
+
+    /// Decrements the ref count for a name and removes it if the count reaches zero.
+    ///
+    /// This recursively untracks `parent_scope` and `nesting` names.
+    pub fn untrack_name_recursive(&mut self, name_id: NameId) {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -191,6 +191,9 @@ impl Graph {
         &self.names
     }
 
+    /// Decrements the ref count for a name and removes it if the count reaches zero.
+    ///
+    /// Note: This does not recursively untrack `parent_scope` or `nesting` names, which are untracked separately.
```
Morriar: Should we add a `untrack_name_recursive` to take care of this?

thomasmarshall: How do you see that being used? We currently track and untrack all names correctly, as far as I understand.

Morriar: If I understand correctly, in https://github.com/Shopify/rubydex/pull/449/changes#diff-916376a194ffa6c30d731493cd92b4a4e71ab20b3b82154f0b41729227fb8c08R91-R93, Vini has to remember all the parts of the name he created to be able to untrack all of them after the search.

 It seems the perfect use-case to untrack recursively?

thomasmarshall: Ah I understand. So in that PR the names are only needed temporarily and we want to remove them afterwards. Yep, having another method would work—should it live in Vini's PR though, rather than here?

PR #470, opened 2026-01-15
Introduce a nested enum for `Declaration::Namespace`

Morriar:
Proposal for https://github.com/Shopify/rubydex/pull/462#discussion_r2695504036.

### DIFF
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -56,9 +56,9 @@ impl<'a> IntoIterator for &'a Ancestors {
 macro_rules! all_declarations {
     ($value:expr, $var:ident => $expr:expr) => {
         match $value {
-            Declaration::Class($var) => $expr,
-            Declaration::SingletonClass($var) => $expr,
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/model/declaration.rs
```diff
@@ -257,6 +244,17 @@ impl Declaration {
...
+    /// # Panics
+    ///
+    /// Panics if the declaration is not a namespace
+    #[must_use]
+    pub fn as_namespace(&self) -> &Namespace {
+        match self {
+            Declaration::Namespace(namespace) => namespace,
+            _ => panic!("Tried to convert a declaration that isn't a namespace to a namespace: {self:?}"),
+        }
+    }
```
vinistock: I _think_ the convention is to return optional and unwrap at the call site.

```suggestion
    pub fn as_namespace(&self) -> Option<&Namespace> {
        match self {
            Declaration::Namespace(namespace) => Some(namespace),
            _ => None,
        }
    }
```

Morriar: Yeah let's do that 👍

PR #471, opened 2026-01-15
Remember attributes as bitflags

vinistock:
### Motivation

This PR is a proposal to change how we handle attributes. Right now, we have distinct definition types for each one. In reality, they are methods and all we're interested in remembering is how they were originally defined.

I think we can simplify our code by quite a bit if we just use bitflags to remember how they were originally defined and reuse the `MethodDefinition` struct to represent them.

The main goal is also to handle attributes correctly without incurring more memory cost. An `attr_accessor` call should result in 2 declarations. We currently have only one definition created, which means that our `definitions_to_declarations` connection fails to model it. Similar issue that we discovered when working on `module_function`.

I don't believe we should change the `definitions_to_declarations` map to store vectors as that will change the size of the values from 2 bytes to vector size + 2 bytes.

Another added benefit, although a lot less important, is that by handling attributes as methods we might also be able to correctly handle the receiver. For example, in a case like this:

```ruby
Foo.attr_reader(:baz)
```

We'd just model it exactly like a method receiver and then in resolution we decide if we go into the singleton level or not based on whether it's a method or an attribute.

### Implementation

I added all attribute types as flags, removed all attribute definitions and fixed all of the code I could find.


Morriar (changes requested):
I don't think this is the right direction, we're trying to work around a huge limitation in the design.

We discovered that we need to be able to encode the fact one definition can lead to multiple declaration: `module_function`, `attr_accessor`, `attr`, and most likely a lot of DSLs (AASM, AcvtiveRecord, StateMachine, etc.).

This invalidate the N:1 definitions:declaration principle we though would work and I think we need to address the fundamental issue rather than introducing hacks.

PR #472, opened 2026-01-15
Save mixins location

Morriar:
As I'm trying to implement diagnostics for mixins, I realized we do not have a location for them.

Consider this:

```rb
class Foo; end

module Bar
  include Foo # diag: trying to include a class
end
```

We currently have no way to attach the diagnostic to the mixin without a location.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1929,73 +1929,13 @@ mod tests {
         }};
     }
 
-    macro_rules! assert_includes_eq {
-        ($context:expr, $def:expr, $expected_names:expr) => {{
+    macro_rules! assert_mixins_eq {
+        ($context:expr, $def:expr, $mixin_type:ident, $expected_names:expr) => {{
             let actual_names = $def
                 .mixins()
                 .iter()
                 .filter_map(|mixin| {
-                    if let Mixin::Include(def) = mixin {
-                        let name = $context
-                            .graph()
-                            .names()
-                            .get(
-                                $context
-                                    .graph()
-                                    .constant_references()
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -832,7 +833,7 @@ impl<'a> RubyIndexer<'a> {
         let parent_nesting_id = self.current_nesting_definition_id();
 
         // Collect all arguments as constant references. Ignore anything that isn't a constant
-        let reference_ids: Vec<_> = arguments
+        let references = arguments
```
st0012: Not directly related to the PR, but can we name this differently? I only know this means "mixin argument constant references" by looking at the comment above.

I think `mixin_constants` or `mixin_arguments` would be better. WDYT?

rust/rubydex/src/resolution.rs
```diff
@@ -832,7 +832,43 @@ impl<'a> Resolver<'a> {
                             // We haven't been able to resolve this name yet, so we push it as a partial linearization to finish
                             // later
                             context.partial = true;
-                            linearized_includes.push_front(Ancestor::Partial(name_id));
+                            linearized_includes.push_front(Ancestor::Partial(*def.name_id()));
+                        }
+                    }
+                }
+                Mixin::Extend(def) => {
```
st0012: I know we need this because def is now different definition structs. But is there anyway to avoid this duplication 😬

Morriar: Removed 👍

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: Renamed to `mixin_arguments` 👍

PR #473, opened 2026-01-15
Save superclass location

Morriar:
As I'm trying to implement diagnostics for class ancestors, I realized we do not have a location for them.

Consider this:

```rb
module Foo; end

class Bar < Foo # diag: trying to subclass a module
end
```

We currently have no way to attach the diagnostic to the superclass without a location.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -583,9 +583,15 @@ impl<'a> RubyIndexer<'a> {
         let offset = Offset::from_prism_location(location);
         let (comments, flags) = self.find_comments_for(offset.start());
         let lexical_nesting_id = self.parent_lexical_scope_id();
-        let superclass = superclass_node
-            .as_ref()
-            .and_then(|n| self.index_constant_reference(n, true));
+        let superclass = superclass_node.as_ref().and_then(|n| {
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -583,9 +583,15 @@ impl<'a> RubyIndexer<'a> {
         let offset = Offset::from_prism_location(location);
         let (comments, flags) = self.find_comments_for(offset.start());
         let lexical_nesting_id = self.parent_lexical_scope_id();
-        let superclass = superclass_node
-            .as_ref()
-            .and_then(|n| self.index_constant_reference(n, true));
+        let superclass = superclass_node.as_ref().and_then(|n| {
+            self.index_constant_reference(n, false).map(|id| {
```
vinistock, line 587: Do we need this `map`? Would it be enough to pass `true` to `index_constant_reference`?

Morriar: The problem is `index_constant_reference` returns a `NameId` and not a `ReferenceId`.

I tried to refactor a bit this part but the result end up messier than what we have currerntly.

I can try again in a follow up to this PR.

vinistock: Aaah, right.

PR #474, opened 2026-01-15
Provide resolution diagnostics

Morriar:
Please review https://github.com/Shopify/rubydex/pull/472 and https://github.com/Shopify/rubydex/pull/473 first.

Closes https://github.com/Shopify/rubydex/pull/435.

Introduce 6 new diagnostics:

* KindRedefinition: redefining a class as a module or vice versa
* ParentRedefinition: redefining the parent of a class
* NonClassSuperclass: subclassing something that is not a class
* CircularDependency: introducing a mixin or parent class cycle
* NonModuleMixin: including something that is not a module
* UnresolvedConstantReference: using a constant that can't be resolved

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -151,6 +151,7 @@ impl<'a> Resolver<'a> {
 
         self.handle_remaining_definitions(other_ids);
 
+        self.validate_constant_references();
         self.validate_declarations();
     }
 
@@ -1696,6 +1697,27 @@ impl<'a> Resolver<'a> {
 
         diagnostics
     }
+
+    fn validate_constant_references(&mut self) {
+        let mut diagnostics = Vec::new();
+
+        for reference in self.graph.constant_references().values() {
+            let name = self.graph.names().get(reference.name_id()).unwrap();
+            if let NameRef::Unresolved(_) = name {
+                diagnostics.push(Diagnostic::new(
+                    Rule::UnresolvedConstantReference,
+                    reference.uri_id(),
+                    reference.offset().clone(),
+                    format!(
+                        "Unresolved constant reference: `{}`",
+                        self.graph.strings().get(name.str()).unwrap().deref()
+                    ),
+                ));
+            }
+        }
+
+        self.graph.diagnostics_mut().extend(diagnostics);
+    }
 }
 
 #[cfg(test)]
@@ -2126,7 +2148,11 @@ mod tests {
         });
         context.resolve();
 
-        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
+        assert_diagnostics_eq!(
+            &context,
+            vec!["unresolved-constant-reference: Unresolved constant reference: `Foo` (1:1-1:4)"],
+            &[Rule::ParseWarning]
+        );
 
         let reference = context.graph().constant_references().values().next().unwrap();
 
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1466,6 +1471,231 @@ impl<'a> Resolver<'a> {
...
+            }
+        }
+
+        diagnostics
+    }
+
+    fn validate_constant_references(&mut self) {
+        let mut diagnostics = Vec::new();
+
+        for reference in self.graph.constant_references().values() {
```
vinistock, line 1704: Instead of looping through the entire list, I think we can just add diagnostics directly to anything that's still in the `unit_queue`, no?

Whatever is left is what we failed to resolve.

rust/rubydex/src/resolution.rs
```diff
@@ -1466,6 +1471,231 @@ impl<'a> Resolver<'a> {
...
+                        self.graph.strings().get(parent_name.name().str()).unwrap()
+                    ),
+                ));
+            }
+        }
+
+        diagnostics
+    }
+
+    fn validate_mixins(&self, declaration: &Declaration) -> Vec<Diagnostic> {
```
vinistock, line 1632: For both of the mixin diagnostics, are we not able to add them during linearization to avoid looping afterwards?

rust/rubydex/src/resolution.rs
```diff
@@ -1466,6 +1471,231 @@ impl<'a> Resolver<'a> {
...
+        let mut diagnostics = Vec::new();
+
+        if declaration.definitions().len() > 1 {
+            let first_definition = self.graph.definitions().get(&declaration.definitions()[0]).unwrap();
+            let first_definition_declaration_kind = DeclarationKind::from_definition_kind(first_definition.kind());
+
+            for definition in declaration.definitions().iter().skip(1) {
+                let definition = self.graph.definitions().get(definition).unwrap();
+                let definition_declaration_kind = DeclarationKind::from_definition_kind(definition.kind());
+                if definition_declaration_kind != first_definition_declaration_kind {
```
vinistock, line 1512: I don't think we can just compare that they're not the same. For example, this is valid code:

```ruby
Foo = Struct.new

class Foo
  include Bar
end
```

I was looking into this specific case as part of the struct work and I believe we need these checks (or maybe some of them) in `Graph#add_declaration`, so that we can auto-promote certain declaration cases like the one in this example.

Another valid case is having a method declaration that has two definitions: one method and one attribute.

Morriar: The trick is in  `from_definition_kind` where we map the output definition kind to declaration kind:

```rs
   pub fn from_definition_kind(definition_kind: DefinitionKind) -> Self {
        match definition_kind {
            DefinitionKind::Class => DeclarationKind::Class,
            DefinitionKind::SingletonClass => DeclarationKind::SingletonClass,
            DefinitionKind::Module => DeclarationKind::Module,
            DefinitionKind::Constant => DeclarationKind::Constant,
            DefinitionKind::ConstantAlias => DeclarationKind::ConstantAlias,
            DefinitionKind::Method
            | DefinitionKind::MethodAlias
            | DefinitionKind::AttrAccessor
            | DefinitionKind::AttrReader
            | DefinitionKind::AttrWriter => DeclarationKind::Method,
            DefinitionKind::InstanceVariable => DeclarationKind::InstanceVariable,
            DefinitionKind::ClassVariable => DeclarationKind::ClassVariable,
            DefinitionKind::GlobalVariable | DefinitionKind::GlobalVariableAlias => DeclarationKind::GlobalVariable,
```

So both an attr accessor and a method definition map to a method declaration so this is correct:

```rb
def foo; end
attr_reader :foo
```

For `Struct`, this would be the same:

```rs
 match definition_kind {
            DefinitionKind::Class | DefinitionKind::Struct => DeclarationKind::Class,
```

Issue #349, opened 2025-12-03
Revisit vector preallocation in `resolution::sorted_units`


vinistock:
We're preallocating the vector sizes based on the number of definitions divided by two. I didn't actually measure the ratio to understand the average in real codebases.

We basically have to understand what is the ratio of classes + modules + singleton classes + constant references vs all of the rest of definitions. Do that for a few codebases, get an average and apply that to the code.

PR #481, opened 2026-01-16
Add partial resolution

thomasmarshall:
Part of #330.

This PR implements partial resolution when a document changes or is removed. We have already implemented the logic that removes definitions, references, names, strings, potentially declarations, and invalidates ancestor chains, whenever a document changes or is removed. This commit ensures the `resolve_all` method only resolves newly added definitions and references, and linearizes ancestors for declarations that were cleared.

I tried a bunch of different approaches to this, such as returning a list of units from `update` and passing them into `resolve_all`, but they were all a little ugly because the first resolution worked differently to subsequent ones. There were lots of conditionals and awkward refactors.

Instead, I quite like the approach here of stashing "pending" IDs on the graph object and then generating sorted units for them. Whenever we add a definition or constant reference to the graph during indexing, we also add its ID to `pending_definitions` or `pending_constant_references`, which are then processed the next time we call `resolve_all`. This way, the first full resolve works exactly the same as subsequent calls.

I think this approach is quite neat and tidy, but very open to other ideas.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -12,6 +12,7 @@ use crate::model::ids::{DeclarationId, DefinitionId, NameId, ReferenceId, String
 use crate::model::name::{Name, NameRef, ResolvedName};
 use crate::model::references::{ConstantReference, MethodRef};
 use crate::model::string_ref::StringRef;
+use crate::resolution::Unit;
 use crate::stats;
 
 pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Object"));
@@ -42,6 +43,8 @@ pub struct Graph {
 
     /// The position encoding used for LSP line/column locations. Not related to the actual encoding of the file
     position_encoding: Encoding,
+
+    pending: Vec<Unit>,
 }
 
 impl Graph {
@@ -57,6 +60,7 @@ impl Graph {
             constant_references: IdentityHashMap::default(),
             method_references: IdentityHashMap::default(),
             position_encoding: Encoding::default(),
+            pending: Vec::new(),
         }
     }
 
@@ -281,6 +285,15 @@ impl Graph {
         &self.names
     }
 
+    #[must_use]
+    pub fn pending(&self) -> &[Unit] {
+        &self.pending
+    }
+
+    pub fn take_pending(&mut self) -> Vec<Unit> {
+        std::mem::take(&mut self.pending)
+    }
+
     /// Decrements the ref count for a name and removes it if the count reaches zero.
     ///
     /// This does not recursively untrack `parent_scope` or `nesting` names.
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1309,84 +1307,100 @@ impl<'a> Resolver<'a> {
...
     /// - The second one contains all other definitions, in no particular order
-    #[must_use]
-    fn sorted_units(&self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
-        let estimated_length = self.graph.definitions().len() / 2;
-        let mut definitions = Vec::with_capacity(estimated_length);
-        let mut others = Vec::with_capacity(estimated_length);
-        let names = self.graph.names();
+    fn sorted_units(&mut self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
+        let pending = self.graph.take_pending();
+        let mut units = Vec::with_capacity(pending.len());
```
vinistock: This also closes #349

vinistock (approved):
I like this approach. @Morriar, this might influence your dependency strategy too.

Honestly, getting this down to 12 seconds, considering that we can parallelize both the constant resolution phase (first step) and the definition creation stage (second step), we might end up getting this even below 4 seconds.

I'd say we can move forward with this for the incremental support (needed for the LSP), but it would be nice to compare both resolution approaches (worklist or dependency tree) with parallelism to understand both performance and maintenance cost/legibility.

rust/rubydex/src/model/graph.rs
```diff
@@ -391,8 +404,11 @@ impl Graph {
                     }
                 }
                 NameRef::Resolved(_) => {
-                    // TODO: consider if this is a valid scenario with the resolution phase design. Either collect
-                    // metrics here or panic if it's never supposed to occur
+                    if let NameRef::Resolved(resolved) = entry.remove() {
+                        let new_resolved =
+                            NameRef::Resolved(Box::new(ResolvedName::new(resolved.name().clone(), declaration_id)));
+                        self.names.insert(name_id, new_resolved);
+                    }
```
thomasmarshall: Is this correct? I had to change it to fix `test_nested_same_names` where there is a reference before another definition with the same name:

```rb
module Foo
end

module Bar
  Foo

  module Foo
  end
end
```

rust/rubydex/src/model/graph.rs
```diff
@@ -42,6 +43,8 @@ pub struct Graph {
 
     /// The position encoding used for LSP line/column locations. Not related to the actual encoding of the file
     position_encoding: Encoding,
+
+    pending: Vec<Unit>,
```
Morriar, line 47: Let's document this

Morriar: I'm not a huge fan of resolution artifacts leaking into the graph, could we store the definitions instead? Do we need to resolve the unit kind ahead of time?

rust/rubydex/src/resolution.rs
```diff
@@ -1309,84 +1306,106 @@ impl<'a> Resolver<'a> {
     /// Returns a tuple of 2 vectors:
     /// - The first one contains all constants, sorted in order for resolution (less complex constant names first)
     /// - The second one contains all other definitions, in no particular order
-    #[must_use]
-    fn sorted_units(&self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
-        let estimated_length = self.graph.definitions().len() / 2;
-        let mut definitions = Vec::with_capacity(estimated_length);
-        let mut others = Vec::with_capacity(estimated_length);
-        let names = self.graph.names();
+    fn sorted_units(&mut self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
```
Morriar, line 1309: Why did we loose the `must_use`?

PR #483, opened 2026-01-16
Remove locks and mutexes from Declaration

Morriar:
I think we can do without them by using `&mut Declaration` references instead.

Let's revisit once we add parallelism.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -476,8 +476,7 @@ impl Graph {
         }
     }
 
-    fn invalidate_ancestor_chains(&self, initial_ids: Vec<DeclarationId>) {
-        let declarations = &self.declarations;
+    fn invalidate_ancestor_chains(&mut self, initial_ids: Vec<DeclarationId>) {
         let mut queue = initial_ids;
         let mut visited = IdentityHashSet::<DeclarationId>::default();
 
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -444,29 +443,37 @@ impl Graph {
...
+                        .get_mut(ancestor_id)
+                        .unwrap()
+                        .as_namespace_mut()
+                        .unwrap()
                         .remove_descendant(&declaration_id);
                 }
-            });
+            }
+
+            let namespace_mut = self
```
vinistock: Nit: 
```suggestion
            let namespace = self
```

rust/rubydex/src/resolution.rs
```diff
@@ -692,7 +692,10 @@ impl<'a> Resolver<'a> {
         } else {
             Ancestors::Complete(ancestors)
         };
-        Self::set_ancestors(self.graph.declarations().get(&declaration_id).unwrap(), result.clone());
+        Self::set_ancestors(
+            self.graph.declarations_mut().get_mut(&declaration_id).unwrap(),
+            result.clone(),
+        );
```
```suggestion
            self.graph.declarations_mut()
              .get_mut(&declaration_id)
              .unwrap()
              .as_namespace_mut()
              .unwrap()
              .set_ancestors(result.clone());
```

rust/rubydex/src/resolution.rs
```diff
@@ -1397,7 +1400,7 @@ impl<'a> Resolver<'a> {
         }
     }
 
-    fn set_ancestors(declaration: &Declaration, ancestors: Ancestors) {
+    fn set_ancestors(declaration: &mut Declaration, ancestors: Ancestors) {
```
vinistock: Let's also remove this method ✂️

Morriar: It's gone now that I rebased on main 👍

PR #484, opened 2026-01-19
Remove diagnostics when documents change

thomasmarshall:
This PR is an alternative approach to #441.

We can add diagnostics to documents, declarations, definitions, and references. Different kinds of diagnostics would be attached to different entities. For example, parse errors would be attached to the document, and resolution errors (like circular dependencies) could be attached to declarations, definitions, or references.

When the document is cleared, each of its definitions and references is removed, along with the diagnostics attached to those entities. Any declaration that is impacted (i.e. because one if its definitions is removed) would have its diagnostics explicitly cleared.

We can access all diagnostics from the top of the graph by iterating through all the entities in the graph and collection the diagnostics. This would be slow, but I'm not sure what the use-case would be. Perhaps it can be improved if we understand when we would want to do this, outside of tests.

We can access all the diagnostics for a particular document via its definitions, references, and associated declarations.

### DIFF
--- a/rust/rubydex/src/diagnostic.rs
+++ b/rust/rubydex/src/diagnostic.rs
@@ -2,7 +2,7 @@
 use crate::model::document::Document;
 use crate::{model::ids::UriId, offset::Offset};
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Diagnostic {
     rule: Rule,
     uri_id: UriId,
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -123,11 +123,11 @@ macro_rules! namespace_declaration {
                 }
             }
 
-            pub fn extend(&mut self, other: Namespace) {
+            pub fn extend(&mut self, mut other: Namespace) {
                 self.definition_ids.extend(other.definitions());
                 self.references.extend(other.references());
                 self.members.extend(other.members());
-                self.diagnostics.extend(other.diagnostics().iter().cloned());
+                self.diagnostics.extend(other.take_diagnostics());
             }
 
             pub fn set_singleton_class_id(&mut self, declaration_id: DeclarationId) {
@@ -226,10 +226,10 @@ macro_rules! simple_declaration {
                 }
             }
 
-            pub fn extend(&mut self, other: Declaration) {
+            pub fn extend(&mut self, mut other: Declaration) {
                 self.definition_ids.extend(other.definitions());
                 self.references.extend(other.references());
-                self.diagnostics.extend(other.diagnostics().iter().cloned());
+                self.diagnostics.extend(other.take_diagnostics());
             }
         }
     };
@@ -354,6 +354,10 @@ impl Declaration {
         all_declarations!(self, it => &it.diagnostics)
     }
 
+    pub fn take_diagnostics(&mut self) -> Vec<Diagnostic> {
+        all_declarations!(self, it => std::mem::take(&mut it.diagnostics))
+    }
+
     pub fn add_diagnostic(&mut self, diagnostic: Diagnostic) {
         all_declarations!(self, it => it.diagnostics.push(diagnostic));
     }
@@ -398,6 +402,10 @@ impl Namespace {
     #[must_use]
     pub fn diagnostics(&self) -> &[Diagnostic] {
         all_namespaces!(self, it => &it.diagnostics)
+    }
+
+    pub fn take_diagnostics(&mut self) -> Vec<Diagnostic> {
+        all_namespaces!(self, it => std::mem::take(&mut it.diagnostics))
     }
 
     /// # Panics


rust/rubydex/src/model/graph.rs
```diff
@@ -173,8 +171,22 @@ impl Graph {
     }
 
     #[must_use]
-    pub fn diagnostics(&self) -> &Vec<Diagnostic> {
-        &self.diagnostics
+    pub fn diagnostics(&self) -> Vec<&Diagnostic> {
```
vinistock: Maybe we can name this `all_diagnostics` to make it clear that this returns the combination of all of them?

rust/rubydex/src/model/declaration.rs
vinistock, line 1: When we extend declarations (i.e.: find a second definition for the same thing), we need to ensure we're also extending the diagnostics.

thomasmarshall: I added a fixup commit for this, but I couldn't find any tests for it, nor where it is called from. Maybe my Rust code navigation skills aren't up to scratch!

vinistock: We might not be doing it yet, but it should happen in `Graph#add_declaration`.

rust/rubydex/src/model/graph.rs
```diff
@@ -173,8 +171,22 @@ impl Graph {
     }
 
     #[must_use]
-    pub fn diagnostics(&self) -> &Vec<Diagnostic> {
-        &self.diagnostics
+    pub fn all_diagnostics(&self) -> Vec<&Diagnostic> {
```
vinistock, line 174: Can we return a slice here?

```suggestion
    pub fn all_diagnostics(&self) -> [&Diagnostic] {
```

rust/rubydex/src/diagnostic.rs
```diff
@@ -2,7 +2,7 @@
 use crate::model::document::Document;
 use crate::{model::ids::UriId, offset::Offset};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
vinistock: Why do we need to clone diagnostics?

rust/rubydex/src/model/graph.rs
thomasmarshall: Hmm I don't know. At the moment it's combining diagnostics from all the different sources together, so it needs to construct a new `Vec` rather than return a slice. Is there an alternative approach we could take here?

thomasmarshall: Could return an iterator to avoid allocating a new `Vec`?

rust/rubydex/src/diagnostic.rs
thomasmarshall: I pushed a fixup that takes them instead of cloning, but it's possible I misunderstood what you were suggesting [here](https://github.com/Shopify/rubydex/pull/484#discussion_r2706028482), in which case neither cloning nor taking might make sense.

rust/rubydex/src/model/graph.rs
vinistock: Oh, right. Sorry, I didn't realize we wouldn't be able to build a slice from multiple vectors. The vec is fine 👍

rust/rubydex/src/diagnostic.rs
vinistock: Maybe my comment was not correct. I wanted to make sure we were not accidentally discarding `Declaration` level diagnostics, but I think this scenario doesn't happen.

And that is probably why we're not using `extend` yet.

rust/rubydex/src/model/definitions.rs
```diff
@@ -149,6 +150,15 @@ impl Definition {
     pub fn is_deprecated(&self) -> bool {
         all_definitions!(self, it => it.flags().is_deprecated())
     }
+
+    #[must_use]
+    pub fn diagnostics(&self) -> &[Diagnostic] {
+        all_definitions!(self, it => &it.diagnostics)
```
Morriar, line 156: This is a costly change when most definitions won't have diagnostics associated.

For declarations I understand the need, but do we need to associate diagnostics to a specific definition, when is it not enough to associate it to the document?

If we do need it, why not store them on the document as a hash?

```rs
document.diagnostics_for_definition(definition) -> Diagnostic[]
```

thomasmarshall: Yeah that's a good point. I opened [a PR here](https://github.com/Shopify/rubydex/pull/516) to move them to the document, as you suggest (and references too).

As for why we need diagnostics linked to specific definitions, I thought there might be some resolution related diagnostics that couldn't be linked to a declaration (maybe in cases where it doesn't exist yet?). Would your circular dependency be one example? @vinistock we discussed this together, but I can't remember the definition-specific example.

vinistock: I think that should be fine. At the end of the day, all diagnostics are in some way related to a document because you have to display them somewhere.

PR #485, opened 2026-01-19
Index Struct.new

vinistock:
Another step for #392

This PR starts handling `Struct.new`, creating all definitions related to it. There are a few things regarding behaviour that's interesting:

- If the first argument is a capitalized string, then Ruby creates an alias for the constant under the `Struct` namespace

```ruby
Foo = Struct.new("Bar")
Foo             # => Foo
Struct::Bar # => Foo
```

- All other arguments become both an attribute accessor and an instance variable
- An `initialize` method is always created (even if there are no fields)
- The `keyword_init` parameter changes the signature of `initialize` to use keyword optional arguments instead of positional optional arguments
- All structs automatically inherit from `Struct`

**Note**: there might be overlap with the handling for `Data.define`, but I didn't want to make this PR too large. I'll refactor if there's an opportunity for it

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -572,14 +572,15 @@ impl<'a> RubyIndexer<'a> {
         Some(definition_id)
     }
 
-    fn handle_class_definition(
+    fn handle_class_definition<F>(
         &mut self,
         location: &ruby_prism::Location,
         name_node: Option<&ruby_prism::Node>,
-        body_node: Option<ruby_prism::Node>,
         superclass_node: Option<ruby_prism::Node>,
-        nesting_type: fn(DefinitionId) -> Nesting,
-    ) {
+        process_body: F,
+    ) where
+        F: FnOnce(&mut Self, DefinitionId),
+    {
         let offset = Offset::from_prism_location(location);
         let (comments, flags) = self.find_comments_for(offset.start());
         let lexical_nesting_id = self.parent_lexical_scope_id();
@@ -625,16 +626,9 @@ impl<'a> RubyIndexer<'a> {
             )));
 
             let definition_id = self.local_graph.add_definition(definition);
-
             self.add_member_to_current_lexical_scope(definition_id);
 
-            if let Some(body) = body_node {
-                self.nesting_stack.push(nesting_type(definition_id));
-                self.visibility_stack.push(Visibility::Public);
-                self.visit(&body);
-                self.visibility_stack.pop();
-                self.nesting_stack.pop();
-            }
+            process_body(self, definition_id);
         }
     }
 
@@ -675,9 +669,7 @@ impl<'a> RubyIndexer<'a> {
 
             if let Some(body) = body_node {
                 self.nesting_stack.push(nesting_type(definition_id));
-                self.visibility_stack.push(Visibility::Public);
-                self.visit(&body);
-                self.visibility_stack.pop();
+                self.visit_with_new_visibility(&body);
                 self.nesting_stack.pop();
             }
         }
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5243,4 +5461,158 @@ mod tests {
...
+        });
+
+        assert_diagnostics_eq!(
+            &context,
+            vec![
+                "dynamic-constant-field: Struct arguments that aren't string or symbol literals will be ignored (1:18-1:31)",
+                "dynamic-constant-field: Struct arguments that aren't string or symbol literals will be ignored (1:33-1:45)"
+            ]
+        );
+    }
```
Morriar, line 5655: I think we can merge these tests

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5243,4 +5461,158 @@ mod tests {
...
+        });
+        assert_no_diagnostics!(&context);
+
+        assert_definition_at!(&context, "1:1-2:4", Class, |foo| {
+            assert_name_id_to_string_eq!(&context, "Foo", foo);
+            let parent = foo.superclass_ref().unwrap();
+            assert_eq!(
+                StringId::from("Struct"),
+                *context.graph().names().get(&parent).unwrap().str()
+            );
```
Morriar: Let's rebase https://github.com/Shopify/rubydex/pull/473 on so you can use `assert_superclass_ref_eq!`.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -703,16 +695,228 @@ impl<'a> RubyIndexer<'a> {
...
+            // First argument is a string, but not a valid constant name. This will crash in the runtime
+            self.local_graph.add_diagnostic(
+                Rule::InvalidConstantName,
+                offset,
+                "When the first argument to Struct.new is a string, it must be a valid constant name".into(),
+            );
+        }
+    }
+
+    fn visit_with_new_visibility(&mut self, node: &ruby_prism::Node) {
```
Morriar, line 918: Let also use it in `visit_singleton_class_node`?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -703,16 +695,228 @@ impl<'a> RubyIndexer<'a> {
             self.handle_class_definition(
                 &node.location(),
                 Some(node),
-                call_node.block(),
                 call_node.arguments().and_then(|args| args.arguments().iter().next()),
-                Nesting::Owner,
+                |indexer, definition_id| {
+                    indexer.nesting_stack.push(Nesting::Owner(definition_id));
```
Morriar, line 707: Can't we leave this in `handle_class_definition` since it's the same in both cases?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -703,16 +695,228 @@ impl<'a> RubyIndexer<'a> {
...
+            );
+
+            return true;
+        }
+
         false
     }
 
+    /// Handles the arguments to `Struct.new` and creates the appropriate definitions
+    fn handle_struct_fields(&mut self, call_node: &ruby_prism::CallNode) {
```
Morriar, line 747: It looks like we're again trying to do some desugaring/rewriting by introducing a synthetic constant alias. Shouldn't we create a `Struct` definition instead and create the proper declaration in the resolver?

I think it's another symptom of https://github.com/Shopify/rubydex/pull/471#pullrequestreview-3679262310 we're trying to work around.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -703,16 +695,228 @@ impl<'a> RubyIndexer<'a> {
...
+
+    fn handle_struct_alias(&mut self, first_node: &ruby_prism::Node, alias_name: &ruby_prism::StringNode) {
+        let offset = Offset::from_prism_location(&first_node.location());
+
+        // Check if the first character is an uppercase letter
+        let name_as_bytes = alias_name.unescaped();
+
+        if name_as_bytes
+            .first()
+            .is_some_and(|char| char.is_ascii_uppercase() && char.is_ascii_alphabetic())
```
thomasmarshall:
```suggestion
            .is_some_and(|char| char.is_ascii_uppercase())
```

Very minor but I think if it's ASCII uppercase (A-Z) the ASCII alphabetic check is redundant.

vinistock: Yes, you're correct!

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: It's not always the same. In `visit_class_node`, this pushes a `Nesting::LexicalScope`.

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: I put a topic for us to discuss in our next meeting.

PR #486, opened 2026-01-19
Correct fully qualified global variable names

### LINKED ISSUE #128: Implement proper full names
* [ ] For constants the full name should be separated by `::`: `A::B::C`
* [ ] For instance methods: `A#foo()`
* [ ] For class methods: `A.foo`
* [ ] For instance variables: `A#@foo`
* [ ] For class variables: `A.@foo`?
* [ ] For global variables: `$foo`

vinistock:
Closes #128

I added a test to verify that all fully qualified names we generate are what we expect. All of them were already correct, with two exceptions:

- Attributes. We're not generating the methods related to attributes yet (proposal in #471)
- Global variables. This is fixed in this PR and there's a slight caveat. I changed the fully qualified to be just the global variable name, but I kept the owner as `Object`. I believe this is fine for now, but we may want to revisit later. Global variables are accessible everywhere, so maybe for completion/go to definition algorithms, it will make more sense to place it in `BasicObject`? Or perhaps we need the magic `<main>`? I'll leave that to a dedicated PR

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -324,9 +324,14 @@ impl<'a> Resolver<'a> {
                 }
                 Definition::GlobalVariable(var) => {
                     let owner_id = *OBJECT_ID;
-                    self.create_declaration(*var.str_id(), id, owner_id, |name| {
+                    let str_id = *var.str_id();
+                    let name = self.graph.strings().get(&str_id).unwrap().clone();
+                    let declaration_id = DeclarationId::from(&name);
+
+                    self.graph.add_declaration(declaration_id, id, || {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -4369,4 +4374,53 @@ mod tests {
...
+
+                def instance_m
+                  @@class_var = 3
+                end
+
+                def self.singleton_m
+                  $global_var = 4
+                end
+
+                def Foo.another_singleton_m; end
```
Morriar, line 4400: Can we add a few tests for tings inside a `class << self` as well?

vinistock: Done!

PR #490, opened 2026-01-19
Keep constant references sorted _after_ the definitions

Morriar:
Also expose a bug in the resolution logic.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1312,7 +1312,7 @@ impl<'a> Resolver<'a> {
     #[must_use]
     fn sorted_units(&self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
         let estimated_length = self.graph.definitions().len() / 2;
-        let mut units = Vec::with_capacity(estimated_length);
+        let mut definitions = Vec::with_capacity(estimated_length);
         let mut others = Vec::with_capacity(estimated_length);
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -4372,4 +4386,34 @@ mod tests {
...
+                  FOO = 42
+                end
+              end
+            "
+        });
+        context.resolve();
+
+        assert_no_diagnostics!(&context, &[Rule::ParseWarning]);
+
+        // FIXME: this is wrong, the reference is not to `Bar::Foo`, but to `Foo`
```
Morriar, line 4410: https://github.com/Shopify/rubydex/issues/491

rust/rubydex/src/resolution.rs
```diff
@@ -1312,7 +1312,7 @@ impl<'a> Resolver<'a> {
     #[must_use]
     fn sorted_units(&self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
         let estimated_length = self.graph.definitions().len() / 2;
-        let mut units = Vec::with_capacity(estimated_length);
+        let mut definitions = Vec::with_capacity(estimated_length);
```
vinistock, line 1315: Let's keep `units`. It's more accurate and the previous `definitions` name as already wrong.

```suggestion
        let mut units = Vec::with_capacity(estimated_length);
```

PR #495, opened 2026-01-20
Bundle third party licenses on release

vinistock:
Since we ship precompiled binaries, we need proper attribution for the Rust dependencies that are involved in the Rubydex gem.

This PR:

- Changes our extconf.rb so that we use `cargo-about` to bundle third party licenses during builds
- Makes sure that we always compile the gem with `RELEASE=true`
- Includes the `THIRD_PARTY_LICENSES.html` file in the packaged Ruby gem


rubydex.gemspec
Morriar, line 5: Do we need to add `spec.licenses`?

PR #502, opened 2026-01-21
Provide resolution diagnostics (inline)

Morriar:
Same than https://github.com/Shopify/rubydex/pull/474 but resolution diagnostics are pushed during the resolution rather than after.

Inline validation appears to be slightly slower, though I _think_ we're in the noise at this point.

```
# Validate after

Timing breakdown
  Listing             0.848s (  2.9%)
  Indexing            4.412s ( 15.1%)
  Resolution         23.282s ( 79.9%)
  Querying            0.585s (  2.0%)
  Total:             29.128s

# Validate inline

Timing breakdown
  Listing             0.784s (  2.6%)
  Indexing            4.809s ( 16.0%)
  Resolution         23.975s ( 79.7%)
  Querying            0.501s (  1.7%)
  Total:             30.070s
```

^ @vinistock

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -280,7 +280,10 @@ impl<'a> Resolver<'a> {
                 self.unit_queue.push_back(unit_id);
             }
             Outcome::Unresolved(None) => {
-                // We couldn't resolve this name. It will be picked up by the drain loop
+                // We couldn't resolve this name. It will be picked up by the drain loop.
+                // TODO: this causes unnecessary retries. `Unresolved(None)` means we had all
+                // dependencies but still failed — it should go to a separate list instead of
+                // being re-queued. See https://github.com/Shopify/rubydex/pull/502#discussion_r2732310351
                 self.unit_queue.push_back(unit_id);
             }
             Outcome::Retry(Some(id_needing_linearization)) | Outcome::Unresolved(Some(id_needing_linearization)) => {
@@ -808,39 +811,23 @@ impl<'a> Resolver<'a> {
         if let Declaration::Namespace(Namespace::SingletonClass(_)) = declaration {
             let attached_decl = self.graph.declarations().get(declaration.owner_id()).unwrap();
 
-            let attached_mixins = attached_decl
-                .definitions()
-                .iter()
-                .filter_map(|definition_id| {
-                    self.mixins_of(*definition_id)
-                        .map(|mixins| mixins.into_iter().map(|mixin| (mixin, *definition_id)))
-                })
-                .flatten()
-                .collect::<Vec<(Mixin, DefinitionId)>>();
-
             mixins.extend(
-                attached_mixins
-                    .into_iter()
-                    .map(|(mixin, _)| mixin)
+                attached_decl
+                    .definitions()
+                    .iter()
+                    .filter_map(|definition_id| self.mixins_of(*definition_id))
+                    .flatten()
                     .filter(|mixin| matches!(mixin, Mixin::Extend(_))),
             );
         }
 
         // Consider only prepends and includes for the current declaration
-        let decl_mixins = declaration
-            .definitions()
-            .iter()
-            .filter_map(|definition_id| {
-                self.mixins_of(*definition_id)
-                    .map(|mixins| mixins.into_iter().map(|mixin| (mixin, *definition_id)))
-            })
-            .flatten()
-            .collect::<Vec<(Mixin, DefinitionId)>>();
-
         mixins.extend(
-            decl_mixins
-                .into_iter()
-                .map(|(mixin, _)| mixin)
+            declaration
+                .definitions()
+                .iter()
+                .filter_map(|definition_id| self.mixins_of(*definition_id))
+                .flatten()
                 .filter(|mixin| matches!(mixin, Mixin::Prepend(_) | Mixin::Include(_))),
         );
 


rust/rubydex/src/resolution.rs
```diff
@@ -146,6 +150,26 @@ impl<'a> Resolver<'a> {
         }
 
         self.handle_remaining_definitions(other_ids);
+
+        for left_unit in unit_queue {
+            if let Unit::Reference(id) = left_unit {
+                let reference = self.graph.constant_references().get(&id).unwrap();
+                let name = self.graph.names().get(reference.name_id()).unwrap();
+                if let NameRef::Unresolved(_) = name {
```
vinistock: We don't need this check. If there's anything left in the queue by the end, it's guaranteed to be unresolved.

rust/rubydex/src/resolution.rs
```diff
@@ -733,6 +781,43 @@ impl<'a> Resolver<'a> {
         result
     }
 
+    fn validate_mixins(&self, mixins: &Vec<(Mixin, DefinitionId)>) -> Vec<Diagnostic> {
```
vinistock: Instead of looping through the mixins twice, could we do this in `linearize_mixins` while we're looping to linearize?

rust/rubydex/src/resolution.rs
```diff
@@ -880,6 +968,28 @@ impl<'a> Resolver<'a> {
             }
         }
 
+        if context.cyclic {
+            for mixin in mixins {
```
vinistock, line 1080: Here instead of looping through the mixins, I think we can add the circular dependency diagnostic directly in `linearize_ancestors` when we conclude that it's a cyclic dependency here: https://github.com/Shopify/rubydex/blob/6b6d1e0ccac39fb500034dbb8927a3d9e3b59d65/rust/rubydex/src/resolution.rs#L642

That way, we automatically add diagnostics for cyclic parent classes too.

rust/rubydex/src/resolution.rs
```diff
@@ -1455,18 +1569,98 @@ impl<'a> Resolver<'a> {
...
+        declaration_id: DeclarationId,
         definition_ids: &[DefinitionId],
         context: &mut LinearizationContext,
     ) -> Ancestors {
-        let (picked_parent, partial) = self.get_parent_class(definition_ids);
+        let (picked_parent, parent_info, partial) = self.get_parent_class(declaration_id, definition_ids);
+
         let result = self.linearize_ancestors(picked_parent, context);
+
+        if matches!(result, Ancestors::Cyclic(_))
```
vinistock, line 1964: This can go away if we move the diagnostic to that spot in `linearize_ancestors`.

rust/rubydex/src/resolution.rs
Morriar: I don't think we can do it here without some involved refactoring as we're missing the information we need to attach the diagnostic to the proper place. At this stage, we don't know if the conflict comes from a parent class or a mixin.

rust/rubydex/src/resolution.rs
Morriar: Same reason than https://github.com/Shopify/rubydex/pull/502#discussion_r2722538513.

rust/rubydex/src/resolution.rs
```diff
@@ -236,6 +266,7 @@ impl<'a> Resolver<'a> {
             }
             Outcome::Unresolved(None) => {
                 // We couldn't resolve this name. Emit a diagnostic
+                unit_queue.push_back(unit_id);
```
vinistock: Just for information, I commented this on Thomas' PR too: there's a bit of a mistake here. The original reason I separated `Retry` from `Unresolved` is to avoid retrying constants that we know for sure are undefined (i.e.: we had all of its dependencies, but still failed to resolve it).

I think we need a list separate from the unit queue for unresolved constants so that we can avoid retrying these.

At the end of the loop, unresolved constants will be the concatenation of the unresolved constants (things we were sure we couldn't resolve) + whatever is left in the queue (stuff we retried, but still failed to resolve).

We can address this in a separate PR, but let's please leave a `TODO`, so that we don't forget. This will increase the number of retries.

rust/rubydex/src/resolution.rs
```diff
@@ -680,28 +712,44 @@ impl<'a> Resolver<'a> {
...
 
+            let attached_mixins = attached_decl
+                .definitions()
+                .iter()
+                .filter_map(|definition_id| {
+                    self.mixins_of(*definition_id)
+                        .map(|mixins| mixins.into_iter().map(|mixin| (mixin, *definition_id)))
+                })
+                .flatten()
+                .collect::<Vec<(Mixin, DefinitionId)>>();
```
vinistock: This is now producing an extra intermediate vector, whereas the code before was extending from an iterator.

Are we able to revert to the original?

rust/rubydex/src/resolution.rs
```diff
@@ -680,28 +712,44 @@ impl<'a> Resolver<'a> {
...
         // Consider only prepends and includes for the current declaration
+        let decl_mixins = declaration
+            .definitions()
+            .iter()
+            .filter_map(|definition_id| {
+                self.mixins_of(*definition_id)
+                    .map(|mixins| mixins.into_iter().map(|mixin| (mixin, *definition_id)))
+            })
+            .flatten()
+            .collect::<Vec<(Mixin, DefinitionId)>>();
```
vinistock: Same thing here.

rust/rubydex/src/resolution.rs
vinistock: Yes, indeed. I think we can change the `seen` from a `HashSet` to a `HashMap`, where the keys are still the declaration IDs and the values are the information you need to add the diagnostic at the right place.

That way we can avoid looping through the mixins again.

I think that will also solve the parent class cyclic check and simplify the code. We can do that in a separate PR though.

rust/rubydex/src/model/declaration.rs
```diff
@@ -54,6 +55,75 @@ impl<'a> IntoIterator for &'a Ancestors {
...
+    Method,
+    GlobalVariable,
+    InstanceVariable,
+    ClassVariable,
+}
+
+impl DeclarationKind {
+    /// Returns the canonical `PascalCase` name used in external APIs and serialization.
+    #[must_use]
+    pub fn as_api_str(self) -> &'static str {
```
vinistock, line 75: Could we name this `as_str` instead?

rust/rubydex/src/model/declaration.rs
```diff
@@ -362,12 +432,20 @@ impl Declaration {
...
     pub fn take_diagnostics(&mut self) -> Vec<Diagnostic> {
         all_declarations!(self, it => std::mem::take(&mut it.diagnostics))
     }
 
     pub fn add_diagnostic(&mut self, diagnostic: Diagnostic) {
-        all_declarations!(self, it => it.diagnostics.push(diagnostic));
+        all_declarations!(self, it => {
+            if !it.diagnostics.iter().any(|d| d.rule() == diagnostic.rule() && d.uri_id() == diagnostic.uri_id() && d.offset() == diagnostic.offset()) {
+                it.diagnostics.push(diagnostic);
+            }
```
vinistock, line 447: Could be a separate PR, but I wonder if we should use a set here to avoid the duplicate check.

rust/rubydex/src/model/graph.rs
```diff
@@ -130,7 +130,37 @@ impl Graph {
...
+                    let declaration = occupied_entry.get_mut();
+                    let definition = self.definitions.get(&definition_id).unwrap();
+                    let definition_declaration_kind = DeclarationKind::from_definition_kind(definition.kind());
+                    // Only emit when the existing declaration is a concrete namespace and the new definition
+                    // is an incompatible kind: a different namespace, or a non-promotable constant.
+                    // We skip constants/aliases becoming namespaces and promotable constants (dynamic values
+                    // that might represent a class).
+                    let is_namespace = matches!(declaration.kind(), DeclarationKind::Class | DeclarationKind::Module);
+                    let is_mismatch = definition_declaration_kind != declaration.kind();
+                    let is_non_promotable_constant =
```
vinistock, line 142: Do we need this considering that this is the else branch of `should_promote`? Or is there a slight distinction in the check?

rust/rubydex/src/model/graph.rs
```diff
@@ -255,6 +285,11 @@ impl Graph {
         &self.documents
     }
 
+    #[must_use]
+    pub fn documents_mut(&mut self) -> &mut IdentityHashMap<UriId, Document> {
```
vinistock, line 289: Not necessarily for this PR, but I'm starting to dislike exposing mutable state like this from the graph publicly.

Maybe we can add more specific APIs and stop doing this. For now, let's at least limit the scope to the crate?


```suggestion
    pub(crate) fn documents_mut(&mut self) -> &mut IdentityHashMap<UriId, Document> {
```

rust/rubydex/src/model/graph.rs
```diff
@@ -366,12 +401,19 @@ impl Graph {
         &self.constant_references
     }
 
+    #[must_use]
+    pub fn constant_references_mut(&mut self) -> &mut IdentityHashMap<ReferenceId, ConstantReference> {
```
vinistock, line 405:
```suggestion
    pub(crate) fn constant_references_mut(&mut self) -> &mut IdentityHashMap<ReferenceId, ConstantReference> {
```

rust/rubydex/src/resolution.rs
```diff
@@ -155,6 +164,38 @@ impl<'a> Resolver<'a> {
...
+                let reference = self.graph.constant_references().get(&id).unwrap();
+                let uri_id = reference.uri_id();
+                let offset = reference.offset().clone();
+                let name = self
+                    .graph
+                    .strings()
+                    .get(self.graph.names().get(reference.name_id()).unwrap().str())
+                    .unwrap()
+                    .deref()
+                    .clone();
```
vinistock, line 179: Do you need these? I think `name` can be a reference since we're only checking it with `starts_with` and `format!` doesn't require ownership of the string.

rust/rubydex/src/resolution.rs
```diff
@@ -869,14 +917,17 @@ impl<'a> Resolver<'a> {
 
     /// Linearize all mixins into a prepend and include list. This function requires the parent ancestors because included
     /// modules are deduplicated against them
+    #[allow(clippy::too_many_lines)]
     fn linearize_mixins(
         &mut self,
+        declaration_id: DeclarationId,
         context: &mut LinearizationContext,
-        mixins: Vec<Mixin>,
+        mixins: &Vec<Mixin>,
```
vinistock, line 925: Could this be `&[Mixin]`?

rust/rubydex/src/resolution.rs
```diff
@@ -892,6 +943,26 @@ impl<'a> Resolver<'a> {
                 Mixin::Prepend(_) => {
                     match self.graph.names().get(constant_reference.name_id()).unwrap() {
                         NameRef::Resolved(resolved) => {
+                            let mixin_declaration = self.graph.declarations().get(resolved.declaration_id()).unwrap();
+
+                            let is_alias_or_promotable = matches!(
+                                mixin_declaration.kind(),
+                                DeclarationKind::ConstantAlias | DeclarationKind::Todo
+                            ) || (mixin_declaration.kind() == DeclarationKind::Constant
+                                && self.graph.all_definitions_promotable(mixin_declaration));
```
vinistock, line 952: Since we need this in a few places, how about we extract a helper?

rust/rubydex/src/resolution.rs
```diff
@@ -5895,4 +6285,132 @@ mod todo_tests {
...
+            vec![
+                "kind-redefinition: Redefining `Foo` as `class`, previously defined as `module` (2:1-2:15)",
+                "kind-redefinition: Redefining `Bar` as `module`, previously defined as `class` (5:1-5:16)",
+                "kind-redefinition: Redefining `Baz` as `constant`, previously defined as `class` (8:1-8:4)",
+            ]
+        );
+    }
+
+    #[test]
+    fn resolution_diagnostics_for_parent_redefinition() {
```
vinistock, line 6331: Can we addd a test for parent classes that use different constant references to refer to the same class?

```ruby
module Foo
  class Bar; end

  class Baz < Bar
  end
end

# There should not be a redefinition diagnostic here since
# it refers to the same thing
class Foo::Baz < Foo::Bar
end
```

PR #503, opened 2026-01-22
Add partial resolution alternative

thomasmarshall:
This PR implements an alternative approach to partial resolution (see #481 for the original approach). There are two main differences:

1. Instead of adding "pending" units to the Graph, we track "dirty" URIs and ancestors. This is preferable because we currently do partial resolution at the granularity of a single document, so we don't need to track individual definitions and references within that document as pending. We can simply reprocess the whole document.
2. This makes it cleaner for `sorted_units` to process either _all_ documents or only dirty ones. If we sometimes operated on all documents (initial resolution) and sometimes operated on a list of pending units (partial updates) then we'd need two different code paths. The previous approach involved us generating pending units even for the initial resolution, which meant a reasonably big increase in memory.

We add a `dirty` property to `Graph` which is either `All` (indicating that we need to resolve everything) or `Partial` which has a set of URIs to reprocess and a set of declarations to relinearize ancestors. The first time we run resolution it processes all documents and then sets dirty to partial with empty sets. When we update a document the URI is added to the dirty URIs set and any declarations that need to have their ancestors relinearized are added to the dirty ancestors set.  The next time we do resolution, we'll only process those dirty documents and ancestors.

### Summary

| Branch | Resolution (avg) | Memory (avg) |
|--------|------------------|--------------|
| `main` | 17.99s | 4920 MB |
| `resolve-single-document` | 12.58s | 5349 MB |
| `resolve-single-document-alt` | 12.38s | 5086 MB |

### Range (5 runs each)

| Branch | Resolution | Memory |
|--------|------------|--------|
| `main` | 17.4 - 18.9s | 4884 - 4915 MB |
| `resolve-single-document` | 12.1 - 13.0s | 5201 - 5464 MB |
| `resolve-single-document-alt` | 11.7 - 13.0s | 5084 - 5091 MB |

### Comparison to main

| Branch | Resolution | Memory |
|--------|-----------|--------|
| `resolve-single-document` | 31% faster | +429 MB (+8.7%) |
| `resolve-single-document-alt` | 31% faster | +166 MB (+3.4%) |

**Why does `resolve-single-document-alt` still have higher Max RSS than `main`?**

On `main` we loop through all `graph.constant_references()` sequentially, whereas on `resolve-single-document-alt` we iterate the `document.constant_references()` and look them up in the graph in a less sequential order. I suspect this causes more pages to be accessed simultaneously within the measurement window.

Both approaches make resolution faster.

### DIFF
--- a/rust/rubydex/src/model/definitions.rs
+++ b/rust/rubydex/src/model/definitions.rs
@@ -142,6 +142,7 @@ impl Definition {
             Definition::SingletonClass(d) => Some(d.name_id()),
             Definition::Module(d) => Some(d.name_id()),
             Definition::Constant(d) => Some(d.name_id()),
+            Definition::ConstantAlias(d) => Some(d.name_id()),
             _ => None,
         }
     }
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -10,7 +10,7 @@ use crate::model::{
         Namespace, SingletonClassDeclaration,
     },
     definitions::{Definition, Mixin},
-    graph::{CLASS_ID, Graph, MODULE_ID, OBJECT_ID},
+    graph::{CLASS_ID, Dirty, Graph, MODULE_ID, OBJECT_ID},
     identity_maps::{IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
     name::{Name, NameRef},
@@ -83,12 +83,8 @@ impl<'a> Resolver<'a> {
     ///
     /// Can panic if there's inconsistent data in the graph
     pub fn resolve_all(&mut self) {
-        // TODO: temporary code while we don't have synchronization. We clear all declarations instead of doing the minimal
-        // amount of work
-        self.graph.clear_declarations();
         // Ensure that Object exists ahead of time so that we can associate top level declarations with the right membership
-
-        {
+        if !self.graph.declarations().contains_key(&OBJECT_ID) {
             self.graph.declarations_mut().insert(
                 *OBJECT_ID,
                 Declaration::Namespace(Namespace::Class(Box::new(ClassDeclaration::new(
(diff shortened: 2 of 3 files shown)


rust/rubydex/src/model/definitions.rs
```diff
@@ -142,6 +142,7 @@ impl Definition {
             Definition::SingletonClass(d) => Some(d.name_id()),
             Definition::Module(d) => Some(d.name_id()),
             Definition::Constant(d) => Some(d.name_id()),
+            Definition::ConstantAlias(d) => Some(d.name_id()),
```
thomasmarshall, line 145: This lets us do use `definition.name_id()` in `sorted_units` instead of matching on definition with an identical arm for each type.

rust/rubydex/src/resolution.rs
```diff
@@ -1310,81 +1306,85 @@ impl<'a> Resolver<'a> {
     /// - The first one contains all constants, sorted in order for resolution (less complex constant names first)
     /// - The second one contains all other definitions, in no particular order
     #[must_use]
-    fn sorted_units(&self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
-        let estimated_length = self.graph.definitions().len() / 2;
-        let mut definitions = Vec::with_capacity(estimated_length);
-        let mut others = Vec::with_capacity(estimated_length);
+    fn sorted_units(&mut self) -> (VecDeque<Unit>, Vec<DefinitionId>) {
+        let dirty = self.graph.take_dirty();
```
thomasmarshall, line 1310: This changes `graph.dirty` from `All` to `Partial` or empties the sets in `Partial`.

rust/rubydex/src/resolution.rs
```diff
@@ -1310,81 +1306,85 @@ impl<'a> Resolver<'a> {
...
+
+        let mut definitions = Vec::new();
```
thomasmarshall, line 1312: I removed estimated capacity—I'm not sure the best way to calculate this given we're sometimes doing partial updates. As indicated in #349 it's not currently correct anyway.

Issue #89, opened 2025-08-12
Start keeping track of visibility for applicable definitions


vinistock:
Some definitions have visibility. For example, methods and constants. Others don't have this concept, like instance variables or global variables.

Define a way to model visibility (should it be through bit flags?) and start keeping track of it for applicable definitions. For example:

```ruby
class Foo
  CONST = 1
  private_constant(:CONST) # private constant

  private def private_method; end

  def now_public; end

  protected

  def now_protected; end

  def self.public_because_protected_does_not_apply_here; end
  def self.make_me_private; end
  private_singleton_method(:make_me_private)
end
```

PR #506, opened 2026-01-22
Index calls to `private_constant` and `public_constant`

Morriar:
One more step towards https://github.com/Shopify/rubydex/issues/89 and https://github.com/Shopify/rubydex/issues/354. We start keeping track of calls to `private_constant` and `public_constant`.

For these ones I chose to keep a new definition around because we'll need to resolve the names to be able to apply the visibility to the right constant.

Take this code for example:

```rb
# file1.rb
class Foo
  Bar = 42
end

# file2.rb
Foo.private_constant(:Bar)
```

there is no way to handle it within the indexer.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -5,10 +5,10 @@ use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
     AttrAccessorDefinition, AttrReaderDefinition, AttrWriterDefinition, ClassDefinition, ClassVariableDefinition,
-    ConstantAliasDefinition, ConstantDefinition, Definition, DefinitionFlags, ExtendDefinition,
-    GlobalVariableAliasDefinition, GlobalVariableDefinition, IncludeDefinition, InstanceVariableDefinition,
-    MethodAliasDefinition, MethodDefinition, Mixin, ModuleDefinition, Parameter, ParameterStruct, PrependDefinition,
-    Receiver, Signatures, SingletonClassDefinition,
+    ConstantAliasDefinition, ConstantDefinition, ConstantVisibilityDefinition, Definition, DefinitionFlags,
+    ExtendDefinition, GlobalVariableAliasDefinition, GlobalVariableDefinition, IncludeDefinition,
+    InstanceVariableDefinition, MethodAliasDefinition, MethodDefinition, Mixin, ModuleDefinition, Parameter,
+    ParameterStruct, PrependDefinition, Receiver, Signatures, SingletonClassDefinition,
 };
 use crate::model::document::Document;
 use crate::model::ids::{DefinitionId, NameId, StringId, UriId};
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1164,6 +1164,84 @@ impl<'a> RubyIndexer<'a> {
...
+            Some(ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. }) => {
+                self.index_constant_reference(&receiver.unwrap(), true)
+            }
+            Some(ruby_prism::Node::SelfNode { .. }) | None => match self.nesting_stack.last() {
+                Some(Nesting::Method(_)) => {
+                    self.local_graph.add_diagnostic(
+                        Rule::InvalidPrivateConstant,
+                        Offset::from_prism_location(&node.location()),
+                        "Private constant called from a method".to_string(),
+                    );
```
vinistock: I think we need to defer this to once we have more inference infrastructure. You can invoke `private_constant` if it's at the class level and this feels like type checking the method call.

```ruby
class Foo
  CONST = 1

  def self.bar
    private_constant(:CONST) # ok!
  end

  def baz
    # Not okay, but not because it's inside a method
    # It's because `private_constant` is defined in `Module`
    private_constant(:CONST)
  end    
end
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1164,6 +1164,84 @@ impl<'a> RubyIndexer<'a> {
...
+                return;
+            }
+        };
+
+        let Some(arguments) = node.arguments() else {
+            return;
+        };
+
+        for argument in &arguments.arguments() {
+            if let Some(symbol_node) = argument.as_symbol_node() {
```
vinistock: It accepts strings too. Maybe we can use `each_string_or_symbol_arg` here?

rust/rubydex/src/model/definitions.rs
```diff
@@ -701,6 +705,80 @@ impl ConstantAliasDefinition {
...
+
+impl ConstantVisibilityDefinition {
+    #[must_use]
+    pub const fn new(
+        name_id: NameId,
+        visibility: Visibility,
+        uri_id: UriId,
+        offset: Offset,
+        comments: Vec<Comment>,
+        flags: DefinitionFlags,
```
vinistock, line 727: Not sure if we need these. Maybe we can get rid of them?

vinistock (approved):

rust/rubydex/src/model/definitions.rs
Morriar: I don't want to create a special case for https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/model/definitions.rs#L174-L177.

I think we need to review this later though, `deprecated` only makes sense for some definitions.

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: Removed the diagnostic for now 👍

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: I added support for strings but didn't use `each_string_or_symbol_arg` as I didn't want to silently ignore the non symbol/sting ones and still produce diagnostics.

vinistock: Would it be possible to add the diagnostic in `each_string_or_symbol_arg`? It might require a more generic message, but I think all indexing parts that use it can't handle dynamic arguments.

rust/rubydex/src/model/definitions.rs
vinistock: Yeah, makes sense.

PR #509, opened 2026-01-22
Remove `Graph#definitions_to_declarations` map

Morriar:
We maintain a huge hash to map a definition to a declaration but in practice we don't use it that much.

It looks like we can do without it and be 1) faster while 2) using less memory:

```
# Before

Timing breakdown
  Listing             0.776s (  2.6%)
  Indexing            5.264s ( 17.5%)
  Resolution         23.504s ( 78.2%)
  Querying            0.502s (  1.7%)
  Total:             30.047s

# After

Timing breakdown
  Listing             0.919s (  3.3%)
  Indexing            4.933s ( 17.8%)
  Resolution         21.274s ( 77.0%)
  Querying            0.512s (  1.9%)
  Total:             27.639s
```

I'm experimenting with this direction so we can replace the N:1 mapping from a definition to a declaration by a N:M without cost.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -29,8 +29,6 @@ pub struct Graph {
     // Map of definition nodes
     definitions: IdentityHashMap<DefinitionId, Definition>,
 
-    definitions_to_declarations: IdentityHashMap<DefinitionId, DeclarationId>,
-
     // Map of unqualified names
     strings: IdentityHashMap<StringId, StringRef>,
     // Map of names
@@ -50,7 +48,6 @@ impl Graph {
         Self {
             declarations: IdentityHashMap::default(),
             definitions: IdentityHashMap::default(),
-            definitions_to_declarations: IdentityHashMap::default(),
             documents: IdentityHashMap::default(),
             strings: IdentityHashMap::default(),
             names: IdentityHashMap::default(),
@@ -79,8 +76,6 @@ impl Graph {
     where
         F: FnOnce() -> Declaration,
     {
-        self.definitions_to_declarations.insert(definition_id, declaration_id);
-
         let declaration = self.declarations.entry(declaration_id).or_insert_with(constructor);
         declaration.add_definition(definition_id);
     }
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1,3 +1,4 @@
+// WIP
```
vinistock: ✂️

rubydex.gemspec
vinistock, line 30: Might need a rebase for this one

rust/rubydex/src/model/graph.rs
```diff
@@ -153,9 +147,104 @@ impl Graph {
...
+                    .and_then(|id| self.definitions().get(&id).unwrap().name_id());
+                (nesting_definition, it.str_id())
+            }
+            Definition::MethodAlias(it) => {
+                let nesting_definition = it
+                    .lexical_nesting_id()
+                    .and_then(|id| self.definitions().get(&id).unwrap().name_id());
+                (nesting_definition, it.new_name_str_id())
+            }
+            _ => unreachable!("Unhandled definition type in definition_to_declaration"),
```
vinistock: Do we need the catchall here? It looks like we're already handling all possibilities.

rust/rubydex/src/model/graph.rs
```diff
@@ -153,9 +147,104 @@ impl Graph {
...
+            .unwrap()
+            .as_namespace()
+            .unwrap()
+            .member(member_str_id)
+    }
+
     #[must_use]
-    pub fn definitions_to_declarations(&self) -> &IdentityHashMap<DefinitionId, DeclarationId> {
-        &self.definitions_to_declarations
+    pub fn resolve_name_id(&self, name_id: NameId) -> Option<&DeclarationId> {
```
vinistock: Personally, I'd rename this slightly since it doesn't actually run resolution, but instead returns what has been already resolved.

Maybe `declaration_for_name_id`?

rust/rubydex/src/model/graph.rs
Morriar: I can remove it if I merge both `match` expression 👍

rust/rubydex/src/model/graph.rs
Morriar: Renamed to `name_id_to_declaration_id` 👍

PR #514, opened 2026-01-22
Fix constant lookup to wait for unresolved ancestors

thomasmarshall:
```rb
module Baz
  X = 1
end

class Bar
  X = 2
end

class Qux < Bar
  X
end
```

Here the `X` reference in `Qux` should be resolved to `Bar::X` (`2`).

But with this second file:

```rb
class Bar
  prepend Baz
end
```

…we prepend `Baz` so the ancestor chain for `Qux` is `[Qux, Baz, Bar, etc]` so the `X` reference should be resolved to `Baz::X` (`1`).

During resolution, some ancestors might not be fully resolved yet ("partial") and we would previously skip past them to return the first constant found.

In the example above, the ancestor chain looks like:

```
[Qux, Partial(Baz), Bar, etc]
```

…we would skip `Partial(Baz)`, find `X` in `Bar`, and incorrectly return `Bar::X`.

My attempt at a fix is to track whether we've seen any unresolved `Partial` before finding a constant. If we have, we return `Unresolved` to defer the lookup until later, when `Baz` is fully resolved.

I couldn't recreate this issue with a single file, it only seemed to happen in a multi-file setup, but I'm not sure exactly what's different—this is unrelated to partial/incremental resolution, it happens for the initial full resolve.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1071,7 +1071,8 @@ impl<'a> Resolver<'a> {
                                             self.graph.record_resolved_name(name_id, declaration_id);
                                             return Outcome::Resolved(declaration_id, None);
                                         }
-                                        Outcome::Unresolved(Some(needs_linearization_id)) => {
+                                        Outcome::Retry(Some(needs_linearization_id))
+                                        | Outcome::Unresolved(Some(needs_linearization_id)) => {
                                             missing_linearization_id.get_or_insert(needs_linearization_id);
                                         }
                                         Outcome::Unresolved(None) => {}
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -197,14 +198,14 @@ impl<'a> Resolver<'a> {
...
-            Outcome::Retry => {
+            Outcome::Retry(None) => {
                 // There might be dependencies we haven't figured out yet, so we need to retry
                 unit_queue.push_back(unit_id);
             }
             Outcome::Unresolved(None) => {
                 // We couldn't resolve this name. Emit a diagnostic
             }
-            Outcome::Unresolved(Some(id_needing_linearization)) => {
+            Outcome::Retry(Some(id_needing_linearization)) | Outcome::Unresolved(Some(id_needing_linearization)) => {
```
vinistock, line 208: Something is funny here, although not related to your PR. We shouldn't be re-enqueueing an `Outcome::Unresolved`. That outcome means: "we had everything we needed to resolve this constant and still failed".

The goal being that we can avoid retrying something we know we cannot resolve.

rust/rubydex/src/resolution.rs
```diff
@@ -1201,23 +1203,33 @@ impl<'a> Resolver<'a> {
...
-                    } else {
-                        None
+            Ancestors::Partial(ids) => {
+                for ancestor_id in ids {
+                    match ancestor_id {
+                        Ancestor::Partial(name_id) => {
+                            // Stop at unresolved ancestors to avoid resolving to a later one.
+                            // Skip if the name matches what we're searching for.
+                            if *self.graph.names().get(&name_id).unwrap().str() != str_id {
+                                return Outcome::Unresolved(Some(declaration_id));
```
vinistock: I think this should be a `Retry` and it worked accidentally because we're also retrying `Unresolved`.
```suggestion
                                return Outcome::Retry(Some(declaration_id));
```

PR #518, opened 2026-01-23
Add Declaration#ancestors to the Ruby API

vinistock:
This PR exposes `Declaration#ancestors` to the Ruby API and demonstrates how it can be used to resolve a method or instance variable reference (necessary for hover, go to definition and signature help).

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -3,6 +3,7 @@
 #include "graph.h"
 #include "handle.h"
 #include "rustbindings.h"
+#include "utils.h"
 
 VALUE cDeclaration;
 VALUE cNamespace;
@@ -208,6 +209,29 @@ static VALUE rdxr_declaration_owner(VALUE self) {
     return rb_class_new_instance(2, argv, decl_class);
 }
 
+// Declaration#ancestors: () -> Enumerator[Declaration]
+static VALUE rdxr_declaration_ancestors(VALUE self) {
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize(self, rb_str_new2("ancestors"), 0, NULL);
+    }
+
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    void *iter = rdx_declaration_ancestors(graph, data->id);
+    if (iter == NULL) {
+        rb_raise(rb_eRuntimeError, "failed to create iterator");
+    }
+
+    VALUE args = rb_ary_new_from_args(2, data->graph_obj, ULL2NUM((uintptr_t)iter));
+    rb_ensure(rdxi_declarations_yield, args, rdxi_declarations_ensure, args);
+
+    return self;
+}
+
 void rdxi_initialize_declaration(VALUE mRubydex) {
     cDeclaration = rb_define_class_under(mRubydex, "Declaration", rb_cObject);
     cNamespace = rb_define_class_under(mRubydex, "Namespace", cDeclaration);
@@ -231,6 +255,7 @@ void rdxi_initialize_declaration(VALUE mRubydex) {
     // Namespace only methods
     rb_define_method(cNamespace, "member", rdxr_declaration_member, 1);
     rb_define_method(cNamespace, "singleton_class", rdxr_declaration_singleton_class, 0);
+    rb_define_method(cNamespace, "ancestors", rdxr_declaration_ancestors, 0);
 
     rb_funcall(rb_singleton_class(cDeclaration), rb_intern("private"), 1, ID2SYM(rb_intern("new")));
 }
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -58,30 +58,6 @@ static VALUE rdxr_graph_index_all(VALUE self, VALUE file_paths) {
     return Qnil;
 }
 
-// Body function for rb_ensure in Graph#declarations
-static VALUE graph_declarations_yield(VALUE args) {
-    VALUE self = rb_ary_entry(args, 0);
-    void *iter = (void *)(uintptr_t)NUM2ULL(rb_ary_entry(args, 1));
-
-    CDeclaration decl;
-    while (rdx_graph_declarations_iter_next(iter, &decl)) {
-        VALUE decl_class = rdxi_declaration_class_for_kind(decl.kind);
-        VALUE argv[] = {self, UINT2NUM(decl.id)};
-        VALUE handle = rb_class_new_instance(2, argv, decl_class);
-        rb_yield(handle);
-    }
-
-    return Qnil;
-}
-
-// Ensure function for rb_ensure in Graph#declarations to always free the iterator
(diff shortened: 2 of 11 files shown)


ext/rubydex/graph.c
```diff
@@ -58,29 +58,6 @@ static VALUE rdxr_graph_index_all(VALUE self, VALUE file_paths) {
     return Qnil;
 }
 
-// Body function for rb_ensure in Graph#declarations
-static VALUE graph_declarations_yield(VALUE args) {
```
vinistock, line 62: These can be reused any time we want to create a declaration iterator, so I  moved to utils.

rust/rubydex-sys/src/declaration_api.rs
```diff
@@ -10,6 +11,49 @@ use crate::graph_api::{GraphPointer, with_graph};
 use crate::utils;
 use rubydex::model::ids::{DeclarationId, StringId};
 
+/// An iterator over declaration IDs
+///
+/// We snapshot the IDs at iterator creation so if the graph is modified, the iterator will not see the changes
+#[derive(Debug)]
+pub struct DeclarationsIter {
```
vinistock, line 68: I moved this so that we can reuse and left it in `declarations_api` since it's a declaration iterator.

ext/rubydex/utils.c
```diff
@@ -25,3 +27,25 @@ void rdxi_check_array_of_strings(VALUE array) {
         Check_Type(item, T_STRING);
     }
 }
+
+// Yield body for iterating over declarations
+VALUE declarations_yield(VALUE args) {
```
Morriar: Shouldn't it be `rdxi` prefixed?

ext/rubydex/utils.c
```diff
@@ -25,3 +27,25 @@ void rdxi_check_array_of_strings(VALUE array) {
...
+        VALUE argv[] = {self, LL2NUM(id)};
+        VALUE handle = rb_class_new_instance(2, argv, cDeclaration);
+        rb_yield(handle);
+    }
+
+    return Qnil;
+}
+
+// Ensure function for iterating over declarations to always free the iterator
+VALUE declarations_ensure(VALUE args) {
```
Morriar: Same about the prefix

rust/rubydex-sys/src/declaration_api.rs
```diff
@@ -112,3 +156,27 @@ pub unsafe extern "C" fn rdx_declaration_definitions_iter_new(
...
+///
+/// # Safety
+///
+/// Assumes that the graph and member pointers are valid
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_declaration_ancestors(pointer: GraphPointer, decl_id: i64) -> *mut DeclarationsIter {
+    let ids = with_graph(pointer, |graph| {
+        let declaration_id = DeclarationId::new(decl_id);
+
+        Resolver::new(graph)
```
Morriar: Outside of singleton ancestors, everything should already be available  through `Namespace#ancestors`: https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/model/declaration.rs#L414-L417, no need for a resolver here.

For the singleton, _if_ we were to create them dynamically, we'll need to compute their ancestors as well so by the time we get here, they should also be available in the namespace.

ext/rubydex/declaration.c
```diff
@@ -227,6 +251,7 @@ void rdxi_initialize_declaration(VALUE mRubydex) {
     rb_define_method(cDeclaration, "unqualified_name", rdxr_declaration_unqualified_name, 0);
     rb_define_method(cDeclaration, "definitions", rdxr_declaration_definitions, 0);
     rb_define_method(cDeclaration, "owner", rdxr_declaration_owner, 0);
+    rb_define_method(cDeclaration, "ancestors", rdxr_declaration_ancestors, 0);
```
Morriar: Should it be only on `Namespace`?

rust/rubydex/src/model/graph.rs
```diff
@@ -896,13 +896,13 @@ mod tests {
 
         let foo_declaration = context.graph().declarations().get(&DeclarationId::from("Foo")).unwrap();
         assert!(matches!(
-            foo_declaration.as_namespace().unwrap().ancestors(),
```
Morriar, line 899: Why do we need to clone just to assert?

rust/rubydex/src/model/graph.rs
```diff
@@ -896,13 +896,13 @@ mod tests {
...
+            foo_declaration.as_namespace().unwrap().clone_ancestors(),
             Ancestors::Complete(_)
         ));
 
         let baz_declaration = context.graph().declarations().get(&DeclarationId::from("Baz")).unwrap();
         assert!(matches!(
-            baz_declaration.as_namespace().unwrap().ancestors(),
+            baz_declaration.as_namespace().unwrap().clone_ancestors(),
```
Morriar: Same question

rust/rubydex/src/model/graph.rs
```diff
@@ -947,7 +947,7 @@ mod tests {
 
         let baz_declaration = context.graph().declarations().get(&DeclarationId::from("Baz")).unwrap();
         assert!(matches!(
-            baz_declaration.as_namespace().unwrap().ancestors(),
```
Morriar, line 950: Same

rust/rubydex/src/model/graph.rs
```diff
@@ -652,7 +652,7 @@ impl Graph {
                 .as_namespace_mut()
                 .expect("expected namespace declaration");
 
-            for ancestor in &namespace.ancestors() {
+            for ancestor in &namespace.clone_ancestors() {
```
Morriar, line 655: Do we need to clone here?

rust/rubydex/src/resolution.rs
```diff
@@ -625,7 +625,7 @@ impl<'a> Resolver<'a> {
             // Return the cached ancestors if we already computed them. If they are partial ancestors, ignore the cache to try
             // again
             if declaration.as_namespace().unwrap().has_complete_ancestors() {
-                let cached = declaration.as_namespace().unwrap().ancestors();
+                let cached = declaration.as_namespace().unwrap().clone_ancestors();
```
Morriar, line 628: Do we need to clone here?

rust/rubydex/src/resolution.rs
```diff
@@ -2117,7 +2117,7 @@ mod tests {
 
         let declaration = context.graph().declarations().get(&DeclarationId::from("Bar")).unwrap();
         assert!(matches!(
-            declaration.as_namespace().unwrap().ancestors(),
+            declaration.as_namespace().unwrap().clone_ancestors(),
```
Morriar, line 2120: Same

rust/rubydex/src/model/graph.rs
vinistock: Yes, because otherwise we try to get a mutable borrow while there's an immutable one.

rust/rubydex/src/resolution.rs
vinistock: It might be possible to stop cloning from the cache, but notice that this was just a rename. We can check this in a new PR.

PR #522, opened 2026-01-26
Incremental resolution tests

thomasmarshall:
This PR adds a set of cross-file and incremental resolution tests exercising more complex examples that could be liable to break if our incremental approach isn't robust. The idea is that these tests all pass when performing full re-indexing/resolution, and should remain passing when we implement a more incremental approach. We can add extra assertions to them as we implement the incremental approach (e.g. making assertions about the intermediate state when entities have been removed/added/invalidated, but before they have been re-resolved).

Some subset of the tests was failing because the temporary `clear_declarations` function was not resetting resolved names to unresolved. This is only supposed to be a temporary measure, but I fixed it for now by resetting the names. This can be deleted/removed potentially when we move to a proper incremental approach.

Claude helped me generate some of the scenarios, and there's some degree of overlap in certain cases. However, I have been using these tests to validate the various incremental resolution prototypes I've been working on and they have proved invaluable. I have annotated the diff with a description of what each test is doing, which should make it relatively straightforward to sense check (I have been through and validated each one as I added the annotations).

TODO: Add an incremental version of the test [here](https://github.com/Shopify/rubydex/pull/514).

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1487,9 +1487,9 @@ mod tests {
     use crate::test_utils::GraphTest;
     use crate::{
         assert_alias_targets_contain, assert_ancestors_eq, assert_constant_alias_target_eq,
-        assert_constant_reference_to, assert_descendants, assert_diagnostics_eq, assert_instance_variables_eq,
-        assert_members_eq, assert_no_constant_alias_target, assert_no_diagnostics, assert_no_members, assert_owner_eq,
-        assert_singleton_class_eq,
+        assert_constant_reference_to, assert_constant_reference_unresolved, assert_descendants, assert_diagnostics_eq,
+        assert_instance_variables_eq, assert_members_eq, assert_no_constant_alias_target, assert_no_diagnostics,
+        assert_no_members, assert_owner_eq, assert_singleton_class_eq,
     };
 
     #[test]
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
         assert_no_members!(context, "Foo");
         assert_members_eq!(context, "Bar::Foo", vec!["FOO"]);
     }
+
+    #[test]
+    fn adding_closer_constant_shadows_outer() {
```
thomasmarshall, line 4103: `BAR` in `Foo::Qux::Baz` initially resolves to `Foo::BAR`. After adding `Foo::Qux::BAR` in a new file, `BAR` resolves to `Foo::Qux::BAR` (closer scope wins).

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Foo::Qux::BAR", "file:///file2.rb:3:6-3:9");
+    }
+
+    #[test]
+    fn removing_shadowing_constant_reveals_outer() {
```
thomasmarshall, line 4143: `BAR` in `Foo::Qux::Baz` resolves to `Foo::Qux::BAR`. After removing `Foo::Qux::BAR`, it resolves to `Foo::BAR` (falls back to outer scope).

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Foo::BAR", "file:///file3.rb:3:6-3:9");
+    }
+
+    #[test]
+    fn adding_mixin_provides_constant() {
```
thomasmarshall, line 4191: `CONST` in `Foo` is unresolved. After adding `include Bar` to `Foo`, `CONST` resolves to `Bar::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file2.rb:2:2-2:7");
+    }
+
+    #[test]
+    fn adding_mixin_in_separate_file_provides_constant() {
```
thomasmarshall, line 4226: Same as above but the `include Bar` is added in a separate file that reopens `Foo`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file2.rb:1:2-1:7");
+    }
+
+    #[test]
+    fn removing_mixin_removes_constant_access() {
```
thomasmarshall, line 4260: `CONST` in `Foo` resolves to `Bar::CONST` via include. After removing the include, `CONST` becomes unresolved.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_unresolved!(context, "file:///file2.rb:1:2-1:7");
+    }
+
+    #[test]
+    fn changing_superclass_affects_resolution() {
```
thomasmarshall, line 4295: `CONST` in `Foo < Bar` resolves to `Bar::CONST`. After changing to `Foo < Baz`, it resolves to `Baz::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+              CONST
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Baz::CONST", "file:///file3.rb:1:2-1:7");
+    }
+    #[test]
+    fn adding_constant_to_ancestor_affects_descendants() {
```
thomasmarshall, line 4335: `INHERITED` in `Child < Parent` is unresolved. After adding `INHERITED` to `Parent`, it resolves to `Parent::INHERITED`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,1025 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Parent::INHERITED", "file:///file2.rb:1:2-1:11");
+    }
+
+    #[test]
+    fn offset_only_changes_preserve_resolution() {
```
thomasmarshall, line 4368: Adding comments to a file (offset-only change) doesn't affect resolution of `Foo::BAR`. In a future commit we could add an assertion that this doesn't result in resolution running.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,995 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Foo::BAR", "file:///file2.rb:0:5-0:8");
+    }
+
+    #[test]
+    fn alias_chains_across_files() {
```
thomasmarshall, line 4402: `Baz::CONST` resolves to `Foo::CONST` through alias chain `Baz = Bar`, `Bar = Foo`. After changing `Bar = Qux`, it resolves to `Qux::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,995 @@ mod tests {
...
+            Bar = Qux
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Qux::CONST", "file:///file4.rb:0:5-0:10");
+    }
+
+    #[test]
+    fn removed_namespace_removes_all_members() {
```
thomasmarshall, line 4445: `Foo::Bar::CONST` and `Foo::BAZ` are resolved. After deleting the file defining `Foo`, both become unresolved.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+
+        context.delete_uri("file:///file1.rb");
+        context.resolve();
+
+        assert_constant_reference_unresolved!(context, "file:///file2.rb:0:10-0:15");
+        assert_constant_reference_unresolved!(context, "file:///file2.rb:1:5-1:8");
+    }
+
+    #[test]
+    fn deeply_nested_shadowing() {
```
thomasmarshall, line 4477: `CONST` in `Foo::Bar::Baz` resolves to `Foo::Bar::CONST`. After removing it, resolves to `Foo::CONST`. After removing that, resolves to top-level `CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "CONST", "file:///file1.rb:5:6-5:11");
+    }
+
+    #[test]
+    fn mixin_added_to_ancestor_affects_descendant() {
```
thomasmarshall, line 4539: `CONST` in `Foo < Bar` is unresolved. After adding `include Baz` to `Bar`, it resolves to `Baz::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Baz::CONST", "file:///file3.rb:1:2-1:7");
+    }
+
+    #[test]
+    fn adding_definition_for_previously_missing_owner() {
```
thomasmarshall, line 4579: `Foo::Bar` with compact syntax is orphaned when `Foo` doesn't exist. After adding `module Foo`, both `Foo::Bar` and `Foo::Bar::CONST` become valid declarations.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            context
+                .graph()
+                .declarations()
+                .get(&DeclarationId::from("Foo::Bar::CONST"))
+                .is_some()
+        );
+    }
+
+    #[test]
+    fn removing_definition_owner_orphans_children() {
```
thomasmarshall, line 4624: `Foo::Bar::CONST` is resolved. After deleting `module Foo`, it becomes unresolved (orphaned).

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+        assert_constant_reference_to!(context, "Foo::Bar::CONST", "file:///file3.rb:0:10-0:15");
+
+        context.delete_uri("file:///file1.rb");
+        context.resolve();
+
+        assert_constant_reference_unresolved!(context, "file:///file3.rb:0:10-0:15");
+    }
+
+    #[test]
+    fn reference_in_updated_file_gets_new_resolution() {
```
thomasmarshall, line 4656: `Foo::A` resolves. After changing the reference to `Foo::B`, it resolves to `Foo::B`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            Foo::B
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Foo::B", "file:///file2.rb:0:5-0:6");
+    }
+
+    #[test]
+    fn multiple_references_to_same_constant_all_update() {
```
thomasmarshall, line 4687: `CONST` in `Foo` resolves to `Foo::CONST`. `CONST` in `Foo::Bar` also resolves to `Foo::CONST`. After adding `Foo::Bar::CONST`, only the nested one changes to `Foo::Bar::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Foo::CONST", "file:///file2.rb:1:2-1:7");
+        assert_constant_reference_to!(context, "Foo::Bar::CONST", "file:///file3.rb:2:4-2:9");
+    }
+
+    #[test]
+    fn extend_adds_singleton_class_methods() {
```
thomasmarshall, line 4734: `CONST` in `class << self` is unresolved. After adding `extend Bar`, it resolves to `Bar::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file2.rb:3:4-3:9");
+    }
+
+    #[test]
+    fn mixin_change_updates_references() {
```
thomasmarshall, line 4773: `CONST` in `Foo` resolves to `Bar::CONST` via include. After changing to `include Baz`, it resolves to `Baz::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Baz::CONST", "file:///file2.rb:3:4-3:9");
+    }
+
+    #[test]
+    fn alias_target_change() {
```
thomasmarshall, line 4816: `Baz::CONST` resolves to `Foo::CONST` when `Baz = Foo`. After changing to `Baz = Bar`, alias target updates.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            "
+        });
+        context.resolve();
+
+        assert_constant_alias_target_eq!(context, "Baz", "Bar");
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file2.rb:1:5-1:10");
+    }
+
+    #[test]
+    fn file_deletion_unresolves_references() {
```
thomasmarshall, line 4852: `CONST` in `Foo` resolves to `Bar::CONST` via include. After deleting the file with `Bar`, it becomes unresolved.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file2.rb:2:14-2:19");
+
+        context.delete_uri("file:///file1.rb");
+        context.resolve();
+
+        assert_constant_reference_unresolved!(context, "file:///file2.rb:2:14-2:19");
+    }
+
+    #[test]
+    fn middle_of_hierarchy_shadow() {
```
thomasmarshall, line 4881: `CONST` in `Foo < Bar < Baz` resolves to `Baz::CONST`. After adding `CONST` to `Bar`, it resolves to `Bar::CONST`.

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///file3.rb:1:14-1:19");
+    }
+
+    #[test]
+    fn namespace_change_invalidates_references() {
```
thomasmarshall, line 4921: `Foo::CONST` resolves. After moving `Foo` inside `Bar`, the reference becomes unresolved (no longer top-level `Foo`).

rust/rubydex/src/resolution.rs
```diff
@@ -4098,4 +4098,891 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_constant_reference_unresolved!(context, "file:///file2.rb:0:5-0:10");
+    }
+
+    #[test]
+    fn new_file_with_shadow() {
```
thomasmarshall, line 4951: `CONST` in `Foo` resolves to `Bar::CONST` via include. After adding `Foo::CONST` in a new file, it resolves to `Foo::CONST` (class constant shadows mixin).

rust/rubydex/src/resolution.rs
thomasmarshall: Maybe this one doesn't belong here yet and could be added later?

rust/rubydex/src/resolution.rs
thomasmarshall: This one seems a bit redundant?

PR #523, opened 2026-01-26
Keep track of original nesting on top level references

vinistock:
We originally took a shortcut for representing top level constant references. We were simply ignoring (and effectively erasing) its nesting information. This works for almost everything, but it unfortunately falls short for the case of completion.

For example:

```ruby
module Foo
  CONST = 1
  
  class ::Bar
    # Completion triggered here without a character (CTRL + space)
  end
end

class Bar
  # Completion triggered here without a character (CTRL + space)
end
```

In this example, both cases of `Bar` refers to the same class, but the lexical scope in which they are inserted in is different. The first `::Bar` has access to `Foo::CONST` through lexical scope and the second `Bar` does not.

We can't support this correctly without remembering the original nesting for constant references while at the same time recording that we found a top level reference.

### Implementation

I tested two approaches:

1. Using a new enum for `ParentScope`. This gives us type safety as the compiler forces us to handle each case.
2. I originally thought the new enum would consume more memory, so I also experimented with using a special `NameId` for a fake name `<!top-level!>`. To my surprise, based on `mem::size_of` and our benchmarks on Core, there's no memory difference with the extra enum

Because of this, I went with the new enum, so that we can have extra type safety.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -13,7 +13,7 @@ use crate::model::{
     graph::{CLASS_ID, Graph, MODULE_ID, OBJECT_ID},
     identity_maps::{IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
-    name::{Name, NameRef},
+    name::{Name, NameRef, ParentScope},
 };
 
 pub enum Unit {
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1779,7 +1797,7 @@ mod tests {
 
         assert_eq!(
             vec![
-                "Top", "Foo", "Qux", "AfterTop", "Bar", "Baz", "Zip", "Zap", "Zop", "Boop"
+                "Top", "Foo", "Qux", "Bar", "AfterTop", "Baz", "Zip", "Zap", "Zop", "Boop"
```
Morriar, line 1800: Why did this flip?

vinistock: I'm honestly not sure, but both `Bar` and `AfterTop` have the exact same score, so I think it's okay.

PR #524, opened 2026-01-27
Fix indexing of attr comments

Morriar:
Consider this example:

```rb
# Comment
attr_reader :foo
```

Before this PR we were looking for the comment from the `:foo` symbol instead of the `attr_reader` call, meaning we didn't find a comment right before.


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -3709,153 +3711,6 @@ mod tests {
         });
     }
 
-    #[test]
-    fn index_comments_attached_to_definitions() {
```
Morriar, line 3713: Just moving all the comments related tests together at the end

PR #525, opened 2026-01-27
Index comments for definitions with an inline visibility modifier

Morriar:
Consider this example:

```rb
# Comment
private def foo; end

# Comment
protected attr_reader :bar
```

Before this PR, as we only looked right before the `def foo` definition or the `attr_reader :bar` call, no comments were associated with the definitions.

Instead, we keep track of the location the visibility modifier was found at and if it was specified inline so we can lookup the documentation from the right offset.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1410,12 +1410,21 @@ impl Visit<'_> for RubyIndexer<'_> {
                 let str_id = self.local_graph.intern_string(format!("{name}()"));
                 let parent_nesting_id = self.parent_nesting_id();
                 let offset = Offset::from_prism_location(&location);
-                let (comments, flags) = self.find_comments_for(call_offset.start());
+
+                let current_visibility = self.current_visibility();
+                let (visibility, offset_for_comments) = if current_visibility.is_inline() {
+                    // If the visibility is inline, we use its offset for the comments
+                    (*current_visibility.visibility(), current_visibility.offset().clone())
+                } else {
+                    (*current_visibility.visibility(), call_offset.clone())
+                };
+
+                let (comments, flags) = self.find_comments_for(offset_for_comments.start());
 
                 // module_function makes attr_* methods private (without creating singleton methods)
-                let visibility = match self.current_visibility().visibility() {
+                let visibility = match visibility {
                     Visibility::ModuleFunction => Visibility::Private,
-                    v => *v,
+                    v => v,
                 };
 
                 let definition = match kind {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -55,7 +55,7 @@ pub struct RubyIndexer<'a> {
     source: &'a str,
     comments: Vec<CommentGroup>,
     nesting_stack: Vec<Nesting>,
-    visibility_stack: Vec<Visibility>,
+    visibility_stack: Vec<(Visibility, bool, Offset)>,
```
alexcrocha: Did you consider a struct for the visibility context? I find the tuple hard to read in the middle of the code.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1375,10 +1375,18 @@ impl Visit<'_> for RubyIndexer<'_> {
...
                 let parent_nesting_id = self.parent_nesting_id();
                 let offset = Offset::from_prism_location(&location);
-                let (comments, flags) = self.find_comments_for(call_offset.start());
+
+                let (visibility, offset_for_comments) = if self.current_visibility().1 {
+                    // If the visibility is inline, we its offset for the comments
+                    (self.current_visibility().0, self.current_visibility().2.clone())
+                } else {
+                    (self.current_visibility().0, call_offset.clone())
+                };
```
alexcrocha: Related to the comment above, the tuple makes it a little harder to read.


What do you think of this:

```
self.current_visibility().visibility
self.current_visibility().is_inline
self.current_visibility().offset
```
vs
```
self.current_visibility().0
self.current_visibility().1
self.current_visibility().2
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1365,16 +1369,26 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+            let call_offset = Offset::from_prism_location(&call.location());
+
             Self::each_string_or_symbol_arg(call, |name, location| {
                 let str_id = self.local_graph.intern_string(format!("{name}()"));
                 let parent_nesting_id = self.parent_nesting_id();
                 let offset = Offset::from_prism_location(&location);
-                let (comments, flags) = self.find_comments_for(offset.start());
+
+                let (visibility, offset_for_comments) = if self.current_visibility().1 {
+                    // If the visibility is inline, we its offset for the comments
```
alexcrocha:
```suggestion
                    // If the visibility is inline, we use its offset for the comments
```

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: Yeah it seemed overkill here, but I agree it's better for readability. Updated 👍

Issue #411, opened 2026-01-06
Design a way to search require paths


vinistock:
In the LSP, we provide require path completion and go to definition with a prefix tree of `require_path => uri`. Let's explore how we can achieve the same behaviour, preferably without paying the price of an extra structure if possible.

Maybe with Rust's parallelism, constructing require paths from all documents using the `$LOAD_PATH` is fast enough that we can just do it every time it's needed.

PR #528, opened 2026-01-27
Add require path resolution for go-to-definition

alexcrocha:
This PR is a step towards https://github.com/Shopify/rubydex/issues/411 and adds the ability to resolve require paths (`"foo/bar"`) to their corresponding file URIs by matching against load paths.

This enables go-to-definition on require statements. When a user clicks on `require "foo/bar"`, we can now find which file that maps to.

  - `Document::require_path(load_paths)` - computes the require path for a document
  - `resolve_require_path(graph, path, load_paths)` - finds the URI ID for a given require path

### DIFF
--- a/rust/rubydex/src/model/document.rs
+++ b/rust/rubydex/src/model/document.rs
@@ -1,4 +1,7 @@
+use std::path::PathBuf;
+
 use line_index::LineIndex;
+use url::Url;
 
 use crate::diagnostic::Diagnostic;
 use crate::model::identity_maps::IdentityHashMap;
@@ -108,11 +111,58 @@ impl Document {
             .chain(self.definition_diagnostics.values().flatten())
             .chain(self.reference_diagnostics.values().flatten())
     }
+
+    /// Computes the require path for this document given load paths.
+    ///
+    /// Returns `None` if:
+    /// - URI is not `file://` scheme
+    /// - URI doesn't end with `.rb`
+    /// - File path doesn't match any load path
+    /// - `Url::to_file_path()` fails
+    #[must_use]
+    pub fn require_path(&self, load_paths: &[PathBuf]) -> Option<String> {
+        let url = Url::parse(&self.uri).ok()?;
+        if url.scheme() != "file" {
+            return None;
+        }
+        let file_path = url.to_file_path().ok()?;
+
+        if file_path.extension().is_none_or(|ext| ext != "rb") {
+            return None;
+        }
+
+        for load_path in load_paths {
+            if let Ok(relative) = file_path.strip_prefix(load_path) {
+                let file_path = relative
+                    .components()
+                    .filter_map(|c| c.as_os_str().to_str())
+                    .collect::<Vec<_>>()
+                    .join("/");
+
+                let require_path = file_path.trim_end_matches(".rb").to_string();
+                return Some(require_path);
+            }
+        }
+        None
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::str::FromStr;
+
     use super::*;
+
+    fn test_root() -> PathBuf {
+        let root = if cfg!(windows) {
+            PathBuf::from_str("C:\\")
+        } else {
+            PathBuf::from_str("/")
+        }
+        .unwrap();
+
+        root.join("test")
+    }
 
     #[test]
     #[should_panic(expected = "Cannot add the same exact definition to a document twice. Duplicate definition IDs")]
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -679,6 +680,22 @@ impl Graph {
...
+    /// Used for go-to-definition.
+    #[must_use]
+    pub fn resolve_require_path(&self, require_path: &str, load_paths: &[PathBuf]) -> Option<UriId> {
+        let normalized = require_path.trim_end_matches(".rb");
+
+        for (uri_id, document) in &self.documents {
+            if let Some(computed) = document.require_path(load_paths)
+                && computed == normalized
+            {
+                return Some(*uri_id);
```
alexcrocha: Hmm, maybe we should be returning the `uri` here instead of `uri_id`?

rust/rubydex/src/model/document.rs
```diff
@@ -108,12 +111,59 @@ impl Document {
...
+    pub fn require_path(&self, load_paths: &[PathBuf]) -> Option<String> {
+        let url = Url::parse(&self.uri).ok()?;
+        if url.scheme() != "file" {
+            return None;
+        }
+        let file_path = url.to_file_path().ok()?;
+
+        if file_path.extension().is_none_or(|ext| ext != "rb") {
+            return None;
+        }
```
Morriar, line 132: I think it's generally nicer to group related statements together. Helps for readability: 

```suggestion

        let file_path = url.to_file_path().ok()?;
        if file_path.extension().is_none_or(|ext| ext != "rb") {
            return None;
        }
```

rust/rubydex/src/model/document.rs
```diff
@@ -108,12 +111,59 @@ impl Document {
...
 
+    fn test_root() -> PathBuf {
+        let root = if cfg!(windows) {
+            PathBuf::from_str("C:\\")
+        } else {
+            PathBuf::from_str("/")
+        }
+        .unwrap();
+
+        root.join("test")
```
Morriar, line 164: You could avoid repeating yourself and simplify a bit:


```suggestion
        let root = if cfg!(windows) {
            "C:\\"
        } else {
            "/"
        }
        
        PathBuf::from_str(root).unwrap().join("test")
```

Morriar: btw, why do we need to join `"test"`?

rust/rubydex/src/model/document.rs
```diff
@@ -138,4 +188,61 @@ mod tests {
...
+
+    #[test]
+    fn require_path() {
+        let root = test_root();
+        let path = root.join("lib").join("foo").join("bar.rb");
+        let uri = Url::from_file_path(path).unwrap().to_string();
+        let load_paths = [root.join("lib")];
+
+        let document = Document::new(uri, "");
+        assert_eq!(Some("foo/bar".to_string()), document.require_path(&load_paths));
```
Morriar, line 200: Keep the expected part as simple as possible:

```suggestion
        assert_eq!("foo/bar", document.require_path(&load_paths).unwrap().to_string());
```

rust/rubydex/src/model/document.rs
```diff
@@ -138,4 +188,61 @@ mod tests {
...
+    }
+
+    #[test]
+    fn require_path_returns_none() {
+        let root = test_root();
+        let load_paths = [root.join("lib")];
+
+        // non-file URI
+        let document = Document::new("untitled:Untitled-1".to_string(), "");
+        assert_eq!(None, document.require_path(&load_paths));
```
Morriar, line 223:
```suggestion
        assert!(document.require_path(&load_paths).is_none());
```

rust/rubydex/src/query.rs
```diff
@@ -59,8 +61,55 @@ fn match_score(query: &str, target: &str) -> usize {
...
 
+/// Resolves a require path to its URI ID. Used for go-to-definition.
+///
+/// # Panics
+///
+/// Panics if one of the search threads panics
+#[must_use]
+pub fn resolve_require_path(graph: &Graph, require_path: &str, load_paths: &[PathBuf]) -> Option<UriId> {
+    let found = AtomicBool::new(false);
+    let found_ref = &found;
```
Morriar, line 72: Do we need the intermediary var?

```suggestion
    let found_ref = &AtomicBool::new(false);
```

Or maybe it needs a `as_ref()`?

rust/rubydex/src/query.rs
```diff
@@ -59,8 +61,55 @@ fn match_score(query: &str, target: &str) -> usize {
...
+    let normalized = require_path.trim_end_matches(".rb");
+    let num_threads = thread::available_parallelism().map(std::num::NonZero::get).unwrap_or(4);
```
Morriar, line 74: Would [rayon](https://docs.rs/rayon/latest/rayon/iter/index.html)'s `par_iter` be easier to use?

rust/rubydex/src/query.rs
```diff
@@ -100,4 +149,40 @@ mod tests {
         context.resolve();
         assert_results_eq!(context, "Fo", ["Foo"]);
     }
+
+    #[test]
+    fn test_resolve_require_path() {
+        let root = if cfg!(windows) {
```
Morriar, line 155: Same comment than earlier

rust/rubydex/src/model/document.rs
alexcrocha: I agree. I'll fine tune my eye to identify these 😄

rust/rubydex/src/model/document.rs
alexcrocha: That's so much better.

We don't need it, I'll remove it in a follow up PR.

rust/rubydex/src/model/document.rs
alexcrocha: 👀📝

rust/rubydex/src/query.rs
alexcrocha: No, we don't need it. We can dereference it directly.

rust/rubydex/src/query.rs
alexcrocha: It would be, yes. That's actually how I started but decided against adding a dependency. If we end up using `rayon` we can simplify this.

PR #505, opened 2026-01-22
Automatically promote constants to namespaces

vinistock:
This PR does two main things:

- Removes all of the todos with the skips we had in resolution for constants
- Handles some meta-programming scenarios without crashing

Although we can't provide first class features when there's meta-programming, we absolutely cannot crash. Consider these examples (or the ones added as tests):

```ruby
Foo = dynamic_module

class Bar
  include Foo
end

Baz = dynamic_class

class Qux < Baz
end

module Zip; end

Boop = dynamic_class do
  include Zip
end
```

This PR takes two strategies to prevent crashing completely in these scenarios:

- The most straightforward one is just skipping things that aren't namespaces during ancestor linearization. If we find an include for something that's not a namespace, we need to skip it
- The second part is auto-promoting constants to namespaces if they get re-opened. This allows us to handle this scenario correctly (which we actually have in Core):

```ruby
# At this point, we have no idea this is a class. We create a Constant declaration for it
Const = dynamic_class

# Then it gets re-opened as a class with proper includes
class Const
  include Foo
end
```

By auto-promoting, we correctly handle ancestors and constant resolution for `Const`. Of course, there's a concession being made to the user here: we trust that they didn't make a mistake by making an incompatible re-opening (i.e.: it was a module, then gets reopened as a class).

PR #529, opened 2026-01-27
Create TODO declarations for unresolved parent scopes

Morriar:
Consider this:

```rb
class Foo::Bar
  def bar; end
end
```

Without this PR, if `Foo` can't be resolved, `Bar` never gets created and `bar()` ends up under `Object`.

This PR introduces `TodoDeclaration`, a placeholder namespace used when a parent scope can't be resolved. Rather than running the resolution queue twice (once normally, then again in a "create Todos" mode), we create Todos eagerly in a single pass by leveraging the existing constant → namespace promotion mechanism from #505.

The logic for the example above now looks like this:

```
1. Try to resolve Bar -> parent scope Foo is Unresolved(None) (genuinely unknown)
2. Create Foo as a Todo declaration under Object
3. Resolve Bar under Foo, enqueue its members
4. Create bar() under Foo::Bar
```

If a real `class Foo` or `module Foo` definition appears later, `Foo` is promoted from `Todo` to `Class`/`Module` in place — same mechanism used to promote constants to namespaces.

Unresolved outcomes with a pending linearization (`Unresolved(Some(id))`) are propagated as-is rather than triggering Todo creation, so we don't create premature Todos for names that might resolve once ancestor chains are fully computed.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1071,10 +1071,20 @@ impl<'a> Resolver<'a> {
                 Outcome::Unresolved(None) => {
                     let parent_name = self.graph.names().get(&parent_scope).unwrap();
                     let parent_str_id = *parent_name.str();
+                    let parent_has_explicit_prefix = parent_name.parent_scope().as_ref().is_some();
+                    // NLL: borrow of parent_name ends here
 
-                    let parent_owner_id = match self.name_owner_id(parent_scope) {
-                        Outcome::Resolved(id, _) => id,
-                        _ => *OBJECT_ID,
+                    // For bare names (no explicit `::` prefix), always use OBJECT_ID as the owner.
+                    // Using nesting here would create "Nesting::Bar" instead of "Bar" for a bare `Bar`
+                    // reference, which is incorrect: if `Bar` can't be found anywhere, the placeholder
+                    // should live at the top level so it can be promoted when `module Bar` appears later.
+                    let parent_owner_id = if parent_has_explicit_prefix {
+                        match self.name_owner_id(parent_scope) {
+                            Outcome::Resolved(id, _) => id,
+                            _ => *OBJECT_ID,
+                        }
+                    } else {
+                        *OBJECT_ID
                     };
 
                     let fully_qualified_name = if parent_owner_id == *OBJECT_ID {
@@ -1718,8 +1728,8 @@ mod tests {
         assert_constant_reference_to, assert_declaration_definitions_count_eq, assert_declaration_does_not_exist,
         assert_declaration_exists, assert_declaration_kind_eq, assert_declaration_references_count_eq,
         assert_descendants, assert_diagnostics_eq, assert_instance_variables_eq, assert_members_eq,
-        assert_no_constant_alias_target, assert_no_diagnostics, assert_no_members,
-        assert_owner_eq, assert_singleton_class_eq,
+        assert_no_constant_alias_target, assert_no_diagnostics, assert_no_members, assert_owner_eq,
+        assert_singleton_class_eq,
     };
 
     #[test]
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -232,10 +232,8 @@ impl Graph {
         }?;
 
         self.declarations
-            .get(nesting_declaration_id)
-            .unwrap()
-            .as_namespace()
-            .unwrap()
+            .get(nesting_declaration_id)?
+            .as_namespace()?
```
vinistock, line 307: Why does this need to change?

rust/rubydex/src/resolution.rs
```diff
@@ -112,40 +123,49 @@ impl<'a> Resolver<'a> {
...
 
-        let (mut unit_queue, other_ids) = self.sorted_units();
+        let other_ids = self.prepare_units();
+
+        self.process_queue();
 
+        // Once we've processed the queue and create all we could, what's remaining is depending on other unresolved units.
+        // We'll create TODO declarations for these and process the queue again to resolve what we can.
+        self.create_todo_declarations = true;
+        self.process_queue();
```
vinistock: Instead of looping through the queue twice, can we not incorporate `Todo` in the regular processing?

If we eagerly create unresolved definitions as `Todo`, while still enqueuing a retry, we could do everything in a single loop.

Maybe in combination with #505

rust/rubydex/src/resolution.rs
```diff
@@ -4967,4 +5015,32 @@ mod tests {
...
+        context.resolve();
+
+        assert_no_diagnostics!(&context);
+
+        assert_todo_declarations_eq!(context, ["Foo"]);
+
+        assert_members_eq!(context, "Object", vec!["Foo"]);
+        assert_members_eq!(context, "Foo", vec!["Bar", "Baz"]);
+        assert_members_eq!(context, "Foo::Bar", vec!["bar()"]);
+        assert_members_eq!(context, "Foo::Baz", vec!["baz()"]);
```
Morriar: Let's also add a test for the Todo → real namespace promotion case, e.g.:

```rb
class Foo::Bar
  def bar; end
end

class Foo
  def foo; end
end
```

The `should_promote` logic in `graph.rs` specifically handles `Namespace::Todo(_) => true` for this. Want to make sure that when `Foo` is first created as a Todo (from `class Foo::Bar`) and then a real `class Foo` definition comes along, it properly promotes and `Foo` is no longer a Todo.

rust/rubydex/src/resolution.rs
```diff
@@ -1027,12 +1027,51 @@ impl<'a> Resolver<'a> {
...
+                    if !self.graph.declarations().contains_key(&declaration_id) {
+                        self.graph.declarations_mut().insert(
+                            declaration_id,
+                            Declaration::Namespace(Namespace::Todo(Box::new(TodoDeclaration::new(
+                                fully_qualified_name,
+                                parent_owner_id,
+                            )))),
+                        );
+                        self.graph.add_member(&parent_owner_id, declaration_id, parent_str_id);
+                    }
```
vinistock: Instead of doing check then insert, could we use the `entry` APIs to do this in one go?

Like

```rust
self.graph.declarations_mut().entry(...).or_insert_with(...)
```

Morriar: Good call, done! The tricky part was that `add_member` still needs to be called only when newly inserted, but `if let Entry::Vacant(e)` handles that — after `e.insert()` consumes the entry, NLL releases the borrow and the `add_member` call works fine.

rust/rubydex/src/resolution.rs
```diff
@@ -5016,249 +5010,114 @@ mod tests {
...
 
-        context.resolve();
-        assert_declaration_kind_eq!(context, "Foo", "Constant");
-        assert_declaration_does_not_exist!(context, "Foo::<Foo>");
-        assert_declaration_does_not_exist!(context, "Foo::<Foo>#bar()");
+        assert_no_diagnostics!(&context);
+        assert_declaration_kind_eq!(context, "Bar", "Module");
+        assert_members_eq!(context, "Bar", vec!["Baz"]);
+        assert_declaration_exists!(context, "Bar::Baz");
+        assert_declaration_does_not_exist!(context, "Foo::Bar");
```
vinistock, line 5426: Can we also assert the fully qualified name of `qux` and that it is a member of `Bar::Baz`?

rust/rubydex/src/resolution.rs
```diff
@@ -1057,12 +1027,50 @@ impl<'a> Resolver<'a> {
...
+                        format!(
+                            "{}::{}",
+                            self.graph.declarations().get(&parent_owner_id).unwrap().name(),
+                            self.graph.strings().get(&parent_str_id).unwrap().as_str()
+                        )
+                    };
+
+                    let declaration_id = DeclarationId::from(&fully_qualified_name);
+
+                    if let std::collections::hash_map::Entry::Vacant(e) =
```
vinistock, line 1102: Nit: let's import `Entry` instead of using the long reference.

rust/rubydex/src/resolution.rs
Morriar: Done, added `assert_members_eq!(context, "Bar::Baz", vec!["qux()"])`.

rust/rubydex/src/resolution.rs
Morriar: Done, imported `hash_map::Entry` in the `use std::collections` block.

PR #530, opened 2026-01-28
Add receiver to method references

vinistock:
To avoid exposing a mutable version of the graph to the Ruby API, we need to ensure that resolution will create the necessary singleton classes and linearize their ancestors.

As a way to achieve this, we can add a receiver to method references that we can resolve to correctly create the expected declarations.

In the future, I would expect method references to be collected during an inference loop since we have to check the type of the receiver. For the time being, we only support constant receivers that we can detect between indexing and resolving.

### Trade offs

This PR stack will eliminate the need to mutate the graph after resolution, but memory usage increases in Core by 1.5 GB.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -496,10 +496,10 @@ impl<'a> RubyIndexer<'a> {
         Some(name_id)
     }
 
-    fn index_method_reference(&mut self, name: String, location: &ruby_prism::Location) {
+    fn index_method_reference(&mut self, name: String, location: &ruby_prism::Location, receiver: Option<NameId>) {
         let offset = Offset::from_prism_location(location);
         let str_id = self.local_graph.intern_string(name);
-        let reference = MethodRef::new(str_id, self.uri_id, offset);
+        let reference = MethodRef::new(str_id, self.uri_id, offset, receiver);
         self.local_graph.add_method_reference(reference);
     }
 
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -496,10 +496,10 @@ impl<'a> RubyIndexer<'a> {
         Some(name_id)
     }
 
-    fn index_method_reference(&mut self, name: String, location: &ruby_prism::Location) {
-        let offset = Offset::from_prism_location(location);
+    fn index_method_reference(&mut self, name: String, location: &ruby_prism::Location, receiver: Option<NameId>) {
+        let offset: Offset = Offset::from_prism_location(location);
```
Morriar: Why do we need this type?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4121,6 +4230,213 @@ mod tests {
...
+        let context = index_source({
+            "
+            Foo.bar
+            "
+        });
+
+        assert_no_diagnostics!(&context);
+
+        let method_ref = context.graph().method_references().values().next().unwrap();
+        assert_eq!(StringId::from("bar"), *method_ref.str());
```
Morriar, line 4254: Let's add a macro for these tests:

```rs
assert_method_reference_at!(&context, "1:4-1:7", |ref| {
  assert_string_eq(ref.str(), "bar")
  assert_name_eq(ref.receiver(), "Foo")
})
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1845,13 +1954,13 @@ impl Visit<'_> for RubyIndexer<'_> {
 
     fn visit_and_node(&mut self, node: &ruby_prism::AndNode) {
         self.visit(&node.left());
-        self.index_method_reference("&&".to_string(), &node.location());
+        self.index_method_reference("&&".to_string(), &node.location(), None);
```
Morriar: Doesn't it depend on the left side?

```rb
Foo && Bar
```

Can't we assume it's `Foo.&&` being called?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1002,6 +1002,92 @@ impl<'a> RubyIndexer<'a> {
...
     }
+
+    fn method_receiver(&mut self, receiver: Option<&ruby_prism::Node>) -> Option<NameId> {
+        let mut is_singleton_name = false;
+
+        let name_id = match receiver {
+            Some(ruby_prism::Node::SelfNode { .. }) | None => {
+                // Implicit or explicit self receiver
+
+                match self.nesting_stack.last() {
```
Morriar, line 1013: Can't you use `current_lexical_scope_name_id` for this?

rust/rubydex/src/model/references.rs
```diff
@@ -62,12 +62,19 @@ pub struct MethodRef {
     uri_id: UriId,
     /// The offsets inside of the document where we found the reference
     offset: Offset,
+    /// The receiver of the method call if it's a constant
+    receiver: Option<NameId>,
```
Morriar, line 68: Can't we use a constant reference here so we don't need to change the resolution?

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: We don't

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: I'll do that in a separate PR. It will require new helpers in local graph test.

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: Yes, we can!

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: No, because that skips methods and we need to take them into consideration to determine the type of `self` inside.

rust/rubydex/src/model/references.rs
vinistock: We can put a constant reference ID there, but that's not enough for resolution. Two things:

- Note that the constant references _are already created_, we're just not storing the ID as the receiver and using a name instead. So resolution is already picking them up
- It's not enough because having the references created won't trigger ancestor linearization for them on its own. We need resolution changes

PR #531, opened 2026-01-28
Resolve method reference receivers

vinistock:
This PR starts resolving method reference receivers, so that the expected singleton classes are created and get their ancestors linearized.

The idea is this:

```ruby
Foo.bar # this is `Foo::<Foo>#bar` and therefore we need <Foo> to exist
Foo.singleton_class.bar # this is `Foo::<Foo>::<<Foo>>#bar` and so we need both <Foo> and <<Foo>> to exist
```

For this reason, we need to loop to create all singleton levels involved in the method reference.

### Trade offs

The resolution time added is actually not terrible, but memory usage is a different story.

This PR stack will eliminate the need to mutate the graph after resolution, but memory usage increases in Core by 1.5 GB.

### DIFF
--- a/rust/rubydex/src/model/name.rs
+++ b/rust/rubydex/src/model/name.rs
@@ -13,6 +13,12 @@ pub enum ParentScope {
     TopLevel,
     /// There's a parent scope in this reference (e.g.: `Foo::Bar`)
     Some(NameId),
+    /// Parent scope representing the class attached to a singleton class context
+    ///
+    ///  `Foo::<Foo>::<<Foo>>`
+    ///  ^ Attached for <Foo>
+    ///        ^ Attached for <<Foo>>
+    Attached(NameId),
 }
 assert_mem_size!(ParentScope, 8);
 
@@ -22,7 +28,7 @@ impl ParentScope {
         F: FnOnce(&NameId) -> T,
     {
         match self {
-            ParentScope::Some(id) => f(id),
+            ParentScope::Some(id) | ParentScope::Attached(id) => f(id),
             _ => default,
         }
     }
@@ -30,7 +36,7 @@ impl ParentScope {
     #[must_use]
     pub fn as_ref(&self) -> Option<&NameId> {
         match self {
-            ParentScope::Some(id) => Some(id),
+            ParentScope::Some(id) | ParentScope::Attached(id) => Some(id),
             _ => None,
         }
     }
@@ -51,7 +57,7 @@ impl ParentScope {
     #[must_use]
     pub fn expect(&self, message: &str) -> NameId {
         match self {
-            ParentScope::Some(id) => *id,
+            ParentScope::Some(id) | ParentScope::Attached(id) => *id,
             _ => panic!("{}", message),
         }
     }
@@ -63,6 +69,7 @@ impl Display for ParentScope {
             ParentScope::None => write!(f, "None"),
             ParentScope::TopLevel => write!(f, "TopLevel"),
             ParentScope::Some(id) => write!(f, "Some({id})"),
+            ParentScope::Attached(id) => write!(f, "Attached({id})"),
         }
     }
 }
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -20,7 +20,7 @@ pub enum Unit {
     /// A definition that defines a constant and might require resolution
     Definition(DefinitionId),
     /// A constant reference that needs to be resolved
-    Reference(ReferenceId),
+    ConstantRef(ReferenceId),
     /// A list of ancestors that have been partially linearized and need to be retried
     Ancestors(DeclarationId),
 }
@@ -140,7 +140,7 @@ impl<'a> Resolver<'a> {
                     Unit::Definition(id) => {
                         self.handle_definition_unit(unit_id, id);
                     }
-                    Unit::Reference(id) => {
+                    Unit::ConstantRef(id) => {
                         self.handle_reference_unit(unit_id, id);
                     }
                     Unit::Ancestors(id) => {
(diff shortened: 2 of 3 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -132,12 +134,15 @@ impl<'a> Resolver<'a> {
...
-                    Unit::Reference(id) => {
+                    Unit::ConstantRef(id) => {
                         self.handle_reference_unit(&mut unit_queue, &mut made_progress, unit_id, id);
                     }
                     Unit::Ancestors(id) => {
                         self.handle_ancestor_unit(&mut unit_queue, &mut made_progress, id);
                     }
+                    Unit::MethodRef(id) => {
+                        self.handle_method_reference_unit(&mut unit_queue, &mut made_progress, id);
+                    }
```
Morriar: See my comment https://github.com/Shopify/rubydex/pull/530/changes#r2742491771, I was wondering if the receiver could be automatically resolved through a constant reference with `Unit::ConstantRef`

rust/rubydex/src/resolution.rs
```diff
@@ -4189,4 +4279,163 @@ mod tests {
...
+        });
+        context.resolve();
+
+        let singleton_foo = context
+            .graph()
+            .declarations()
+            .get(&DeclarationId::from("Foo::<Foo>::<<Foo>>::<<<Foo>>>"))
+            .unwrap();
+
+        assert_eq!(1, singleton_foo.references().len());
```
Morriar:
```suggestion
        assert_references_to!("Foo::<Foo>::<<Foo>>::<<<Foo>>>", 1)
```

rust/rubydex/src/resolution.rs
vinistock: You were actually right! I found a way to not have this new method and incorporate everything into constant resolution.

rust/rubydex/src/resolution.rs
```diff
@@ -4089,4 +4097,135 @@ mod tests {
...
+
+    #[test]
+    fn method_calls_automatically_trigger_singleton_creation() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            "
+            class Foo
+            end
+
+            Foo.
```
st0012, line 4109: Do we need to simulate method completion situation in rubydex tests? This is incorrect syntax and IMO is confusing to the reader. `Foo.bar` should probably work too?

st0012: Also, should we have `assert_no_diagnostics` in the new tests?

rust/rubydex/src/model/name.rs
```diff
@@ -13,6 +13,9 @@ pub enum ParentScope {
     TopLevel,
     /// There's a parent scope in this reference (e.g.: `Foo::Bar`)
     Some(NameId),
+    /// Parent scope representing an attached singleton class (e.g., `<Foo>` in `Foo::<Foo>`).  Unlike `Some`, which
+    /// triggers ancestor search during resolution, `Attached` triggers singleton class creation.
+    Attached(NameId),
```
Morriar, line 21: This name is confusing because it's the opposite of Sorbet `attached_class`: https://sorbet.org/docs/attached-class

Why not `Singleton`?

rust/rubydex/src/resolution.rs
vinistock: The thing is, if we don't want to allow lazy creation of singletons, we have to ensure that `Foo.` is properly creating the singleton entity or else completion won't work.

I think we have to keep this test or else it's easy to forget this particular use case.

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -465,6 +465,21 @@ macro_rules! assert_no_diagnostics {
     }};
 }
 
+/// Asserts the number of references registered to a specific declaration
+#[cfg(test)]
+#[macro_export]
+macro_rules! assert_reference_count {
```
Morriar: Looks like you're duplicating https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/test_utils/graph_test.rs#L249. I'm open to a renaming if you want.

rust/rubydex/src/model/name.rs
vinistock: Maybe the comment is confusing because that's not what it means. This is what it means (which matches Sorbet):

```
Foo::<Foo>
^ ParentScope::Attached
      ^ Just a name

Foo::<Foo>::<<Foo>>
^ Attached for <Foo>
      ^ Attached for <<Foo>>
```

rust/rubydex/src/resolution.rs
```diff
@@ -4089,4 +4097,135 @@ mod tests {
...
+        context.resolve();
+
+        assert!(
+            context
+                .graph()
+                .declarations()
+                .get(&DeclarationId::from("Foo::<Foo>"))
+                .is_none()
+        );
+    }
```
Morriar:
```suggestion
        assert_declaration_does_not_exist!("Foo::<Foo>");
    }
```

rust/rubydex/src/model/name.rs
Morriar: Yeah the comment got me confused.

Maybe something like:

```rs
/// Parent scope representing the class attached to a singleton class context
///
///  Foo::<Foo>::<<Foo>>
///  ^ Attached for <Foo>
///        ^ Attached for <<Foo>>
```

rust/rubydex/src/resolution.rs
st0012: So the behaviour between seeing `Foo.` and `Foo.b` is different? Can you point me to the code that makes the distinction?

vinistock: It's not different, the point is we want to ensure that it's not. In `Foo.b`, you have no syntax errors and a proper method call. This scenario is naturally expected to create the singleton class because you're invoking a singleton method.

`Foo.` on the other hand is a bit of an edge case. You have an incomplete method call for which the message is empty. The tests is just to ensure that even in this case we still produce the singleton class, otherwise we can't provide completion.

rust/rubydex/src/test_utils/graph_test.rs
vinistock: Nah, it's all good. I just hadn't found it.

rust/rubydex/src/resolution.rs
vinistock: Added the diagnostic assertions.

st0012: Can we indicate this intention in the test name and a comment on this line? Something like `incomplete_method_calls_automatically_trigger_singleton_creation`

vinistock: Renamed it 👍

PR #537, opened 2026-01-29
Add require path search for completion

alexcrocha:
This PR is a step towards #411 and adds finding all require paths. This enables editors to suggest completions when typing require statements.

For example, when a user types `require "foo/`, the editor can filter the returned paths to show suggestions starting with "foo/".

  - `search_require_paths(graph, load_paths)` - finds and returns all require paths
  - When the same require path exists in multiple load paths, the earliest load path wins (matching Ruby's `require` behaviour)

### DIFF
--- a/rust/rubydex/src/model/document.rs
+++ b/rust/rubydex/src/model/document.rs
@@ -119,8 +119,12 @@ impl Document {
     /// - URI doesn't end with `.rb`
     /// - File path doesn't match any load path
     /// - `Url::to_file_path()` fails
+    ///
+    /// # Panics
+    ///
+    /// Panics if load path entries exceed u16.
     #[must_use]
-    pub fn require_path(&self, load_paths: &[PathBuf]) -> Option<String> {
+    pub fn require_path(&self, load_paths: &[PathBuf]) -> Option<(String, u16)> {
         let url = Url::parse(&self.uri).ok()?;
         if url.scheme() != "file" {
             return None;
@@ -131,7 +135,7 @@ impl Document {
             return None;
         }
 
-        for load_path in load_paths {
+        for (load_path_index, load_path) in load_paths.iter().enumerate() {
             if let Ok(relative) = file_path.strip_prefix(load_path) {
                 let file_path = relative
                     .components()
@@ -140,7 +144,10 @@ impl Document {
                     .join("/");
 
                 let require_path = file_path.trim_end_matches(".rb").to_string();
-                return Some(require_path);
+                return Some((
+                    require_path,
+                    load_path_index.try_into().expect("Load path entries exceed u16"),
+                ));
             }
         }
         None
@@ -197,7 +204,7 @@ mod tests {
         let load_paths = [root.join("lib")];
 
         let document = Document::new(uri, "");
-        assert_eq!(Some("foo/bar".to_string()), document.require_path(&load_paths));
+        assert_eq!(Some(("foo/bar".to_string(), 0)), document.require_path(&load_paths));
     }
 
     #[test]
@@ -210,7 +217,7 @@ mod tests {
         let load_paths = [load_path_a, load_path_ab];
 
         let document = Document::new(uri, "");
-        assert_eq!(Some("b/foo".to_string()), document.require_path(&load_paths));
+        assert_eq!(Some(("b/foo".to_string(), 0)), document.require_path(&load_paths));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/query.rs
```diff
@@ -105,6 +106,53 @@ pub fn resolve_require_path(graph: &Graph, require_path: &str, load_paths: &[Pat
     })
 }
 
+/// Returns all require paths matching the given prefix. Used for completion.
+///
+/// # Panics
+///
+/// Panics if one of the search threads panics
+#[must_use]
+pub fn search_require_paths(graph: &Graph, prefix: &str, load_paths: &[PathBuf]) -> Vec<(String, UriId)> {
```
vinistock: Here, we can return all require paths and offload the filtering to the editor side. Also, we don't need to return the UriId.


```suggestion
pub fn require_paths(graph: &Graph, load_paths: &[PathBuf]) -> Vec<String> {
```

rust/rubydex/src/query.rs
```diff
@@ -105,6 +106,53 @@ pub fn resolve_require_path(graph: &Graph, require_path: &str, load_paths: &[Pat
...
+            .collect();
+
+        handles
+            .into_iter()
+            .flat_map(|handle| handle.join().unwrap())
+            .collect::<Vec<_>>()
+    });
+
+    // Deduplicate by require_path, keeping first occurrence
+    let mut seen = HashSet::new();
```
vinistock, line 149: We're deduping based on thread termination order, whereas what we want (to match the Ruby behaviour) is to dedup based on the index where we found the require path in the `$LOAD_PATH`.

We should make `document.require_path` return a tuple of `(String, u16)` for (require_path, index_in_load_path_where_we_found_it)` and then we use the index for deduplication.

rust/rubydex/src/query.rs
```diff
@@ -105,6 +106,53 @@ pub fn resolve_require_path(graph: &Graph, require_path: &str, load_paths: &[Pat
...
+                scope.spawn(move || {
+                    let mut results = Vec::new();
+                    for (uri_id, document) in chunk {
+                        if let Some(require_path) = document.require_path(load_paths)
+                            && require_path.starts_with(prefix)
+                        {
+                            results.push((require_path, **uri_id));
+                        }
+                    }
+                    results
```
vinistock: After the other changes, this can become:


```suggestion
                    chunk.iter().filter_map(|document| {
                      document.require_path(load_paths)
                    }).collect::<Vec<_>>()
```

rust/rubydex/src/query.rs
alexcrocha: Nice, I didn't realize we could offload the filtering to the editor here

rust/rubydex/src/query.rs
alexcrocha: Thanks for catching this 👍 I'll adjust

rust/rubydex/src/model/document.rs
```diff
@@ -140,7 +140,8 @@ impl Document {
                     .join("/");
 
                 let require_path = file_path.trim_end_matches(".rb").to_string();
-                return Some(require_path);
+                #[expect(clippy::cast_possible_truncation)]
+                return Some((require_path, load_path_index as u16));
```
vinistock: Let's try the conversion and panic if we fail, so that we can easily discover if u16 is the wrong assumption about how long the `$LOAD_PATH` might be.
```suggestion
                return Some((require_path, load_path_index.try_into().expect("Load path entries exceed u16"))); 
```

rust/rubydex/src/query.rs
```diff
@@ -105,6 +106,54 @@ pub fn resolve_require_path(graph: &Graph, require_path: &str, load_paths: &[Pat
...
+/// Returns all require paths. Used for completion.
+///
+/// When multiple files resolve to the same require path (e.g., `foo.rb` exists in multiple
+/// load paths), the one from the earliest load path wins. This matches Ruby's `require` behavior.
+///
+/// # Panics
+///
+/// Panics if one of the search threads panics
+#[must_use]
+pub fn require_paths(graph: &Graph, load_paths: &[PathBuf]) -> Vec<String> {
```
Morriar, line 118: Looks like there is a lot of duplication with `resolve_require_path`. Have you considered merging some of the logic in a function that would accept a block?

alexcrocha: Addressed this in #580. Both functions now share a `collect_require_paths` helper

PR #542, opened 2026-01-29
Make IDs 32bits

vinistock:
Step towards #141

When thinking about using u32 IDs, I was getting really caught up by the fact that Rust's `HashMap` implementation still requires storing a u64 for the key digests. Of course that consumes more memory and ideally we can have a custom `HashMap` implementation that uses half the amount of bits.

However, we actually use IDs quite extensively outside of just `HashMap` keys. For example, many definitions store StringId, names store multiple NameIds and ancestors is a list of DeclarationId. We can actually have very substantial memory savings right now even without the 32 bit `HashMap`.

### Implementation

This PR reduces the size of our IDs to u32 from i64. I recommend reviewing per commit:

1. Reduce the ID size to u32
2. Allow `DefinitionId` and `ReferenceId` to be tagged with their kind. This helps us reduce the changes of collision by encoding the kind in the lower bits of the ID. Essentially, we have 28 bits for the digest + 4 for the kind. Definitions and references are the things we have the most in Core and the only ones I got conflicts for running the release mode, so I think this is enough for the time being.

    As an added benefit, this allows to check the kind of definitions and references without having to retrieve them from the graph. That information is directly encoded in the ID, we can just invoke `kind`. Note that the `kind` addition conflicts with #502, we can let that PR go first.
3. Adjust the C side to use u32

### Impact

Despite still using 64 bit `HashMaps`, this PR still reduces our memory used from `~3810 MB` to 
`~2880 MB` (25% reduction).

I believe we can further reduce this with a custom `HashMap` implementation that stores u32 internally.

### Why now?

I believe this is the right time to do this for the following reason: we're starting adoption of Rubydex in our tools. It will be easier for us to verify that we can indeed get away with u32 and lower memory usage _and then_ increase to u64 if necessary than the other way around.

Also, we're almost getting to 4GB and a reduction is definitely welcome.

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -52,10 +52,10 @@ static VALUE declaration_definitions_yield(VALUE args) {
     HandleData *data;
     TypedData_Get_Struct(self, HandleData, &handle_type, data);
 
-    int64_t id = 0;
+    uint32_t id = 0;
     DefinitionKind kind;
     while (rdx_definitions_iter_next(iter, &id, &kind)) {
-        VALUE argv[] = {data->graph_obj, LL2NUM(id)};
+        VALUE argv[] = {data->graph_obj, UINT2NUM(id)};
         VALUE defn_class = rdxi_definition_class_for_kind(kind);
         VALUE handle = rb_class_new_instance(2, argv, defn_class);
         rb_yield(handle);
@@ -120,14 +120,14 @@ static VALUE rdxr_declaration_member(VALUE self, VALUE name) {
         rb_raise(rb_eTypeError, "expected String");
     }
 
-    const int64_t *id_ptr = rdx_declaration_member(graph, data->id, StringValueCStr(name));
+    const uint32_t *id_ptr = rdx_declaration_member(graph, data->id, StringValueCStr(name));
     if (id_ptr == NULL) {
         return Qnil;
     }
 
-    int64_t id = *id_ptr;
-    free_i64(id_ptr);
-    VALUE argv[] = {data->graph_obj, LL2NUM(id)};
+    uint32_t id = *id_ptr;
+    free_u32(id_ptr);
+    VALUE argv[] = {data->graph_obj, UINT2NUM(id)};
 
     return rb_class_new_instance(2, argv, cDeclaration);
 }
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/model/ids.rs
Morriar, line 13: Should we also tag declarations?

rust/rubydex/src/model/definitions.rs
```diff
@@ -49,6 +49,53 @@ impl DefinitionFlags {
     }
 }
 
+#[repr(u8)]
```
Morriar, line 52: You should be able to fit this in 4 bits?

rust/rubydex/src/model/references.rs
```diff
@@ -1,8 +1,32 @@
 use crate::{
+    assert_mem_size,
     model::ids::{NameId, ReferenceId, StringId, UriId},
     offset::Offset,
 };
 
+#[repr(u8)]
```
Morriar, line 7: 4 bits should be enough?

rust/rubydex/src/resolution.rs
```diff
@@ -1800,7 +1800,7 @@ mod tests {
 
         assert_eq!(
             [
-                "Top", "Foo", "Qux", "Bar", "AfterTop", "Baz", "Zip", "Zap", "Zop", "Boop"
+                "Top", "Foo", "Bar", "AfterTop", "Qux", "Baz", "Zip", "Zap", "Zop", "Boop"
```
Morriar, line 1803: Why did this change again?

rust/rubydex/src/model/ids.rs
vinistock: I haven't figure out a good way to do it. Consider our Ruby API:

```ruby
graph["Foo"]
```

The Ruby API has no way of knowing what `Foo` is before looking it up, so we wouldn't be able to construct the right `DeclarationId` to search.

For now, I think it's okay because we have a lot fewer declarations than definitions.

rust/rubydex/src/model/definitions.rs
vinistock: Is there a way to do this? u8 is the smallest integer storage we can use no?

rust/rubydex/src/resolution.rs
vinistock: This time I know! Because we changed the hashing algorithm and now entries are inserted in different ordering inside of the hashmap.

Once again, these changes don't actually represent behaviour change because the names have the same score.

PR #545, opened 2026-01-30
Expose Declaration#descendants to Ruby API

vinistock:
Expose the descendants information to the Ruby API, which will allow the Ruby LSP to finally implement the sub-types feature.

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -232,6 +232,29 @@ static VALUE rdxr_declaration_ancestors(VALUE self) {
     return self;
 }
 
+// Declaration#descendants: () -> Enumerator[Declaration]
+static VALUE rdxr_declaration_descendants(VALUE self) {
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize(self, rb_str_new2("descendants"), 0, NULL);
+    }
+
+    HandleData *data;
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/model/declaration.rs
```diff
@@ -412,6 +412,11 @@ impl Namespace {
         all_namespaces!(self, it => std::mem::take(&mut it.diagnostics))
     }
 
+    #[must_use]
+    pub fn descendants(&self) -> &IdentityHashSet<DeclarationId> {
```
Morriar: Let's put it closer from `add_descendant` and friends

ext/rubydex/declaration.c
```diff
@@ -252,6 +275,7 @@ void rdxi_initialize_declaration(VALUE mRubydex) {
     rb_define_method(cDeclaration, "definitions", rdxr_declaration_definitions, 0);
     rb_define_method(cDeclaration, "owner", rdxr_declaration_owner, 0);
     rb_define_method(cDeclaration, "ancestors", rdxr_declaration_ancestors, 0);
+    rb_define_method(cDeclaration, "descendants", rdxr_declaration_descendants, 0);
```
Morriar: Should it be only on `Namespace`?

vinistock: Ah, yeah! Missed that in the rebase.

PR #546, opened 2026-02-02
Expose Declaration#owner to Ruby API

vinistock:
Similar to exposing a declaration's singleton class, consumers of the API should be able to read the owner.

Just a note: for all declarations, owner means what namespace owns that declaration, except for singleton classes where owner means the attached object. There's no implementation difference or multiple code paths. It's just a conceptual difference where the same field represents slightly different things to avoid extra memory cost.

### DIFF
--- a/test/declaration_test.rb
+++ b/test/declaration_test.rb
@@ -129,4 +129,27 @@ class DeclarationTest < Minitest::Test
       assert_equal("Foo::Bar::<Bar>", bar.singleton_class.name)
     end
   end
+
+  def test_declaration_owner
+    with_context do |context|
+      context.write!("file1.rb", <<~RUBY)
+        module Foo
+          class Bar
+            class << self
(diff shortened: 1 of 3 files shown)


test/declaration_test.rb
```diff
@@ -129,4 +129,27 @@ def something; end
...
+      RUBY
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      assert_nil(graph["Foo"].singleton_class)
+
+      bar = graph["Foo::Bar"]
+      assert_equal("Foo::Bar::<Bar>", bar.singleton_class.name)
```
st0012: Why are we still asserting singleton_class here and above?

vinistock: Ooops. That was totally just wrong.

PR #551, opened 2026-02-03
[DOC] Document anonymous class and module behaviors

st0012:


docs/ruby-behaviors.md
```diff
@@ -1214,3 +1215,172 @@ Foo.singleton_class.on_singleton_singleton
...
+```
+
+### Namespaces and lexical nesting
+
+The block does **not** introduce lexical nesting for `class`/`module` definitions:
+
+```ruby
+class Foo
+  Bar = Class.new do
+    class Baz; end
```
Morriar, line 1308: Should we repeat the `self::` trick for classes: `class self::Baz; end`?

docs/ruby-behaviors.md
```diff
@@ -1214,3 +1215,172 @@ Foo.singleton_class.on_singleton_singleton
...
+
+# Baz is Foo::Baz, not Foo::Bar::Baz
+```
+
+You cannot reliably reference the assigned constant inside the block because the assignment happens after the block runs. `Bar::Baz` inside the block refers to an outer `Bar` (if any) or raises `NameError`.
+
+```ruby
+class Foo
+  Bar = Class.new do
+    class Baz < Bar; end # causes uninitialized constant Foo::Bar (NameError)
```
Morriar, line 1320: And `class Baz < self::Bar` won't work either

PR #555, opened 2026-02-03
Expose require path resolution to Ruby API

alexcrocha:
Follow-up to #528, and a step towards #411. 

This PR adds FFI and C bindings to expose `resolve_require_path` to Ruby, returning a `Document` that can be used directly for go-to-definition on require statements.

### DIFF
--- a/rust/rubydex-sys/src/graph_api.rs
+++ b/rust/rubydex-sys/src/graph_api.rs
@@ -11,6 +11,7 @@ use rubydex::model::ids::DeclarationId;
 use rubydex::resolution::Resolver;
 use rubydex::{indexing, listing, query};
 use std::ffi::CString;
+use std::path::PathBuf;
 use std::{mem, ptr};
 
 pub type GraphPointer = *mut c_void;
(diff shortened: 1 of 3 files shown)


rust/rubydex-sys/src/graph_api.rs
```diff
@@ -429,6 +430,51 @@ pub unsafe extern "C" fn rdx_graph_method_references_iter_new(pointer: GraphPoin
...
+    })
+}
+
+/// Frees a URI ID pointer returned by `rdx_resolve_require_path`.
+///
+/// # Safety
+/// - `uri_id` must be a pointer returned by `rdx_resolve_require_path`, or NULL.
+/// - `uri_id` must not be used after being freed.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_uri_id_free(uri_id: *const u32) {
```
vinistock: We already have `free_u32` for this, so we can just reuse it.

alexcrocha: Nice!

test/graph_test.rb
```diff
@@ -368,6 +368,21 @@ def self.other_thing
...
 
+  def test_resolve_require_path
+    with_context do |context|
+      context.write!("lib/foo/bar.rb", "class Bar; end")
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+
+      load_paths = [context.absolute_path_to("lib")]
+      document = graph.resolve_require_path("foo/bar", load_paths)
```
Morriar, line 379: What would happen with this scenario?

```rb
context.write!("lib/foo/bar.rb", "class Bar; end")
context.write!("test/foo/bar.rb", "class Bar; end")

load_paths = [context.absolute_path_to("lib"), context.absolute_path_to("test")]
graph.resolve_require_path("foo/bar", load_paths)
```

alexcrocha: Good catch, this was a race condition 😄 earliest path should win, not whichever thread completes first. Fixed in #580 and added a test with your scenario too. 

Also took the chance to dedup the shared logic with `require_paths` as you suggested in #537.

PR #558, opened 2026-02-04
Fix duplicate constant reference insertions

vinistock:
As part of working to detect ID collisions, I ended up noticing we were pushing duplicate constant references in two situations:

- When indexing a mixin operation with `self`, we were creating a new constant reference, but we already do that immediately after. We just need to return the `name_id` from the closure
- For constant aliases, we were indexing the `node.value()` and also visiting it, which also pushed duplicates. I started visiting the value conditionally and ensured that recursing on constant alias targets creates the expected structures recursively

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -801,16 +801,28 @@ impl<'a> RubyIndexer<'a> {
                 self.index_constant_reference(value, true)
             }
             ruby_prism::Node::ConstantWriteNode { .. } => {
-                self.index_constant_alias_target(&value.as_constant_write_node().unwrap().value())
+                let node = value.as_constant_write_node().unwrap();
+                let target_name_id = self.index_constant_alias_target(&node.value())?;
+                self.add_constant_alias_definition(value, target_name_id, false);
+                Some(target_name_id)
             }
             ruby_prism::Node::ConstantOrWriteNode { .. } => {
-                self.index_constant_alias_target(&value.as_constant_or_write_node().unwrap().value())
+                let node = value.as_constant_or_write_node().unwrap();
+                let target_name_id = self.index_constant_alias_target(&node.value())?;
+                self.add_constant_alias_definition(value, target_name_id, false);
+                Some(target_name_id)
             }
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4484,7 +4490,12 @@ mod tests {
...
         assert_no_diagnostics!(&context);
 
-        let method_ref = context.graph().method_references().values().next().unwrap();
+        let method_ref = context
+            .graph()
+            .method_references()
+            .values()
+            .find(|r| *r.str() == StringId::from("&&"))
+            .unwrap();
         assert_eq!(StringId::from("&&"), *method_ref.str());
```
Morriar: Are we asserting twice on the equality between the `find` and `assert_eq`?


```suggestion
        assert!(context
            .graph()
            .method_references()
            .values()
            .find(|r| *r.str() == StringId::from("&&"))
            .is_some());
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4511,7 +4522,12 @@ mod tests {
...
         assert_no_diagnostics!(&context);
 
-        let method_ref = context.graph().method_references().values().next().unwrap();
+        let method_ref = context
+            .graph()
+            .method_references()
+            .values()
+            .find(|r| *r.str() == StringId::from("||"))
+            .unwrap();
         assert_eq!(StringId::from("||"), *method_ref.str());
```
Morriar:
```suggestion
        assert!(context
            .graph()
            .method_references()
            .values()
            .find(|r| *r.str() == StringId::from("||"))
            .is_some());
```

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: We still use the variable below, but I removed the duplicate assertion 👍

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: We still use the variable below, but I removed the duplicate assertion 👍

PR #559, opened 2026-02-04
Incremental infrastructure

thomasmarshall:
This PR sets up the basic infrastructure and API for incremental updates:

```rb
result = graph.index_all(file_paths)
graph.resolve(result)
```

For now we can continue using:

```rb
graph.index_all(file_paths)
graph.resolve
```

…because `graph.resolve(nil)` will resolve everything, but we might want to switch to making the parameter required.

@vinistock suggested that we should be returning a list of units, but by returning a result struct containing two sets of IDs, it's easier to reuse the existing infrastructure for sorting and processing the units. I'm not sure I fully understood the benefit to constructing the units at index time rather than during resolution, but I'm happy to revisit if you think that would be preferable?

As of this PR, incremental updates **don't work**—we should continue using `graph.resolve(nil)` for now.

## Feedback

I'm not confident in the FFI C side of things, particularly when the result will be freed (whether we pass it into `#resolve` or not). Claude Code helped me with this and I sense-checked to the best of my current abilities. Any feedback on this would be much appreciated!

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -6,6 +6,7 @@
 #include "reference.h"
 #include "ruby/internal/globals.h"
 #include "rustbindings.h"
+#include "index_result.h"
 #include "utils.h"
 
 static VALUE cGraph;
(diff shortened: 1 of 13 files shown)


Morriar (commented):
<deleted>

st0012:
We've implemented incremental resolution and it's working fine (tho still need perf tuning). @thomasmarshall Thank you for the early explorations 🫡

PR #560, opened 2026-02-04
Expose require paths to Ruby API

### LINKED ISSUE #411: Design a way to search require paths

alexcrocha:
Closes #411 

This PR adds FFI and C bindings to expose `require_paths` to Ruby, returning all indexed require paths for a given load path. This enables completion functionality for require statements in editors.

Also adds `free_c_string_array` as a generic utility for freeing arrays of C strings allocated by Rust.

### DIFF
--- a/rust/rubydex-sys/src/graph_api.rs
+++ b/rust/rubydex-sys/src/graph_api.rs
@@ -443,6 +443,44 @@ pub unsafe extern "C" fn rdx_resolve_require_path(
     })
 }
 
+/// Returns all require paths for completion.
+/// Returns array of C strings and writes count to `out_count`.
+/// Returns null if `load_path` contain invalid UTF-8.
+/// Caller must free with `free_c_string_array`.
+///
+/// # Safety
+/// - `pointer` must be a valid `GraphPointer` previously returned by this crate.
+/// - `load_path` must be an array of `load_path_count` valid, null-terminated UTF-8 strings.
+/// - `out_count` must be a valid, writable pointer.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_require_paths(
+    pointer: GraphPointer,
+    load_path: *const *const c_char,
+    load_path_count: usize,
+    out_count: *mut usize,
+) -> *const *const c_char {
+    let Ok(paths_vec) = (unsafe { utils::convert_double_pointer_to_vec(load_path, load_path_count) }) else {
+        return ptr::null_mut();
+    };
+    let paths_vec = paths_vec.into_iter().map(PathBuf::from).collect::<Vec<_>>();
+
+    let results = with_graph(pointer, |graph| query::require_paths(graph, &paths_vec));
+
+    let c_strings: Vec<*const c_char> = results
+        .into_iter()
+        .filter_map(|string| {
+            CString::new(string)
+                .ok()
+                .map(|c_string| c_string.into_raw().cast_const())
+        })
+        .collect();
+
+    unsafe { *out_count = c_strings.len() };
+
+    let boxed = c_strings.into_boxed_slice();
+    Box::into_raw(boxed).cast::<*const c_char>()
+}
+
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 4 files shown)


rust/rubydex-sys/src/utils.rs
```diff
@@ -48,3 +48,23 @@ pub extern "C" fn free_u32(ptr: *const u32) {
...
+    if ptr.is_null() {
+        return;
+    }
+
+    let slice = unsafe { Box::from_raw(std::ptr::slice_from_raw_parts_mut(ptr, count)) };
+    for pointer in &slice {
+        if !pointer.is_null() {
+            free_c_string(*pointer);
+        }
+    }
```
vinistock: If you instantiate the slice from the pointer, does Rust not automatically free it at the end of this function?

In other words, maybe we don't need the for loop here because once `slice` goes out of scope, Rust will immediately drop it.

rust/rubydex-sys/src/graph_api.rs
```diff
@@ -443,6 +443,44 @@ pub unsafe extern "C" fn rdx_resolve_require_path(
...
+/// - `pointer` must be a valid `GraphPointer` previously returned by this crate.
+/// - `load_path` must be an array of `load_path_count` valid, null-terminated UTF-8 strings.
+/// - `out_count` must be a valid, writable pointer.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_require_paths(
+    pointer: GraphPointer,
+    load_path: *const *const c_char,
+    load_path_count: usize,
+    out_count: *mut usize,
+) -> *mut *const c_char {
```
vinistock: Let's make this a `*const *const c_char` instead. We don't expect consumers of the API to be able to mutate this pointer.

rust/rubydex-sys/src/utils.rs
alexcrocha: Our CI memleak check detects a memory leak without us freeing each pointer. I will add it back

vinistock: I think this would work:

```rust
unsafe {
  let _ = convert_double_pointer_to_vec(ptr, count);
}
```

alexcrocha: It also [fails](https://github.com/Shopify/rubydex/actions/runs/21733102985/job/62692382829) the memleak check.
There doesn't seem to be a way around freeing each `CString` ourselves. [CString::into_raw](https://doc.rust-lang.org/stable/alloc/ffi/struct.CString.html#method.into_raw) does say this: 
> Failure to call [CString::from_raw](https://doc.rust-lang.org/stable/alloc/ffi/struct.CString.html#method.from_raw) will lead to a memory leak.

Is there a code smell catching your attention here?

vinistock: Sorry, I didn't realize that `convert_double_pointer_to_vec` operates on `CStr` and not `CString`.

The only weird thing to me is for us to manually loop freeing each part. I don't know what's more conventional, looping through the slice or doing this (which is what I originally intended with my previous comment):

```rust
let _ = unsafe {
  slice::from_raw_parts(data, len)
    .iter()
    .map(|arg| CString::from_raw(*arg).to_str().map(ToString::to_string))
    .collect()
}
```

This just instantiates the array of strings back and lets Rust automatically drop it vs looping through each element.

alexcrocha: Ah, I understand it now. I believe the key difference is that `slice::from_raw_parts` returns a reference `&[T]` and doesn't take ownership. 

I replaced the explicit loop with the iterator, but I have to keep the `Box::from_raw` for Rust to take ownership of the outer pointer.

Issue #87, opened 2025-08-12
Implement core definitions with RBS


vinistock:
Using the work from #8, let's start indexing core definitions (e.g.: `String` or `Integer`). We basically need to find out where the `rbs` gem is installed (since it contains all of the RBS files), then we glob the stdlib/core definitions and index each file.

**Note**: one interesting aspect that we could experiment with is shipping a pre-populated DB with all RBS declarations. These don't change until there's a new release of RBS, so we could simply index them and include the DB file as part of the distributed gem, which would prevent having all developers from needing to do the work from scratch.

In this case, "creating" a fresh DB would mean just copying the bundled DB over with the base RBS definitions.

To match the Ruby LSP's current level of RBS indexing, we need to

- [x] Change `MethodDefinitions` to store a vector of signatures. We currently only allow one set of parameters, but RBS allows for overloads, so we must track a list of possible parameters
- [x] Index classes
- [x] Index modules
- [x] Index methods
- [x] Index globals
- [x] Index constants
- [x] Index aliases
- [x] Index includes
- [x] Index prepends
- [x] Index extends

PR #562, opened 2026-02-04
Add RBS indexer scaffolding

alexcrocha:
Part of https://github.com/Shopify/rubydex/issues/87

This PR introduces the basic structure for indexing RBS files. The indexer doesn't extract any definitions yet. It is intentionally minimal to establish the pattern and enable parallel development.

The `RBSIndexer` follows the same interface as `RubyIndexer`:
 - `new(uri, source)` - creates the indexer
 - `index()` - processes the file (currently no-op)
 - `local_graph()` - returns the `LocalGraph`

### DIFF
--- a/rust/rubydex/src/indexing.rs
+++ b/rust/rubydex/src/indexing.rs
@@ -1,6 +1,6 @@
 use crate::{
     errors::Errors,
-    indexing::{local_graph::LocalGraph, ruby_indexer::RubyIndexer},
+    indexing::{local_graph::LocalGraph, rbs_indexer::RBSIndexer, ruby_indexer::RubyIndexer},
     job_queue::{Job, JobQueue},
     model::graph::Graph,
 };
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing.rs
```diff
@@ -55,11 +56,20 @@ impl Job for IndexingRubyFileJob {
             return;
         };
 
-        let mut ruby_indexer = RubyIndexer::new(url.to_string(), &source);
-        ruby_indexer.index();
+        let local_graph = if let Some(ext) = self.path.extension()
```
soutaro, line 59: It would be better to add `IndexingRBSFileJob` (or something similar). Processing RBS files within `IndexingRubyFileJob` might look a bit odd.

soutaro: Or, I think we can simply skip doing this part. We need some API on Ruby extension to specify the path where RBS files are read from.

vinistock: The idea of the indexing job is for it to accept any URI and be able to index it. Yes, we'll need to pass RBS file paths to indexing, but I'm not sure we need a different job for it?

The jobs would end up being identical, so I'd rather we share the infrastructure.

That said, we can probably rename the job to `IndexingJob` and remove `Ruby` and `File` from its name (since it can index any type of URI).

alexcrocha: Renamed it to `IndexingJob` for now. If necessary, we can change it down the line

PR #515, opened 2026-01-22
Capture DSL definitions and support `Class|Module.new`

st0012:
# DSL Support for Class.new and Module.new

### Supported Ruby Patterns

This implementation adds support for dynamically created classes and modules:

```ruby
# Named dynamic class with inheritance and mixins
Foo = Class.new(Parent) do
  include SomeMixin

  def instance_method
    @ivar = 1
  end

  def self.class_method; end

  CONST = "value"
end

# Named dynamic module
Bar = Module.new do
  def helper; end
end

# Nested dynamic classes
Outer = Class.new do
  Inner = Class.new { def foo; end }
end

# Anonymous (indexed but no declaration)
Class.new { def bar; end }


# Singleton class inside anonymous classes
Baz = Class.new do
  class << self
    def self.qux; end
  end
end
```

### High-Level Approach

```text
┌────────────────────────────────────┐      ┌────────────────────────────────────┐
│             INDEXING               │      │            RESOLUTION              │
├────────────────────────────────────┤      ├────────────────────────────────────┤
│                                    │      │                                    │
│  Foo = Class.new(Parent) { ... }   │      │  1. Resolve receiver "Class"       │
│            │                       │      │     → CLASS_ID                     │
│            ▼                       │      │                                    │
│  ┌──────────────────────────────┐  │      │  2. Match DSL processor            │
│  │ ConstantDefinition(Foo)      │  │      │     class_new_matches() = true     │
│  └──────────────────────────────┘  │      │                                    │
│  ┌──────────────────────────────┐  │      │  3. Create DynamicClassDefinition  │
│  │ DslDefinition                │  │ ───► │     - name_id from Constant        │
│  │   receiver: Class (NameId)   │  │      │     - superclass from args(Parent) │
│  │   method: "new"              │  │      │     - mixins from block            │
│  │   args: [Parent]             │  │      │                                    │
│  │   assigned_to: Foo.id        │  │      │  4. Remove ConstantDefinition      │
│  └──────────────────────────────┘  │      │                                    │
│  ┌──────────────────────────────┐  │      │  5. Resolve DynamicClassDefinition │
│  │ Block contents assigned with │  │      │     → ClassDeclaration(Foo)        │
│  │ lexical_nesting_id = DSL     │  │      │     → Members resolved with        │
│  └──────────────────────────────┘  │      │       owner = Foo                  │
│                                    │      │                                    │
└────────────────────────────────────┘      └────────────────────────────────────┘
```

### Named vs Anonymous

```text
[...]

PR #566, opened 2026-02-05
Add `--report-orphans` flag to produce orphan definition report

st0012:
When used, rubydex would list orphan definitions under `/tmp/rubydex-orphan-report.txt` (can be changed by supplying arg to the flag).

Example output:

```
Constant	::Rack::Cache::MetaStore::RAILS	/Users/hung-wulo/src/github.com/rails/rails/actionpack/lib/action_dispatch/http/rack_cache.rb:29:31-29:35
Constant	::Rack::Cache::EntityStore::RAILS	/Users/hung-wulo/src/github.com/rails/rails/actionpack/lib/action_dispatch/http/rack_cache.rb:61:33-61:37
InstanceVariable	@adapter	/Users/hung-wulo/src/github.com/rails/rails/activejob/test/helper.rb:8:1-8:9
SingletonClass	Benchmark::<Benchmark>	/Users/hung-wulo/src/github.com/rails/rails/activesupport/lib/active_support/core_ext/benchmark.rb:5:1-5:13
```

This feature is helpful for investigating issues. ~~For example, some definitions in Core become orphan due to `NameId` collision and that's shown in the report (I've shown that to @vinistock in pairing).~~ (This is fixed now) 
Another example is that this helps me investigate incorrect resolution in #515.

However, it's also worth noting that detecting orphans don't necessarily mean there's a but in the code or rubydex. For example, due to the lack of `<main>` representation, top level instance variables would be considered orphans now.

### DIFF
--- a/rust/rubydex/src/main.rs
+++ b/rust/rubydex/src/main.rs
@@ -27,6 +27,16 @@ struct Args {
 
     #[arg(long = "stats", help = "Show detailed performance statistics")]
     stats: bool,
+
+    #[arg(
+        long = "report-orphans",
+        value_name = "PATH",
+        num_args = 0..=1,
+        require_equals = true,
+        default_missing_value = "/tmp/rubydex-orphan-report.txt",
+        help = "Write orphan definitions report to specified file"
+    )]
+    report_orphans: Option<String>,
 }
 
 #[derive(Debug, Clone, ValueEnum)]
@@ -99,6 +109,20 @@ fn main() {
     if args.stats {
         Timer::print_breakdown();
         MemoryStats::print_memory_usage();
+    }
+
+    // Orphan report
+    if let Some(ref path) = args.report_orphans {
+        match std::fs::File::create(path) {
+            Ok(mut file) => {
+                if let Err(e) = graph.write_orphan_report(&mut file) {
+                    eprintln!("Failed to write orphan report: {e}");
+                } else {
+                    println!("Orphan report written to {path}");
+                }
+            }
+            Err(e) => eprintln!("Failed to create orphan report file: {e}"),
+        }
     }
 
     // Generate visualization or print statistics
--- a/rust/rubydex/src/model/definitions.rs
+++ b/rust/rubydex/src/model/definitions.rs
@@ -142,7 +142,16 @@ impl Definition {
             Definition::SingletonClass(d) => Some(d.name_id()),
             Definition::Module(d) => Some(d.name_id()),
             Definition::Constant(d) => Some(d.name_id()),
-            _ => None,
+            Definition::ConstantAlias(d) => Some(d.name_id()),
+            Definition::GlobalVariable(_)
+            | Definition::InstanceVariable(_)
+            | Definition::ClassVariable(_)
+            | Definition::AttrAccessor(_)
+            | Definition::AttrReader(_)
+            | Definition::AttrWriter(_)
+            | Definition::Method(_)
+            | Definition::MethodAlias(_)
+            | Definition::GlobalVariableAlias(_) => None,
         }
     }
 
--- a/rust/rubydex/src/stats.rs
+++ b/rust/rubydex/src/stats.rs
@@ -1,4 +1,6 @@
 pub mod memory;
+// TODO: When the rubydex is stable enough, turn this into a debug-only feature or revisit if we still need it.
+pub mod orphan_report;
 pub mod timer;
 
 /// Helper function to compute percentage
--- a/rust/rubydex/src/stats/orphan_report.rs
+++ b/rust/rubydex/src/stats/orphan_report.rs
@@ -1,0 +1,262 @@
+use std::collections::HashSet;
+use std::io::Write;
+
+use crate::model::declaration::Declaration;
+use crate::model::definitions::Definition;
+use crate::model::graph::Graph;
+use crate::model::ids::{DefinitionId, NameId, StringId};
+use crate::model::name::{NameRef, ParentScope};
+
+impl Graph {
+    /// Writes a report of orphan definitions (definitions not linked to any declaration).
+    ///
+    /// Format: `type\tconcatenated_name\tlocation` (TSV)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if writing fails.
+    pub fn write_orphan_report(&self, writer: &mut impl Write) -> std::io::Result<()> {
+        // Collect all definition IDs that are linked to declarations
+        let linked_definition_ids: HashSet<&DefinitionId> = self
+            .declarations()
+            .values()
+            .flat_map(Declaration::definitions)
+            .collect();
+
+        // Find orphan definitions
+        let mut orphans: Vec<_> = self
+            .definitions()
+            .iter()
+            .filter(|(id, _)| !linked_definition_ids.contains(id))
+            .collect();
+
+        // Sort by type, then by location for consistent output
+        orphans.sort_by(|(_, a), (_, b)| {
+            a.kind()
+                .cmp(b.kind())
+                .then_with(|| a.uri_id().cmp(b.uri_id()))
+                .then_with(|| a.offset().cmp(b.offset()))
+        });
+
+        for (_, definition) in orphans {
+            let kind = definition.kind();
+            let name = match definition.name_id().copied() {
+                Some(id) => self.build_concatenated_name_from_name(id),
+                None => self.build_concatenated_name_from_lexical_nesting(definition),
+            };
+            let location = self.definition_location(definition);
+
+            writeln!(writer, "{kind}\t{name}\t{location}")?;
+        }
+
+        Ok(())
+    }
+
+    /// Walks the Name system's `parent_scope` chain to reconstruct the constant path.
+    /// Falls back to `nesting` for enclosing scope context when there is no explicit parent scope.
+    ///
+    /// Note: this produces a concatenated name by piecing together name parts, not a properly
+    /// resolved qualified name.
+    pub(crate) fn build_concatenated_name_from_name(&self, name_id: NameId) -> String {
+        let Some(name_ref) = self.names().get(&name_id) else {
+            return "<unknown>".to_string();
+        };
+        let simple_name = self.string_id_to_string(*name_ref.str());
+
+        match name_ref.parent_scope() {
+            ParentScope::Some(parent_id) | ParentScope::Attached(parent_id) => {
+                let parent_name = self.build_concatenated_name_from_name(*parent_id);
+                format!("{parent_name}::{simple_name}")
+            }
+            ParentScope::TopLevel => format!("::{simple_name}"),
+            ParentScope::None => {
+                let prefix = name_ref
+                    .nesting()
+                    .as_ref()
+                    .map(|nesting_id| self.build_nesting_prefix(*nesting_id))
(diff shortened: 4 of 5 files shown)


rust/rubydex/src/main.rs
```diff
@@ -27,6 +27,16 @@ struct Args {
 
     #[arg(long = "stats", help = "Show detailed performance statistics")]
     stats: bool,
+
+    #[arg(
+        long = "show-orphan",
```
Morriar: It's not really `show` if you write it to a file, what about `write-orphan`?

Another approach would be to print it to stdout and let the user pipe the result.

rust/rubydex/src/model/graph.rs
```diff
@@ -689,6 +689,167 @@ impl Graph {
...
+            last
+        } else {
+            format!("{}{separator}{last}", parts.join("::"))
+        }
+    }
+
+    /// Converts a `StringId` to its string value.
+    fn string_id_to_string(&self, string_id: StringId) -> String {
+        // Handle special "<anonymous>" StringId that isn't in the strings table
+        let anonymous_id = StringId::from("<anonymous>");
```
Morriar: When do we actually ask for this exact name? Is the actual string something like `{}:{}<anonymous>`? What kind we just look for the string id in the string table?

rust/rubydex/src/model/graph.rs
```diff
@@ -689,6 +689,167 @@ impl Graph {
...
+        if string_id == anonymous_id {
+            return "<anonymous>".to_string();
+        }
+
+        self.strings
+            .get(&string_id)
+            .map_or_else(|| format!("<unknown:{string_id}>"), |s| s.to_string())
```
Morriar: When can this happen? Feels like we should unwrap and fail violently?

rust/rubydex/src/model/graph.rs
```diff
@@ -689,6 +689,167 @@ impl Graph {
...
+
+        let offset = definition.offset();
+        let line_index = document.line_index();
+        let start = line_index.line_col(offset.start().into());
+
+        // Convert file:// URI to path
+        let uri_str = document.uri();
+        let path = uri_str.strip_prefix("file://").unwrap_or(uri_str);
+
+        format!("{}:{}:{}", path, start.line + 1, start.col + 1)
```
Morriar: Let's make https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/offset.rs#L62-L69 a default feature and reuse it

rust/rubydex/src/model/graph.rs
```diff
@@ -689,6 +689,167 @@ impl Graph {
...
+
+            writeln!(file, "{kind} | {name} | {location}")?;
+        }
+
+        Ok(())
+    }
+
+    /// Walks the Name system's `parent_scope` chain to reconstruct the constant path.
+    /// Falls back to `nesting` for enclosing scope context when there is no explicit parent scope.
+    fn build_qualified_name_from_name(&self, name_id: NameId) -> String {
```
Morriar: Let's add a few tests for this, there is a lot of logic here

rust/rubydex/src/model/graph.rs
```diff
@@ -689,6 +689,167 @@ impl Graph {
...
+
+        for (_, definition) in orphans {
+            let kind = definition.kind();
+            let name = match definition.name_id().copied() {
+                Some(id) => self.build_qualified_name_from_name(id),
+                None => self.build_qualified_name_from_lexical_nesting(definition),
+            };
+            let location = self.definition_location(definition);
+
+            writeln!(file, "{kind} | {name} | {location}")?;
```
Morriar: Looks like you could almost print a [`.tsv` file](https://en.wikipedia.org/wiki/Tab-separated_values)? Makes it easier to import as a spreadsheet and as easy to read in raw format.


```suggestion
            writeln!(file, "{kind}\t{name}\t{location}")?;
```

rust/rubydex/src/main.rs
st0012: I named it `report-orphans` because `write-orphans` could be interpreted as we're modifying something.

rust/rubydex/src/model/definitions.rs
```diff
@@ -142,6 +142,7 @@ impl Definition {
             Definition::SingletonClass(d) => Some(d.name_id()),
             Definition::Module(d) => Some(d.name_id()),
             Definition::Constant(d) => Some(d.name_id()),
+            Definition::ConstantAlias(d) => Some(d.name_id()),
```
vinistock, line 145: Good catch. I think we should strive to avoid using the catchall arm and instead always list all possibilities explicitly.

It's bit ugly, but the advantage is that whenever a new variant is added, everyone is forced to consider how it should be handled everywhere.

rust/rubydex/src/model/graph.rs
```diff
@@ -731,6 +731,156 @@ impl Graph {
...
     }
 
+    /// Writes a report of orphan definitions (definitions not linked to any declaration) to a file.
+    ///
+    /// Format: `type\tfully_qualified_name\tlocation` (TSV)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the file cannot be created or written to.
+    pub fn write_orphan_report(&self, path: &std::path::Path) -> std::io::Result<()> {
```
vinistock: We may end up with a bunch of different reports, statistics and visualizations that operate on the graph. I think it would be worth moving these to their dedicated modules.

rust/rubydex/src/model/graph.rs
```diff
@@ -731,6 +731,156 @@ impl Graph {
...
+
+        // Sort by type, then by location for consistent output
+        orphans.sort_by(|(_, a), (_, b)| {
+            a.kind()
+                .cmp(b.kind())
+                .then_with(|| a.uri_id().cmp(b.uri_id()))
+                .then_with(|| a.offset().cmp(b.offset()))
+        });
+
+        let mut file = std::fs::File::create(path)?;
```
vinistock: Nitpick: since we're already using `Write`, how about we pass the `io` object? It makes the API easier to use with stdout if desired.

rust/rubydex/src/model/graph.rs
```diff
@@ -731,6 +731,156 @@ impl Graph {
...
+
+            writeln!(file, "{kind}\t{name}\t{location}")?;
+        }
+
+        Ok(())
+    }
+
+    /// Walks the Name system's `parent_scope` chain to reconstruct the constant path.
+    /// Falls back to `nesting` for enclosing scope context when there is no explicit parent scope.
+    pub(crate) fn build_qualified_name_from_name(&self, name_id: NameId) -> String {
```
vinistock: Not a strong opinion, but I would probably simplify the code and only keep URI + line/col information. That way, whoever is investigating the inconsistent data can open the file in one click and take a look at what pattern causes it.

rust/rubydex/src/model/definitions.rs
st0012: Updated

rust/rubydex/src/model/graph.rs
st0012: I still think it'd helpful to have the type and name. The collision cases were found because the generated FQN was weird. If all we have is the location, then we wouldn't be able to tell they became orphan because of the name collision.

rust/rubydex/src/stats.rs
```diff
@@ -1,4 +1,5 @@
 pub mod memory;
+pub mod orphan_report;
```
vinistock, line 3: Not a blocker and it may complicate the code a bit, but it would be nice to only include this in debug builds.

rust/rubydex/src/stats/orphan_report.rs
```diff
@@ -0,0 +1,233 @@
...
+
+impl Graph {
+    /// Writes a report of orphan definitions (definitions not linked to any declaration).
+    ///
+    /// Format: `type\tfully_qualified_name\tlocation` (TSV)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if writing fails.
+    pub fn write_orphan_report(&self, writer: &mut impl Write) -> std::io::Result<()> {
```
vinistock, line 18: Since this is a debugging thing, maybe let's avoid exposing it as part of the lib crate.
```suggestion
    pub(crate) fn write_orphan_report(&self, writer: &mut impl Write) -> std::io::Result<()> {
```

rust/rubydex/src/stats/orphan_report.rs
```diff
@@ -0,0 +1,233 @@
...
+            NameRef::Resolved(resolved) => self
+                .declarations()
+                .get(resolved.declaration_id())
+                .map_or_else(String::new, |decl| decl.name().to_string()),
+            NameRef::Unresolved(_) => self.build_qualified_name_from_name(nesting_id),
+        }
+    }
+
+    /// Builds a qualified name for non-constant definitions by walking the `lexical_nesting_id` chain.
+    pub(crate) fn build_qualified_name_from_lexical_nesting(&self, definition: &Definition) -> String {
```
vinistock: This is a nitpick, but I would avoid calling this qualified name. We can only truly know the right qualified name if we properly resolved involved constants.

This is more of a "concantenated name", where we just piece together all parts rather than running more complex logic.

rust/rubydex/src/stats/orphan_report.rs
```diff
@@ -0,0 +1,233 @@
...
+    fn definition_location(&self, definition: &Definition) -> String {
+        let uri_id = definition.uri_id();
+
+        let Some(document) = self.documents().get(uri_id) else {
+            return format!("{uri_id}:<unknown>");
+        };
+
+        // Convert file:// URI to path
+        let uri_str = document.uri();
+        let path = uri_str.strip_prefix("file://").unwrap_or(uri_str);
```
vinistock: If we want complete correctness, this should actually use the `Url` crate since stripping `file://` does not produce the correct file paths on all platforms (e.g.: Windows).

But tbh, why not just use the full URI and avoid any conversions?


```suggestion
        let path = Url::parse(uri_str).to_file_path().unwrap_or(uri_str);
```

rust/rubydex/src/stats/orphan_report.rs
st0012: Other functions are crate only already, but this one is called in `main.rs` so I don't think it's possible to hide?

rust/rubydex/src/stats.rs
st0012: Can we include it in the release build for a bit longer, at least until things are stable enough that we don't need this?
I just tried debug build against Core and the time it takes to run went from ~75 to ~330 seconds, which is painfully slow.
I can add a TODO comment for us to hide it or even remove it later if we think we won't need it.

vinistock: Sure 👍

rust/rubydex/src/stats/orphan_report.rs
vinistock: Oh, I didn't realize the executable would be considered external to the crate.

rust/rubydex/src/stats/orphan_report.rs
st0012: The `file:///foo.rb#L1` format allows us to cmd+click to the line. So I'll use it instead and add a test case to ensure it.

PR #569, opened 2026-02-06
Add debug assertion for ID collisions

vinistock:
This PR adds debug assertions (so that we don't pay the comparison price in release mode) to verify that we don't get collisions either in the local or global graph.

There are two slightly different strategies depending on the entity:

- For unique entities like definitions and references, we can directly check if there was a value in the `HashMap` when we try to insert
- For non unique entities like `StringRef`, `NameRef` or `Declaration`, where we increment ref counts or add extra information, we can't simply check if there was a value in the `HashMap` already. We need to verify if the value in there is exactly the same as the one we're inserting. If they are different, then we have a collision

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -69,12 +69,33 @@ impl Graph {
         &mut self.declarations
     }
 
-    pub fn add_declaration<F>(&mut self, declaration_id: DeclarationId, definition_id: DefinitionId, constructor: F)
+    pub fn add_declaration<F>(
+        &mut self,
+        definition_id: DefinitionId,
+        fully_qualified_name: String,
+        constructor: F,
+    ) -> DeclarationId
     where
-        F: FnOnce() -> Declaration,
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -69,12 +69,29 @@ impl Graph {
...
     }
 
-    pub fn add_declaration<F>(&mut self, declaration_id: DeclarationId, definition_id: DefinitionId, constructor: F)
-    where
-        F: FnOnce() -> Declaration,
+    pub fn add_declaration<F>(
+        &mut self,
+        declaration_id: DeclarationId,
+        definition_id: DefinitionId,
+        fully_qualified_name: String,
```
Morriar, line 75: Can't we build the `declaration_id` inside `add_declaration` if we already pass the `fully_qualified_name`?

vinistock: Yeah, we can.

PR #573, opened 2026-02-06
Introduce Receiver enum to distinguish self vs constant method receivers

st0012:
## Summary

Prerequisite for https://github.com/Shopify/rubydex/pull/515 (anonymous Class/Module). Anonymous classes have no names, so `def self.foo` inside them can't be represented as `Option<NameId>`. `SelfReceiver(DefinitionId)` points directly to the enclosing definition without needing a name.

- Replaces `Option<NameId>` on `MethodDefinition.receiver` with `Receiver::SelfReceiver(DefinitionId) | ConstantReceiver(NameId)`, making the distinction between `def self.foo` and `def Foo.foo` explicit at the type level — previously both were `NameId` and indistinguishable by the time resolution runs.
- SelfReceiver enables a direct `definition_id_to_declaration_id` lookup instead of going through name resolution.
- Fixed a panic where `def self.baz` inside an unresolvable class (e.g., `class Foo::Bar` where `Foo` is undefined) would crash on `.expect()` — now silently skips orphaned methods.

### DIFF
--- a/rust/rubydex/src/model/definitions.rs
+++ b/rust/rubydex/src/model/definitions.rs
@@ -712,9 +712,21 @@ pub struct MethodDefinition {
     lexical_nesting_id: Option<DefinitionId>,
     parameters: Vec<Parameter>,
     visibility: Visibility,
-    receiver: Option<NameId>,
+    receiver: Option<Receiver>,
 }
+
 assert_mem_size!(MethodDefinition, 112);
+
+/// The receiver of a singleton method definition.
+#[derive(Debug)]
+pub enum Receiver {
+    /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
+    SelfReceiver(DefinitionId),
+    /// `def Foo.bar` - receiver is an explicit constant that needs resolution
+    ConstantReceiver(NameId),
+}
+
+assert_mem_size!(Receiver, 16);
 
 impl MethodDefinition {
     #[allow(clippy::too_many_arguments)]
@@ -728,7 +740,7 @@ impl MethodDefinition {
         lexical_nesting_id: Option<DefinitionId>,
         parameters: Vec<Parameter>,
         visibility: Visibility,
-        receiver: Option<NameId>,
+        receiver: Option<Receiver>,
     ) -> Self {
         Self {
             str_id,
@@ -747,8 +759,13 @@ impl MethodDefinition {
     pub fn id(&self) -> DefinitionId {
         let mut formatted_id = format!("{}{}{}", *self.uri_id, self.offset.start(), *self.str_id);
 
-        if let Some(receiver) = self.receiver {
-            formatted_id.push_str(&receiver.to_string());
+        if let Some(receiver) = &self.receiver {
+            match receiver {
+                Receiver::SelfReceiver(def_id) => formatted_id.push_str(&def_id.to_string()),
+                Receiver::ConstantReceiver(name_id) => {
+                    formatted_id.push_str(&name_id.to_string());
+                }
+            }
         }
 
         DefinitionId::from(&formatted_id)
@@ -785,7 +802,7 @@ impl MethodDefinition {
     }
 
     #[must_use]
-    pub fn receiver(&self) -> &Option<NameId> {
+    pub fn receiver(&self) -> &Option<Receiver> {
         &self.receiver
     }
 
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -9,7 +9,7 @@ use crate::model::{
         Declaration, GlobalVariableDeclaration, InstanceVariableDeclaration, MethodDeclaration, ModuleDeclaration,
         Namespace, SingletonClassDeclaration,
     },
-    definitions::{Definition, Mixin},
+    definitions::{Definition, Mixin, Receiver},
     graph::{CLASS_ID, Graph, MODULE_ID, OBJECT_ID},
     identity_maps::{IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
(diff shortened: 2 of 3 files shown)


rust/rubydex/src/model/definitions.rs
```diff
@@ -703,7 +703,16 @@ pub struct MethodDefinition {
...
+
+/// The receiver of a singleton method definition.
+#[derive(Debug, Clone, Copy)]
+pub enum Receiver {
+    /// `def self.foo` - receiver is the enclosing definition (class, module, or DSL)
+    SelfReceiver(DefinitionId),
+    /// `def Foo.bar` - receiver is an explicit constant that needs resolution
+    ConstantReceiver(NameId),
 }
 assert_mem_size!(MethodDefinition, 112);
```
Morriar: Let's keep this line attached to `MethodDefinition`

Maybe let's add one for `Receiver`

rust/rubydex/src/model/definitions.rs
```diff
@@ -703,7 +703,16 @@ pub struct MethodDefinition {
     lexical_nesting_id: Option<DefinitionId>,
     parameters: Vec<Parameter>,
     visibility: Visibility,
-    receiver: Option<NameId>,
+    receiver: Option<Receiver>,
+}
+
+/// The receiver of a singleton method definition.
+#[derive(Debug, Clone, Copy)]
```
vinistock: Does this really need clone and copy?

rust/rubydex/src/diagnostic.rs
```diff
@@ -105,4 +105,6 @@ rules! {
     TopLevelMixinSelf;
 
     // Resolution
+    UnresolvedMethodReceiver;
+    OrphanInstanceVariable;
```
vinistock: Do we need these two rules? I feel like they are both the same problem: an unresolved constant.

For example:

```ruby
module Foo
  class Bar::Baz
    def initialize
      @var = 1
    end
  end
end
```

If we fail to resolve the `Bar` reference, we also can't correctly fully qualify the `initialize` method or `@var` instance variable. However, the root cause is only one problem (one diagnostic): the unresolved `Bar` reference. We wouldn't want to push diagnostics for everything defined inside, but rather only for the root cause.

The ivar case is where this is more obvious to me: you can't have an orphaned instance variable in Ruby. They always belong to something. If there's an orphaned instance variable, then it means we failed to resolve a constant reference, which is the root cause.

rust/rubydex/src/resolution.rs
```diff
@@ -272,23 +273,48 @@ impl<'a> Resolver<'a> {
...
-                            NameRef::Resolved(resolved) => *resolved.declaration_id(),
-                            NameRef::Unresolved(_) => {
-                                // Error diagnostic: if we couldn't resolve the constant being used, then we don't know
-                                // where this method is being defined. For example:
-                                //
-                                // def Foo.bar; end
-                                //
-                                // where Foo is undefined
+                    let uri_id = *method_definition.uri_id();
+                    let offset = method_definition.offset().clone();
```
vinistock: Why do we need to clone the offset here?

rust/rubydex/src/resolution.rs
```diff
@@ -331,6 +357,8 @@ impl<'a> Resolver<'a> {
                 }
                 Definition::InstanceVariable(var) => {
                     let str_id = *var.str_id();
+                    let uri_id = *var.uri_id();
+                    let offset = var.offset().clone();
```
vinistock: Same question.

rust/rubydex/src/diagnostic.rs
st0012: > you can't have an orphaned instance variable in Ruby.

Yeah that's fair.
There is one case where you sort of can have orphan ivars tho: if they're assigned in frozen classes, like `Integer`

```rb
class Integer
  def foo
    @foo = 1 # 'Integer#foo': can't modify frozen Integer: 1 (FrozenError)
  end
end

a = 1
puts a.foo
```

But in that case, a rule like "mutating frozen instances" will be more helpful. So either way this can be removed.

rust/rubydex/src/resolution.rs
st0012: `method_definition.offset()` returns `&Offset` but `Diagnostic::new` expects a `Offset` value. So from my understanding we have to `clone` it to be able to pass it?

Morriar: Yeah the diagnostic needs to take ownership of the `offset` because it has an independent lifetime from the node. 

Though, it may be clearer to just clone in the Diagnostic call, see lower

rust/rubydex/src/resolution.rs
```diff
@@ -272,23 +273,48 @@ impl<'a> Resolver<'a> {
...
+                            let Some(&owner_decl_id) = self.graph.definition_id_to_declaration_id(*def_id) else {
+                                // The enclosing class/module couldn't be resolved (e.g., `class Foo::Bar` where
+                                // `Foo` is undefined). The method is orphaned as a consequence.
+                                self.graph.add_definition_diagnostic(
+                                    &uri_id,
+                                    id,
+                                    Diagnostic::new(
+                                        Rule::UnresolvedMethodReceiver,
+                                        uri_id,
+                                        offset,
```
Morriar: That way it makes it clear we clone it for the diagnostic:
```suggestion
                                        method_definition.offset().clone(),
```

rust/rubydex/src/resolution.rs
```diff
@@ -272,23 +273,48 @@ impl<'a> Resolver<'a> {
...
+                            let receiver_decl_id = match self.graph.names().get(name_id).unwrap() {
+                                NameRef::Resolved(resolved) => *resolved.declaration_id(),
+                                NameRef::Unresolved(_) => {
+                                    self.graph.add_definition_diagnostic(
+                                        &uri_id,
+                                        id,
+                                        Diagnostic::new(
+                                            Rule::UnresolvedMethodReceiver,
+                                            uri_id,
+                                            offset,
```
Morriar:
```suggestion
                                            method_definition.offset().clone(),
```

rust/rubydex/src/model/definitions.rs
```diff
@@ -712,10 +712,22 @@ pub struct MethodDefinition {
...
-    receiver: Option<NameId>,
+    receiver: Option<Receiver>,
 }
+
 assert_mem_size!(MethodDefinition, 112);
 
+/// The receiver of a singleton method definition.
+#[derive(Debug)]
+pub enum Receiver {
+    /// `def self.foo` - receiver is the enclosing definition (class, module, or DSL)
```
vinistock:
```suggestion
    /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
```

rust/rubydex/src/resolution.rs
```diff
@@ -272,23 +273,47 @@ impl<'a> Resolver<'a> {
...
-                                // def Foo.bar; end
-                                //
-                                // where Foo is undefined
+                    let uri_id = *method_definition.uri_id();
+                    let owner_id = match method_definition.receiver() {
+                        Some(Receiver::SelfReceiver(def_id)) => {
+                            let Some(&owner_decl_id) = self.graph.definition_id_to_declaration_id(*def_id) else {
+                                // The enclosing class/module couldn't be resolved (e.g., `class Foo::Bar` where
+                                // `Foo` is undefined). The method is orphaned as a consequence.
+                                self.graph.add_definition_diagnostic(
```
vinistock: I don't think this is a diagnostic that should exist. The underlying problem is: `UnresolvedConstantReference`. The fact that this impacts a method definition is a consequence of the root cause.

I would still vote for us to decouple this pull request from resolution diagnostics, which is probably easier to design by taking a more complete look at all problems that can occur in #354.

rust/rubydex/src/model/definitions.rs
vinistock:
```suggestion
    /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
```

vinistock (commented):
I think this PR is good, but let's not include resolution diagnostics as a part of it yet.

We need to think of resolution diagnostics in a complete way to avoid over notifying the user with several errors that are in fact just one underlying problem.

rust/rubydex/src/resolution.rs
st0012: Removed

rust/rubydex/src/resolution.rs
```diff
@@ -346,25 +349,29 @@ impl<'a> Resolver<'a> {
...
-                                    continue;
+                            if let Some(receiver) = method.receiver() {
+                                let receiver_decl_id = match receiver {
+                                    Receiver::SelfReceiver(def_id) => *self
+                                        .graph
+                                        .definition_id_to_declaration_id(*def_id)
+                                        .expect("SelfReceiver definition should have a declaration"),
+                                    Receiver::ConstantReceiver(name_id) => {
+                                        let Some(NameRef::Resolved(resolved)) = self.graph.names().get(name_id) else {
+                                            // TODO: add diagnostic for unresolved receiver
```
vinistock: ✂️

vinistock (approved):

PR #575, opened 2026-02-09
Search for closest namespace when connecting definition to declaration

vinistock:
Finding the declaration for a definition that was nested inside something that isn't a namespace was crashing. For example, this scenario:

```ruby
class Foo
  def bar
    # This ivar is nested under the `bar` method. To find its declaration, we need
    # to go all the way to `Foo`
    @something = 1
  end
end
```

The fix is to keep searching for the closest namespace and only default to `Object` if the definition is not nested under anything.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -191,60 +191,42 @@ impl Graph {
             Definition::ConstantAlias(it) => {
                 return self.name_id_to_declaration_id(*it.name_id());
             }
-            Definition::GlobalVariable(it) => {
-                let nesting_definition = it
-                    .lexical_nesting_id()
-                    .and_then(|id| self.definitions().get(&id).unwrap().name_id());
-                (nesting_definition, it.str_id())
-            }
-            Definition::GlobalVariableAlias(it) => {
-                let nesting_definition = it
-                    .lexical_nesting_id()
-                    .and_then(|id| self.definitions().get(&id).unwrap().name_id());
-                (nesting_definition, it.new_name_str_id())
-            }
-            Definition::InstanceVariable(it) => {
-                let nesting_definition = it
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -1502,4 +1499,73 @@ mod tests {
         assert!(context.graph().get("Foo::<Foo>").is_none());
         assert!(context.graph().get("Foo::<Foo>::<<Foo>>").is_none());
     }
+
+    #[test]
+    fn indexing_the_same_document_twice() {
```
alexcrocha, line 1504: We're not asserting anything in this test. Is your intention to make sure we can index the same document twice and not panic?

vinistock: Indeed, I was initially just making sure we don't panic, but I added assertions for the number of entities we end up with after the double indexing (which should be identical).

rust/rubydex/src/model/graph.rs
```diff
@@ -253,13 +235,28 @@ impl Graph {
...
-            .unwrap()
-            .as_namespace()
-            .unwrap()
+            .get(nesting_declaration_id)?
+            .as_namespace()?
             .member(member_str_id)
     }
 
+    /// Finds the closest namespace name ID to connect a definition to its declaration
+    fn find_enclosing_namespace_name_id(&self, starting_id: Option<&DefinitionId>) -> Option<&NameId> {
```
Morriar, line 244: Naming between `lexical_nesting` and `namespace` is getting confusing, we may want to revisit it at some point

vinistock: Yes, indeed. What indexing collects isn't really the lexical nesting since methods don't produce lexical scopes. We'll need to review the nomenclature.

PR #576, opened 2026-02-09
Improve index workspace path discovery

vinistock:
This PR includes a few improvements to how we list paths for indexing a workspace:

1. Excluding `/extensions/` directories. These are inserted by RubyGems as absolute paths in the gem's `require_paths`. These directories get automatically inserted for any native extension even if the directory doesn't exist. Additionally, they never contain any Ruby files
2. Ensuring paths do not contain duplicates. When indexing a gem's source code, its paths will appear both in the `workspace_path` and as part of `Bundler.locked_gems` because the gem depends on itself
3. Excluding common directories that are not relevant for Ruby projects, like `.git`, `node_modules`, `.bundle` or `tmp`. We know from the Ruby LSP that globbing these directories can consume a significant amount of time, so we can exclude them and let the Rust side descend only for relevant folders

### DIFF
--- a/lib/rubydex/graph.rb
+++ b/lib/rubydex/graph.rb
@@ -5,6 +5,14 @@ module Rubydex
   #
   # Note: this class is partially defined in C to integrate with the Rust backend
   class Graph
+    IGNORED_DIRECTORIES = [
+      ".bundle",
+      ".git",
+      ".github",
+      "node_modules",
+      "tmp",
+    ].freeze
+
     #: (?workspace_path: String) -> void
     def initialize(workspace_path: Dir.pwd)
       @workspace_path = workspace_path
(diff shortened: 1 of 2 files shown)


test/graph_test.rb
```diff
@@ -414,6 +414,41 @@ def test_require_paths_with_invalid_arguments
...
+      context.write!("node_modules/pkg/index.js", "")
+      context.write!("top_level.rb", "class TopLevel; end")
+
+      graph = Rubydex::Graph.new(workspace_path: context.absolute_path)
+      paths = graph.workspace_paths
+
+      # Includes workspace directories
+      assert_includes(paths, context.absolute_path_to("lib"))
+      assert_includes(paths, context.absolute_path_to("app"))
+
```
alexcrocha, line 431: Just missing the top level `.rb` file
```suggestion

# Includes top level .rb files
assert_includes(paths, context.absolute_path_to("top_level.rb"))
```

vinistock: Good catch, I forgot that assertion.

lib/rubydex/graph.rb
```diff
@@ -13,32 +21,52 @@ def initialize(workspace_path: Dir.pwd)
...
+        if File.directory?(full_path)
+          paths << full_path unless IGNORED_DIRECTORIES.include?(entry)
+        elsif File.extname(entry) == ".rb"
+          paths << full_path
+        end
+      end
+
+      add_workspace_dependency_paths(paths)
+      paths.uniq!
+      paths
```
Morriar, line 46: It may be faster to move this to the Rust side at some point and then append the dependency path from Ruby

vinistock: Probably, but I'm not too worried since the descending of directories is already happening on the Rust side. This is only looking at gem require paths + the first level of the workspace.

PR #577, opened 2026-02-09
Return array of IO errors from indexing

vinistock:
This PR changes the return of `Graph#index_all` so that we return an array of error messages rather than raising.

Raising is weird for this API because indexing didn't actually fail completely. It just encountered some IO errors (like file permission problems), but likely succeeded in indexing everything else in the codebase.

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -9,7 +9,6 @@
 #include "utils.h"
 
 static VALUE cGraph;
-static VALUE eIndexingError;
 
 // Free function for the custom Graph allocator. We always have to call into Rust to free data allocated by it
 static void graph_free(void *ptr) {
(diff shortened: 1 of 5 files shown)


ext/rubydex/graph.c
```diff
@@ -27,35 +26,39 @@ static VALUE rdxr_graph_alloc(VALUE klass) {
     return TypedData_Wrap_Struct(klass, &graph_type, graph);
 }
 
-// Graph#index_all: (Array[String] file_paths) -> nil
-// Raises IndexingError if anything failed during indexing
+// Graph#index_all: (Array[String] file_paths) -> Array[String]
```
Morriar, line 29: It would be nicer to return an array of error objects than bare strings.

vinistock: Do you mean instantiating `StandardError` descendants despite not raising?

Morriar: Not necessarily. Take a look at how `Prism.parse` returns a `Result` object that contains an array of `Prims::ParseError`.

PR #579, opened 2026-02-09
Expose `Graph#delete_document` to Ruby API

vinistock:
Another piece that we're missing in the Ruby API for the LSP: this PR exposes `Graph#delete_document`, which not only removes data related to the document, but also removes the document itself. This is used when handling file events (e.g.: switching branches).

### DIFF
--- a/test/graph_test.rb
+++ b/test/graph_test.rb
@@ -414,7 +414,8 @@ class GraphTest < Minitest::Test
       assert_equal(2, graph.documents.count)
       foo = graph["Foo"]
 
-      graph.delete_document(context.uri_to("foo.rb"))
+      deleted = graph.delete_document(context.uri_to("foo.rb"))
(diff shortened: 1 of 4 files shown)


test/graph_test.rb
```diff
@@ -402,6 +402,42 @@ def test_require_paths
...
+      assert_nil(graph["Foo"])
+
+      assert_equal(1, graph.documents.count)
+      assert_equal("Bar", graph.documents.first.definitions.first.name)
+    end
+  end
+
+  def test_delete_uri_with_non_existing_uri
+    graph = Rubydex::Graph.new
+    graph.delete_document("file:///non_existing.rb")
```
Morriar: I wonder if we should return the document if it existed or nil if it didn't, a bit like `Array#delete`?

vinistock: Yes! I like that.

PR #580, opened 2026-02-10
Fix resolve_require_path to respect load path ordering

alexcrocha:
This PR addresses post-merge review feedback from https://github.com/Shopify/rubydex/pull/555#discussion_r2783119531 (non-deterministic behaviour when a require path matches files under multiple load paths).

Ruby's `require` searches `$LOAD_PATH` in order and loads the first match. `resolve_require_path` (used for go-to-definition) didn't respect this and was returning whichever thread found a match first, making the result depend on thread scheduling rather than load path order. 

Instead of iterating all documents and checking if their path matches the require path, `resolve_require_path` now goes the other way. It iterates the load path, constructs the expected file URI from each entry, and does a direct hash map lookup via `UriId`. First match wins, which gives us correct load path priority and is O(load_path) instead of O(documents).

### DIFF
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -1,7 +1,8 @@
 use std::collections::HashSet;
 use std::path::PathBuf;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread;
+
+use url::Url;
 
 use crate::model::graph::Graph;
 use crate::model::ids::{DeclarationId, UriId};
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/query.rs
```diff
@@ -62,59 +61,53 @@ fn match_score(query: &str, target: &str) -> usize {
...
-                        }
-                    }
-                    None
-                })
-            })
-            .collect();
-
-        handles.into_iter().find_map(|handle| handle.join().unwrap())
-    })
+    collect_require_paths(graph, load_paths, |path| path == normalized)
```
vinistock: Does this mean we're now collecting all require paths instead of halting the search once we find the first match?

It might be a bit slow.

alexcrocha: We are only collecting paths that match the require path, but you're right that we're now iterating all documents to find them. I am looking into how we can avoid doing that while respecting the load path order

alexcrocha: So, instead of iterating all documents and checking if their path matches the require path, why don't we go the other way? Iterate the load paths, construct the expected file path from each, and check if a document exists with that URI. First match wins.

Let me know your thoughts on this approach.

vinistock: I think we're discussing slightly different things. Based on my understanding, you're asking if we should build a full URI from the require path we're searching for and then iterate documents to find it. I'm not sure this will work tbh and I think it's more straight forward to reason about the current approach.

My main concern is about having to iterate over all documents. The previous implementation halts the search the moment we find the match, without having to go through every single document.

alexcrocha: > you're asking if we should build a full URI from the require path we're searching for and then iterate documents to find it

Not exactly.  We can avoid iterating through the documents by iterating through the load path entries instead, constructing the URI, and doing a hash lookup instead.

I am pushing a new commit demonstrating this new approach (and will squash the original one once approved)

PR #582, opened 2026-02-10
Trigger release on tag pushes

vinistock:
I noticed that our `cibuildgem` action is not creating tags or GitHub releases. In addition to not creating those, this is the only gem that's not released with the new approach of pushing tags.

This PR standardizes our action to follow the same approach as our other gems:

- Pushing a `v1.2.3` tag cuts a stable release
- Pushing a `v1.2.3-beta1` tag cuts a pre-release
- Trigger the action manually performs a dry-run

And if a release is made, the GH release is auto generated.

### DIFF
--- a/.github/workflows/cibuildgem.yaml
+++ b/.github/workflows/cibuildgem.yaml
@@ -1,12 +1,13 @@
 name: "Package and release gems with precompiled binaries"
 on:
+  # Manually trigger a dry-run that doesn't publish to RubyGems
   workflow_dispatch:
-    inputs:
-      release:
-        description: "If the whole build passes on all platforms, release the gems on RubyGems.org"
-        required: false
-        type: boolean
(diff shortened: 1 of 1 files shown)


.github/workflows/cibuildgem.yaml
```diff
@@ -1,12 +1,11 @@
 name: "Package and release gems with precompiled binaries"
 on:
   workflow_dispatch:
```
alexcrocha, line 4: Since `workflow_dispatch` is empty now, it was not obvious to understand its purpose as the dry run trigger without reading the rest of the config first. 

Should we add a comment here like `# Manual trigger for dry run`? That will probably help me in a couple of months 😄

vinistock: Good call.

PR #585, opened 2026-02-13
Add completion for expressions

vinistock:
First step for #59

This PR adds a completion API with support for expression completion. That means, completion when there's nothing before the cursor.

The idea of the implementation is to have an enum to describe all of the possible contexts we may see completion being triggered since they all collect slightly different declarations. I tried to document all scenarios, despite having only implemented expression.

**Notes**

The expression completion is currently missing keywords. I'll propose something for those in a different PR.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -320,6 +320,31 @@ impl Graph {
         name_id
     }
 
+    /// Searches for the initial attached object for an arbitrarily nested singleton class.
+    /// Walks up the owner chain until finding a non-singleton namespace.
+    ///
+    /// # Example
+    /// For `Foo::<Foo>::<<Foo>>`, returns `Foo`
+    ///
+    /// # Panics
+    ///
+    /// Panics if we attached a singleton class to something that isn't a namespace
+    #[must_use]
+    pub fn attached_object<'a>(&'a self, maybe_singleton: &'a Namespace) -> &'a Namespace {
+        let mut attached_object = maybe_singleton;
+
+        while matches!(attached_object, Namespace::SingletonClass(_)) {
+            attached_object = self
+                .declarations
+                .get(attached_object.owner_id())
+                .unwrap()
+                .as_namespace()
+                .unwrap();
+        }
+
+        attached_object
+    }
+
     #[must_use]
     pub fn get(&self, name: &str) -> Option<Vec<&Definition>> {
         let declaration_id = DeclarationId::from(name);
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -1,11 +1,15 @@
 use std::collections::HashSet;
+use std::error::Error;
 use std::path::PathBuf;
 use std::thread;
 
 use url::Url;
 
-use crate::model::graph::Graph;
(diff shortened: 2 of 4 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -320,6 +320,27 @@ impl Graph {
         name_id
     }
 
+    /// Searches for the initial attached object for an arbitrarily nested singleton class
+    ///
+    /// # Panics
+    ///
+    /// Only panics if we attached a singleton class to something that isn't a namespace
+    #[must_use]
+    pub fn attached_object<'a>(&'a self, maybe_singleton: &'a Namespace) -> &'a Namespace {
```
Morriar, line 333: Let's add a doc comment here to clarify the purpose. Also, should we validate that we eventually find a non-singleton class?

```suggestion
    /// Searches for the initial attached object for an arbitrarily nested singleton class.
    /// Walks up the owner chain until finding a non-singleton namespace.
    ///
    /// # Example
    /// For `Foo::<Foo>::<<Foo>>`, returns `Foo`
    ///
    /// # Panics
    ///
    /// Panics if we attached a singleton class to something that isn't a namespace
    /// or if the owner chain is broken.
    #[must_use]
    pub fn attached_object<'a>(&'a self, maybe_singleton: &'a Namespace) -> &'a Namespace {
```

My only concern is - what happens if we have a broken chain where the loop never exits? Should we add a safety check or assert that `attached_object` is actually different from the input?

rust/rubydex/src/query.rs
```diff
@@ -153,13 +157,180 @@ pub fn require_paths(graph: &Graph, load_paths: &[PathBuf]) -> Vec<String> {
...
+/// - Relative require path completion collects all require paths accessible from the directory of the current file
+///
+/// # Panics
+///
+/// Will panic if we incorrectly inserted non namespace declarations as ancestors
+///
+/// # Errors
+///
+/// Will error if the given `self_name_id` does not point to a namespace declaration
+pub fn completion_candidates<'a>(
```
Morriar, line 210: I wonder if we could make these errors more structured?

Right now we're using string-based errors which makes it harder to handle specific cases. Any thoughts on creating an enum for completion errors?

```rust
pub enum CompletionError {
    NameNotFound(NameId),
    UnresolvedName(NameId),
    InvalidDeclarationType,
}
```

This would make error handling more explicit and easier to test. We can do this later if needed 👍

vinistock: Added another task in #59. I agree, we should use specific errors.

rust/rubydex/src/model/graph.rs
vinistock: I added the docs you suggested. Regarding the broken chain, I'm trying to think if we can add some debug asserts to verify that owners are always namespaces and ultimately connected to `Object`.

vinistock: I think we can have a sanity check about the resolution data, so that we can avoid paying the price of checking in the completion loop. I'll prototype this in a different PR.

vinistock: https://github.com/Shopify/rubydex/pull/592

PR #586, opened 2026-02-17
Add MCP server for AI-assisted Ruby code intelligence

st0012:
## Summary

Adds an MCP (Model Context Protocol) server to rubydex as a separate workspace crate (`rubydex-mcp`), allowing AI assistants like Claude Code to semantically query indexed Ruby programs. This is a prototype/experiment exploring how code intelligence can improve AI agent effectiveness compared to plain grep-based approaches.

**What it does:** When launched with `rubydex_mcp <path>`, the binary indexes the Ruby project and serves semantic queries over JSON-RPC stdio. AI agents can then use these tools instead of grep for structural Ruby queries.

### 6 MCP tools

| Tool | What it does |
|------|-------------|
| `search_declarations` | Fuzzy search for classes, modules, methods, constants |
| `get_declaration` | Full details by fully qualified name with docs, ancestors, members |
| `get_descendants` | What classes/modules inherit from or include this one |
| `find_constant_references` | All precise, resolved constant references across the codebase |
| `get_file_declarations` | Structural overview of a file's declarations |
| `codebase_stats` | Indexed codebase statistics |

### Key design decisions

- **Separate crate** — `rubydex-mcp` is its own workspace crate to keep MCP dependencies (42 unique crates: rmcp, tokio, schemars, etc.) out of the rubydex library consumed by rubydex-sys
- **Read-only at runtime** — the server owns the `Graph` directly (no `Arc`, no locks); all tools are pure reads
- **Precise references only** — `find_constant_references` returns only resolved constant references, not fuzzy name matches; agents use Grep for method call sites
- **Structured JSON errors** — all error responses include machine-readable error type, message, and recovery suggestion
- **Result limits** — `find_constant_references` defaults to 50 results (max 200) with a `truncated` flag to prevent context window blowout
- **No source embedding** — tools return file paths and line numbers; agents use their built-in Read tool to fetch source when needed

### Also in this PR

- `Location` struct added to `rubydex::offset` for reusable offset-to-line/col conversion
- `.mcp.json` committed to repo root for zero-config setup
- `dev up` step to auto-install the MCP binary
- Benchmarking skill (`.claude/skills/rubydex-mcp-benchmark/`) for A/B comparing MCP-assisted vs grep-based development

### Early benchmark results (IRB project)

| Scenario | Metric | With MCP | Without MCP |
|----------|--------|----------|-------------|
| Rename (IRB::Locale, 16 refs) | Tokens | 22k | 31.8k |
| Extract module (IRB::Context) | Tokens | 33.2k | 63.0k |

## Usage

The `.mcp.json` is committed to the repo root. Install the binary:

```bash
cargo install --path rust/rubydex-mcp
```

Or just run `dev up` — it installs automatically.

To use rubydex MCP in other Ruby projects, add a `.mcp.json` to that project's root:

```json
{
  "mcpServers": {
    "rubydex": {
      "command": "${HOME}/.cargo/bin/rubydex_mcp"
    }
  }
}
```

Then start Claude Code from that project directory. Verify with `/mcp` in the session.

### DIFF
--- a/.claude/skills/rubydex-mcp-benchmark/SKILL.md
+++ b/.claude/skills/rubydex-mcp-benchmark/SKILL.md
@@ -1,0 +1,124 @@
+---
+name: rubydex-mcp-benchmark
+description: Use when benchmarking rubydex MCP against plain Grep/Glob for Ruby code tasks. Guides A/B comparison setup, candidate selection, and metrics collection.
+---
+
+# Rubydex MCP Benchmark
+
+Compare rubydex MCP-assisted development vs standard Grep/Glob for Ruby code tasks.
+
+## Prerequisites
+
+Ensure the `rubydex_mcp` binary is installed:
+
+```bash
+# From the rubydex repo root
+cargo install --path rust/rubydex-mcp
+```
+
+This places `rubydex_mcp` in `~/.cargo/bin/`. Verify: `rubydex_mcp --help`
+
+## Step 1: Find a Good Benchmark Candidate
+
+Use rubydex MCP tools to find a class or module suitable for testing:
+
+1. Run `codebase_stats` to get an overview of the target project
+2. Run `search_declarations` with kind "Class" or "Module" to list candidates
+3. For each candidate, run `find_constant_references` and count references
+4. Pick a target with **10-30 references across 4+ files** that has test coverage
+
+Good candidates are:
+- Self-contained (not deeply entangled with the whole codebase)
+- Referenced in both lib and test directories
+- Have a dedicated test file
+
+## Step 2: Set Up the Experiment
+
+Create two git worktrees for the target Ruby project:
+
+```bash
+cd /path/to/target-project
+git worktree add ../project-with-mcp -b bench-with-mcp HEAD
+git worktree add ../project-without-mcp -b bench-without-mcp HEAD
+```
+
+Add `.mcp.json` to the with-mcp worktree only:
+
+```bash
+cat > ../project-with-mcp/.mcp.json << 'EOF'
+{
+  "mcpServers": {
+    "rubydex": {
+      "command": "${HOME}/.cargo/bin/rubydex_mcp"
+    }
+  }
+}
+EOF
+```
+
+## Step 3: Generate Commands for the Developer
+
+Pick a testing prompt based on the scenario. Keep prompts **vague and natural** - like a lazy developer would type. Don't specify full namespaces, file paths, or implementation details. Let the agent figure it out. This tests the MCP advantage fairly since the MCP agent can resolve ambiguity semantically while the non-MCP agent has to grep around.
+
+**Guidelines for writing prompts:**
+- Use short class names, not fully qualified (`Context` not `IRB::Context`)
+- Don't specify file locations or paths
+- Don't tell the agent how to do it, just what to do
+- Always end with how to run tests
+
+**Examples:**
+
+Rename:
+```
+Rename Locale to LocaleManager in IRB. All existing tests must continue to pass. Run tests with: bundle exec rake test
+```
+
+Extract module:
+```
+Extract all history-related methods from Context into a new module Context::HistoryManager. All existing tests must continue to pass. Run tests with: bundle exec rake test
+```
+
+Then generate two ready-to-copy commands and present them to the developer:
+
+**With MCP** (run in terminal 1):
+```bash
+cd /path/to/project-with-mcp && claude "<testing prompt>" --dangerously-skip-permissions
+```
+
+**Without MCP** (run in terminal 2):
+```bash
+cd /path/to/project-without-mcp && claude "<testing prompt>" --dangerously-skip-permissions
+```
+
+Tell the developer to copy and run these commands in separate terminals simultaneously.
+
+## Step 4: Compare Results
+
+After both sessions complete, compare:
+
+| Metric | With MCP | Without MCP |
+|--------|----------|-------------|
+| Duration | (from session) | (from session) |
+| Tokens | (from session stats) | (from session stats) |
+| Files changed | `git diff --stat` in with-mcp worktree | `git diff --stat` in without-mcp worktree |
+| Tests | pass/fail | pass/fail |
+
+Run `git diff --stat` in each worktree to compare the changes made.
+
+## Step 5: Cleanup
+
+```bash
+cd /path/to/target-project
+git worktree remove ../project-with-mcp
+git worktree remove ../project-without-mcp
+git branch -D bench-with-mcp bench-without-mcp
+```
+
+## Reference: Our Benchmark Results
+
+| Scenario | Metric | With MCP | Without MCP | MCP Advantage |
+|----------|--------|----------|-------------|---------------|
+| Rename (IRB::Locale, 16 refs) | Tokens | 22k | 31.8k | 30% fewer |
+| Rename (IRB::Locale, 16 refs) | Time | 2m 9s | 2m 44s | 25% faster |
+| Extract module (IRB::Context) | Tokens | 33.2k | 63.0k | 47% fewer |
+| Extract module (IRB::Context) | Time | 1m 49s | 2m 39s | 31% faster |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,10 +57,12 @@ When necessary, commands can be executed for the Ruby code.
 
 ## Rust workspace
 
-The Rust workspace under the `rust` directory contains two crates:
+The Rust workspace under the `rust` directory contains three crates:
 
 - `rubydex`: this crate implements the entire indexing and static analysis logic. The implementation aims to be optimized
 to achieve maximum performance in super large codebases while maintaining memory usage to a minimum
+- `rubydex-mcp`: an MCP (Model Context Protocol) server that exposes rubydex's code intelligence as tools for AI
+assistants. Communicates over stdio using JSON-RPC
 - `rubydex-sys`: this crate provides bindings for C, so that the logic from `rubydex` can be called through FFI
 
 The workspace's goal is to provide all indexing and static analysis capabilities to power tools such as language servers,
@@ -85,9 +87,11 @@ When necessary, commands can be executed for the Rust code.
 - `cargo run -- <directory> --stats`: runs the indexer with detailed performance breakdown
 - `cargo run -- <directory> --stop-after <stage>`: stops after the specified stage (Listing, Indexing, or Resolution)
 - `cargo run -- <directory> --visualize`: generates a DOT visualization of the graph
-- `cargo test`: runs Rust tests
+- `cargo test`: runs Rust tests (all workspace crates)
+- `cargo test -p rubydex-mcp`: runs MCP server tests only
 - `cargo test test_name`: runs a specific tests example
 - `cargo fmt`: auto formats the Rust code
 - `cargo clippy`: lints the Rust code
+- `cargo install --path rust/rubydex-mcp`: installs the MCP server binary
 - `bundle exec rake lint_rust`: lints the Rust code
 - `bundle exec rake format_rust`: auto formats the Rust code
--- a/README.md
+++ b/README.md
@@ -77,6 +77,49 @@ diagnostic.message
 diagnostic.location
 ```
 
+## MCP Server (Experimental)
+
+Rubydex can run as an MCP (Model Context Protocol) server, enabling AI assistants
+like Claude to semantically query your Ruby codebase.
+
+### Setup
+
+1. Install the binary:
+   ```bash
+   cargo install --path rust/rubydex-mcp
+   ```
+
+2. Add rubydex to the Ruby project you want to index:
+   ```bash
+   claude mcp add --scope project rubydex "\${HOME}/.cargo/bin/rubydex_mcp"
+   ```
+
+   Or manually create a `.mcp.json` in the project root:
+   ```json
+   {
+     "mcpServers": {
+       "rubydex": {
+         "command": "${HOME}/.cargo/bin/rubydex_mcp"
+       }
+     }
+   }
+   ```
+
+3. Start Claude Code from that project directory. The MCP server indexes
+   the project at startup and provides semantic code intelligence tools.
+   Verify with `/mcp` in the session.
+
+### Available MCP Tools
+
+| Tool | Description |
+|------|-------------|
+| `search_declarations` | Fuzzy search for classes, modules, methods, constants |
+| `get_declaration` | Full details by fully qualified name with docs, ancestors, members |
+| `get_descendants` | What classes/modules inherit from or include this one |
+| `find_constant_references` | All precise, resolved constant references across the codebase |
+| `get_file_declarations` | List declarations defined in a specific file |
+| `codebase_stats` | High-level statistics about the indexed codebase |
+
 ## Contributing
 
 See [the contributing documentation](CONTRIBUTING.md).
--- a/rust/rubydex-mcp/src/main.rs
+++ b/rust/rubydex-mcp/src/main.rs
@@ -1,0 +1,48 @@
+use clap::Parser;
+
+mod server;
+mod tools;
+
+#[derive(Parser, Debug)]
+#[command(
+    name = "rubydex_mcp",
+    about = "Rubydex MCP server for AI-assisted Ruby code intelligence",
+    version
+)]
+struct Args {
+    #[arg(value_name = "PATH", default_value = ".")]
+    path: String,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    let root = match std::fs::canonicalize(&args.path) {
+        Ok(p) => p
+            .into_os_string()
+            .into_string()
+            .expect("Project path is not valid UTF-8"),
+        Err(e) => {
+            eprintln!("Warning: failed to canonicalize '{}': {e}", args.path);
+            args.path
+        }
+    };
+
+    // Create the server and start indexing in the background.
+    let server = server::RubydexServer::new(root.clone());
+    server.spawn_indexer(root);
+
+    // Serve MCP over stdio immediately while indexing runs.
+    // We need to do this because Claude Code's default MCP server timeout is 30 seconds,
+    // And in big codebases it's possible to exceed that and Claude Code would just consider
+    // the server fail to connect.
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .expect("Failed to build tokio runtime");
+
+    if let Err(e) = rt.block_on(server.serve()) {
+        eprintln!("MCP server error: {e}");
+        std::process::exit(1);
+    }
+}
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -1,0 +1,533 @@
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, RwLock};
+
+use crate::tools::{
+    FindConstantReferencesParams, GetDeclarationParams, GetDescendantsParams, GetFileDeclarationsParams,
+    SearchDeclarationsParams,
+};
+use rmcp::{
+    ServerHandler,
+    handler::server::{router::tool::ToolRouter, wrapper::Parameters},
+    model::{ServerCapabilities, ServerInfo},
+    tool, tool_handler, tool_router,
+    transport::io::stdio,
+};
+use rubydex::model::ids::{DeclarationId, UriId};
+use rubydex::model::{
+    declaration::{Ancestor, Ancestors},
+    graph::Graph,
+};
+use url::Url;
+
+struct ServerState {
+    graph: Option<Graph>,
+    error: Option<String>,
+}
+
+pub struct RubydexServer {
+    state: Arc<RwLock<ServerState>>,
+    root_path: PathBuf,
+    tool_router: ToolRouter<Self>,
+}
+
+impl RubydexServer {
+    pub fn new(root: String) -> Self {
+        Self {
+            state: Arc::new(RwLock::new(ServerState {
+                graph: None,
+                error: None,
+            })),
+            root_path: PathBuf::from(root),
+            tool_router: Self::tool_router(),
+        }
+    }
+
+    /// Spawns a background thread that indexes the codebase and marks the server as ready.
+    pub fn spawn_indexer(&self, path: String) {
+        let state = Arc::clone(&self.state);
+        std::thread::spawn(move || {
+            let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+                let (file_paths, errors) = rubydex::listing::collect_file_paths(vec![path]);
+                for error in &errors {
+                    eprintln!("Listing error: {error}");
+                }
+
+                let mut graph = Graph::new();
+                let errors = rubydex::indexing::index_files(&mut graph, file_paths);
+                for error in &errors {
+                    eprintln!("Indexing error: {error}");
+                }
+
+                let mut resolver = rubydex::resolution::Resolver::new(&mut graph);
+                resolver.resolve_all();
+
+                eprintln!(
+                    "Rubydex indexed {} files, {} declarations",
+                    graph.documents().len(),
+                    graph.declarations().len()
+                );
+                graph
+            }));
+
+            let mut state = state.write().expect("state lock poisoned");
+            match result {
+                Ok(graph) => {
+                    state.graph = Some(graph);
+                }
+                Err(panic) => {
+                    let msg = panic
+                        .downcast_ref::<String>()
+                        .map(String::as_str)
+                        .or_else(|| panic.downcast_ref::<&str>().copied())
+                        .unwrap_or("unknown error");
+                    eprintln!("Rubydex indexing failed: {msg}");
+                    state.error = Some(msg.to_string());
+                }
+            }
+        });
+    }
+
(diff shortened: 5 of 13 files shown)


.claude/skills/rubydex-mcp-benchmark/SKILL.md
```diff
@@ -0,0 +1,125 @@
+---
```
st0012, line 1: This skill helps benchmarking process quite a bit. It's not needed to use the MCP server, but IMO we should keep this while we're still evaluating what requests to keep/implement.

rust/rubydex/Cargo.toml
```diff
@@ -30,13 +30,19 @@ tempfile = { version = "3.0", optional = true }
 crossbeam-deque = "0.8"
 crossbeam-utils = "0.8"
 crossbeam-channel = "0.5"
+rmcp = { version = "0.15", features = ["server", "macros", "transport-io", "schemars"] }
+tokio = { version = "1", features = ["macros", "rt", "io-std"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+schemars = "1"
```
st0012: All of these dependencies are pulled by `rmcp`, which is the official MCP SDK for Rust. It's the only direct dependency added for this feature.

rust/rubydex/src/main.rs
```diff
@@ -98,6 +104,19 @@ fn main() {
         return exit(args.stats);
     }
 
+    if args.mcp {
```
vinistock: I wonder if this should be a separate executable instead since most of the options for this one don't apply exactly to the MCP server.

rust/rubydex/Cargo.toml
vinistock: In that case, why do we need to list all of them? Also, I really think we should split the MCP into its own separate executable crate.

Otherwise, we're including the entire MCP implementation + all dependencies even for consumers that are using Rubydex as a library and don't need all of it.

rust/rubydex/src/mcp/mod.rs
vinistock, line 1: My understanding is that `mod.rs` naming convention is deprecated in favour of declaring the module as its own file (i.e.: `mcp.rs`).

Although, I think this can go away if we write a dedicated executable for the MCP in its own crate.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,529 @@
...
+/// returning a relative path suitable for agent consumption.
+fn format_path(uri: &str, root: &str) -> String {
+    let path = uri.strip_prefix("file://").unwrap_or(uri);
+    path.strip_prefix(root)
+        .map_or(path, |p| p.strip_prefix('/').unwrap_or(p))
+        .to_string()
+}
+
+/// A resolved location in the codebase: relative path, 1-indexed line, and 0-indexed column.
+struct Location {
```
vinistock: This should probably become a first class citizen in Rubydex itself and not in the MCP.

rust/rubydex/src/mcp/server.rs
```diff
@@ -0,0 +1,529 @@
...
+            path: format_path(doc.uri(), root),
+            line: line_col.line + 1,
+            col: line_col.col + 1,
+        }
+    }
+}
+
+/// Reads source lines around a given line number (0-indexed) from a file URI.
+/// Returns `context` lines before and after the target line.
+fn read_source_context(uri: &str, line: u32, ctx: u32) -> Option<String> {
```
vinistock: Considering that the idea is for the MCP to be available to answer any questions about the code, why should we provide a way to see the source code? Also, can't the LLM already read it with its built-in tools?

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,529 @@
...
+    let snippet: Vec<String> = lines[start_line..end_line]
+        .iter()
+        .enumerate()
+        .map(|(i, l)| format!("{:>4} | {}", start_line + i + 1, l))
+        .collect();
+    Some(snippet.join("\n"))
+}
+
+/// Formats an ancestor chain into a JSON array of `{"name": ..., "kind": ...}` objects.
+fn format_ancestors(graph: &Graph, ancestors: &Ancestors) -> Vec<serde_json::Value> {
```
vinistock, line 184: If we separate the MCP into its own crate, I think we can just implement serialization for the things we want to make JSON and get rid of all of these formatting methods.

E.g.:

```rust
impl Serialize for Ancestors {
  // define how to serialize a list of ancestors
}

impl Serialize for Ancestor {
  // define how to serialize a single ancestor
}
```

rust/rubydex/src/mcp/server.rs
st0012: I assumed agents can only read files on full file, but then I learned that CC's `Read` tool can use `offset` + `limit`. So this is not needed indeed.

rust/rubydex/Cargo.toml
st0012: Because we do directly use all of them in the codebase, we have to declare them here.

rust/rubydex-mcp/src/server.rs
st0012: Because we still need access to Graph to get the `NameRef`, we can't actually simplify the code like this.

rust/rubydex/src/main.rs
st0012: Separated

CLAUDE.md
```diff
@@ -1,3 +1,25 @@
 # Instructions
 
 For detailed project documentation and agent instructions, see [AGENTS.md](AGENTS.md).
+
+## MCP Server Setup
```
vinistock: Do these instructions make sense? Like, these instructions are for this repo (Rubydex), so Claude wouldn't see these when working elsewhere.

How about we commit the `.mcp.json` file and add a task to `dev up` for installing/updating the MCP?

That way, running `dev up` is all anyone needs to get the MCP and we don't need Claude to figure out how to install it.

README.md
```diff
@@ -77,6 +77,45 @@ diagnostic.message
...
+3. Start Claude Code from that project directory. The MCP server indexes
+   the project at startup and provides semantic code intelligence tools.
+   Verify with `/mcp` in the session.
+
+### Available MCP Tools
+
+| Tool | Description |
+|------|-------------|
+| `search_declarations` | Fuzzy search for classes, modules, methods, constants |
+| `get_declaration` | Full details by FQN with docs, ancestors, members |
```
vinistock: This says it includes ancestors, but we also have a `get_ancestors` tool. Is the documentation correct?

```suggestion
| `get_declaration` | Full details by fully qualified name with docs, ancestors, members |
```

README.md
```diff
@@ -77,6 +77,45 @@ diagnostic.message
...
+| `get_ancestors` | C3-linearized ancestor chain (method resolution order) |
```
vinistock: What does C3-linearized mean?

README.md
```diff
@@ -77,6 +77,45 @@ diagnostic.message
...
+| `get_descendants` | Reverse inheritance: what inherits from this class/module |
+| `find_constant_references` | All precise, resolved constant references across the codebase |
+| `get_file_declarations` | List declarations defined in a specific file |
```
vinistock, line 120: Is this declarations or definitions?

rust/rubydex/src/offset.rs
```diff
@@ -65,4 +65,11 @@ impl Offset {
         let end = line_index.line_col(self.end().into());
         format!("{}:{}-{}:{}", start.line + 1, start.col + 1, end.line + 1, end.col + 1)
     }
+
+    /// Converts the start of an offset to 1-indexed line and column numbers.
+    #[must_use]
+    pub fn to_location(&self, document: &Document) -> (u32, u32) {
```
vinistock: This only returns the line and column for the start, despite being called `to_location`, which is a bit confusing.

How about we extract this into a small PR that introduces a `Location` struct with start and end line/col information that we can use?

rust/rubydex-mcp/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[derive(Parser, Debug)]
+#[command(
+    name = "rubydex_mcp",
+    about = "Rubydex MCP server for AI-assisted Ruby code intelligence",
+    version
+)]
+struct Args {
+    #[arg(value_name = "PATHS", default_value = ".")]
+    paths: Vec<String>,
```
vinistock: Is this ever different than the workspace path? Maybe this can be a string rather than a vector.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,457 @@
...
+use rubydex::model::ids::DeclarationId;
+use rubydex::model::{
+    declaration::{Ancestor, Ancestors},
+    document::Document,
+    graph::Graph,
+    name::NameRef,
+};
+use rubydex::offset::Offset;
+
+#[derive(Clone)]
```
vinistock: I'm curious, why does the server need to be cloned?

rust/rubydex-mcp/src/server.rs
vinistock: My comment stands. This struct should be defined in the Rubydex crate, should hold start/end line/col information and only be used by the MCP.

README.md
st0012: It's declaration. The idea is that:
- It's not uncommon for LLMs to list files and try to get information out of particular file(s)
- Instead of providing `get_file_definition`, which is not immediately helpful in most cases, we just return the declarations associated with those definitions instead

CLAUDE.md
st0012: I added `dev up` support and `.mcp.json` but having a simplified instructions here will help external contributors after we open sourced it.

README.md
st0012: I dropped this request as we already return ancestors in `get_declaration`

rust/rubydex-mcp/src/server.rs
st0012: It was a left-over from previous implementation. I've dropped it now.

CLAUDE.md
st0012: Actually, this should do into `AGENTS.md`

rust/rubydex/src/offset.rs
```diff
@@ -60,9 +60,30 @@ impl Offset {
...
     }
 }
+
+/// A resolved 1-indexed location within a file, with start and end positions.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Location {
+    pub start_line: u32,
+    pub start_col: u32,
+    pub end_line: u32,
+    pub end_col: u32,
```
vinistock: We've been avoiding making any struct's internal state public. Let's add readers for these and make them private?

rust/rubydex/src/offset.rs
```diff
@@ -60,9 +60,30 @@ impl Offset {
...
+    pub fn to_location(&self, document: &Document) -> Location {
         let line_index = document.line_index();
         let start = line_index.line_col(self.start().into());
         let end = line_index.line_col(self.end().into());
-        format!("{}:{}-{}:{}", start.line + 1, start.col + 1, end.line + 1, end.col + 1)
+        Location {
+            start_line: start.line + 1,
+            start_col: start.col + 1,
+            end_line: end.line + 1,
+            end_col: end.col + 1,
```
vinistock: Let's not make this 1 indexed. The LSP and most tool-to-tool communication tends to use zero index and then it gets a +1 only on presentation.

We can even have a `to_presentation` method in `Location` that returns a new one with the +1.

AGENTS.md
```diff
@@ -85,9 +87,21 @@ When necessary, commands can be executed for the Rust code.
...
+- `cargo test`: runs Rust tests (all workspace crates)
+- `cargo test -p rubydex-mcp`: runs MCP server tests only
 - `cargo test test_name`: runs a specific tests example
 - `cargo fmt`: auto formats the Rust code
 - `cargo clippy`: lints the Rust code
+- `cargo install --path rust/rubydex-mcp`: installs the MCP server binary
 - `bundle exec rake lint_rust`: lints the Rust code
 - `bundle exec rake format_rust`: auto formats the Rust code
+
+### MCP Server Setup
```
vinistock: I'm still a bit torn about this. Considering that installing the MCP server is a one time action and from then on you only ask Claude to implement stuff, is it really worth including this in the context?

Like, the only action that requires this information is the first time repository setup. Literally everything you do afterwards does not require this information.

README.md
vinistock: Got it 👍

rust/rubydex-mcp/tests/mcp.rs
vinistock, line 1: Not something we need to address for this first version, but I feel like it's a bit odd to have all of the MCP tests be real integration (actually spawn the server, perform IO...).

We definitely want to have at least some of these, but in future iterations it might be better to have more unit tests for the logic and just a handful of integration tests.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,418 @@
...
+            "declarations": declarations,
+        });
+
+        serde_json::to_string_pretty(&result).unwrap_or_else(|_| "{}".to_string())
+    }
+
+    #[tool(
+        description = "Get an overview of the indexed Ruby codebase: total file count, declaration counts, and breakdown by kind (classes, modules, methods, constants). Use this to understand codebase size and composition, or to verify that indexing completed successfully."
+    )]
+    fn codebase_stats(&self) -> String {
```
vinistock, line 481: Let's start like this and audit the tools later, but would knowing aggregate numbers for the codebase help Claude in any task?

This feels more relevant to us (the tool devs) than for a Ruby dev using the MCP.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,418 @@
...
+                }));
+            }
+        }
+
+        let result = serde_json::json!({
+            "file": format_path(doc.uri(), &self.root),
+            "declarations": declarations,
+        });
+
+        serde_json::to_string_pretty(&result).unwrap_or_else(|_| "{}".to_string())
```
vinistock: I'm not familiar with this, but is `to_string_pretty` similar to Ruby's `JSON.pretty_generate`?

I ask because prettifying the JSON consumes more characters that have to be sent through IO. Considering that no one actually looks at these JSONs, we may want to use simple `to_string` and avoid the extra line breaks and indentation spaces.

rust/rubydex-mcp/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+struct Args {
+    #[arg(value_name = "PATH", default_value = ".")]
+    path: String,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    let root = match std::fs::canonicalize(&args.path) {
+        Ok(p) => p.to_string_lossy().to_string(),
```
vinistock: Probably an uncommon edge case, but if the result of canonicalizing the path has invalid data for the encoding, using `to_string_lossy` will return the path with those unicode fillers—which I assume will likely crash the server when trying to use `root`.

I'd vote for us to unwrap/expect here. Realistically, if we can't properly canonicalize a path to a valid UTF-8 string, we can't really launch the server.

rust/rubydex-mcp/src/server.rs
vinistock: Ah, true. Yeah, this is an unfortunate consequence of all of the interning of the adjacency graph.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+}
+
+#[tool_handler]
+impl ServerHandler for RubydexServer {
+    fn get_info(&self) -> ServerInfo {
+        ServerInfo {
+            instructions: Some("Rubydex provides semantic Ruby code intelligence. Use these tools INSTEAD OF Grep when working with Ruby code structure.\n\nDecision guide:\n- Know a name? -> search_declarations (fuzzy search by name)\n- Have an exact FQN? -> get_declaration (full details with docs, ancestors, members)\n- Need reverse hierarchy? -> get_descendants (what inherits from this class/module)\n- Refactoring a class/module/constant? -> find_constant_references (all precise usages across codebase)\n- Exploring a file? -> get_file_declarations (structural overview)\n- First time in codebase? -> codebase_stats (size and composition)\n\nTypical workflow: search_declarations -> get_declaration -> find_constant_references.\n\nFQN format: \"Foo::Bar\" for classes/modules/constants, \"Foo::Bar#method_name\" for instance methods.\n\nUse Grep instead for: literal string search, log messages, comments, non-Ruby files, or content search rather than structural queries.".into()),
```
vinistock: Nit: can we use a multiline string here so that this is easier to read?

rust/rubydex-mcp/src/server.rs
vinistock, line 1: Probably not a concern at the moment, but going forward, we should split the request implementations into separate files. If we implement all of them here this file is going to be gigantic.

Also, that helps us unit tests each request separately without needing integration.

rust/rubydex-mcp/src/server.rs
st0012: I kept it because Claude likes to call it first to get an overview. I don't know how that helps but it's also a good way to verify the codebase is indexed?

st0012: Oh it's actually guided by instructions. But thinking about it, using stats to understand the size and composition of the codebase makes sense. So I'd keep it.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn find_constant_references(&self, Parameters(params): Parameters<FindConstantReferencesParams>) -> String {
+        let graph = &self.graph;
+        let (declaration_id, _decl) = lookup_declaration!(graph, &params.name);
+
+        let limit = params.limit.unwrap_or(50).min(200);
+        let mut references: Vec<serde_json::Value> = Vec::new();
+        let mut truncated = false;
+
+        for const_ref in graph.constant_references().values() {
+            if references.len() >= limit {
```
Morriar, line 385: This iterates over all constant references in the graph on every call — O(n) over the whole codebase to find matches for a single declaration. For large Ruby projects this could be slow.

I wonder if we could build a reverse index at startup (declaration_id → Vec<const_ref_offset>) in `RubydexServer::new`? The graph is read-only at runtime, so building it once would be cheap and make lookups O(k) where k is the result count.

README.md
```diff
@@ -77,6 +77,44 @@ diagnostic.message
...
+like Claude to semantically query your Ruby codebase.
+
+### Setup
+
+1. Install the binary:
+   ```bash
+   cargo install --path rust/rubydex-mcp
+   ```
+
+2. Add a `.mcp.json` to the Ruby project you want to index:
```
Morriar: Doesn't Claude has a `mcp add` command, can't we suggest that?

.claude/skills/rubydex-mcp-benchmark/SKILL.md
```diff
@@ -0,0 +1,124 @@
...
+
+This places `rubydex_mcp` in `~/.cargo/bin/`. Verify: `rubydex_mcp --help`
+
+## Step 1: Find a Good Benchmark Candidate
+
+Use rubydex MCP tools to find a class or module suitable for testing:
+
+1. Run `codebase_stats` to get an overview of the target project
+2. Run `search_declarations` with kind "Class" or "Module" to list candidates
+3. For each candidate, run `find_references` and count references
```
Morriar:
```suggestion
3. For each candidate, run `find_constant_references` and count references
```

rust/rubydex-mcp/src/tools.rs
Morriar, line 1: I wonder if we should split this file into one file per tool?

rust/rubydex-mcp/tests/mcp.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+#[test]
+#[allow(clippy::too_many_lines)]
+fn mcp_server_e2e() {
+    with_context(|context| {
+        context.write(
+            "app.rb",
+            r#"
+class Animal
```
Morriar: You should be able to indent this code properly, we normalize the indent inside `Context#write`: https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/test_utils/context.rs#L120

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,436 @@
...
+        description = "List all Ruby classes, modules, methods, and constants defined in a specific file. Returns a structural overview with names, kinds, and line numbers. Use this to understand a file's structure before reading it, or to see what a file contributes to the codebase. Accepts relative or absolute paths."
+    )]
+    fn get_file_declarations(&self, Parameters(params): Parameters<GetFileDeclarationsParams>) -> String {
+        let graph = &self.graph;
+
+        let target_path = &params.file_path;
+        let matching_doc = graph.documents().iter().find(|(_, doc)| {
+            let uri = doc.uri();
+            let path = uri.strip_prefix("file://").unwrap_or(uri);
+            path == target_path || path.ends_with(&format!("/{target_path}"))
```
Morriar: `ends_with` can produce false positives when two files share a common suffix — e.g. both `app/user.rb` and `lib/app/user.rb` would match a query for `app/user.rb`.

I wonder if we could normalize to absolute paths and use exact matching instead?

```suggestion
            path == target_path
                || (!std::path::Path::new(target_path).is_absolute()
                    && path == format!("{}/{}", self.root, target_path))
```

rust/rubydex-mcp/src/tools.rs
st0012: Given the list is still short, I wouldn't do that just yet

rust/rubydex-mcp/src/server.rs
vinistock: We already have this. After resolution, `declaration.references()` should return all references that got resolved to it.

We need this for the LSP as well for find references and rename, so it's already built.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,539 @@
...
+        let graph = state.graph.as_ref().unwrap();
+
+        let absolute_target = if Path::new(&params.file_path).is_absolute() {
+            PathBuf::from(&params.file_path)
+        } else {
+            self.root_path.join(&params.file_path)
+        };
+        let normalized_target = normalize_path_for_compare(&absolute_target);
+
+        let matching_doc = graph.documents().iter().find(|(_, doc)| {
```
vinistock: We don't need to iterate over all documents. We can just:

- Turn the file path into a URI with `Url::parse`
- Turn the URI into a `UriId`
- Access the hash directly with that ID

rust/rubydex-mcp/src/server.rs
vinistock: We can keep it for now, but I can't think of any tasks where knowing the number of classes would help Claude.

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,539 @@
...
+
+        serde_json::to_string(&result).unwrap_or_else(|_| "{}".to_string())
+    }
+}
+
+const SERVER_INSTRUCTIONS: &str = r#"Rubydex provides semantic Ruby code intelligence. Use these tools INSTEAD OF Grep when working with Ruby code structure.
+
+Decision guide:
+- Know a name? -> search_declarations (fuzzy search by name)
+- Have an exact FQN? -> get_declaration (full details with docs, ancestors, members)
```
vinistock:
```suggestion
- Have an exact fully qualified name? -> get_declaration (full details with docs, ancestors, members)
```

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,539 @@
...
+- Need reverse hierarchy? -> get_descendants (what inherits from this class/module)
+- Refactoring a class/module/constant? -> find_constant_references (all precise usages across codebase)
+- Exploring a file? -> get_file_declarations (structural overview)
+- First time in codebase? -> codebase_stats (size and composition)
```
vinistock: Let's at least not direct Claude to use codebase stats if we're not sure how it can be helpful.
```suggestion
- Want general statistics? -> codebase_stats (size and composition)
```

rust/rubydex-mcp/src/server.rs
```diff
@@ -0,0 +1,539 @@
...
+
+Typical workflow: search_declarations -> get_declaration -> find_constant_references.
+
+FQN format: "Foo::Bar" for classes/modules/constants, "Foo::Bar#method_name" for instance methods.
```
vinistock:
```suggestion
Fully qualified name format: "Foo::Bar" for classes/modules/constants, "Foo::Bar#method_name" for instance methods.
```

rust/rubydex/src/offset.rs
```diff
@@ -77,9 +60,71 @@ impl Offset {
...
+    }
+
+    #[must_use]
+    pub fn end_col(&self) -> u32 {
+        self.end_col
+    }
+
+    /// Returns a new `Location` with 1-indexed values for display purposes.
+    #[must_use]
+    pub fn to_presentation(&self) -> Self {
```
vinistock, line 122: I think this is good for now, but we may want to follow up with a PR similar to this one https://github.com/Shopify/rubydex/pull/607.

It prevents someone from doing `to_presentation` multiple times and accidentally bump the location in a way that doesn't make sense.

st0012: Added to #608

rust/rubydex-mcp/src/server.rs
st0012: Updated 👍

PR #587, opened 2026-02-17
Add API to index in-memory sources

vinistock:
This PR allows us to index documents that are stored in-memory (like unsaved files) rather than on disk.

I was starting the work on the Ruby LSP adoption and realized that all of our tests need to be able to index in-memory sources. Rather than re-writing the entire test suite, I figured we ought to add this API already since we need it for the regular LSP operations anyway.

My proposal is that we create a separate API for in-memory sources that is disconnected from indexing all files from disk. The reason is because we have these scenarios:

- Boot: we need to index everything from disk based on the workspace path + dependency paths
- Watched files change notifications: these notifications are received when a file modification was committed to disk and we always receive a list of modifications. Using the existing `index_all` API, which accepts an array and runs in parallel is the adequate choice
- Regular file modifications: these are the only ones that happen in-memory and they are always sent to the LSP one by one

### DIFF
--- a/rust/rubydex/src/indexing.rs
+++ b/rust/rubydex/src/indexing.rs
@@ -5,12 +5,39 @@ use crate::{
     model::graph::Graph,
 };
 use crossbeam_channel::{Sender, unbounded};
-use std::{fs, path::PathBuf, sync::Arc};
+use std::{ffi::OsStr, fs, path::PathBuf, sync::Arc};
 use url::Url;
 
 pub mod local_graph;
 pub mod rbs_indexer;
 pub mod ruby_indexer;
+
+/// The language of a source document, used to dispatch to the appropriate indexer
+pub enum LanguageId {
+    Ruby,
+    Rbs,
+}
+
+impl From<&OsStr> for LanguageId {
+    fn from(ext: &OsStr) -> Self {
+        if ext == "rbs" { Self::Rbs } else { Self::Ruby }
+    }
+}
+
+impl LanguageId {
+    /// Determines the language from an LSP language ID string.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the language ID is not recognized.
+    pub fn from_language_id(language_id: &str) -> Result<Self, Errors> {
+        match language_id {
+            "ruby" => Ok(Self::Ruby),
+            "rbs" => Ok(Self::Rbs),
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -22,17 +21,10 @@ impl GraphTest {
...
-        indexer.index();
-        indexer.local_graph()
-    }
-
+    /// Indexes a Ruby source
     pub fn index_uri(&mut self, uri: &str, source: &str) {
         let source = normalize_indentation(source);
-        let local_index = Self::index_source(uri, &source);
-        self.graph.update(local_index);
+        indexing::index_source(&mut self.graph, uri, &source, &LanguageId::Ruby);
```
vinistock, line 27: For tests, I would rather expose an `index_rbs_uri` rather than modifying the signature to accept a new language_id parameter. Let me know if you disagree.

rust/rubydex/src/indexing.rs
```diff
@@ -56,24 +83,21 @@ impl Job for IndexingJob {
...
-            let mut indexer = RBSIndexer::new(url.to_string(), &source);
-            indexer.index();
-            indexer.local_graph()
-        } else {
-            let mut indexer = RubyIndexer::new(url.to_string(), &source);
-            indexer.index();
-            indexer.local_graph()
-        };
+        let language = self.path.extension().map_or(LanguageId::Ruby, LanguageId::from);
+        let local_graph = build_local_graph(url.to_string(), &source, &language);
```
amomchilov, line 87: Can these 2 lines be replaced with just a call to your new `index_source` function?

vinistock: Not quite. There are two subtle differences:

- `index_source` expects the language ID to be an enum since that gets passed by the editor to us. This is reading and transforming from the file's extension
- `index_source` immediately updates the graph, but we don't want that to happen in the parallel indexing case. The threads send the indexing result through a channel and the main thread is responsible for the merge

rust/rubydex/src/test_utils/graph_test.rs
alexcrocha: A separate helper for indexing rbs uris is fine for testing 👍

PR #589, opened 2026-02-19
Add incremental invalidation/resolution

st0012:
## Summary

Replace the full re-resolution strategy (`clear_declarations` + resolve everything from scratch) with incremental invalidation. Graph mutations (`update`/`delete_document`) now compute the minimal set of definitions, references, and ancestor chains that need re-resolution, and the resolver processes only that subset.

### How it works

Graph mutations follow a three-step pipeline:

1. **`invalidate`** — Detaches old definitions from declarations. Identifies affected declarations (definition removed, new definition/reference touching existing names). Feeds them into `invalidate_graph`, a unified worklist that processes declarations and names.
2. **`remove_document_data`** — Removes old refs/defs/names/strings from maps. Cleans up empty declarations.
3. **`extend`** — Merges new `LocalGraph` data and queues new definitions/references for resolution.

Work items are accumulated as `Unit` (definitions, references, ancestors) and drained by the resolver. For the initial full index, this contains everything. For incremental updates, only the invalidated subset.

### Reverse index: `name_dependents`

A single reverse index `name_dependents: NameId → Vec<NameDependent>` maps each name to the definitions and references that depend on it. Each dependent is stored under its own `name_id` **and** under the `nesting`/`parent_scope` of that name, so invalidation can trace from any scope level directly to affected refs/defs without an intermediate name-to-name layer.

### Invalidation worklist

`invalidate_graph` processes two kinds of items:

- **Declaration** — Two modes: *removal* (no definitions left or orphaned owner) removes the declaration tree and unresolves all resolved names; *ancestor-stale* (still has definitions) clears ancestors/descendants and cascades to dependent names.
- **Name** — Two modes: *structural cascade* (nesting or parent_scope dependency broken) unresolves the name and detaches its refs/defs from the old declaration; *ancestor-only* (dependencies intact but ancestor context changed) unresolves references for re-evaluation.

### Other changes

- **`without_resolution` option** — Skip accumulating pending work for tools that only need definitions.
- **`name_dependents` populated during indexing** — `LocalGraph` builds `name_dependents` entries in `add_definition` and `add_constant_reference`, then merged into the global graph during `extend`.
- **Resolution refactored** — `resolve_all` renamed to `resolve`, drains `pending_work` via `Unit` enum instead of scanning all names.

### TODO: smarter ancestor invalidation

Currently, any definition change on a declaration enters ancestor-stale mode, which conservatively unresolves all nested references. A future optimization: compare old vs new definitions for ancestor-affecting fields (`mixins`, `parent_class`). If they haven't changed, skip ancestor invalidation entirely — only re-resolve the definition itself.

## Compared to `main`

**Correctness: identical** — all declaration counts, definition counts, orphan rates, and linked/orphan breakdowns match exactly between main and the branch.

**Performance (initial full index on 94,036 files):**

| Stage | main | branch | Delta |
|---|---|---|---|
| Listing | 0.645s | 0.701s | +0.056s (+8.7%) |
| Indexing | 10.309s | 10.261s | -0.048s (-0.5%) |
| **Resolution** | **41.570s** | **25.275s** | **-16.295s (-39.2%)** |
| Querying | 0.647s | 0.696s | +0.049s (+7.6%) |
| **Total** | **53.171s** | **36.933s** | **-16.238s (-30.5%)** |

**Memory:**

| | main | branch | Delta |
|---|---|---|---|
| Max RSS | 3,909 MB | 4,437 MB | +528 MB (+13.5%) |

Resolution is **39% faster** at the cost of **13.5% more memory** from the reverse index and pending work accumulation. (Note: benchmarks predate some of the recent changes — numbers may have shifted slightly.)

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -356,6 +356,24 @@ static VALUE rdxr_graph_set_encoding(VALUE self, VALUE encoding) {
     return Qnil;
 }
 
+// Graph#without_resolution=: (Boolean) -> void
+// Configures the graph to skip accumulating resolution work items
+static VALUE rdxr_graph_set_without_resolution(VALUE self, VALUE without_resolution) {
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+    rdx_graph_set_without_resolution(graph, RTEST(without_resolution));
+    return Qnil;
+}
+
+// Graph#without_resolution?: () -> Boolean
+// Returns whether the graph is configured to skip accumulating resolution work items
+static VALUE rdxr_graph_without_resolution(VALUE self) {
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    return rdx_graph_without_resolution(graph) ? Qtrue : Qfalse;
+}
+
 // Graph#resolve_constant: (String, Array[String]) -> Declaration?
 // Runs the resolver on a single constant reference to determine what it points to
 static VALUE rdxr_graph_resolve_constant(VALUE self, VALUE const_name, VALUE nesting) {
@@ -498,6 +516,8 @@ void rdxi_initialize_graph(VALUE mRubydex) {
     rb_define_method(cGraph, "[]", rdxr_graph_aref, 1);
     rb_define_method(cGraph, "search", rdxr_graph_search, 1);
     rb_define_method(cGraph, "encoding=", rdxr_graph_set_encoding, 1);
+    rb_define_method(cGraph, "without_resolution=", rdxr_graph_set_without_resolution, 1);
+    rb_define_method(cGraph, "without_resolution?", rdxr_graph_without_resolution, 0);
     rb_define_method(cGraph, "resolve_require_path", rdxr_graph_resolve_require_path, 2);
     rb_define_method(cGraph, "require_paths", rdxr_graph_require_paths, 1);
 }
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -60,7 +60,7 @@ impl RubydexServer {
                 }
 
                 let mut resolver = rubydex::resolution::Resolver::new(&mut graph);
-                resolver.resolve_all();
+                resolver.resolve();
 
                 eprintln!(
                     "Rubydex indexed {} files, {} declarations",
--- a/rust/rubydex/src/indexing/local_graph.rs
+++ b/rust/rubydex/src/indexing/local_graph.rs
@@ -3,6 +3,7 @@ use std::collections::hash_map::Entry;
 use crate::diagnostic::{Diagnostic, Rule};
 use crate::model::definitions::Definition;
 use crate::model::document::Document;
+use crate::model::graph::NameDependent;
 use crate::model::identity_maps::IdentityHashMap;
 use crate::model::ids::{DefinitionId, NameId, ReferenceId, StringId, UriId};
 use crate::model::name::{Name, NameRef};
@@ -18,6 +19,7 @@ type LocalGraphParts = (
     IdentityHashMap<NameId, NameRef>,
     IdentityHashMap<ReferenceId, ConstantReference>,
     IdentityHashMap<ReferenceId, MethodRef>,
+    IdentityHashMap<NameId, Vec<NameDependent>>,
 );
 
 #[derive(Debug)]
@@ -29,6 +31,7 @@ pub struct LocalGraph {
     names: IdentityHashMap<NameId, NameRef>,
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
+    name_dependents: IdentityHashMap<NameId, Vec<NameDependent>>,
 }
 
 impl LocalGraph {
@@ -42,6 +45,7 @@ impl LocalGraph {
             names: IdentityHashMap::default(),
             constant_references: IdentityHashMap::default(),
             method_references: IdentityHashMap::default(),
+            name_dependents: IdentityHashMap::default(),
         }
     }
 
@@ -69,6 +73,10 @@ impl LocalGraph {
 
     pub fn add_definition(&mut self, definition: Definition) -> DefinitionId {
         let definition_id = definition.id();
+
+        if let Some(name_id) = definition.name_id() {
+            self.insert_name_dependent(*name_id, NameDependent::Definition(definition_id));
+        }
 
         if self.definitions.insert(definition_id, definition).is_some() {
             debug_assert!(false, "DefinitionId collision in local graph");
@@ -133,6 +141,7 @@ impl LocalGraph {
 
     pub fn add_constant_reference(&mut self, reference: ConstantReference) -> ReferenceId {
         let reference_id = reference.id();
+        self.insert_name_dependent(*reference.name_id(), NameDependent::Reference(reference_id));
 
         if self.constant_references.insert(reference_id, reference).is_some() {
             debug_assert!(false, "ReferenceId collision in local graph");
@@ -172,6 +181,26 @@ impl LocalGraph {
         self.document.add_diagnostic(diagnostic);
     }
 
+    // Name dependents
+
+    /// Registers a dependent under its own `name_id` and under the `nesting`/`parent_scope`
+    /// of that name, so invalidation can trace from any of those scopes.
+    fn insert_name_dependent(&mut self, name_id: NameId, dep: NameDependent) {
+        self.name_dependents.entry(name_id).or_default().push(dep);
+
+        if let Some(name_ref) = self.names.get(&name_id) {
+            for owner_id in [name_ref.nesting().as_ref().copied(), name_ref.parent_scope().as_ref().copied()]
+                .into_iter()
+                .flatten()
+            {
+                let deps = self.name_dependents.entry(owner_id).or_default();
+                if !deps.contains(&dep) {
+                    deps.push(dep);
+                }
+            }
+        }
+    }
+
     // Into parts
 
     #[must_use]
@@ -184,6 +213,7 @@ impl LocalGraph {
             self.names,
             self.constant_references,
             self.method_references,
+            self.name_dependents,
         )
     }
 }
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -14,9 +14,40 @@ use crate::model::references::{ConstantReference, MethodRef};
 use crate::model::string_ref::StringRef;
 use crate::stats;
 
+/// An entity whose validity depends on a particular `NameId`.
+/// Used as the value type in the `name_dependents` reverse index.
+///
+/// Each dependent is stored under its own `name_id` **and** under the `nesting`/`parent_scope`
+/// of that name. This lets us trace from a changed name directly to the affected refs/defs
+/// without an intermediate `Name → Name` layer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NameDependent {
+    Definition(DefinitionId),
+    Reference(ReferenceId),
+}
+
+/// Items processed by the unified invalidation worklist.
+enum InvalidationItem {
+    /// A declaration whose ancestor chain is stale, or that has become empty and needs removal.
+    Declaration(DeclarationId),
+    /// A name whose dependencies may have changed, needing cascade or reference re-evaluation.
+    Name(NameId),
+}
+
 pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Object"));
 pub static MODULE_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Module"));
 pub static CLASS_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Class"));
+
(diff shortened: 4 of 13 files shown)


rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -6,11 +6,16 @@ use crate::indexing::ruby_indexer::RubyIndexer;
 use crate::model::graph::Graph;
 use crate::resolution::Resolver;
 
-#[derive(Default)]
 pub struct GraphTest {
     graph: Graph,
 }
 
+impl Default for GraphTest {
```
vinistock: Can we go back to the derive instead?

rust/rubydex/src/model/graph.rs
```diff
@@ -18,6 +18,26 @@ pub static OBJECT_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::
 pub static MODULE_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Module"));
 pub static CLASS_ID: LazyLock<DeclarationId> = LazyLock::new(|| DeclarationId::from("Class"));
 
+/// Work items produced by graph mutations (update/delete) that need resolution.
+/// Drained internally by the resolver.
+#[derive(Default, Debug)]
+pub(crate) struct PendingWork {
```
vinistock: This looks like a `Vec<Unit>`. Can we not that instead? It naturally fits into resolution and avoids having to allocate the units during resolution (since indexing already did it upfront).

rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +60,26 @@ pub struct Graph {
...
+    /// Used during incremental invalidation to find names that need unresolving when a declaration is removed.
+    declaration_to_names: IdentityHashMap<DeclarationId, IdentityHashSet<NameId>>,
+
+    /// Reverse index: for each `NameId`, which constant references use it.
+    /// Eliminates O(R) scans of `constant_references` during invalidation.
+    name_to_references: IdentityHashMap<NameId, Vec<ReferenceId>>,
+
+    /// Reverse index: for each `NameId`, which other names depend on it
+    /// (via nesting or `parent_scope`). Used for cascade invalidation.
+    name_dependents: IdentityHashMap<NameId, Vec<NameId>>,
```
vinistock: Names are always associated to either definitions or constant references. Instead of having 3 separate maps, did you experiment using an enum for `NameDependent`?

Something like:

```rust
// The name ids are already stored in the definitions and references, so
// we can follow them to get to the names
enum NameDependent {
  Definition(DefinitionId),
  Reference(ReferenceId)
}
```

Also, I suspect that `name_dependents` will have a ton of duplicates as we may discover dependencies between names multiple times across the codebase. We should use an `IdentityHashSet` instead of a `Vec` here.

rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +60,26 @@ pub struct Graph {
...
+
+    /// Reverse index: for each `NameId`, which definitions use it as their name.
+    /// Eliminates O(D) scans of definitions during cascade.
+    name_to_definitions: IdentityHashMap<NameId, Vec<DefinitionId>>,
+
+    /// Accumulated pending work from update/delete operations.
+    /// Drained internally by the resolver.
+    pending_work: PendingWork,
```
vinistock: Instead of accumulating pending work in the global graph, can we not accumulate this within `index_files`?

We want to allow consumers of the toolkit to decide if they need resolution or not. Keeping this state as part of the graph means that any clients that don't run resolution will have a memory leak from indexing files multiple times.

If we return a `Vec<Unit>` from `indexing::index_files`, then consumers of the API can decide whether to ignore it (which gets freed immediately) or pass it to resolution. We can also make `delete_document` return both the deleted document and a `Vec<Unit>`.

Something like:

```ruby
graph = Rubydex::Graph.new

# Initial indexing
units = graph.index_workspace
graph.resolve(units)

# Incremental LSP indexing
units = graph.index_source(...)
graph.resolve(units)

# Document deletion
document, units = graph.delete_document(uri)
graph.resolve(units)
```

rust/rubydex/src/model/graph.rs
```diff
@@ -1581,4 +2041,988 @@ mod tests {
...
+
+        assert!(
+            matches!(reference_name, NameRef::Unresolved(_)),
+            "Deleting include file should invalidate constant reference"
+        );
+        assert_declaration_references_count_eq!(context, "Bar::CONST", 0);
+    }
+
+    #[test]
+    fn re_indexing_preserves_constant_singleton_classes() {
```
vinistock, line 2979: This scenario is a bug in main. Are we able to test the incremental behaviour without asserting the current bug?

rust/rubydex/src/model/graph.rs
st0012: From Vini: if we keep this internal state approach, we need to be mindful that the vector needs to be resized manually or we could leak memory even with an empty vector.

rust/rubydex/src/model/graph.rs
```diff
@@ -401,19 +444,96 @@ impl Graph {
         &self.names
     }
 
+    /// Drains the accumulated work items, returning them for use by the resolver.
+    pub fn take_pending_work(&mut self) -> Vec<Unit> {
+        std::mem::take(&mut self.pending_work)
```
vinistock, line 541: Let's double check if this shrinks the vector or keeps the current allocated capacity to avoid wasting memory.

rust/rubydex/src/model/graph.rs
```diff
@@ -401,19 +444,96 @@ impl Graph {
...
+                let declaration_id = *resolved.declaration_id();
+                let name = resolved.name().clone();
+                self.names.insert(name_id, NameRef::Unresolved(Box::new(name)));
+
+                if let Some(name_set) = self.declaration_to_names.get_mut(&declaration_id) {
+                    name_set.remove(&name_id);
+                    if name_set.is_empty() {
+                        self.declaration_to_names.remove(&declaration_id);
+                    }
+                }
```
vinistock: This is one of the reasons why I'm not a fan of the multiple hashmaps. We need to ensure that the data on the auxiliary maps is also consistent and the benefit is just avoiding tracing the graph from `declaration -> definitions -> names`. I'm not convinced tbh.

Also, to make sure we're making progress, I think you can probably ship a separate PR with just the `unresolve_name` and `unresolve_reference` methods separately.

rust/rubydex/src/model/graph.rs
```diff
@@ -575,18 +709,40 @@ impl Graph {
                     entry.get_mut().increment_ref_count(name_ref.ref_count());
                 }
                 Entry::Vacant(entry) => {
-                    entry.insert(name_ref);
+                    // Track nesting and parent_scope dependents for new names only
+                    if let Some(nesting_id) = entry.insert(name_ref).nesting() {
+                        self.name_dependents.entry(*nesting_id).or_default().insert(name_id);
```
vinistock: I think populating the `name_dependents` should happen during indexing in a local graph and then we just merge the hashmaps here (similar to what we do for every other hashmap).

rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +67,22 @@ pub struct Graph {
...
+    /// Used during incremental invalidation to find names that need unresolving when a declaration is removed.
+    declaration_to_names: IdentityHashMap<DeclarationId, IdentityHashSet<NameId>>,
+
+    /// Reverse index: for each `NameId`, which definitions and constant references use it.
+    /// Eliminates O(D+R) scans during invalidation.
+    name_users: IdentityHashMap<NameId, Vec<NameUser>>,
+
+    /// Reverse index: for each `NameId`, which other names depend on it
+    /// (via nesting or `parent_scope`). Used for cascade invalidation.
+    name_dependents: IdentityHashMap<NameId, IdentityHashSet<NameId>>,
```
vinistock: Why do we need these two maps? The idea of having an enum for name dependents is so that you can go `NameId -> ReferenceId | DefinitionId -> NameId`.

rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +67,22 @@ pub struct Graph {
...
+
+    /// Accumulated work items from update/delete operations.
+    /// Drained by `take_pending_work()` before resolution.
+    pending_work: Vec<Unit>,
```
vinistock, line 86: We still need an answer for the ever growing memory if users don't call resolve.

rust/rubydex/src/model/graph.rs
st0012: `take` replaces the field with an empty vector, so there will be no leaks. From its doc:

> take allows taking ownership of a struct field by replacing it with an "empty" value

rust/rubydex/src/model/graph.rs
st0012: I merged the maps but I think the value should also include NameId. So it'd be `NameId -> ReferenceId | DefinitionId | NameId`.

I prototyped using reference/definition to look up name but it doesn't work well with parent_scope/nesting cases. For example:

```ruby
class Baz
  include Foo
  CONST
end
```

The reference `CONST` creates a name with `nesting=baz_name`, but it doesn't create a member declaration under `Baz`. So `Baz.members()` is empty — there's no path from `baz_name` to the reference's name through the declaration/definitions. When `Baz`'s ancestors change (e.g., a new `prepend Bar` is added), we need to re-evaluate `CONST`, but without the explicit `Name(NameId)` entry under `baz_name`, we can't discover it.

rust/rubydex/src/model/graph.rs
st0012: I'll address all the feedback first and then see how we can split this PR.

rust/rubydex/src/model/graph.rs
vinistock: > there's no path from baz_name to the reference's name through the declaration/definitions

This is the only reason why we need `name_dependents`. In your example, this is what I would expect the hashmap to look like:

```
# name_dependents

{
  NameId(Baz) => Set[ReferenceId(Foo), ReferenceId(CONST)]
}
```

This allows the graph to remember which references and definitions will be potentially impacted by a name change. We then trace `name_dependents` and the rest of the graph to unresolve names.

In this case, if we had to unresolve all names due to a change to `Baz`, I would expect the algorithm to do something like this:

1. `Baz` changed. Loop through name dependents
2. Regardless whether the dependent is a reference or definition, get its `name_id`, pull the name from the graph and unresolve it
3. Now, unresolving the definition and reference may invalidate other things. Go back to 1. and invalidate the `name_id` for the reference/definition

This also involves invalidating ancestors, but you get the idea.

rust/rubydex/src/model/graph.rs
st0012: First one: https://github.com/Shopify/rubydex/pull/627

ext/rubydex/graph.c
```diff
@@ -356,6 +356,24 @@ static VALUE rdxr_graph_set_encoding(VALUE self, VALUE encoding) {
     return Qnil;
 }
 
+// Graph#without_resolution=: (Boolean) -> void
+// Configures the graph to skip accumulating resolution work items
+static VALUE rdxr_graph_set_without_resolution(VALUE self, VALUE without_resolution) {
```
vinistock, line 361: IMO, this makes for a weirder API than returning the units that API consumers can just ignore.

Basically this:

```ruby
g = Rubydex::Graph.new
g.without_resolution = true
g.index_workspace
```

vs

```ruby
g = Rubydex::Graph.new
_ignored_units = g.index_workspace
# Consumers that want resolution do
# g.resolve(units)
```

@Morriar what do you think?

rust/rubydex-mcp/src/server.rs
```diff
@@ -60,7 +60,7 @@ impl RubydexServer {
                 }
 
                 let mut resolver = rubydex::resolution::Resolver::new(&mut graph);
-                resolver.resolve_all();
+                resolver.resolve();
```
vinistock, line 63: If we want to rename this, can we move it to a separate PR? This is a complex change, so I'm trying to make sure we're focused on the core aspects of the implementation and that we lower the barrier for better reviews.

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -42,6 +45,7 @@ impl LocalGraph {
             names: IdentityHashMap::default(),
             constant_references: IdentityHashMap::default(),
             method_references: IdentityHashMap::default(),
+            name_dependents: IdentityHashMap::default(),
```
vinistock, line 48: Populating and managing the `name_dependents` can be a separate PR too.

rust/rubydex/src/query.rs
```diff
@@ -881,11 +881,26 @@ mod tests {
             Some(Name::new(StringId::from("Foo"), ParentScope::None, None).id()),
         )
         .id();
-        assert_completion_eq!(
-            context,
-            CompletionReceiver::Expression(name_id),
-            ["Foo::Bar", "$var", "Foo", "$var2", "Foo::Bar#bar_m()"]
-        );
+        let mut candidates = completion_candidates(
```
vinistock, line 884: Why is this changing? We should still use the `assert_completion_eq`.

ext/rubydex/graph.c
st0012: No the idea is to pass it as constructor arg: `Rubydex::Graph.new(without_resolution: true)`. I just haven't found a good way to set the state and past it to the Rust side.

PR #591, opened 2026-02-21
Index modules with RBS indexer

alexcrocha:
One more step towards #87. 

This PR starts indexing modules with the RBS indexer.

 - `.rbs` files are now collected during file discovery (previously, only `.rb` files were collected)
 - Added `new_rbs()` to `LocalGraphTest` so `rbs_indexer` tests can use the same assertion pattern as `ruby_indexer`
 - Added `index_rbs_uri()` to `GraphTest` as a separate method from `index_uri()` due to in-memory sources
 - Extracted shared test macros from `ruby_indexer` to `local_graph_test.rs` so both indexers can reuse them
   - Diagnostic macros were renamed to `assert_local_diagnostics_eq!` / `assert_no_local_diagnostics!` to avoid conflicting with the `GraphTest` versions in `graph_test.rs`

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -155,41 +155,10 @@ impl Visit for RBSIndexer<'_> {
 #[cfg(test)]
 mod tests {
     use crate::test_utils::LocalGraphTest;
-    use crate::{assert_def_name_eq, assert_def_name_offset_eq, assert_definition_at};
-
-    // Diagnostic assertions
-
-    fn format_diagnostics(context: &LocalGraphTest) -> Vec<String> {
-        let mut diagnostics = context.graph().diagnostics().iter().collect::<Vec<_>>();
-
-        diagnostics.sort_by_key(|d| d.offset());
-        diagnostics
-            .iter()
-            .map(|d| d.formatted(context.graph().document()))
-            .collect()
-    }
-
-    macro_rules! assert_diagnostics_eq {
-        ($context:expr, $expected_diagnostics:expr) => {{
-            assert_eq!(
-                $expected_diagnostics,
-                format_diagnostics($context).as_slice(),
-                "diagnostics mismatch: expected `{:?}`, got `{:?}`",
-                $expected_diagnostics,
-                format_diagnostics($context)
-            );
-        }};
-    }
-
-    macro_rules! assert_no_diagnostics {
-        ($context:expr) => {{
-            assert!(
-                $context.graph().diagnostics().is_empty(),
-                "expected no diagnostics, got {:?}",
-                format_diagnostics($context)
-            );
-        }};
-    }
+    use crate::{
+        assert_def_name_eq, assert_def_name_offset_eq, assert_definition_at, assert_local_diagnostics_eq,
+        assert_no_local_diagnostics,
+    };
 
     fn index_source(source: &str) -> LocalGraphTest {
         LocalGraphTest::new_rbs("file:///foo.rbs", source)
@@ -199,7 +168,7 @@ mod tests {
     fn index_source_with_errors() {
         let context = index_source("module");
 
-        assert_diagnostics_eq!(&context, ["parse-error: Failed to parse RBS document (1:1-1:1)"]);
+        assert_local_diagnostics_eq!(&context, ["parse-error: Failed to parse RBS document (1:1-1:1)"]);
 
         assert!(context.graph().definitions().is_empty());
     }
@@ -215,7 +184,7 @@ mod tests {
             "
         });
 
-        assert_no_diagnostics!(&context);
+        assert_no_local_diagnostics!(&context);
         assert_eq!(context.graph().definitions().len(), 2);
 
         assert_definition_at!(&context, "1:1-4:4", Module, |def| {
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/offset.rs
```diff
@@ -47,6 +47,23 @@ impl Offset {
...
+    /// # Panics
+    ///
+    /// This function can panic if the RBS location offsets are negative
+    #[must_use]
+    pub fn from_rbs_location(location: &ruby_rbs::node::RBSLocationRange) -> Self {
+        Self::new(
+            location
+                .start()
+                .try_into()
+                .expect("RBS location start offset should not be negative"),
```
vinistock: I would make the message a little more generic because this can both fail if it's negative or if it exceeds 32 bits.

```suggestion
                .expect("RBS location start offset should fit into u32"),
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,217 @@
...
+        if module_node.members().iter().next().is_some() {
+            self.nesting_stack.push(definition_id);
+            node::visit_module_node(self, module_node);
+            self.nesting_stack.pop();
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vinistock, line 156: The test scenarios here are valid, but they are mixing expectations of indexing and resolution. Let's try to keep these tests exclusively about indexing data and move resolution data (ownership, fully qualified names, owned members) to `resolution.rs`.

These tests should never invoke `context.resolve()` and should instead assert the structure of the indexed data. Take a look at the tests in `ruby_indexer`, there are some handy assertion macros to verify the data.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,217 @@
...
+        let definition_id = self.local_graph.add_definition(definition);
+        if let Some(id) = lexical_nesting_id {
+            self.add_member_to_current_lexical_scope(id, definition_id);
+        }
+
+        if module_node.members().iter().next().is_some() {
+            self.nesting_stack.push(definition_id);
+            node::visit_module_node(self, module_node);
+            self.nesting_stack.pop();
+        }
```
vinistock: Could we do something like this instead?

```suggestion
        self.nesting_stack.push(definition_id);
        
        for member in &module_node.members() {
            self.visit(member);
        }
        
        self.nesting_stack.pop();
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,217 @@
...
+
+        let context = index("");
+        assert_eq!(baseline_count, context.graph().declarations().len());
+    }
+
+    #[test]
+    fn qualified_name_inside_nested_module() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rbs", "module Outer\n  module Foo\n  end\nend");
+        context.index_uri("file:///test.rbs", "module Outer\n  module Foo::Bar\n  end\nend");
```
vinistock: The `index_uri` method has special logic to erase indentation, so that we can write these examples using multiline strings ([example](https://github.com/Shopify/rubydex/blob/77d639f951a2bc8b6248e766b21bb1274a64372a/rust/rubydex/src/indexing/ruby_indexer.rs#L2509)).

Let's use multiline strings so tests are more readable.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,217 @@
...
+
+    #[test]
+    fn parse_error() {
+        let context = index("module");
+        assert_declaration_does_not_exist!(context, "module");
+        assert!(!context.format_diagnostics(&[]).is_empty());
+    }
+
+    #[test]
+    fn empty_source() {
```
vinistock: This test can probably be removed.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,217 @@
...
+
+    fn index(source: &str) -> GraphTest {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///test.rbs", source);
+        context.resolve();
+        context
+    }
+
+    #[test]
+    fn simple_module() {
```
vinistock: Let's make sure we're testing these two cases:

```ruby
# The name Foo has no parent scope or nesting
module Foo
  # The name Bar has no parent scope and nesting Foo
  # The name Baz has parent scope Bar and nesting Foo
  class Bar::Baz
  end
end
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -1,29 +1,270 @@
...
+
+    macro_rules! assert_no_diagnostics {
+        ($context:expr) => {{
+            assert!(
+                $context.graph().diagnostics().is_empty(),
+                "expected no diagnostics, got {:?}",
+                format_diagnostics($context)
+            );
+        }};
+    }
```
vinistock: We already have these assertions and helper method in `ruby_indexer`. Can we move this to `local_graph_test` like we did with the rest of the assertions so that it can be shared by both?

alexcrocha: They're conflicting in name with `graph_test.rb`'s exported macros, so I chose to duplicate and let `macro_rules!` shadow them locally. 

But I can extract to `local_graph_test` with a new names (`assert_no_local_graph_diagnostics`?) and update all the places we use them.

Let me know what you think.

alexcrocha: I extracted the macros to deduplicate the logic. Also inlined the `format_diagnostics` helper into the macros to avoid having to `pub(crate)`the module.

PR #592, opened 2026-02-23
Add integrity check for owner relationships

vinistock:
**Note**: this PR starts running resolution on our top 100 gems check, which fails until #505 is shipped.

This PR adds an integrity check back, so that we can perform more expensive validation of the graph to catch bugs early.

This first iteration only validates ownership relations. The implemented rules are:

- Only `Object` and `BasicObject` can own themselves
- All owners are namespaces
- All singleton classes are eventually attached to a namespace that isn't a singleton class (which would create infinite recursion in certain parts of the algorithms)

I exposed this API to Ruby, so that we can start running it against the top 100 gems.

### DIFF
--- a/lib/rubydex/failures.rb
+++ b/lib/rubydex/failures.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Rubydex
+  class Failure
+    #: String
+    attr_reader :message
+
+    #: (String) -> void
+    def initialize(message)
+      @message = message
+    end
+  end
+
+  class IntegrityFailure < Failure; end
+end
--- a/rbi/rubydex.rbi
+++ b/rbi/rubydex.rbi
@@ -204,6 +204,9 @@ module Rubydex
     sig { returns(T::Array[String]) }
     def workspace_paths; end
 
+    sig { returns(T::Array[Failure]) }
+    def check_integrity; end
+
     private
 
     # Gathers the paths we have to index for all workspace dependencies
@@ -235,7 +238,7 @@ module Rubydex
     def end_line; end
 
     sig { returns(String) }
-    def path; end
+    def to_file_path; end
 
     sig { returns(Integer) }
     def start_column; end
--- a/rust/rubydex/src/integrity.rs
+++ b/rust/rubydex/src/integrity.rs
@@ -1,0 +1,278 @@
+use std::fmt;
+
+use crate::model::{
+    declaration::{Declaration, Namespace},
+    graph::{BASIC_OBJECT_ID, Graph, OBJECT_ID},
+    ids::DeclarationId,
+};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum IntegrityErrorKind {
+    /// A declaration's owner is not a namespace (module, class, or singleton class)
+    OwnerIsNotNamespace,
+    /// A declaration's owner does not exist in the graph
+    OwnerDoesNotExist,
+    /// A singleton class chain never resolves to a non-singleton namespace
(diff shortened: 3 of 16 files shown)


rust/rubydex/src/integrity.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        write!(f, "{message}. Defined in: {}", self.uris.join(", "))
+    }
+}
+
+impl std::error::Error for IntegrityError {}
+
+/// Checks the integrity of the graph data
+#[must_use]
+pub fn check_integrity(graph: &Graph) -> Vec<IntegrityError> {
```
st0012, line 60: Should we have some tests for this?

rust/rubydex/src/main.rs
```diff
@@ -98,6 +101,21 @@ fn main() {
...
     }
 
+    // Integrity check
+    if args.check_integrity {
+        let errors = time_it!(integrity_check, { integrity::check_integrity(&graph) });
+
+        if errors.is_empty() {
+            println!("Integrity check passed: no issues found");
+        } else {
+            eprintln!("Integrity check found {} issue(s):", errors.len());
```
st0012, line 111: Should we exit with status 1 in this case? We can make it build-type dependent too (e.g. only fail at debug build)

lib/rubydex/failures.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Rubydex
+  class Error
+    #: String
+    attr_reader :message
+
+    #: (String) -> void
+    def initialize(message)
+      @message = message
```
st0012, line 10: Should we call `super(message)` here?

lib/rubydex/errors.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Rubydex
+  class Error
```
Morriar: This is reopening https://github.com/Shopify/rubydex/blob/main/lib/rubydex.rb#L7

vinistock: Yeah, not intentionally. I removed the one inheriting from `StandardError`.

lib/rubydex/failures.rb
vinistock: This class doesn't inherit from anything.

lib/rubydex/errors.rb
st0012: I think it's not conventional to have a class named `Error` but not inherit from `StandardError` and don't act like a Ruby exception.

Maybe rename it to like `Failure` and use that as the postfix, like `IntegrityCheckFailure`?

vinistock: Yeah, that's fair. I renamed to `Failure`.

lib/rubydex/failures.rb
st0012, line 1: After the rename, this should be `failure.rb`?

vinistock: Ah, yeah, indeed.

rbi/rubydex.rbi
```diff
@@ -204,6 +204,9 @@ module Rubydex
     sig { returns(T::Array[String]) }
     def workspace_paths; end
 
+    sig { void }
+    def check_integrity; end
```
alexcrocha: Hmm, shouldn't we be returning an array of IntegrityFailure objects or an empty array here?

alexcrocha (approved):

rbi/rubydex.rbi
vinistock: Yes, you're correct!

PR #595, opened 2026-02-24
Add completion for namespace access

vinistock:
This PR implements completion candidates for namespace access (e.g.: `Foo::`). This scenario returns all constants and singleton methods accessible in the ancestor chain.

**Note**: there's an interesting behaviour in Ruby concerning inherited constants from `Object` and forward. They are not available when accessing namespaces even if present in the ancestor chain.

For example:

```ruby
class Foo
end

Foo::String
```

You would expect `Foo::String` to be valid. `Foo` inherits from `Object` and `String` belongs to `Object`. However, Ruby actually cuts the ancestor chain short and does not include inherited constants for ancestor after `Object`. This is captured in the implementation and in tests.

### DIFF
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -142,6 +142,9 @@ pub enum CompletionReceiver {
     /// Completion requested for an expression with no previous token (e.g.: at the start of a line with nothing before)
     /// Includes: all keywords, all global variables and reacheable instance variables, class variables, constants and methods
     Expression(NameId),
+    /// Completion requested after a namespace access operator (e.g.: `Foo::`)
+    /// Includes: all constants and singleton methods for the namespace and its ancestors
+    NamespaceAccess(DeclarationId),
 }
 
 pub struct CompletionContext<'a> {
@@ -197,8 +200,6 @@ macro_rules! collect_candidates {
 ///   resolves to
 /// - Method calls on anything (e.g.: `foo.`, `@bar.`, `@@baz.`, `Qux.`) collects all methods that exist on the type
 ///   returned by the receiver
-/// - Require path completion collects all require paths accessible from the `$LOAD_PATH`
-/// - Relative require path completion collects all require paths accessible from the directory of the current file
 ///
 /// # Panics
 ///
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/query.rs
```diff
@@ -787,4 +837,214 @@ mod tests {
...
+    }
+
+    #[test]
+    fn namespace_access_completion_includes_inherited_members() {
+        let mut context = GraphTest::new();
+
+        context.index_uri(
+            "file:///foo.rb",
+            "
+            class Parent
```
st0012, line 882: Can we have a test case similar to this but for module?

```rb
module Helpers
  HELPER_CONST = 1

  class << self
    def helper_method; end
  end
end

module Foo
  include Helpers

  CONST = 1

  class << self
    def class_method; end
  end
end

# Foo's completion candidates should have Foo::CONST, Foo::HELPER_CONST, Foo.class_method
# But it should not include Foo.helper_method
```

rust/rubydex/src/query.rs
```diff
@@ -213,9 +214,58 @@ pub fn completion_candidates<'a>(
...
+                Declaration::Namespace(_) | Declaration::Constant(_) | Declaration::ConstantAlias(_)
+            );
+        }
+    }
+
+    // Collect singleton methods from the singleton class and its ancestors
+    if let Some(singleton_id) = namespace.singleton_class() {
+        let singleton = graph.declarations().get(singleton_id).unwrap().as_namespace().unwrap();
+
+        for ancestor in singleton.ancestors() {
```
Morriar, line 258: I wonder if the singleton method ancestor walk should also stop at `Object`'s singleton class?

The constant collection loop above breaks at `Object` to avoid surfacing inherited constants (e.g. `Foo::String`). Singleton methods inherited from `Object`'s singleton (like `new`, `allocate`, `superclass`, etc.) would still be collected here since there's no equivalent cutoff. Is that intentional? If so, a short comment explaining the asymmetry would help.

vinistock: This is intentional because of Ruby semantics. Ruby prohibits you from referencing a constant inherited from `Object` (a.k.a. a top level level constant).

This crashes:
```ruby
class Foo
end

# Foo inherits from Object and Object owns String
# but this is not allowed anyway
Foo::String
```

This is not the case for singleton methods. You can invoke singleton methods defined in `Object` or higher up the chain. I think the current behaviour does make sense.

PR #596, opened 2026-02-24
Add method call with receiver completion

vinistock:
This PR adds completion for method calls with a receiver. We can only provide this type of completion if we know what's the type of the receiver, so this scenario expects a declaration ID.

If the receiver is of an unknown type, there's no way to provide any completion.

### DIFF
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -145,6 +145,10 @@ pub enum CompletionReceiver {
     /// Completion requested after a namespace access operator (e.g.: `Foo::`)
     /// Includes: all constants and singleton methods for the namespace and its ancestors
     NamespaceAccess(DeclarationId),
+    /// Completion requested after a method call operator (e.g.: `foo.`, `@bar.`, `@@baz.`, `Qux.`).
+    /// In the case of singleton completion (e.g.: `Foo.`), the declaration ID should be for the singleton class (i.e.: `Foo::<Foo>`)
+    /// Includes: all methods that exist on the type of the receiver and its ancestors
+    MethodCall(DeclarationId),
 }
 
 pub struct CompletionContext<'a> {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/query.rs
```diff
@@ -145,6 +145,9 @@ pub enum CompletionReceiver {
     /// Completion requested after a namespace access operator (e.g.: `Foo::`)
     /// Includes: all constants and singleton methods for the namespace and its ancestors
     NamespaceAccess(DeclarationId),
+    /// Completion requested after a method call operator (e.g.: `foo.`, `@bar.`, `@@baz.`, `Qux.`)
```
Morriar: The doc comment mentions `Qux.` as an example for `MethodCall`, but `Qux.` in Ruby typically means calling a **class (singleton) method**, not an instance method. The function only collects instance methods from the namespace's ancestors — so for `Qux.something` to work correctly, the caller would need to pass the singleton class `DeclarationId`, not `Qux`'s `DeclarationId` directly.

I wonder if we should clarify this in the doc comment? Something like:

```suggestion
    /// Completion requested after a method call operator (e.g.: `foo.`, `@bar.`, `@@baz.`)
    /// Includes: all methods that exist on the type of the receiver and its ancestors.
    /// Note: for constant receivers like `Qux.`, the caller must pass the singleton class DeclarationId
    /// to collect singleton/class methods.
    MethodCall(DeclarationId),
```

Otherwise someone passing `Qux`'s `DeclarationId` for a `Qux.bar_class_method` completion will get instance methods, which is likely not what they want.

Morriar (approved):
We should to clarify what happens on singleton class receiver and possibly add a test showing the current behavior even if incorrect.

rust/rubydex/src/query.rs
vinistock: The behaviour was already correct, but I clarified in the docs that we expect the declaration ID for the singleton and added a new test.

PR #597, opened 2026-02-24
Simplify completion tests with assertion macro

vinistock:
Our tests were getting long, with repeated logic. This PR adds an `assert_completion_eq!` macro and simplifies all tests.

### DIFF
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -412,6 +412,26 @@ mod tests {
         };
     }
 
+    /// Asserts that the completion candidates match the expected list
+    macro_rules! assert_completion_eq {
+        ($context:expr, $receiver:expr, $expected:expr) => {
+            assert_eq!(
+                $expected,
+                *completion_candidates($context.graph(), CompletionContext::new($receiver))
+                    .unwrap()
+                    .iter()
+                    .map(|id| $context
+                        .graph()
+                        .declarations()
+                        .get(id)
+                        .unwrap()
+                        .name()
+                        .to_string())
+                    .collect::<Vec<_>>()
+            );
+        };
+    }
+
     #[test]
     fn fuzzy_search_returns_partial_matches() {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/query.rs
```diff
@@ -807,20 +779,11 @@ mod tests {
...
-        .map(|id| context.graph().declarations().get(id).unwrap().name().to_string())
-        .collect::<Vec<_>>();
-
-        // Foo::CONST is reachable from Foo::Bar through lexical scoping, so it must
-        // appear as a completion candidate when the user types the unqualified name CONST
-        assert!(
-            candidates.contains(&"Foo::CONST".to_string()),
-            "Expected Foo::CONST in candidates, got: {candidates:?}",
+        // Foo::CONST is reachable from Foo::Bar through lexical scoping
+        assert_completion_eq!(
```
alexcrocha, line 785: We're losing some context by truncating this comment. Flagging just in case this wasn't intentional

rust/rubydex/src/query.rs
```diff
@@ -565,16 +585,9 @@ mod tests {
...
-        .unwrap()
-        .iter()
-        .map(|id| context.graph().declarations().get(id).unwrap().name().to_string())
-        .collect::<Vec<_>>();
-
-        assert_eq!(
+        assert_completion_eq!(
+            context,
+            CompletionReceiver::Expression(name_id),
             vec![
```
Morriar: You can also simplify the `vec!` like in https://github.com/Shopify/rubydex/pull/526

rust/rubydex/src/query.rs
vinistock: Good catch. Added it back.

rust/rubydex/src/query.rs
vinistock: Nice. Turned them all into simple arrays.

PR #598, opened 2026-02-24
Provide completion for expressions inside call arguments

vinistock:
This PR adds completion for expressions inside of call arguments. This is a special case of expression completion. It includes everything that you would expect from expression completion + any keyword parameters that the user may refer to.

### DIFF
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -6,6 +6,7 @@ use std::thread;
 use url::Url;
 
 use crate::model::declaration::{Ancestor, Declaration};
+use crate::model::definitions::{Definition, Parameter};
 use crate::model::graph::{Graph, OBJECT_ID};
 use crate::model::identity_maps::IdentityHashSet;
 use crate::model::ids::{DeclarationId, NameId, StringId, UriId};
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/query.rs
```diff
@@ -373,6 +390,36 @@ fn expression_completion<'a>(
...
+    method_decl_id: DeclarationId,
+    context: CompletionContext<'a>,
+) -> Result<Vec<CompletionCandidate>, Box<dyn Error>> {
+    let mut candidates = expression_completion(graph, self_name_id, context)?;
+    let Some(method_decl) = graph.declarations().get(&method_decl_id) else {
+        return Ok(candidates);
+    };
+
+    // Find the first Method definition to extract keyword parameters
+    for def_id in method_decl.definitions() {
```
Morriar, line 407: Let's add a few tests showing we only consider the "first" definitions we indexed:

```rb
def foo(a, b, c); end # we show this one
def foo(x, y); end 
```

Also, are sure the indexing order for the same method in different document is stable?

```rb
# file1.rb
def foo(a, b, c); end # we show this one
```

```rb
# file2.rb
def foo(x, y); end 
```

We should always pick the same. Do we need to sort the `definitions()` by document URI?

vinistock: Added a test for this. We always return the same definition because the hashing of the ID is deterministic, which means that we find the same entry in the hashmap when iterating. I don't believe we need to sort.

rust/rubydex/src/query.rs
```diff
@@ -1187,4 +1239,112 @@ mod tests {
...
+        context.index_uri(
+            "file:///foo.rb",
+            "
+            class Foo
+              def bar(first:, second:); end
+            end
+            ",
+        );
+        context.index_uri(
+            "file:///foo.rb",
```
st0012: Should this be `bar.rb`? Or we want to test same-file override?

vinistock: Yes, that's right. Good catch.

vinistock:
### Merge activity

* **Mar 3, 9:55 PM UTC**: @vinistock merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/598).

PR #599, opened 2026-02-24
Fix tag matchers for beta release

vinistock:
Our versioning format is actually `0.1.0.beta5` and not `0.1.0-beta5`.


.github/workflows/cibuildgem.yaml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     tags:
       - v[0-9]+.[0-9]+.[0-9]+
-      - v[0-9]+.[0-9]+.[0-9]+-beta[0-9]*
+      - v[0-9]+.[0-9]+.[0-9]+.beta[0-9]*
```
amomchilov, line 9: That'll be a regex wildcard, which will still match `0.1.0-beta5`.

If you want a literal dot, you need to escape it:

```suggestion
      - v[0-9]+.[0-9]+.[0-9]+\.beta[0-9]*
```

st0012: GH workflow uses glob instead of regexp ([learned this recently](https://github.com/Shopify/ruby-lsp/commit/ec5eba7bd7121ac014b37cc286b99658dd10005a))

amomchilov: Ah, I didn't know `+` was valid in globs, too.

PR #600, opened 2026-02-24
Follow aliases for completion

vinistock:
This PR starts following alias targets for completion. If we have `ALIAS.` or `ALIAS::`, we need to search the alias target for candidates, not the `ConstantAlias` itself.

My idea was to add a method that keeps tracing alias targets until it finds something that isn't an alias. Then we use that for completion.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -457,6 +457,38 @@ impl Graph {
         Some(targets)
     }
 
+    /// Resolves a constant alias chain to the final non-alias declaration.
+    ///
+    /// Returns `None` if the declaration is not a constant alias, the chain is circular, or the chain leads to an
+    /// unresolved name.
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -1606,4 +1638,51 @@ mod tests {
...
+    }
+
+    #[test]
+    fn resolve_alias_returns_none_for_non_alias() {
+        let mut context = GraphTest::new();
+
+        context.index_uri("file:///foo.rb", "class Foo; end");
+        context.resolve();
+
+        assert_eq!(context.graph().resolve_alias(&DeclarationId::from("Foo")), None);
```
Morriar:
```suggestion
        assert!(context.graph().resolve_alias(&DeclarationId::from("Foo")).is_none());
```

PR #603, opened 2026-02-25
Make locations zero based with `to_display` method

vinistock:
While trying to adopt Rubydex in the LSP, I noticed that all of the location objects are adding `+1` to all location values, which would require the LSP to then `-1` all of them back.

This PR makes locations zero based and instead provides a `to_display` method which returns a one based version of them. That way, tools that need display location can still get it easily.

### DIFF
--- a/lib/rubydex/location.rb
+++ b/lib/rubydex/location.rb
@@ -39,6 +39,19 @@ module Rubydex
       a <=> b
     end
 
+    # Turns this zero based location into a one based location for display purposes.
+    #
+    #: () -> Location
+    def to_display
+      self.class.new(
+        uri: @uri,
+        start_line: @start_line + 1,
+        end_line: @end_line + 1,
+        start_column: @start_column + 1,
+        end_column: @end_column + 1,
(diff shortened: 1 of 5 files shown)


lib/rubydex/location.rb
```diff
@@ -39,6 +39,19 @@ def <=>(other)
       a <=> b
     end
 
+    # Turns this zero based location into a one based location for display purposes.
+    #
+    #: () -> Location
```
Morriar, line 44: After this PR, `to_s` outputs zero-based values, which makes it not display-friendly anymore. I wonder if we should either:

1. Update `to_s` to use `to_display` internally (so it stays human-readable by default), or
2. Document clearly that `to_s` is now zero-based and intended for internal use

Right now callers expecting `path:line:col` for display will silently get off-by-one output if they use `to_s` directly.

vinistock: I agree. What do you think of this approach? https://github.com/Shopify/rubydex/pull/607

PR #604, opened 2026-02-25
Fix handling of null byte like terminators

vinistock:
To my surprise you can use `\0` as a string terminator in Ruby. I found out when running Ruby LSP tests against Prism fixtures, which made `index_source` crash.

To be able to handle this case correctly, we need the pointer and the length to instantiate the string on the Rust side, otherwise it considers the first `\0` as the string's terminator and we crash.

I also improved the error messages. We were always showing the same message no matter what problem happened, which isn't a nice API.

### DIFF
--- a/test/graph_test.rb
+++ b/test/graph_test.rb
@@ -495,6 +495,21 @@ class GraphTest < Minitest::Test
     refute_nil(graph["Bar"])
   end
 
+  def test_index_source_with_invalid_source_encoding
+    graph = Rubydex::Graph.new
+    error = assert_raises(ArgumentError) do
+      graph.index_source("file:///test.rb", "\xFF\xFE".b, "ruby")
+    end
+    assert_match(/source is not valid UTF-8/, error.message)
+  end
+
+  def test_index_source_with_null_bytes_in_source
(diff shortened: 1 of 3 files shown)


test/graph_test.rb
```diff
@@ -545,6 +545,29 @@ def test_workspace_paths
...
 
+  def test_index_source_with_invalid_source_encoding
+    graph = Rubydex::Graph.new
+    error = assert_raises(ArgumentError) do
+      graph.index_source("file:///test.rb", "\xFF\xFE".b, "ruby")
+    end
+    assert_match(/source is not valid UTF-8/, error.message)
+  end
+
+  def test_index_source_with_invalid_language_id
```
Morriar: I think this duplicates `test_index_source_with_unknown_language_id` (~line 456) — both use `"python"` and assert the same error message. Any way we could either remove this one or make it test something distinct (e.g. a `language_id` that actually fails UTF-8 conversion rather than just being unrecognised)?

vinistock: You're right, I missed the original test for this. I removed this one and grouped all of the `index_source` tests together.

PR #605, opened 2026-02-25
Add `Declaration#resolve_member` to follow ancestors

vinistock:
This PR adds `Declaration#resolve_member`, which searches the ancestors of the declaration for the given member. This is a super common operation in the LSP, so we might as well avoid so much duplication and just provide a way to do this neatly.

We need the `only_inherited` option to search only ancestors after the main namespace, which is what we use to provide go to definition for `super`.

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -169,6 +169,42 @@ static VALUE rdxr_declaration_member(VALUE self, VALUE name) {
     return rb_class_new_instance(2, argv, decl_class);
 }
 
+// Namespace#find_member: (String member, only_inherited: false) -> Declaration?
+// Searches for a member in the ancestor chain of the declaration
+static VALUE rdxr_declaration_find_member(int argc, VALUE *argv, VALUE self) {
+    VALUE member, opts;
+    rb_scan_args(argc, argv, "1:", &member, &opts);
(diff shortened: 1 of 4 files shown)


ext/rubydex/declaration.c
```diff
@@ -169,6 +169,42 @@ static VALUE rdxr_declaration_member(VALUE self, VALUE name) {
     return rb_class_new_instance(2, argv, decl_class);
 }
 
+// Namespace#resolve_member: (String member, only_inherited: false) -> Declaration?
+// Searches for a member in the ancestor chain of the declaration
+static VALUE rdxr_declaration_resolve_member(int argc, VALUE *argv, VALUE self) {
```
vinistock: Rename to `find_member`.

ext/rubydex/declaration.c
```diff
@@ -277,6 +313,7 @@ void rdxi_initialize_declaration(VALUE mRubydex) {
 
     // Namespace only methods
     rb_define_method(cNamespace, "member", rdxr_declaration_member, 1);
+    rb_define_method(cNamespace, "resolve_member", rdxr_declaration_resolve_member, -1);
```
st0012: Let's rename it to `find_member` as we're not actually resolving anything?
(Discussed in pairing)

PR #606, opened 2026-02-26
Index classes with RBS indexer

alexcrocha:
One more step towards #87.

This PR adds class support to the RBS indexer, including superclass references. `visit_class_node` follows the same structure as `visit_module_node`, the only difference is that we're also building a `ConstantReference` for the superclass if it is present.

Along the way:
  - Extracted `nesting_name_id` and `register_definition` helpers shared by both visitors
  - Moved `assert_def_superclass_ref_eq!` to `local_graph_test.rs` so both indexers can use it
  - Modified the RBS resolution tests with scenarios that mix modules, classes, and superclass resolution

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -1,14 +1,15 @@
 //! Visit the RBS AST and create type definitions.
 
-use ruby_rbs::node::{self, ModuleNode, Node, TypeNameNode, Visit};
+use ruby_rbs::node::{self, ClassNode, ModuleNode, Node, TypeNameNode, Visit};
 
 use crate::diagnostic::Rule;
 use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
-use crate::model::definitions::{Definition, DefinitionFlags, ModuleDefinition};
+use crate::model::definitions::{ClassDefinition, Definition, DefinitionFlags, ModuleDefinition};
 use crate::model::document::Document;
 use crate::model::ids::{DefinitionId, NameId, UriId};
 use crate::model::name::{Name, ParentScope};
+use crate::model::references::ConstantReference;
 use crate::offset::Offset;
 
 pub struct RBSIndexer<'a> {
@@ -87,6 +88,17 @@ impl<'a> RBSIndexer<'a> {
         self.nesting_stack.last().copied()
     }
 
+    fn nesting_name_id(&self, lexical_nesting_id: Option<DefinitionId>) -> Option<NameId> {
+        lexical_nesting_id.map(|id| {
+            let owner = self
+                .local_graph
+                .definitions()
+                .get(&id)
+                .expect("owner definition should exist");
+            *owner.name_id().expect("nesting definition should have a name")
+        })
+    }
+
     fn add_member_to_current_lexical_scope(&mut self, owner_id: DefinitionId, member_id: DefinitionId) {
         let owner = self
             .local_graph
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -99,20 +111,72 @@ impl<'a> RBSIndexer<'a> {
...
+        if let Some(id) = lexical_nesting_id {
+            self.add_member_to_current_lexical_scope(id, definition_id);
+        }
+        definition_id
+    }
 }
 
 impl Visit for RBSIndexer<'_> {
-    fn visit_module_node(&mut self, module_node: &ModuleNode) {
+    fn visit_class_node(&mut self, class_node: &ClassNode) {
```
alexcrocha, line 129: Ignore this Github diff, I am not renaming this. Github is showing it getting added back below 🤦

rust/rubydex/src/test_utils/local_graph_test.rs
```diff
@@ -251,6 +251,42 @@ macro_rules! assert_def_name_eq {
     }};
 }
 
+/// Asserts that a definition's superclass reference matches the expected name.
+///
+/// Usage:
+/// - `assert_def_superclass_ref_eq!(ctx, def, "Bar::Baz")`
```
Morriar, line 257: The doc comment says `"Bar::Baz"` is supported, but the implementation only resolves the leaf string via `.str()` — it won't match a qualified path like `Bar::Baz`, only the leaf `Baz`.

I wonder if we should use `assert_name_path_eq!` here instead to actually support qualified names?

```suggestion
/// - `assert_def_superclass_ref_eq!(ctx, def, "Bar::Baz")` - asserts the full path `Bar::Baz`
```

And the macro body could delegate:

```rust
macro_rules! assert_def_superclass_ref_eq {
    ($context:expr, $def:expr, $expected_name:expr) => {{
        let name_id = *$context
            .graph()
            .constant_references()
            .get($def.superclass_ref().unwrap())
            .unwrap()
            .name_id();
        $crate::assert_name_path_eq!($context, $expected_name, name_id);
    }};
}
```

This would also remove all the nested `.get()` chains and make it consistent with how `assert_def_name_eq!` works.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -236,4 +296,80 @@ mod tests {
...
+            "
+            class Foo < Bar
+            end
+            "
+        });
+
+        assert_no_local_diagnostics!(&context);
+
+        assert_definition_at!(&context, "1:1-2:4", Class, |def| {
+            assert_def_name_eq!(&context, def, "Foo");
```
Morriar, line 344: The `assert!(def.superclass_ref().is_some())` check is redundant — `assert_def_superclass_ref_eq!` already calls `.unwrap()` on the superclass_ref and will panic with a clear message if it's `None`.

```suggestion
        assert_definition_at!(&context, "1:1-2:4", Class, |def| {
            assert_def_name_eq!(&context, def, "Foo");
            assert_def_superclass_ref_eq!(&context, def, "Bar");
        });
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -99,20 +111,72 @@ impl<'a> RBSIndexer<'a> {
...
+    ) -> DefinitionId {
+        let definition_id = self.local_graph.add_definition(definition);
+        if let Some(id) = lexical_nesting_id {
+            self.add_member_to_current_lexical_scope(id, definition_id);
+        }
+        definition_id
+    }
 }
 
 impl Visit for RBSIndexer<'_> {
```
Morriar, line 128: The comment collection logic is identical between `visit_class_node` and `visit_module_node`. Any way we could extract it to a helper?

```rust
fn collect_comments(&self, comment: Option<impl AsRef<[u8]> + ...>) -> Vec<Comment> {
    // ...
}
```

Not a blocker, just noting the duplication for a follow-up.

alexcrocha: I'll extract it to a helper next PR I use it 👍

rust/rubydex/src/test_utils/local_graph_test.rs
alexcrocha: Opened a quick PR to address this https://github.com/Shopify/rubydex/pull/612

PR #607, opened 2026-02-26
Create `DisplayLocation`

vinistock:
Addressing https://github.com/Shopify/rubydex/pull/603#discussion_r2859879577

I agree that that it's not very easy to reason about zero based locations. I think one way to improve the behaviour is with a `DisplayLocation` object, which is essentially a regular location with overrides.

This allows us to:

- Always make `to_s` display user friendly locations regardless of whether the underlying data is zero or one based
- Prevents silly API mistakes like invoking `to_display` more than once (which will keep bumping all data)

### DIFF
--- a/test/location_test.rb
+++ b/test/location_test.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class LocationTest < Minitest::Test
+  def test_location_equality
+    a = Rubydex::Location.new(uri: "file:///foo.rb", start_line: 0, end_line: 0, start_column: 0, end_column: 5)
+    b = Rubydex::Location.new(uri: "file:///foo.rb", start_line: 0, end_line: 0, start_column: 0, end_column: 5)
+
+    assert_equal(0, a <=> b)
+    assert_equal(a, b)
+  end
+
+  def test_location_ordering_by_uri
+    a = Rubydex::Location.new(uri: "file:///a.rb", start_line: 0, end_line: 0, start_column: 0, end_column: 0)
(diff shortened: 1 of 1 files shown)


lib/rubydex/location.rb
```diff
@@ -52,6 +54,22 @@ def to_display
...
 
+    #: -> String
+    def to_s
+      "#{path}:#{@start_line + 1}:#{@start_column + 1}-#{@end_line + 1}:#{@end_column + 1}"
+    end
+  end
+
+  # A one based location intended for display purposes. This is what should be used when displaying a location to users,
+  # like in CLIs
+  class DisplayLocation < Location
```
Morriar, line 68: Let's add dedicated tests for `DisplayLocation` please:

1. That `to_display` is idempotent (returns `self`)
2. That values aren't double-bumped when calling `to_display` on an existing `DisplayLocation`
3. `to_s` output format

vinistock: Done!

lib/rubydex/location.rb
```diff
@@ -52,6 +57,29 @@ def to_display
...
+    #
+    #: () -> DisplayLocation
+    def to_display
+      self
+    end
+
+    # Normalize to zero-based for comparison with Location
+    #
+    #: () -> [String, Integer, Integer, Integer, Integer]
+    def comparable_values
```
Morriar: I'm wondering if we should validate that `DisplayLocation` is always constructed with 1-based values. Right now `comparable_values` subtracts 1 unconditionally, so if someone directly constructs `DisplayLocation.new(start_line: 0, ...)`, the comparison values go negative and sorting breaks.

Is direct construction of `DisplayLocation` intended to be supported, or should it only come from `Location#to_display`? If the latter, maybe a private constructor or at least a note in the class comment would help?

PR #609, opened 2026-02-26
Include RBS paths when indexing workspace

vinistock:
Now that we have some RBS indexing, we can automatically include the RBS paths for `core` and `stdlib` to get the complete data when using `index_workspace`. In the future, we may want to ship with a fully indexed RBS cache, but for now let's just include the paths.

One design note: we should not require that `rbs` be on the bundle. Instead of using `Gem::Specification.find` or `Gem.find_files` (both of which are scoped to the bundle), I just searched `Gem.path` manually for the latest `rbs` installation.

### DIFF
--- a/test/graph_test.rb
+++ b/test/graph_test.rb
@@ -560,6 +560,35 @@ class GraphTest < Minitest::Test
     end
   end
 
+  def test_index_workspace_includes_rbs_core_definitions
+    graph = Rubydex::Graph.new
+    graph.index_workspace
+    graph.resolve
+
+    ["Kernel", "Object", "BasicObject", "Integer"].each do |core_namespace|
+      rbs_kernel = graph[core_namespace].definitions.find do |definition|
+        uri = URI(definition.location.uri)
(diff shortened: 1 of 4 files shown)


test/graph_test.rb
```diff
@@ -560,6 +560,23 @@ def test_workspace_paths
     end
   end
 
+  def test_index_workspace_includes_rbs_core_definitions
+    # Ensure the `rbs` gem is installed on CI
+    Gem.install("rbs") if ENV["CI"]
```
Morriar: I'm wondering if `Gem.install` in a test is the right approach here — it's a side effect that modifies the system state during a test run.

Should we add it to the `tests` group in the Gemfile instead?

Another solution is to only install the gem in the `context` as we do in Tapioca: https://github.com/Shopify/tapioca/blob/main/spec/tapioca/cli/gem_spec.rb#L120-L125

vinistock: Added to the Gemfile. Groups don't do anything for gems, so I just added it in the default scope.

PR #614, opened 2026-02-27
Add benchmarking instructions to agents

vinistock:
I had good results asking Claude to continuously optimize certain parts of the analysis. To make this process easier, I think we can make the default behaviour of `utils/bench` to use a `DEFAULT_BENCH_WORKSPACE` path.

That way, we can have the instructions be generic without exposing internal paths.

### DIFF
--- a/utils/bench
+++ b/utils/bench
@@ -4,13 +4,16 @@
 # improvements or regressions
 #
 # Usage:
-#   utils/bench              # defaults to huge corpus
+#   utils/bench              # uses $DEFAULT_BENCH_WORKSPACE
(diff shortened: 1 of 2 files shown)


utils/bench
```diff
@@ -69,6 +93,6 @@ fi
 echo "Benchmarking with corpus: $CORPUS_PATH"
 echo "----------------------------------------"
 
-cd "$PROJECT_ROOT/rust" || exit 1
+cd "$PROJECT_ROOT/rust/rubydex" || exit 1
```
vinistock, line 96: This is just to speed up compilation. The MCP server takes a while to compile in release mode and it is not required for this bench script.

PR #615, opened 2026-02-27
Index constants with RBS indexer

alexcrocha:

This PR adds `visit_constant_node` so top-level, qualified, and class-nested constants are indexed.

Also extracts `collect_comments` to deduplicate comment collection across visitors, as suggested in https://github.com/Shopify/rubydex/pull/606#discussion_r2859884590, and moves `assert_def_comments_eq!` to the shared test macros in `local_graph_test`.

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -1,11 +1,11 @@
 //! Visit the RBS AST and create type definitions.
 
-use ruby_rbs::node::{self, ClassNode, ModuleNode, Node, TypeNameNode, Visit};
+use ruby_rbs::node::{self, ClassNode, CommentNode, ConstantNode, ModuleNode, Node, TypeNameNode, Visit};
 
 use crate::diagnostic::Rule;
 use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
-use crate::model::definitions::{ClassDefinition, Definition, DefinitionFlags, ModuleDefinition};
+use crate::model::definitions::{ClassDefinition, ConstantDefinition, Definition, DefinitionFlags, ModuleDefinition};
 use crate::model::document::Document;
 use crate::model::ids::{DefinitionId, NameId, UriId};
 use crate::model::name::{Name, ParentScope};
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -112,6 +112,16 @@ impl<'a> RBSIndexer<'a> {
         }
     }
 
+    fn collect_comments(comment_node: Option<CommentNode>) -> Vec<Comment> {
```
soutaro, line 115: I think we should split the comment into lines. A RBS comment has multi line contents, but a Rubydex comment only has single-line content.

I think I have an implementation for this. Will share it later.

vinistock: Do you mean splitting into multiple lines at the RBS parser level?

The challenge with splitting during indexing is when the comment's content includes a line break, which ends up requiring more careful parsing. E.g.:

```ruby
# something \n
```

rust/rubydex/src/resolution.rs
```diff
@@ -4484,6 +4484,35 @@ mod tests {
...
+
+            Bar::QUX: String
+            "
+        });
+        context.resolve();
+
+        assert_no_diagnostics!(&context);
+
+        assert_declaration_exists!(context, "FOO");
+        assert_declaration_kind_eq!(context, "FOO", "Constant");
```
vinistock, line 4506: Let's assert that this one is owned by `Object` (i.e.: top level).

rust/rubydex/src/resolution.rs
```diff
@@ -4484,6 +4484,35 @@ mod tests {
...
+
+        assert_declaration_exists!(context, "Bar::BAZ");
+        assert_declaration_kind_eq!(context, "Bar::BAZ", "Constant");
+        assert_owner_eq!(context, "Bar::BAZ", "Bar");
+
+        assert_declaration_exists!(context, "Bar::QUX");
+        assert_declaration_kind_eq!(context, "Bar::QUX", "Constant");
```
vinistock, line 4514: Let's assert that `QUX` is owned by `Bar`.

rust/rubydex/src/resolution.rs
```diff
@@ -4484,6 +4484,35 @@ mod tests {
...
+        let mut context = GraphTest::new();
+        context.index_rbs_uri("file:///test.rbs", {
+            r"
+            FOO: String
+
+            class Bar
+              BAZ: Integer
+            end
+
+            Bar::QUX: String
```
Morriar: Just to cover more cases:

```suggestion
            Bar::QUX: ::String
```

PR #616, opened 2026-02-27
Index globals with RBS indexer

alexcrocha:
One more step towards https://github.com/Shopify/rubydex/issues/87.

This PR adds `visit_global_node` so global variable declarations (`$foo: String`) are discovered and resolved as members of `Object`, matching Ruby indexer's behaviour.

PR #617, opened 2026-02-28
Index includes, prepends, and extends with RBS indexer

alexcrocha:

This PR adds `visit_include_node`, `visit_prepend_node`, and `visit_extend_node` with a shared `index_mixin` helper. Mixins resolve into the ancestor chain matching the Ruby indexer.

Also moves `assert_def_mixins_eq!` to the shared test macros in `local_graph_test`.

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -1,17 +1,19 @@
 //! Visit the RBS AST and create type definitions.
 
 use ruby_rbs::node::{
-    self, ClassNode, CommentNode, ConstantNode, GlobalNode, ModuleNode, Node, NodeList, TypeNameNode, Visit,
+    self, ClassNode, CommentNode, ConstantNode, ExtendNode, GlobalNode, IncludeNode, ModuleNode, Node, NodeList,
+    PrependNode, TypeNameNode, Visit,
 };
 
 use crate::diagnostic::Rule;
 use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
-    ClassDefinition, ConstantDefinition, Definition, DefinitionFlags, GlobalVariableDefinition, ModuleDefinition,
+    ClassDefinition, ConstantDefinition, Definition, DefinitionFlags, ExtendDefinition, GlobalVariableDefinition,
+    IncludeDefinition, Mixin, ModuleDefinition, PrependDefinition,
 };
 use crate::model::document::Document;
-use crate::model::ids::{DefinitionId, NameId, UriId};
+use crate::model::ids::{DefinitionId, NameId, ReferenceId, UriId};
 use crate::model::name::{Name, ParentScope};
 use crate::model::references::ConstantReference;
 use crate::offset::Offset;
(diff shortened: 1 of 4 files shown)


alexcrocha:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/rubydex/617?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#617** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#616** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#615** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -486,6 +537,62 @@ mod tests {
...
+
+    #[test]
+    fn index_include_qualified_name() {
+        let context = index_source({
+            "
+            class Foo
+              include Bar::Baz
+            end
+            "
+        });
```
Morriar, line 605: The test name says "qualified name" but `assert_def_mixins_eq\!` only checks the leaf segment (`"Baz"`), not the full `Bar::Baz` path. I wonder if we should assert the full path here to confirm the qualified name resolves correctly — maybe using `assert_name_path_eq\!` on the mixin's reference?

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -486,6 +537,62 @@ mod tests {
         });
```
Morriar, line 555: The Ruby indexer has `index_includes_at_top_level` to confirm top-level mixins are silently ignored. Should we add an equivalent here for the RBS indexer to document that behavior explicitly?

vinistock: Wait, is that the case? If so, we have a bug. Mixin operations at the top level impact `Object`:

```ruby
module Foo; end

include Foo

puts Object.ancestors.inspect
# => [Object, Foo, Kernel, BasicObject]
```

Regardless, we should add the test.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -486,6 +537,62 @@ mod tests {
...
+        });
+    }
+
+    #[test]
+    fn index_multiple_includes() {
+        let context = index_source({
+            "
+            module Foo
+              include Bar
+              include Baz
```
vinistock, line 585: One important test we have in `ruby_indexer` is `include A, B`. The order of inclusion is actually in reverse, so we should test this here too (same for prepend and extend).

rust/rubydex/src/indexing/rbs_indexer.rs
vinistock: Yeah, we should assert on the entire string `Bar::Baz` and not just `Baz`. Although, that's an existing limitation. If this involves updating too many tests, let's do it in a different PR.

alexcrocha: I'll take the suggestion and fix the macro to assert the full qualified path in a follow-up PR 👍

rust/rubydex/src/indexing/rbs_indexer.rs
alexcrocha: Including at the top level returns a parse error. 

In RBS docs, mixins are not part of the [declarations](https://github.com/ruby/rbs/blob/master/docs/syntax.md#declarations) and I believe we're hitting the default case [here](https://github.com/ruby/rbs/blob/3a9c363ea23cf0c4c0f4b2485ccadda11764b272/src/parser.c#L3007).

Doesn't look like this is an issue for the RBS indexer

rust/rubydex/src/indexing/rbs_indexer.rs
alexcrocha: I double checked and RBS mixins don't support multi-argument. It needs separate include lines ([doc](https://github.com/ruby/rbs/blob/master/docs/syntax.md#mixin-include-mixin-extend-mixin-prepend), [impl](https://github.com/ruby/rbs/blob/3a9c363ea23cf0c4c0f4b2485ccadda11764b272/src/parser.c#L2061))

vinistock: Nice.

PR #618, opened 2026-03-02
Support `deprecated` annotation

soutaro:
RBS has `%a{deprecated}` and `%a{deprecated: message}` annotation. This PR is to import the flags for class and module declarations.

#87

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -1,6 +1,8 @@
 //! Visit the RBS AST and create type definitions.
 
-use ruby_rbs::node::{self, ClassNode, CommentNode, ConstantNode, GlobalNode, ModuleNode, Node, TypeNameNode, Visit};
+use ruby_rbs::node::{
+    self, ClassNode, CommentNode, ConstantNode, GlobalNode, ModuleNode, Node, NodeList, TypeNameNode, Visit,
+};
 
 use crate::diagnostic::Rule;
 use crate::indexing::local_graph::LocalGraph;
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -371,4 +388,40 @@ mod tests {
...
+        assert_definition_at!(&context, "6:1-7:4", Module, |def| {
+            assert_def_name_eq!(&context, def, "Bar");
+            assert!(def.flags().contains(DefinitionFlags::DEPRECATED));
+        });
+
+        assert_definition_at!(&context, "10:1-11:4", Module, |def| {
+            assert_def_name_eq!(&context, def, "Baz");
+            assert!(!def.flags().contains(DefinitionFlags::DEPRECATED));
+        });
+    }
```
Morriar, line 611: Let's also add a test for classes with deprecation, similar to the module test — just to confirm both paths are covered.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -123,6 +123,22 @@ impl<'a> RBSIndexer<'a> {
         }
         definition_id
     }
+
+    fn flags(list: &NodeList) -> DefinitionFlags {
+        let mut flags = DefinitionFlags::empty();
+
+        for node in list.iter() {
+            if let Node::Annotation(annotation) = node {
```
vinistock, line 148: Does `node.annotations()` always return a `NodeList` of `Node::Annotation`? Or can it return some other AST node?

Maybe, as a future improvement to the RBS crate API, we can make `NodeList` generic and then we might be able to not need to narrow the type.

Something like:

```rust
fn annotations() -> NodeList<Node::Annotation> {
}
```

vinistock (approved):
Nice

Morriar (approved):

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: Changed the second example from `module` to `class` so that it tests class declaration too.

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: Added `panic()` since the annotations node should not have any node other than annotation.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -504,4 +531,82 @@ mod tests {
             });
         });
     }
+
+    #[test]
+    fn flags_test_deprecation() {
```
soutaro, line 536: Extracted tests for deprecation annotation detection.

PR #620, opened 2026-03-02
Follow aliases when they are used in mixins or parent classes

vinistock:
When a constant alias is used in a mixin operation or a parent class, we need to follow the alias to figure out what namespace it points to.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -286,31 +286,21 @@ impl<'a> Resolver<'a> {
                                 // `Foo` is undefined). The method is orphaned as a consequence.
                                 continue;
                             };
-                            self.get_or_create_singleton_class(owner_decl_id)
-                                .expect("Methods with self receiver should always be inside a namespace")
+                            let Some(singleton_id) = self.get_or_create_singleton_class(owner_decl_id) else {
+                                // The enclosing declaration is a constant that can't be promoted to a namespace
+                                // (e.g., `CONST = 1; module CONST; def self.bar; end; end`). The method is
+                                // orphaned as a consequence.
+                                continue;
+                            };
+                            singleton_id
                         }
                         Some(Receiver::ConstantReceiver(name_id)) => {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -852,15 +835,11 @@ impl<'a> Resolver<'a> {
                 Mixin::Prepend(_) => {
                     match self.graph.names().get(constant_reference.name_id()).unwrap() {
                         NameRef::Resolved(resolved) => {
-                            let declaration_id = resolved.declaration_id();
-
-                            // Skip if the mixin is not a namespace (e.g.: dynamically defined class or module that we
-                            // misidentified as a constant)
-                            if !self.graph.is_namespace(declaration_id) {
+                            let Some(module_id) = self.resolve_to_namespace(*resolved.declaration_id()) else {
                                 continue;
```
Morriar, line 839: I'm wondering if we should set `context.partial = true` here when `resolve_to_namespace` returns `None` because the alias target isn't resolved yet (as opposed to being a non-namespace constant).

What would happen with this test?

```rs
  #[test]
  fn include_via_alias_when_target_resolves_later() {
      context.index_uri("file:///a.rb", {
          r"
          AliasedModule = SomeModule

          class Foo
            include AliasedModule
          end

          module SomeModule
          end
          "
      });
      context.resolve();

      assert_ancestors_eq!(context, "Foo", ["SomeModule", "Foo", "Object"]);
  }
```

vinistock: This specific example already works as we figure out what the alias points to. I prototyped making the ancestors partial in this case, but I'm not 100% sure it's worth it to be honest.

Making the ancestors partial means we enqueue retries for linearization, which in the case of an unresolved alias will never complete and only slows down resolution.

Also, it makes the code a bit more complicated because partial ancestors store a `NameId`, so then we need to follow aliases to trace back what `NameId` are associated with them.

Considering that we will add a diagnostic pointing to the fact that the alias was not successfully resolved, I would vote to not do this. At least not for now.

PR #627, opened 2026-03-02
Add unresolve functions and `declaration_to_names` reverse index

st0012:
Extracted from #589 

Add methods for unwinding name resolution during incremental updates:

- `unresolve_name`: converts a `Resolved` name back to `Unresolved`
- `unresolve_reference`: detaches a reference from its declaration and unresolves the underlying name
- `remove_name`: removes a name and cleans up reverse indices

Add `declaration_to_names` reverse index that tracks which names resolve to each declaration. Populated during `record_resolved_name`, cleaned up by the unresolve/remove primitives.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -494,6 +494,43 @@ impl Graph {
         &self.names
     }
 
+    /// Converts a `Resolved` `NameRef` back to `Unresolved`, preserving the original `Name` data.
+    /// Returns the `DeclarationId` it was previously resolved to, if any.
+    fn unresolve_name(&mut self, name_id: NameId) -> Option<DeclarationId> {
+        let name_ref = self.names.get(&name_id)?;
+
+        match name_ref {
+            NameRef::Resolved(resolved) => {
+                let declaration_id = *resolved.declaration_id();
+                let name = resolved.name().clone();
+                self.names.insert(name_id, NameRef::Unresolved(Box::new(name)));
+                Some(declaration_id)
+            }
+            NameRef::Unresolved(_) => None,
+        }
+    }
+
+    /// Unresolves a constant reference: removes it from the target declaration's reference set
+    /// and unresolves its underlying name.
+    fn unresolve_reference(&mut self, reference_id: ReferenceId) -> Option<DeclarationId> {
+        let constant_ref = self.constant_references.get(&reference_id)?;
+        let name_id = *constant_ref.name_id();
+
+        if let Some(old_decl_id) = self.unresolve_name(name_id) {
+            if let Some(declaration) = self.declarations.get_mut(&old_decl_id) {
+                declaration.remove_reference(&reference_id);
+            }
+            Some(old_decl_id)
+        } else {
+            None
+        }
+    }
+
+    /// Removes a name from the graph entirely.
+    fn remove_name(&mut self, name_id: NameId) {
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -40,6 +40,10 @@ pub struct Graph {
 
     /// The position encoding used for LSP line/column locations. Not related to the actual encoding of the file
     position_encoding: Encoding,
+
+    /// Reverse index: for each declaration, which `NameId`s resolved to it.
+    /// Used during incremental invalidation to find names that need unresolving when a declaration is removed.
+    declaration_to_names: IdentityHashMap<DeclarationId, IdentityHashSet<NameId>>,
```
vinistock: Is this map intended to be the `name_dependents`? If so, can't we already make this `NameId -> Set<Dependent>` with the enum of reference or definition ID?

st0012: No it's another map. My next PR will introduce `name_dependents: IdentityHashMap<NameId, Vec<NameDependent>>`, and these will be the only reverse indecies we need.

vinistock: Okay, but in what scenario do we need this map? I'm trying to understand how it is used.

We currently have:

- `Definition -> Name -> Declaration`
- `Definition -> Declaration`
- `Reference -> Name -> Declaration`

and `name_dependents` adds

- `Name -> Definition | Reference`

It's not clear to me why the current graph structure + `name_dependents` isn't enough to follow the data in all directions.

st0012: We can remove it by doing just a bit more lookups in invalidation. I've dropped it here and #589

rust/rubydex/src/model/graph.rs
```diff
@@ -462,14 +462,51 @@ impl Graph {
...
+                declaration.remove_reference(&reference_id);
+            }
+            Some(old_decl_id)
+        } else {
+            None
+        }
+    }
+
+    /// Removes a name from the graph entirely.
+    fn remove_name(&mut self, name_id: NameId) {
```
st0012, line 530: Creating this method as we'll also update `name_dependents` here in follow up PRs.

rust/rubydex/src/model/graph.rs
```diff
@@ -682,12 +719,9 @@ impl Graph {
         }
 
         for ref_id in document.constant_references() {
+            self.unresolve_reference(*ref_id);
```
vinistock, line 754: This is okay for now, but be careful with mixing the removal and the invalidation of the data.

The fact that a constant reference was removed doesn't necessarily mean that the `NameRef` needs to become unresolved. For example:

```ruby
# file1.rb
class Foo
  # NameRef(CONST, parent_scope: None, nesting: Foo)
  CONST
end

# Delete
# file2.rb
class Foo
  # NameRef(CONST, parent_scope: None, nesting: Foo)
  CONST
end
```

st0012: My plan is for this and #589 to ALWAYS unresolve `CONST` when a `Foo` definition is removed to make sure this will be correct:

```rb
class Foo                                                                                                                                                               
  include Bar  # Bar defines CONST
  CONST        # resolves to Bar::CONST via ancestor lookup                                                                                                             
end           
```
And then we create a follow up PR to make this and other invalidation more efficient.

PR #630, opened 2026-03-03
Include `rbi` directory in published gem

vinistock:
Tapioca won't find our type annotations if we don't include it as part of the published gem 😅.


vinistock:
### Merge activity

* **Mar 3, 10:49 PM UTC**: @vinistock merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/630).

PR #631, opened 2026-03-04
Index method aliases with RBS indexer

alexcrocha:

This PR adds `visit_alias_node` so method aliases in RBS files are indexed. Both instance aliases (`alias bar baz`) and singleton aliases (`alias self.x self.y`) are supported.

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -367,8 +367,8 @@ mod tests {
     use crate::test_utils::LocalGraphTest;
     use crate::{
         assert_def_comments_eq, assert_def_mixins_eq, assert_def_name_eq, assert_def_name_offset_eq, assert_def_str_eq,
-        assert_def_superclass_ref_eq, assert_definition_at, assert_local_diagnostics_eq, assert_no_local_diagnostics,
-        assert_string_eq,
+        assert_def_superclass_ref_eq, assert_definition_at, assert_local_diagnostics_eq, assert_method_has_receiver,
+        assert_no_local_diagnostics, assert_string_eq,
     };
 
     fn index_source(source: &str) -> LocalGraphTest {
@@ -807,7 +807,7 @@ mod tests {
         assert_definition_at!(&context, "2:3-2:26", MethodAlias, |def| {
             assert_string_eq!(&context, def.new_name_str_id(), "bar()");
             assert_string_eq!(&context, def.old_name_str_id(), "baz()");
-            assert!(def.receiver().is_some());
+            assert_method_has_receiver!(&context, def, "Foo");
         });
     }
 
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -2070,8 +2070,8 @@ impl Visit<'_> for RubyIndexer<'_> {
 mod tests {
     use crate::{
         assert_def_comments_eq, assert_def_mixins_eq, assert_def_name_eq, assert_def_name_offset_eq, assert_def_str_eq,
-        assert_def_superclass_ref_eq, assert_definition_at, assert_local_diagnostics_eq, assert_name_path_eq,
-        assert_no_local_diagnostics, assert_string_eq,
+        assert_def_superclass_ref_eq, assert_definition_at, assert_local_diagnostics_eq, assert_method_has_receiver,
+        assert_name_path_eq, assert_no_local_diagnostics, assert_string_eq,
         model::{
             definitions::{Definition, Parameter, Receiver, Signatures},
             ids::{StringId, UriId},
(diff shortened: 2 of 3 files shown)


alexcrocha:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/rubydex/631?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#631** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/631?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/631?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#617** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/617?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#616** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#615** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -324,6 +324,40 @@ impl Visit for RBSIndexer<'_> {
...
+        let mut new_name = Self::bytes_to_string(alias_node.new_name().name());
+        let mut old_name = Self::bytes_to_string(alias_node.old_name().name());
+        new_name.push_str("()");
+        old_name.push_str("()");
+
+        let new_name_str_id = self.local_graph.intern_string(new_name);
+        let old_name_str_id = self.local_graph.intern_string(old_name);
+
+        let offset = Offset::from_rbs_location(&alias_node.location());
+        let comments = Self::collect_comments(alias_node.comment());
```
Morriar: Heads-up: PR #633 is changing `collect_comments` from a static method to an instance method (`&self` for indent-aware multi-line comment splitting). If #633 merges before this one, this `Self::collect_comments` call will fail to compile and will need to become `self.collect_comments`.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -324,6 +324,40 @@ impl Visit for RBSIndexer<'_> {
             Mixin::Extend(ExtendDefinition::new(ref_id))
         });
     }
+
+    fn visit_alias_node(&mut self, alias_node: &node::AliasNode) {
+        // Singleton aliases (alias self.x self.y) require singleton class modeling
+        // that the RBS indexer doesn't currently support. Skip until we add
+        // SingletonClass definitions to the RBS nesting stack.
```
vinistock: Based on this, I think that we're missing a step before to tie all of this together.

Basically, we need to remember whether it was an instance or singleton level alias as part of the definition, so that resolution can use that information to create the `Method` in the right owner.

Similarly, we're not currently handling these cases for the Ruby indexer:

```ruby
class Bar; end

class Foo
  Bar.alias_method(:new_to_s, :to_s)
end
```

I think that we might need an optional receiver for aliases, similar to what we have for [methods](https://github.com/Shopify/rubydex/blob/88a4d0157144fbb8770043616520ed736a0674c2/rust/rubydex/src/model/definitions.rs#L749).

That way, we can handle both the Ruby and the RBS case:

- For the Ruby `Bar.alias_method(:new_to_s, :to_s)`, the receiver is a `ConstantReceiver(Bar)`
- For the RBS `alias self.x self.y`, the receiver is a `SelfReceiver`

This is assuming that you can't do something like `alias self.x y` or `alias x self.y`, which I don't believe are allowed.

alexcrocha: I opened a [Draft PR](https://github.com/Shopify/rubydex/pull/645) based on your feedback. Let me know if this is what you have in mind

vinistock: Yes, that's the idea. I left a couple of thoughts there and we still need some tests. Also, let's ensure we confirm RBS' behaviour here and if it matches our understanding.

alexcrocha: Thank you, I'll clean it up. I wanted to make sure I was going in the right direction first

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -714,4 +746,73 @@ mod tests {
...
+
+        assert_no_local_diagnostics!(&context);
+
+        assert_definition_at!(&context, "3:3-3:16", MethodAlias, |def| {
+            assert!(def.flags().contains(DefinitionFlags::DEPRECATED));
+        });
+    }
+
+    #[test]
+    fn index_alias_node_singleton() {
```
vinistock, line 795: Even if it is a syntax error, could we add a test for `alias self.x y` and `alias x self.y`?

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -714,4 +746,99 @@ mod tests {
...
     }
+
+    #[test]
+    fn index_alias_node() {
+        let context = index_source({
+            "
+            class Foo
+              # Some documentation
+              alias bar baz
+              alias qux quux
```
vinistock: If we're not going to assert the empty comments on `qux`, then there's no need to have this as part of the test. It's the same exact thing as `bar`.

rust/rubydex/src/model/definitions.rs
```diff
@@ -724,7 +724,7 @@ pub struct MethodDefinition {
 assert_mem_size!(MethodDefinition, 112);
 
 /// The receiver of a singleton method definition.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
```
vinistock: Do we really need the clone and copy? If we can get rid of it, better.

alexcrocha: Not needed, removed 👍

rust/rubydex/src/resolution.rs
```diff
@@ -511,11 +511,18 @@ impl<'a> Resolver<'a> {
                 Definition::MethodAlias(alias) => {
                     // Method aliases operate on instance methods. The SelfReceiver arm is for
                     // RBS `alias self.x self.y`.
+                    let new_name_str_id = *alias.new_name_str_id();
                     let owner_id = match alias.receiver() {
-                        Some(Receiver::SelfReceiver(def_id)) => *self
-                            .graph
-                            .definition_id_to_declaration_id(*def_id)
-                            .expect("SelfReceiver definition should have a declaration"),
+                        Some(Receiver::SelfReceiver(def_id)) => {
```
vinistock, line 516: Just double-checking, we don't confuse this case here, right? Might be worth adding a test

```ruby
class Foo
  self.alias_method :new_to_s, :to_s

  # This should create
  # Foo#new_to_s() and not Foo::<Foo>#new_to_s()
end
```

alexcrocha: We added a [test](https://github.com/Shopify/rubydex/pull/645/changes#diff-2c1da190ea79ca7c9b449ed8efe95bb004edd2618254fbcc08aafc3f7fe051f9R2561) in #645 for this case 👍

PR #632, opened 2026-03-04
Index multiline comments

soutaro:
https://github.com/Shopify/rubydex/issues/87

An RBS `CommentNode` object represents sequence of comment lines, but the Rubydex `Comment` object represents a line of comment. This PR implements the conversion from one `CommentNode` to list of `Comment` objects.

### DIFF
--- a/.claude/skills/benchmark-rbs/SKILL.md
+++ b/.claude/skills/benchmark-rbs/SKILL.md
@@ -1,0 +1,99 @@
+---
+name: benchmark-rbs
+description: >
+  Benchmark Rubydex indexing performance on RBS core/stdlib, comparing the current branch
+  against main. Measures maximum RSS and execution time.
+---
+
+# Benchmark RBS Indexing
+
+Compare Rubydex indexing performance between `main` and the current topic branch using
+the RBS core and stdlib directories.
+
+## Target paths
+
+Clone the `https://github.com/ruby/rbs.git` repository under the repository root and use
+the following paths as benchmark targets:
+
+- `tmp/rbs/core`
+- `tmp/rbs/stdlib`
+
+The `tmp/` directory is at the rubydex repository root. Create it if it doesn't exist.
+
+## Steps
+
+### 0. Record branch names and commit SHAs
+
+Before building, record the current branch name and commit SHA, and the main branch SHA.
+These will be used in the results table, e.g. `main (abc1234)` and `my-branch (def5678)`.
+
+```bash
+TOPIC_BRANCH=$(git branch --show-current)
+TOPIC_SHA=$(git rev-parse --short HEAD)
+MAIN_SHA=$(git rev-parse --short main)
+```
+
+Also record the RBS repository branch and SHA after cloning:
+
+```bash
+RBS_BRANCH=$(git -C tmp/rbs branch --show-current)
+RBS_SHA=$(git -C tmp/rbs rev-parse --short HEAD)
+```
+
+Print the RBS version in the results, e.g. `rbs: master (abc1234)`.
+
+### 1. Build the current branch (release)
+
+```bash
+cargo build --release
+cp rust/target/release/rubydex_cli tmp/rubydex_cli_branch
+```
+
+### 2. Build main (release)
+
+```bash
+git stash
+git checkout main
+cargo build --release
+cp rust/target/release/rubydex_cli tmp/rubydex_cli_main
+git checkout -
+git stash pop
+```
+
+### 3. Run benchmarks
+
+Run each binary 3 times and take the median. Use `utils/mem-use` for memory and time
+measurement. Pass both `core` and `stdlib` together as a single benchmark run.
+
+```bash
+# Main
+utils/mem-use tmp/rubydex_cli_main tmp/rbs/core tmp/rbs/stdlib --stats
+
+# Branch
+utils/mem-use tmp/rubydex_cli_branch tmp/rbs/core tmp/rbs/stdlib --stats
+```
+
+### 4. Compare results
+
+Present a table comparing:
+
+| Branch | Max RSS (MB) | Indexing (s) | Resolution (s) |
+|--------|-------------|-------------|----------------|
+| main (MAIN_SHA)           | ...  | ...  | ...  |
+| TOPIC_BRANCH (TOPIC_SHA)  | ...  | ...  | ...  |
+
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -714,4 +750,25 @@ mod tests {
             assert!(def.flags().contains(DefinitionFlags::DEPRECATED));
         });
     }
+
+    #[test]
+    fn split_multiline_comments() {
```
vinistock, line 878: Let's add a test with an escaped line break as part of the comment's content. It's easy for this to get confused.

```ruby
# splits strings at the \\n char
def foo; end
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -147,14 +147,49 @@ impl<'a> RBSIndexer<'a> {
...
+        let lines: Vec<&str> = comment_block.split('\n').collect();
+
+        let mut comments = Vec::with_capacity(lines.len());
+        let mut current_offset = start as u32;
+
+        for (i, line) in lines.iter().enumerate() {
+            let line_text = if i == 0 {
+                line.to_string()
+            } else {
+                line[indent_len..].to_string()
```
Morriar: Two thoughts here:

1. `line[indent_len..]` will panic if a subsequent line in the block is shorter than `indent_len` bytes. Could this happen if RBS allows a blank or near-blank line within a `CommentNode`? Might be worth a bounds check: `if line.len() >= indent_len { line[indent_len..].to_string() } else { line.to_string() }`

2. The new `split_multiline_comments` test only covers unindented comments (`indent_len = 0`). Could we add a test case with indented comments to exercise the `indent_len` stripping and offset math?

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -714,4 +750,25 @@ mod tests {
             assert!(def.flags().contains(DefinitionFlags::DEPRECATED));
         });
     }
+
+    #[test]
+    fn split_multiline_comments() {
+        let source = "# First line\n# Second line\n# Third line\nclass Foo\nend";
```
Morriar: Can't we use `index_source` and nicely indent the example like we do in other tests?

Morriar: Should we also have a test that shows how it behaves with an initial indent?

```rb
module Outer
  # First line
  # Second line
  class Inner
  end
end
```

rust/rubydex/src/model/graph.rs
```diff
@@ -896,7 +896,7 @@ impl Graph {
                 let has_docs = definitions.iter().any(|def| !def.comments().is_empty());
                 if has_docs {
                     declarations_with_docs += 1;
-                    let doc_size: usize = definitions.iter().map(|def| def.comments().len()).sum();
+                    let doc_size: usize = definitions.iter().map(|def| def.comments().iter().map(|c| c.string().len()).sum::<usize>()).sum();
```
soutaro: It says `bytes`, but it actually was *lines*. (*lines* would also make sense, but this is a fix to make it *bytes*.)

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -147,14 +147,49 @@ impl<'a> RBSIndexer<'a> {
...
+
+        let mut comments = Vec::with_capacity(lines.len());
+        let mut current_offset = start as u32;
+
+        for (i, line) in lines.iter().enumerate() {
+            let line_text = if i == 0 {
+                line.to_string()
+            } else {
+                line[indent_len..].to_string()
+            };
```
vinistock: Is this assuming that all comment lines are indented the same? What happens in this case? Let's add a test to document the behaviour.

```rbs
class Foo
  # my first comment starts out great
# now I mess up the indent
  def bar: () -> void

# this one is the opposite
  # I fix it later
  def baz: () -> void
end
```

soutaro: My bad. 🙇‍♂️ 
I was thinking the RBS parser rejects the differently indented comments, but it actually accepts. I need to fix the implementation.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -147,14 +147,44 @@ impl<'a> RBSIndexer<'a> {
...
+        let Some(comment_node) = comment_node else {
+            return Vec::new();
+        };
+
+        let location = comment_node.location();
+        let start = location.start().cast_unsigned() as usize;
+        let end = location.end().cast_unsigned() as usize;
+
+        let comment_block = &self.source[start..end];
+        let lines: Vec<&str> = comment_block.split('\n').collect();
```
st0012, line 161: - If the comment ends with `\n`, this could include an empty string, is that acceptable?
- What if the newline is `\r\n` (on windows)?

soutaro: Comment doesn't end with `\n`. This is ensured by the RBS parser.

For `\r\n`, the comment content will be broken. File I/O functions in Rust don't convert `\r\n` automatically, unlike Ruby. Handling only `\n` leaves its predecessor `\r`. I should update the implementation. 🙇‍♂️

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -147,15 +147,52 @@ impl<'a> RBSIndexer<'a> {
...
+
+        let mut comments = Vec::with_capacity(lines.len());
+        let mut current_offset = start as u32;
+
+        for (i, line) in lines.iter().enumerate() {
+            let mut size = 1;
+            let mut line = *line;
+
+            if line.ends_with('\r') {
+                line = line.trim_end_matches('\r');
```
Morriar, line 171: Isn't `trim_end_matches('\r')` stripping *all* trailing `\r`, not just one?

For a normal CRLF file there's always exactly one, so in practice it's fine. But if a line ended in `\r\r`, both would get stripped while `size` only increments by 1, throwing off `current_offset`.

We could use `strip_suffix` to be explicit:

```suggestion
            if let Some(stripped) = line.strip_suffix('\r') {
                line = stripped;
                size += 1;
            }
```

PR #633, opened 2026-03-04
Index method definitions from RBS

soutaro:
This PR implements indexing method definitions `def` from RBS.

- Support all parameter kinds: required/optional positional, rest, post, required/optional keyword, rest keyword, block, and *untyped parameters* `(?) -> void`
- Support overloaded method signatures (`def foo: (String) -> Integer | (Integer) -> String`)
- Handle visibility (`public`/`private`), singleton methods (`def self.foo`), and module_function (`def self?.foo`)
- Generate default names for unnamed RBS parameters (`arg0`, `arg1`, ..., `args`, `kwargs`, `block`)

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -1,8 +1,10 @@
 //! Visit the RBS AST and create type definitions.
 
+use core::panic;
+
 use ruby_rbs::node::{
-    self, AliasKind, ClassNode, CommentNode, ConstantNode, ExtendNode, GlobalNode, IncludeNode, ModuleNode, Node,
-    NodeList, PrependNode, TypeNameNode, Visit,
+    self, AliasKind, ClassNode, CommentNode, ConstantNode, ExtendNode, FunctionTypeNode, GlobalNode, IncludeNode,
+    ModuleNode, Node, NodeList, PrependNode, TypeNameNode, Visit,
 };
 
 use crate::diagnostic::Rule;
@@ -10,12 +12,14 @@ use crate::indexing::local_graph::LocalGraph;
 use crate::model::comment::Comment;
 use crate::model::definitions::{
     ClassDefinition, ConstantDefinition, Definition, DefinitionFlags, ExtendDefinition, GlobalVariableDefinition,
-    IncludeDefinition, MethodAliasDefinition, Mixin, ModuleDefinition, PrependDefinition, Receiver,
+    IncludeDefinition, MethodAliasDefinition, MethodDefinition, Mixin, ModuleDefinition, Parameter, ParameterStruct,
+    PrependDefinition, Receiver, Signature, Signatures,
 };
 use crate::model::document::Document;
-use crate::model::ids::{DefinitionId, NameId, ReferenceId, UriId};
+use crate::model::ids::{DefinitionId, NameId, ReferenceId, StringId, UriId};
 use crate::model::name::{Name, ParentScope};
 use crate::model::references::ConstantReference;
+use crate::model::visibility::Visibility;
 use crate::offset::Offset;
 
 pub struct RBSIndexer<'a> {
@@ -23,6 +27,7 @@ pub struct RBSIndexer<'a> {
     local_graph: LocalGraph,
     source: &'a str,
     nesting_stack: Vec<DefinitionId>,
+    current_visibility: Visibility,
 }
 
 impl<'a> RBSIndexer<'a> {
@@ -36,6 +41,7 @@ impl<'a> RBSIndexer<'a> {
             local_graph,
             source,
             nesting_stack: Vec::new(),
+            current_visibility: Visibility::Public,
         }
     }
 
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -147,14 +148,49 @@ impl<'a> RBSIndexer<'a> {
...
+        let lines: Vec<&str> = comment_block.split('\n').collect();
+
+        let mut comments = Vec::with_capacity(lines.len());
+        let mut current_offset = start as u32;
+
+        for (i, line) in lines.iter().enumerate() {
+            let line_text = if i == 0 {
+                line.to_string()
+            } else {
+                line[indent_len..].to_string()
```
Morriar: This can panic if a subsequent comment line has fewer leading spaces than `indent_len`. For example, if the first line has 4 spaces of indentation but a later line accidentally starts at column 0, `line[4..]` would panic with an out-of-bounds string slice.

Might be safer to use `line.get(indent_len..).unwrap_or(line)` here:

```suggestion
                line.get(indent_len..).unwrap_or(line).to_string()
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -324,6 +360,42 @@ impl Visit for RBSIndexer<'_> {
             Mixin::Extend(ExtendDefinition::new(ref_id))
         });
     }
+
+    fn visit_method_definition_node(&mut self, def_node: &node::MethodDefinitionNode) {
+        // Singleton and singleton_instance methods are not indexed for now.
```
vinistock: We should have the mechanisms for indexing singleton methods. Not sure what `singleton_instance` method means.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -324,6 +360,42 @@ impl Visit for RBSIndexer<'_> {
...
+        };
+
+        let definition = Definition::Method(Box::new(MethodDefinition::new(
+            str_id,
+            self.uri_id,
+            offset,
+            comments,
+            flags,
+            lexical_nesting_id,
+            Vec::new(),
```
vinistock: I was reflecting on this for the parameters. I think we have a few options, but my suggestion is that we make `parameters` in `MethodDefinition` an enum.

```rust
enum MethodParameters {
  Simple(Signature),
  Overloaded(Vec<Signature>),
}

// where Signature just holds the vector of parameters for convenience
```

The reason I feel like this is the best option is because a lot of methods in RBS will not have overloads. If we were to create a separate `RBSMethodDefinition` that always accepted a vector of signatures, we would pay the price of allocating vectors for a bunch of RBS methods that don't have overloads.

I believe we'll reduce our memory usage and keep the implementation simple by simply allowing method defs to accept a single signature or overloads.

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: I’ve been working on instance method support because it's easier to finish.

I’ll work on singleton method support once I finish indexing instance methods.
`singleton_instance` is for `module_function`. (Yeah, that makes things even more complicated. 😄)

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: @vinistock 

Introducing a new enum like `MethodParameters` sounds better to me, because we can add more variants when we need, like overloaded method definitions or *untyped* method types (`(?) -> T`) in RBS.

However the memory footprint will be slightly bigger than `Vec<Signature>` (24 bytes vs 32 bytes with the enum) with fewer heap allocations. Does it make sense?

Quote from claude-code analysis:

```
  ┌────────────────────────┬────────────────┬───────────────────────┐
  │                        │ Vec<Signature> │ MethodParameters enum │
  ├────────────────────────┼────────────────┼───────────────────────┤
  │ Inline size            │ 24 bytes       │ 32 bytes              │
  ├────────────────────────┼────────────────┼───────────────────────┤
  │ Single-sig heap allocs │ 2              │ 1                     │
  ├────────────────────────┼────────────────┼───────────────────────┤
  │ Multi-sig heap allocs  │ 2              │ 2                     │
  ├────────────────────────┼────────────────┼───────────────────────┤
  │ MethodDefinition size  │ 112 (current)  │ 120                   │
  └────────────────────────┴────────────────┴───────────────────────┘

  The enum saves one heap allocation for every single-signature method but costs 8 bytes more per MethodDefinition. For multi-signature methods the alloc count is the same.
```

st0012: @soutaro can you calculate how that'll accumulate to be in our monolith?

soutaro: @st0012 Good point!

In `core/shopify`, there are `376,097` method definitions, and the difference between the two data structures is about **3 MB**. The max RSS is `4830.97 MB`, so `3 / 4830.97` is about **0.06%**.

I don’t think this difference matters, so we can go with the enum definition. 💪

Screenshot 2026-03-13 at 10 32 45

rust/rubydex/src/resolution.rs
```diff
@@ -5463,4 +5463,21 @@ mod tests {
...
         assert_declaration_does_not_exist!(context, "Foo::Bar");
     }
+
+    #[test]
+    fn rbs_method_definition_instance() {
+        let mut context = GraphTest::new();
+        context.index_rbs_uri("file:///foo.rbs", {
+            r"
+            class Foo
+                def foo: () -> void
```
```suggestion
              def foo: () -> void
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -714,4 +892,188 @@ mod tests {
...
+            assert!(matches!(def.signatures(), Signatures::Overloaded(_)));
+        });
+    }
+
+    #[test]
+    fn index_method_definition_with_dot_dot_dot() {
+        let context = index_source({
+            "
+            class Foo
+              def to_s: ...
```
vinistock, line 1365: I didn't know about this syntax. TIL. What does this mean exactly?

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -169,6 +173,93 @@ impl<'a> RBSIndexer<'a> {
...
+                Offset::from_rbs_location(&value.location()).end(),
+            );
+            let str_id = self.local_graph.intern_string(name);
+            parameters.push(Parameter::OptionalKeyword(ParameterStruct::new(offset, str_id)));
+        }
+
+        if let Some(node) = function_node.rest_keywords() {
+            let location = node.location();
+            // Include the ** prefix (2 bytes before the node location)
+            let start = location.start() - 2;
```
vinistock: Are we doing the same for the Ruby indexer? I have the gut feeling that we would actually prefer not including the `**`.

For example, if we want to rename the parameter, we wouldn't edit the `**`, but only what comes after.

Anyway, let's just ensure it's consistent between both indexers.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -169,6 +173,93 @@ impl<'a> RBSIndexer<'a> {
...
+            );
+            let str_id = self.local_graph.intern_string(name);
+            parameters.push(Parameter::RequiredKeyword(ParameterStruct::new(offset, str_id)));
+        }
+
+        for (key, value) in function_node.optional_keywords().iter() {
+            let name = format!(
+                "{}: {}",
+                self.source_at(&key.location()),
+                self.source_at(&value.location())
```
vinistock: This seems like something we should probably add to the RBS crate for convenience (which we can do separately if preferred).

Essentially a `name` method that returns the byte slice.

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: It copies the method type from the ancestor. To mark that there is a method definition in the class, or to add some documentation.

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: I'm fine to drop the `**`. Will fix the code to align with Ruby indexer. 👍 

The problem is RBS type declaration may not have the name of parameter, and the `&block` parameter syntax is not supported.

```rbs
def foo: (String, *Integer) { (Integer) -> String } -> void
          ^^^^^^   ^^^^^^^                                     ==> Unnamed parameters
                            ^^^^^^^^^^^^^^^^^^^^^^^            ==> No &block parameter
```

Assigning parameter names automatically like `arg0` and `&block` would make sense for now.

vinistock: Interesting, I hadn't considered that. Let's indeed use `arg0` and `block` for now, but later we may need to reconsider how we handle parameters.

I guess there are two main scenarios:

- An RBS file defines a method we did not know about. This case may need to default to `arg0` and `block`, although it's a bit unfortunate. It would be nice if we could have proper names for them
- An RBS file only adds types to an existing method we discovered from Ruby. In this case, the RBS annotations are essentially only assigning types, so maybe this one is handled a bit differently

Anyway, a problem for later.

rust/rubydex/src/test_utils/local_graph_test.rs
```diff
@@ -53,6 +55,16 @@ impl LocalGraphTest {
         &self.uri
     }
 
+    #[must_use]
+    pub fn source(&self) -> &str {
+        &self.source
+    }
+
+    #[must_use]
+    pub fn source_at(&self, offset: &Offset) -> &str {
```
soutaro, line 64: I want this utility to test where an offset points to.

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: Hmm. Mixing RBS and Ruby method definitions is a bit complicated.

I think there are two directions for using method signatures:

1. For callers — type checking call expressions, showing documentation, parameter hints, ...
2. For `def` implementation — type checking the implementation body (not sure yet what rubydex will do here)

In Steep:

1. For callers RBS is the source of truth. (Steep simply ignores Ruby methods if RBS is missing.)
2. For def implementations, it reasons the type of each parameter in the Ruby `def` syntax from RBS method types to type check the method body.

I think rubydex can take a similar approach for callers: use RBS information when it exists, and fall back to signatures from the Ruby method definition when RBS is missing. When RBS type definition exists, the absence of parameter names may not be a big problem, since the parameter types itself would help very well.

PR #635, opened 2026-03-04
Expose `Declaration#references` to the Ruby API

vinistock:
This PR exposes the `Declaration#references` Ruby API, which is what the LSP will use to implement find references and rename.

Similar to the `declarations` iterator, I extracted the base infrastructure for iterating over references to `utils.c`, so that we can more easily reuse.

### DIFF
--- a/rust/rubydex-sys/src/declaration_api.rs
+++ b/rust/rubydex-sys/src/declaration_api.rs
@@ -7,6 +7,7 @@ use std::ptr;
 
 use crate::definition_api::{DefinitionsIter, rdx_definitions_iter_new_from_ids};
 use crate::graph_api::{GraphPointer, with_graph};
+use crate::reference_api::{ReferenceKind, ReferencesIter};
 use crate::utils;
 use rubydex::model::ids::{DeclarationId, StringId};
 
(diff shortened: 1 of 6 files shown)


vinistock:
* **#636** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/636?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#635** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/declaration_test.rb
```diff
@@ -380,6 +380,54 @@ class Child < Parent; end
...
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      declaration = graph["A"]
+      references = []
+      declaration.references do |ref|
+        references << ref
+      end
+
+      assert_equal(2, references.size)
```
Morriar, line 421: The tests only cover constant references (both `B < A` and `A.new` resolve to `ConstantReference`). Could we add a test case that exercises the method reference path too?

rust/rubydex-sys/src/declaration_api.rs
```diff
@@ -397,3 +398,39 @@ pub unsafe extern "C" fn rdx_declaration_descendants(pointer: GraphPointer, decl
...
+            return ReferencesIter::new(Vec::new().into_boxed_slice());
+        };
+
+        let entries: Vec<(u64, ReferenceKind)> = decl
+            .references()
+            .iter()
+            .filter_map(|ref_id| {
+                if graph.constant_references().contains_key(ref_id) {
+                    Some((**ref_id, ReferenceKind::Constant))
+                } else if graph.method_references().contains_key(ref_id) {
```
Morriar: Can't we already guess the kind of reference depending on the kind of declaration it references?

Could a `MethodDeclaration` be referenced by a `ConstantReference` and inversely could a `ClassDeclaration` be referenced by a `MethodReference`?

Morriar: In any the need to check 2 collections to kind the matching ID is unfortunate

vinistock: It's definitely weird. I removed the hash lookups, but to solve the more general problem I think we need https://github.com/Shopify/rubydex/issues/637.

test/declaration_test.rb
vinistock: Well, the thing is, we don't resolve method references, so they don't get associated with declarations.

We could associate certain method calls with declarations (the ones that happen on a constant or on `self`), but complete and correct association would require type inference.

Since we don't know for sure all method references, the existing consumers (Spoom and Ruby LSP) will probably not escape looping through references anyway.

I added a test to document the current behaviour.

Morriar: Ah right we have no way to test this at the moment. Maybe we should make the related branch in `rdx_declaration_references_iter_new` as `unreachable` for the moment?

vinistock: That will make the code a bit weird since we compute the `kind` of reference eagerly now (which means doing `graph["foo()"].references` would segfault immediately).

Let's keep the empty array as the return for now since we have the test documenting the behaviour.

PR #638, opened 2026-03-04
Add pagination support rubydex mcp

st0012:
## Summary

- Add `offset` and `total` pagination to `search_declarations`, `get_descendants`, and `find_constant_references`
- `limit=0` treated as "use default" to prevent empty pages
- Add 26 unit tests calling tool handlers directly via graph injection 
- Document pagination and ordering in `docs/architecture.md`

## Design decisions

**Result ordering**

Ordering is stable within a server session (graph is immutable after indexing) but may change across restarts due to parallel indexing affecting insertion order.

**No caching**

Each call re-walks the full filtered collection. This is O(N) per call and acceptable for expected cardinalities (hundreds to low thousands). Pagination is introduced now to support large codebases so we can start testing.

Caching will be introduced when we finished the incremental invalidation/resolution infrastructure.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -108,6 +108,29 @@ Connections between nodes use hashed IDs defined in `ids.rs`:
 - `StringId`: ID for interned string values
 - `ReferenceId`: ID for constant or method reference occurrences (combines reference kind, URI, and offset)
 
+## MCP Server
+
+The `rubydex-mcp` crate exposes rubydex's code intelligence as MCP tools over stdio JSON-RPC. The server indexes the codebase on startup, then serves tool requests against the immutable graph.
+
+### Pagination
+
+Tools that may return a high number of results accept `offset` and `limit` parameters and return a `total` count to support pagination.
+
+Pagination uses a two-pass approach: first collect all entries that pass filtering into a `Vec`, then apply `skip(offset).take(limit)`. This ensures `total` accurately reflects the number of results the caller can page through.
+
+### Result Ordering
+
+All collection-returning tools iterate over `IdentityHashMap` or `IdentityHashSet` structures. These use a deterministic hasher, so iteration order is fixed for a given map state.
+
+- **Within a server session**: Order is consistent between requests as long as the graph has not been re-indexed. Incremental re-indexing (e.g., after a file save) may change the graph between paginated requests, causing items to shift, appear, or disappear. Callers should not assume pagination stability across graph changes.
+- **Across server restarts**: Order may change. Indexing is parallelized, so thread scheduling affects insertion order into the graph, which determines HashMap/HashSet bucket layout.
+
+### Key Files
+
+- `rubydex-mcp/src/server.rs`: Tool handler implementations and pagination logic
+- `rubydex-mcp/src/tools.rs`: Parameter struct definitions with JSON Schema annotations
+- `rubydex-mcp/tests/mcp.rs`: Integration tests (full MCP protocol over stdio)
+
 ## FFI Layer
 
 The Rust crate exposes a C-compatible FFI API through `rubydex-sys`. The C extension in `ext/rubydex/` wraps this API for Ruby.


rust/rubydex-mcp/src/server.rs
```diff
@@ -206,54 +206,69 @@ fn format_ancestors(graph: &Graph, ancestors: &Ancestors) -> Vec<serde_json::Val
 #[tool_router]
 impl RubydexServer {
     #[tool(
-        description = "Search for Ruby classes, modules, methods, or constants by name. Use this INSTEAD OF Grep when you know part of a Ruby identifier name and want to find its definition. Returns fully qualified names, kinds, and file locations. Use the `kind` filter (\"Class\", \"Module\", \"Method\", \"Constant\") to narrow results."
+        description = "Search for Ruby classes, modules, methods, or constants by name. Use this INSTEAD OF Grep when you know part of a Ruby identifier name and want to find its definition. Returns fully qualified names, kinds, and file locations. Use the `kind` filter (\"Class\", \"Module\", \"Method\", \"Constant\") to narrow results. Results are paginated: the response includes `total` (the full count of matches). If `total` exceeds the number of returned results, use `offset` to fetch subsequent pages."
```
st0012, line 225: I hope I didn't need to do this, but during testing I found that Claude didn't immediately realize it can pull more data through pagination.

rust/rubydex-mcp/src/server.rs
```diff
@@ -531,3 +567,558 @@ impl ServerHandler for RubydexServer {
...
+            class B; end
+            class C; end
+            ",
+        );
+
+        let res = search_declarations!(s, query: String::new(), kind: None, limit: Some(2), offset: Some(0));
+        assert_eq!(array!(res, "results").len(), 2);
+        let total = res["total"].as_u64().unwrap();
+
+        let res = search_declarations!(s, query: String::new(), kind: None, limit: Some(2), offset: Some(9999));
```
Morriar, line 760: The pagination tests check count and out-of-bounds offset, but don't verify that page 2 returns *different* items than page 1. If `skip(offset)` were off-by-one or broken, these would still pass.

Maybe add something like:

```rust
let page1 = search_declarations!(s, query: String::new(), kind: None, limit: Some(1), offset: Some(0));
let page2 = search_declarations!(s, query: String::new(), kind: None, limit: Some(1), offset: Some(1));
let name1 = array!(page1, "results")[0]["name"].as_str().unwrap();
let name2 = array!(page2, "results")[0]["name"].as_str().unwrap();
assert_ne!(name1, name2);
```

rust/rubydex-mcp/src/server.rs
```diff
@@ -364,51 +391,58 @@ impl RubydexServer {
...
     )]
     fn find_constant_references(&self, Parameters(params): Parameters<FindConstantReferencesParams>) -> String {
         let state = ensure_graph_ready!(self);
         let graph = state.graph.as_ref().unwrap();
         let (_, decl) = lookup_declaration!(graph, &params.name);
 
-        let limit = params.limit.unwrap_or(50).min(200);
-        let mut references: Vec<serde_json::Value> = Vec::with_capacity(limit.min(decl.references().len()));
-        let mut truncated = false;
+        let limit = params.limit.filter(|&l| l > 0).unwrap_or(50).min(200); // default 50, max 200
```
Morriar, line 409: Looks like a lot of these logic could be extracted into a more generic `paginate` function?

rust/rubydex-mcp/src/server.rs
```diff
@@ -206,54 +206,69 @@ fn format_ancestors(graph: &Graph, ancestors: &Ancestors) -> Vec<serde_json::Val
...
+        description = "Search for Ruby classes, modules, methods, or constants by name. Use this INSTEAD OF Grep when you know part of a Ruby identifier name and want to find its definition. Returns fully qualified names, kinds, and file locations. Use the `kind` filter (\"Class\", \"Module\", \"Method\", \"Constant\") to narrow results. Results are paginated: the response includes `total` (the full count of matches). If `total` exceeds the number of returned results, use `offset` to fetch subsequent pages."
     )]
     fn search_declarations(&self, Parameters(params): Parameters<SearchDeclarationsParams>) -> String {
         let state = ensure_graph_ready!(self);
         let graph = state.graph.as_ref().unwrap();
         let ids = rubydex::query::declaration_search(graph, &params.query);
 
-        let limit = params.limit.unwrap_or(25).min(100);
+        let limit = params.limit.filter(|&l| l > 0).unwrap_or(25).min(100); // default 25, max 100
```
Morriar: Does that mean we treat `limit = 0` as `limit = 25`?

rust/rubydex-mcp/src/server.rs
```diff
@@ -531,3 +567,558 @@ impl ServerHandler for RubydexServer {
...
+        for (uri, source) in sources {
+            gt.index_uri(uri, source);
+        }
+        gt.resolve();
+
+        let (root, _) = test_root();
+        let server = RubydexServer::new(root.to_string());
+        {
+            let mut state = server.state.write().unwrap();
+            state.graph = Some(gt.into_graph());
```
Morriar, line 659: Does it really need to take ownership of the graph? Can't it just take a reference with a matching lifetime?

rust/rubydex-mcp/src/server.rs
Morriar, line 760: Let's add a test where we actually access the second page and check the results are what we expected.

docs/architecture.md
```diff
@@ -108,6 +108,31 @@ Connections between nodes use hashed IDs defined in `ids.rs`:
...
+## MCP Server
+
+The `rubydex-mcp` crate exposes rubydex's code intelligence as MCP tools over stdio JSON-RPC. The server indexes the codebase on startup, then serves tool requests against the immutable graph.
+
+### Pagination
+
+Three tools return paginated collections: `search_declarations`, `get_descendants`, and `find_constant_references`. Each accepts `offset` and `limit` parameters and returns a `total` count.
+
+Pagination uses a two-pass approach: first collect all entries that pass filtering into a `Vec`, then apply `skip(offset).take(limit)`. This ensures `total` accurately reflects the number of results the caller can page through.
+
```
Morriar, line 120: Should we callout somewhere that the pagination is only valid for the same resolved graph. If it changes incrementally in between 2 page accesses, things will appear/disappear.

rust/rubydex-mcp/src/server.rs
st0012: I don't understand why LLMs would call a tool with 0 and expect it to return nothing?

rust/rubydex-mcp/src/server.rs
```diff
@@ -364,51 +391,58 @@ impl RubydexServer {
...
+                    "column": loc.start_col(),
+                }))
+            })
+            .collect();
 
-        let mut result = serde_json::json!({
+        let result = serde_json::json!({
             "name": params.name,
             "references": references,
+            "total": total,
```
st0012, line 438: What callers do you refer to here? The same Claude Code session that could run multiple versions of the MCP? I don't think tools should or already are depending on MCP tools as they're fixed API endpoint.
Yes this is intentional because truncated isn't necessary when LLMs can now calculate it by themselves with simple arithmetic operations.

docs/architecture.md
```diff
@@ -108,6 +108,29 @@ Connections between nodes use hashed IDs defined in `ids.rs`:
 - `StringId`: ID for interned string values
 - `ReferenceId`: ID for constant or method reference occurrences (combines reference kind, URI, and offset)
 
+## MCP Server
+
+The `rubydex-mcp` crate exposes rubydex's code intelligence as MCP tools over stdio JSON-RPC. The server indexes the codebase on startup, then serves tool requests against the immutable graph.
+
+### Pagination
+
+Three tools return paginated collections: `search_declarations`, `get_descendants`, and `find_constant_references`. Each accepts `offset` and `limit` parameters and returns a `total` count.
```
vinistock: This is almost guaranteed to go stale if we add/remove tools that involve pagination. Could we be more generic here and say that certain tools accept pagination to support large scale applications?

docs/architecture.md
```diff
@@ -108,6 +108,29 @@ Connections between nodes use hashed IDs defined in `ids.rs`:
...
+
+Pagination uses a two-pass approach: first collect all entries that pass filtering into a `Vec`, then apply `skip(offset).take(limit)`. This ensures `total` accurately reflects the number of results the caller can page through.
+
+### Result Ordering
+
+All collection-returning tools iterate over `IdentityHashMap` or `IdentityHashSet` structures. These use a deterministic hasher (xxh3, no random seed), so iteration order is fixed for a given map state.
```
vinistock: Similarly, if we change the hasher we'll not remember to update this.

```suggestion
All collection-returning tools iterate over `IdentityHashMap` or `IdentityHashSet` structures. These use a deterministic hasher, so iteration order is fixed for a given map state.
```

rust/rubydex-mcp/src/server.rs
vinistock: Can we also take the opportunity to add something like `ONLY USE THIS MCP FOR RUBY RELATED TASKS`?

I continue to see Claude repeatedly invoke it for Rust files 😅

rust/rubydex-mcp/src/server.rs
```diff
@@ -341,74 +364,79 @@ impl RubydexServer {
     }
 
     #[tool(
-        description = "Find all classes and modules that inherit from or include a given class/module. Returns known descendants, including transitive relationships, for impact analysis before modifying a base class or module."
+        description = "Find all classes and modules that inherit from or include a given class/module. Returns known descendants (including itself and transitive relationships) for impact analysis before modifying a base class or module. Results are paginated: the response includes `total`. If `total` exceeds the number of returned results, use `offset` to fetch subsequent pages."
```
vinistock:
```suggestion
        description = "Returns all known descendants for the given namespace including itself and all transitive descendants. Can be used to understand how a module/class is used across the codebase. Results are paginated: the response includes `total`. If `total` exceeds the number of returned results, use `offset` to fetch subsequent pages."
```

rust/rubydex-mcp/src/server.rs
```diff
@@ -341,74 +364,79 @@ impl RubydexServer {
...
         let graph = state.graph.as_ref().unwrap();
         let (_, decl) = lookup_declaration!(graph, &params.name);
         let namespace = require_namespace!(decl, &params.name, "get_descendants");
 
-        let descendants: Vec<serde_json::Value> = namespace
-            .descendants()
-            .iter()
-            .filter_map(|desc_id| {
-                let desc_decl = graph.declarations().get(desc_id)?;
+        let limit = params.limit.filter(|&l| l > 0).unwrap_or(100).min(500); // default 100, max 500
```
vinistock: Why is the default 100 for descendants and 25 for declaration search (both just return declarations)? Could we simplify the pagination logic if we just used the same defaults for everything?

rust/rubydex-mcp/src/server.rs
```diff
@@ -519,6 +547,8 @@ Typical workflow: search_declarations -> get_declaration -> find_constant_refere
 
 Fully qualified name format: "Foo::Bar" for classes/modules/constants, "Foo::Bar#method_name" for instance methods.
 
+Pagination: search_declarations, get_descendants, and find_constant_references return paginated results. The response includes a `total` count. When `total` exceeds the number of returned items, use `offset` to fetch the next page.
```
vinistock: Here too, I think we should be more generic and not list the specific tools or else we're guaranteed to forget to update it.

Something like:


```suggestion
Pagination: tools that may return a high number of results include `total` for pagination. When `total` exceeds the number of returned items, use `offset` to fetch the next page.
```

rust/rubydex-mcp/src/server.rs
```diff
@@ -531,3 +561,570 @@ impl ServerHandler for RubydexServer {
...
+    macro_rules! array {
+        ($json:expr, $field:literal) => {
+            $json[$field]
+                .as_array()
+                .expect(concat!("expected '", $field, "' to be an array"))
+        };
+    }
+
+    /// Assert a JSON field equals the expected u64 value.
+    macro_rules! assert_json_int {
```
vinistock, line 606: The MCP's architecture still makes these tests a bit weird. It would be nice to have a separation of logic and presentation, so that we could verify logic with proper Rust structures and not have to worry about JSON.

rust/rubydex-mcp/src/server.rs
alexcrocha: Highjacking this resolved conversation for my own curiosity. 

Is there a specific reason we have 50 and not the max as the default when no value is provided? I noticed the other 2 tools have different maximums (200 and 500) which leads me to assume the payload is being taken into account there, but their defaults is also kept at 50.

Follow up Q, if payload is the similar, should the maximums be similar?

alexcrocha (approved):
Looks good to me 👍  

Left a non-blocking question about page size defaults and maximums

rust/rubydex-mcp/src/server.rs
st0012: We currently don't know what optimal default and max look like yet. But in practice the default is much less relevant because if agents are using these tools without limit param, we probably need to improve our descrptions/instructions first.
Wrt max, I think we should just get pagination out and start using the mcp server in Core and other large codebases to know how to tune it.

rust/rubydex-mcp/src/server.rs
st0012: Updated

rust/rubydex-mcp/src/server.rs
st0012: I'll attempt it in https://github.com/Shopify/rubydex/pull/653

rust/rubydex-mcp/src/server.rs
```diff
@@ -203,57 +203,80 @@ fn format_ancestors(graph: &Graph, ancestors: &Ancestors) -> Vec<serde_json::Val
...
+                        }))
+                    })
+                    .collect();
 
-            if let Some(kind) = kind_filter
-                && !decl.kind().eq_ignore_ascii_case(kind)
-            {
-                continue;
-            }
+                Some(serde_json::json!({
```
Morriar, line 266: I'm wondering if it's worth adding a `next_offset` to the response? Something like:

```json
{ "results": [...], "total": 42, "next_offset": 25 }
```

Where `next_offset = offset + results.len()`, absent when `offset + len >= total`. That way AI clients don't have to track `limit` themselves to compute the next page. Not a blocker though — `total` alone is enough to paginate.

PR #639, opened 2026-03-04
Ensure `find_member` collects main namespace members

vinistock:
There's a bug in the logic that makes `find_member` ignore the members defined in the main workspace.

### DIFF
--- a/test/declaration_test.rb
+++ b/test/declaration_test.rb
@@ -413,4 +413,59 @@ class DeclarationTest < Minitest::Test
       assert_equal("Parent#initialize()", decl.name)
     end
   end
+
+  def test_find_member_returns_prepended_members
+    with_context do |context|
+      context.write!("file1.rb", <<~RUBY)
+        module Prepended
+          def initialize
+            @name = "John"
+          end
+        end
+
+        class Foo
+          prepend Prepended
+        end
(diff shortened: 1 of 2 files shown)


test/declaration_test.rb
```diff
@@ -413,4 +413,59 @@ def initialize
...
+      graph.resolve
+
+      foo = graph["Foo"]
+      decl = foo.find_member("@name")
+      assert_equal("Prepended\#@name", decl.name)
+
+      decl = foo.find_member("initialize()")
+      assert_equal("Prepended#initialize()", decl.name)
+    end
+  end
```
Morriar, line 442: I wonder if we should add a test for `only_inherited: true` with a prepended module? The new `test_find_member_returns_prepended_members` test only covers the default (`only_inherited: false`) case. It'd be good to also verify that when `only_inherited: true`, prepended modules are correctly skipped and the search starts from the ancestor after self.

Something like:
```ruby
decl = foo.find_member("initialize()", only_inherited: true)
assert_nil(decl) # or assert it falls through to the next ancestor
```

vinistock: Yeah, that makes sense https://github.com/Shopify/rubydex/pull/687

PR #640, opened 2026-03-04
Use compact name style for rbi

vinistock:
I completely forgot that we can't use unqualified constant references in type signatures. Tapioca turns all RBIs into the compact style and then all references fail to type check.

For example:

```ruby
# This
module Rubydex
  class Comment
    sig { returns(Location) }
    def location; end
  end
end

# becomes this
class Rubydex::Comment
  sig { returns(Location) }
  def location; end
end
```

which fails type checking with undefined constant `Location`.

This PR reformats the RBI to use the compact style and ensures that all types are fully qualified.

### DIFF
--- a/rbi/rubydex.rbi
+++ b/rbi/rubydex.rbi
@@ -1,273 +1,280 @@
 # frozen_string_literal: true
 # typed: strict
 
-module Rubydex
-  class Comment
-    sig { params(string: String, location: Location).void }
-    def initialize(string:, location:); end
+module Rubydex; end
 
-    sig { returns(Location) }
-    def location; end
+class Rubydex::Comment
+  sig { params(string: String, location: Rubydex::Location).void }
+  def initialize(string:, location:); end
 
-    sig { returns(String) }
-    def string; end
-  end
+  sig { returns(Rubydex::Location) }
+  def location; end
 
-  class ConstantReference < Reference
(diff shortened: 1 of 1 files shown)


vinistock:
* **#640** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rbi/rubydex.rbi
```diff
@@ -1,273 +1,280 @@
...
+    def new(*args); end
   end
+end
 
-  class Error < StandardError; end
+class Rubydex::Error < StandardError; end
 
-  # The global graph representing all declarations and their relationships for the workspace
-  #
-  # Note: this class is partially defined in C to integrate with the Rust backend
```
alexcrocha, line 142: We are losing these comments. Is it intentional?

alexcrocha (approved):

rbi/rubydex.rbi
vinistock: Yeah, we had initially copied this from the Ruby file, but there's no need.

PR #641, opened 2026-03-04
Add incremental invalidation engine

st0012:
Replaces the old `remove_definitions_for_document` + `invalidate_ancestor_chains` approach with a targeted invalidation engine. When a file is updated or deleted, the engine traces through the `name_dependents` reverse index to invalidate only the affected declarations, names, and references — instead of requiring a full graph rebuild.

## How it works

`consume_document_changes()` (renamed from `update()`) and `delete_document()` run a three-step pipeline:

1. **`invalidate`** — read-only scan that seeds a worklist from old/new definitions and references, building a `pending_detachments` side table for definitions that need to be detached from their declarations
2. **`remove_document_data`** — cleans up phase 1 data (definitions, references, names, strings) from maps
3. **`extend`** — merges new content and queues work items for the resolver

The resolver still does `clear_declarations` + full rebuild. Wiring it to drain `pending_work` incrementally is a follow-up.

### Invalidation worklist

The worklist processes three item types:

- **`Declaration`** — two modes:
  - *Remove*: no definitions remain or owner was already removed (orphaned). Cascades to members, singleton class, and descendants. Orphaned definitions are re-queued for re-resolution (e.g. `class Foo::Bar` survives even if `Foo` changes from a module to an alias).
  - *Update*: declaration survives but its ancestor chain may have changed (mixin added/removed, superclass changed). Clears ancestors/descendants and re-queues ancestor resolution.
- **`Name`** — structural dependency broken (name's nesting or parent scope removed). Unresolves the name and cascades to all dependents.
- **`References`** — ancestor context changed, but the name itself is still valid. Needed for mixin-related invalidation:

```ruby
class Foo < Bar
  CONST
end

# Another file adds:
class Foo
  include Baz # Foo's ancestors changed, so references like CONST need re-evaluation
end
```

### Cascade differentiation

The `name_dependents` reverse index distinguishes `ChildName` (compact syntax `Foo::Bar`) from `NestedName` (nested syntax `module Foo; class Bar; end; end`):

- **Structural cascade** (name removed): both `ChildName` and `NestedName` → `Name`
- **Ancestor-triggered cascade** (mixin changed): `ChildName` → `Name` (resolves through parent), `NestedName` → `References` (only references need rechecking)

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -3177,4 +3177,185 @@ mod invalidation_tests {
         assert_members_eq!(context, "Baz", ["Bar"]);
         assert_members_eq!(context, "Baz::Bar", ["bar()"]);
     }
+    #[test]
+    fn switching_include_target_invalidates_ancestors_and_references() {
+        let mut context = GraphTest::new();
+
+        context.index_uri(
+            "file:///m.rb",
+            r"
+            module M1
+              CONST = 1
+            end
+            module M2
+              CONST = 2
+            end
+            ",
+        );
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo
+              include M1
+              CONST
+            end
+            ",
+        );
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "M1", "Object"]);
+        assert_constant_reference_to!(context, "M1::CONST", "file:///foo.rb:3:3-3:8");
+
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo
+              include M2
+              CONST
+            end
+            ",
+        );
+
+        // Middle state: Foo's only definition was in foo.rb, so the declaration is removed.
+        // CONST reference is unresolved.
+        assert_declaration_does_not_exist!(context, "Foo");
+        assert_constant_reference_unresolved!(context, "CONST");
+
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "M2", "Object"]);
+        assert_constant_reference_to!(context, "M2::CONST", "file:///foo.rb:3:3-3:8");
+    }
+
+    #[test]
+    fn removing_superclass_invalidates_ancestors() {
+        let mut context = GraphTest::new();
+
+        context.index_uri(
+            "file:///bar.rb",
+            r"
+            class Bar
+              CONST = 1
+            end
+            ",
+        );
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo < Bar
+              CONST
+            end
+            ",
+        );
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "Bar", "Object"]);
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///foo.rb:2:3-2:8");
+
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo
+              CONST
+            end
+            ",
+        );
+
+        // Middle state: Foo's only definition was in foo.rb, so the declaration is removed.
+        assert_declaration_does_not_exist!(context, "Foo");
+        assert_constant_reference_unresolved!(context, "CONST");
+
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "Object"]);
+        assert_constant_reference_unresolved!(context, "CONST");
+    }
+
+    #[test]
+    fn changing_alias_target_invalidates_dependents() {
+        let mut context = GraphTest::new();
+
+        context.index_uri(
+            "file:///targets.rb",
+            r"
+            class Bar
+              CONST = 1
+            end
+            class Baz
+              CONST = 2
+            end
+            ",
+        );
+        context.index_uri(
+            "file:///alias.rb",
+            r"
+            Foo = Bar
+            ",
+        );
+        context.index_uri(
+            "file:///ref.rb",
+            r"
+            Foo::CONST
+            ",
+        );
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Bar::CONST", "file:///ref.rb:1:6-1:11");
+
+        context.index_uri(
+            "file:///alias.rb",
+            r"
+            Foo = Baz
+            ",
+        );
+
+        // Middle state: old Foo alias declaration removed, CONST ref unresolved
+        assert_constant_reference_unresolved!(context, "CONST");
+
+        context.resolve();
+
+        assert_constant_reference_to!(context, "Baz::CONST", "file:///ref.rb:1:6-1:11");
+    }
+
+    #[test]
+    fn switching_mixin_order_invalidates_ancestor_chain() {
+        let mut context = GraphTest::new();
+
+        context.index_uri(
+            "file:///m.rb",
+            r"
+            module Bar; end
+            module Baz; end
+            ",
+        );
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo
+              include Bar
+              include Baz
+            end
+            ",
+        );
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "Baz", "Bar", "Object"]);
+
+        context.index_uri(
+            "file:///foo.rb",
+            r"
+            class Foo
+              include Baz
+              include Bar
+            end
+            ",
+        );
+
+        // Middle state: Foo's only definition was in foo.rb, so the declaration is removed.
+        assert_declaration_does_not_exist!(context, "Foo");
+
+        context.resolve();
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "Bar", "Baz", "Object"]);
+    }
 } // mod invalidation_tests


rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -172,6 +181,29 @@ impl LocalGraph {
...
+                name_ref.nesting().as_ref().copied(),
+                name_ref.parent_scope().as_ref().copied(),
+            ]
+            .into_iter()
+            .flatten()
+            {
+                let deps = self.name_dependents.entry(owner_id).or_default();
+                if !deps.contains(&dep) {
+                    deps.push(dep);
+                }
```
vinistock: We don't need this if we use an `IdentityHashSet` instead of a `Vec`.

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -172,6 +181,29 @@ impl LocalGraph {
...
+    fn insert_name_dependent(&mut self, name_id: NameId, dep: NameDependent) {
+        self.name_dependents.entry(name_id).or_default().push(dep);
+
+        if let Some(name_ref) = self.names.get(&name_id) {
+            for owner_id in [
+                name_ref.nesting().as_ref().copied(),
+                name_ref.parent_scope().as_ref().copied(),
+            ]
+            .into_iter()
+            .flatten()
```
vinistock: Isn't this array already flat?

rust/rubydex/src/indexing/local_graph.rs
vinistock, line 1: I think it would be nice to have some indexing tests to ensure that we get the right data structures initially.

For example:

```ruby
# Bar and CONST don't depend on anything. There are no code changes that
# can modify what they mean other than changing them directly
module Bar; end
CONST = 1

# Foo has no dependencies
module Foo
  # This Bar depends on Foo
  # The Baz depends on both Bar and Foo
  class Bar::Baz
    # CONST depends on Foo and Baz
    CONST
  end
end
```

Note that the key is ensuring that the chain of dependents can be followed. For example, since `Baz` already depends on `Bar`, remembering that `CONST` depends on `Baz` is enough. We don't need to tie `CONST` to `Bar`. If `Bar` gets invalidated, it will invalidate `Baz` and that will in turn invalidate `CONST`.

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
-    /// updates to existing entries
+    /// updates to existing entries.
+    ///
+    /// Runs incremental invalidation to cascade changes through the graph and
+    /// accumulates pending work items for the resolver to process.
+    ///
+    /// The three steps must run in this order:
+    /// 1. `invalidate` -- reads resolved names and declaration state to determine what to invalidate
+    /// 2. `remove_document_data` -- removes old refs/defs/names/strings from maps
+    /// 3. `extend` -- merges the new `LocalGraph` into the now-clean graph
```
vinistock, line 854: No need to do it in this PR, but I think this method now deserves a more descriptive name. Maybe `consume_document_changes` or `ingest_local_graph`. It does a whole lot more than just "update".

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
     pub fn update(&mut self, other: LocalGraph) {
-        // For each URI that was indexed through `other`, check what was discovered and update our current global
-        // representation
         let uri_id = other.uri_id();
-        if let Some(document) = self.documents.remove(&uri_id) {
-            self.remove_definitions_for_document(&document);
+        let old_document = self.documents.remove(&uri_id);
+
+        self.invalidate(old_document.as_ref(), Some(&other));
```
vinistock: It might be worth adding a fast path for the initial indexing on boot (which can trigger no invalidation and no removal of data).

Maybe a boolean flag for skipping invalidation.

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
+        let mut items: Vec<InvalidationItem> = Vec::new();
+
+        // Detach old definitions from their declarations, queue affected directly
+        if let Some(document) = old_document {
+            for def_id in document.definitions() {
+                if let Some(declaration_id) = self.definition_id_to_declaration_id(*def_id).copied()
+                    && let Some(declaration) = self.declarations.get_mut(&declaration_id)
+                    && declaration.remove_definition(def_id)
+                {
+                    declaration.clear_diagnostics();
```
vinistock: This is doing two data removal operations: removing a definition and clearing diagnostics. Does it make sense to move this to `remove_document_data`? Or will that result in duplicate work?

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
+                    && let Some(NameRef::Resolved(resolved)) = self.names.get(name_id)
+                {
+                    items.push(InvalidationItem::Declaration(*resolved.declaration_id()));
+                }
+            }
+
+            for cr in lg.constant_references().values() {
+                if let Some(name_ref) = self.names.get(cr.name_id())
+                    && let Some(nesting_id) = name_ref.nesting()
+                    && let Some(NameRef::Resolved(resolved)) = self.names.get(nesting_id)
```
vinistock, line 912: Can you clarify that document change this is accounting for? I'm having a hard time understanding it.

A constant reference was changed and we're enqueuing invalidation for the reference's nesting declaration.

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
+                if let Some(owner) = self.declarations.get_mut(&owner_id)
+                    && let Some(ns) = owner.as_namespace_mut()
+                {
+                    ns.remove_member(&unqualified_str_id);
+                }
+            }
+
+            self.declarations.remove(&decl_id);
+        } else {
+            // Ancestor-stale mode
```
vinistock: What does this mean?

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
+                }
+
+                if let Some(owner) = self.declarations.get_mut(&owner_id)
+                    && let Some(ns) = owner.as_namespace_mut()
+                {
+                    ns.remove_member(&unqualified_str_id);
+                }
+            }
+
+            self.declarations.remove(&decl_id);
```
vinistock, line 1067: This is also doing data removal. Maybe this is fine, but I'm calling it out because the method documentation mentions a separation between invalidation and removal.

rust/rubydex/src/model/graph.rs
```diff
@@ -725,148 +792,440 @@ impl Graph {
...
+            // Ancestor-stale mode
+            if !visited_declarations.insert(decl_id) {
+                return;
+            }
+
+            let Some(namespace) = self.declarations.get_mut(&decl_id).and_then(|d| d.as_namespace_mut()) else {
+                return;
+            };
+
+            // Remove self from each ancestor's descendant set
```
vinistock, line 1080: I could be misunderstanding and maybe this was an existing bug, but removing `self` is not enough. The entire ancestor chain of `self` must be removed from descendants. However, we should absolutely not try to perform this removal because there are module deduping rules that you cannot possibly account for with a removal.

Whenever a declaration gets invalidated, we always need to invalidate the ancestors of all descendants. In both branches of this method.

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -70,6 +74,10 @@ impl LocalGraph {
     pub fn add_definition(&mut self, definition: Definition) -> DefinitionId {
         let definition_id = definition.id();
 
+        if let Some(name_id) = definition.name_id() {
+            self.insert_name_dependent(*name_id, NameDependent::Definition(definition_id));
```
st0012: We should move these into Name construction logic directly. When we build names with nesting and parent_scope, we should immediately add those reverse relationships. We can ship this in a separate PR too.

rust/rubydex/src/model/graph.rs
st0012: Moved

rust/rubydex/src/model/graph.rs
st0012: Added more comments. Basically, it's triggered in

```rb
class Foo
  include Bar # this is added/removed
end
```

So we simply change ancestors/descendents update in this branch.

rust/rubydex/src/model/graph.rs
st0012: This is for

```rb
class Foo
  include Bar # constant reference, when added we invalidate Foo entirely for now
end
```

I've added comments for it.

rust/rubydex/src/model/graph.rs
st0012: I currently treat declaration removal as invalidation, kinda similar to unresolving a name. We remove (unresolve) the declaration here if we found it has no underlying definitions anymore.

The underlying materials (definitions, constant references, names...etc.) are only removed in `remove_document_data`.

rust/rubydex/src/model/graph.rs
st0012: Both paths will update the ancestors. I've updated the document to make it more clear: `invalidate_declaration` either "remove/rebuild" or "update" a declaration. In both paths we update ancestors.

This also means there are optimization opportunities in both paths we can do later, which I also included in comments.

rust/rubydex/src/model/graph.rs
st0012: Updated

rust/rubydex/src/model/graph.rs
```diff
@@ -800,147 +843,375 @@ impl Graph {
...
+                {
+                    items.push(InvalidationItem::Declaration(*resolved.declaration_id()));
+                }
+            }
+
+            // Constant references include `include`/`prepend`/`extend` targets.
+            // A new mixin changes the nesting declaration's ancestor chain, so we
+            // invalidate the nesting declaration.
+            // We can optimize this later by checking where the constant reference is used.
+            for cr in lg.constant_references().values() {
```
vinistock: Nit: I would avoid abbreviations that are too short. I'd change this to `reference`, `const_reference`, `const_ref` or even `constant_reference`.

rust/rubydex/src/model/graph.rs
```diff
@@ -800,147 +843,375 @@ impl Graph {
...
+                if let Some(name_ref) = self.names.get(cr.name_id())
```
vinistock: I would vote for doing this here

```rust
let name_ref = self.names.get(cr.name_id()).unwrap();
// if statement
```

A constant reference that holds a `name_id` for a name that doesn't exist in the graph is a significant data consistency issue. It's best to crash immediately and discover those bugs early than to silently ignore.

rust/rubydex/src/model/graph.rs
vinistock: Got it. Similar to the other, I would say it's best to unwrap the option here to avoid hiding data inconsistencies in the graph.

rust/rubydex/src/model/graph.rs
```diff
@@ -800,147 +843,375 @@ impl Graph {
...
             }
         }
 
         for ref_id in document.constant_references() {
-            self.unresolve_reference(*ref_id);
-
             if let Some(constant_ref) = self.constant_references.remove(ref_id) {
+                // Detach from target declaration. References unresolved during invalidation
+                // were already detached; this catches the rest.
+                if let Some(NameRef::Resolved(resolved)) = self.names.get(constant_ref.name_id())
```
vinistock: Same about unwrapping the option.

rust/rubydex/src/model/graph.rs

rust/rubydex/src/model/graph.rs
st0012: We can't panic here because it's possible both the const reference and the name exists only in the local graph and haven't been extended into global graph yet.

rust/rubydex/src/model/graph.rs
st0012: Renamed

PR #642, opened 2026-03-05
Add method alias inside re-opened alias test

vinistock:
Add the missing method alias to our test.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -5194,6 +5194,9 @@ mod tests {
               def initialize
                 @instance_ivar = 456
               end
+
+              def bar; end
+              alias new_bar bar
             end
             "
         });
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -5206,6 +5209,7 @@ mod tests {
         assert_declaration_exists!(context, "Foo#some_attr()");
         assert_declaration_exists!(context, "Foo::<Foo>#class_method()");
         assert_declaration_exists!(context, "Foo#initialize()");
+        assert_declaration_exists!(context, "Foo#new_bar()");
```
Morriar, line 5212: Should we also assert `Foo#bar()` (the original method)? The test asserts all other members in the namespace, so I'd expect `bar` to also be there.

vinistock: I think that's overkill. The functionality is already being tested by `initialize`. I just added `bar` because aliasing `initialize` is not common.

PR #645, opened 2026-03-05
Add receiver to MethodAliasDefinition

alexcrocha:
This PR resolves a gap identified in https://github.com/Shopify/rubydex/pull/631#discussion_r2890589175

`alias_method` can be called on a specific class, like `Foo.alias_method(:new, :old)`. When that happens, the alias belongs to `Foo` rather than whatever class happens to be the enclosing scope. Adding a `receiver` field to `MethodAliasDefinition` lets the resolver figure out the right owner.

Unlike `def Foo.bar` which defines a singleton method, `Foo.alias_method` always creates instance aliases. So we resolve constant receivers to the class itself, not its singleton class.

This PR also extracts `resolve_singleton_owner`, `create_singleton_name_id`, and `Receiver::append_to_id` to reduce duplication, and moves the receiver guard out of `each_string_or_symbol_arg` into `index_attr` since `alias_method` legitimately accepts constant receivers now.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -4578,6 +4578,7 @@ mod tests {
 
         assert_definition_at!(&context, "1:1-5:4", Class, |foo| {
             assert_definition_at!(&context, "3:5-3:24", MethodAlias, |alias_method| {
+                assert!(alias_method.receiver().is_none());
                 assert_eq!(foo.id(), alias_method.lexical_nesting_id().unwrap());
             });
         });
@@ -4996,7 +4997,7 @@ mod tests {
             let old_name = context.graph().strings().get(def.old_name_str_id()).unwrap();
             assert_eq!(new_name.as_str(), "foo()");
             assert_eq!(old_name.as_str(), "bar()");
-
+            assert!(def.receiver().is_none());
             assert!(def.lexical_nesting_id().is_none());
         });
 
@@ -5168,6 +5169,29 @@ mod tests {
             assert_method_has_receiver!(&context, def, "B");
             assert!(def.lexical_nesting_id().is_none());
         });
+    }
+
+    #[test]
+    fn index_alias_method_with_dynamic_receiver_not_indexed() {
+        let context = index_source({
+            "
+            class Foo
+              def original; end
+            end
+
+            foo.alias_method :new_name, :original
+            "
+        });
+
+        assert_no_local_diagnostics!(&context);
+
+        let alias_count = context
+            .graph()
+            .definitions()
+            .values()
+            .filter(|def| matches!(def, Definition::MethodAlias(_)))
+            .count();
+        assert_eq!(0, alias_count);
     }
 
     #[test]


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1657,6 +1657,13 @@ impl Visit<'_> for RubyIndexer<'_> {
                 let reference = MethodRef::new(old_name_str_id, self.uri_id, old_offset.clone(), method_receiver);
                 self.local_graph.add_method_reference(reference);
 
+                let receiver = node.receiver().and_then(|recv_node| match recv_node {
```
vinistock: Are we able to reuse `method_receiver` from a few lines above? We need to be careful not to double-index the same constant reference or do the same work twice.

rust/rubydex/src/resolution.rs
```diff
@@ -509,11 +509,41 @@ impl<'a> Resolver<'a> {
                     }
                 }
                 Definition::MethodAlias(alias) => {
-                    let Some(owner_id) = self.resolve_lexical_owner(*alias.lexical_nesting_id()) else {
-                        continue;
+                    let new_name_str_id = *alias.new_name_str_id();
+                    let lexical_nesting_id = *alias.lexical_nesting_id();
+                    let owner_id = match alias.receiver() {
```
vinistock, line 514: This receiver handling is starting to spread. I think we have a similar thing for methods and instance variables.

It would be nice to see if we can extract a helper to simplify the code.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1653,8 +1659,23 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                    _ => None,
+                });
+
+                // alias_method references instance methods, so we use the class NameId directly
+                let method_ref_receiver = match &receiver {
+                    Some(Receiver::ConstantReceiver(name_id)) => Some(*name_id),
+                    Some(Receiver::SelfReceiver(_)) | None => {
+                        self.method_receiver(receiver_node.as_ref(), node.location())
+                    }
+                };
```
vinistock: Why do we need these two separate paths for the receiver? It seems you're handling constant receiver in one and self in the other. Doesn't `method_receiver` already handle both?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4909,6 +4932,105 @@ mod tests {
...
+                Some(Receiver::ConstantReceiver(name_id)) => {
+                    let name = context.graph().names().get(name_id).unwrap();
+                    let actual_name = context.graph().strings().get(name.str()).unwrap().as_str();
+                    assert_eq!(actual_name, "Foo");
+                }
+                other => panic!("Expected ConstantReceiver, got {other:?}"),
+            }
+        });
+
+        let method_ref = context
```
vinistock: I think method references are already part of other tests, so we may not need to test it again here.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4909,6 +4932,105 @@ mod tests {
...
+
+        assert_no_local_diagnostics!(&context);
+
+        assert_definition_at!(&context, "5:1-5:28", MethodAlias, |def| {
+            let new_name = context.graph().strings().get(def.new_name_str_id()).unwrap();
+            let old_name = context.graph().strings().get(def.old_name_str_id()).unwrap();
+            assert_eq!(new_name.as_str(), "baz()");
+            assert_eq!(old_name.as_str(), "bar()");
+
+            match def.receiver() {
```
vinistock: It would be nice to have an assertion macro for checking the receiver.

rust/rubydex/src/model/definitions.rs
```diff
@@ -724,7 +724,7 @@ pub struct MethodDefinition {
 assert_mem_size!(MethodDefinition, 112);
 
 /// The receiver of a singleton method definition.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
```
vinistock: Why do we need to clone and copy receivers?

rust/rubydex/src/model/definitions.rs
```diff
@@ -734,6 +734,16 @@ pub enum Receiver {
 
 assert_mem_size!(Receiver, 16);
 
+impl Receiver {
+    /// Appends the receiver's inner ID to the given string for `DefinitionId` hashing.
+    pub fn append_to_id(&self, formatted_id: &mut String) {
```
vinistock: This function is a bit of an odd inversion of responsibility. Can we instead implement `Display` for the receiver and just include it directly in the formatted ID?

rust/rubydex/src/resolution.rs
```diff
@@ -2540,6 +2554,64 @@ mod tests {
...
+    #[test]
+    fn resolving_method_alias_with_constant_receiver() {
+        let mut context = GraphTest::new();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Foo
+              def bar; end
+            end
+
+            Foo.alias_method :baz, :bar
```
vinistock: Let's add a method similar to this one, but with the `alias_method` inside of the namespace rather than outside.

rust/rubydex/src/resolution.rs
```diff
@@ -279,42 +279,11 @@ impl<'a> Resolver<'a> {
...
-                            singleton_id
-                        }
-                        None => {
-                            let Some(resolved) = self.resolve_lexical_owner(*method_definition.lexical_nesting_id())
-                            else {
-                                continue;
-                            };
-                            resolved
-                        }
+                    let Some(owner_id) = self.resolve_singleton_owner(receiver.as_ref(), lexical_nesting_id) else {
```
vinistock: This simplification is great. Can we use it in the instance variable branch, which needs similar handling?

rust/rubydex/src/model/definitions.rs
alexcrocha: Implemented. Much nicer 👍

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: The old `method_receiver()` wrapped constant receivers into a singleton class, so we couldn't reuse it directly for `alias_method`. 

Taking a deeper look at the problem, I split the responsibilities: `method_receiver()` now only classifies what kind of receiver it is (returning a `Receiver` enum), and extracted the rest into `index_method_receiver()`, which creates the singleton class name and indexes the constant reference.

I wasn't happy with how the code was turning out and noticed there were a few concepts tied together, which led me to refactor the `Receiver` enum to better handle the different scenarios. 

Since this is an unexpected refactor, it's in a different commit for an easier review.
Let me know if you have any questions.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Addressed. See my reply in https://github.com/Shopify/rubydex/pull/645#discussion_r2899016659 for the full context.

rust/rubydex/src/model/definitions.rs
alexcrocha: Before this PR, the receiver handling was inline within each match arm, so there was no borrow conflict.

Extracting `resolve_singleton_owner` as a shared helper to remove duplication means we now need to pass the receiver to a `&mut self` method while `self.graph.definitions()` is still borrowed. 

`Copy` lets us dereference the receiver out of the borrow (`let receiver = *method_definition.receiver()`), which I find cleaner than `.clone()`

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -45,6 +45,32 @@ impl Nesting {
...
+/// to store.
+enum ReceiverKind {
+    /// `self` or no receiver — resolved from the lexical nesting.
+    Implicit(NameId),
+    /// Explicit constant (e.g. `Foo`, `Foo::Bar`). Singleton-wrapped for method
+    /// definitions (`def Foo.bar`), but used directly for `alias_method`.
+    Constant(NameId),
+    /// `singleton_class` call or method nested inside a singleton method.
+    SingletonClass(NameId),
+}
```
st0012: We already have `Receiver` enum in `definitions.rs` to represent receiver types. Why can't we use them here?

alexcrocha: Stan and I pair-reviewed this PR and revisited my original choice to create a separate enum. Thanks to his feedback, we're now reusing the existing Receiver enum directly. Simpler and avoids the extra enum.

@st0012, let me know what you think of the changes.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1029,14 +1026,10 @@ impl<'a> RubyIndexer<'a> {
...
-    fn method_receiver(
-        &mut self,
-        receiver: Option<&ruby_prism::Node>,
-        fallback_location: ruby_prism::Location,
-    ) -> Option<NameId> {
-        let mut is_singleton_name = false;
-
-        let name_id = match receiver {
+    /// Classifies a method receiver AST node into a [`Receiver`] without applying
+    /// singleton wrapping or registering constant references.
```
st0012: What does the `singleton wrapping` mean?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1063,52 +1058,70 @@ impl<'a> RubyIndexer<'a> {
...
-                        Some(self.local_graph.add_name(Name::new(str_id, ParentScope::None, None)))
+                        let name_id = self.local_graph.add_name(Name::new(str_id, ParentScope::None, None));
+                        Some(Receiver::SelfReceiver(name_id))
                     }
                 }
             }
             Some(ruby_prism::Node::CallNode { .. }) => {
-                // Check if the receiver is `singleton_class`
+                // For `Foo.singleton_class.bar`, we need the inner receiver (`Foo`) with
+                // singleton wrapping and constant reference registration. This calls
```
st0012: This seems to contradict what the method comment says?

rust/rubydex/src/model/definitions.rs
```diff
@@ -723,17 +723,32 @@ pub struct MethodDefinition {
...
+///
+/// All three variants carry a [`NameId`]. `SingletonClassReceiver` represents receivers
+/// resolved through a singleton method nesting or `.singleton_class` call chain.
+#[derive(Debug, Clone, Copy)]
 pub enum Receiver {
-    /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
-    SelfReceiver(DefinitionId),
-    /// `def Foo.bar` - receiver is an explicit constant that needs resolution
+    /// `self` or no receiver — resolved from the lexical nesting.
+    SelfReceiver(NameId),
```
st0012: Why are we changing this? We use `DefinitionId` because when the receiver relies on lexical scope, we can't immediately know the name when indexing. It could also have no name at all, for example:

```rb
foo do
  def self.bar; end
end
```

We will need to preserve these definitions too for DSL handling.

rust/rubydex/src/model/definitions.rs
```diff
@@ -723,17 +723,32 @@ pub struct MethodDefinition {
...
+/// Classifies the receiver of a method or method alias definition.
+///
+/// All three variants carry a [`NameId`]. `SingletonClassReceiver` represents receivers
+/// resolved through a singleton method nesting or `.singleton_class` call chain.
+#[derive(Debug, Clone, Copy)]
 pub enum Receiver {
-    /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
-    SelfReceiver(DefinitionId),
-    /// `def Foo.bar` - receiver is an explicit constant that needs resolution
+    /// `self` or no receiver — resolved from the lexical nesting.
```
st0012: When there's no receiver, it's reflected by having `None` in the receiver attribute.

alexcrocha:
Based on the feedback received, decided to rewrite from scratch. 

The previous implementation was based on a flawed mental model of the problem, and the complexity reflected that. With a better understanding of how Ruby aliasing actually works, the new version is much simpler.

Previous commits will be squashed. Please use the "Files changed" tab for the current state.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Wrong choice of words. No longer applicable, please ignore.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: I agree. This is no longer applicable as the implementation was simplified.

rust/rubydex/src/model/definitions.rs
alexcrocha: Thank you for this context. With my new understanding of the problem I restarted from scratch and we're no longer changing this.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: No longer applicable. See https://github.com/Shopify/rubydex/pull/645#issuecomment-4043757189

rust/rubydex/src/model/definitions.rs
alexcrocha: No longer applicable as the implementation changed.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -361,17 +361,12 @@ impl<'a> RubyIndexer<'a> {
...
+    // Runs the given closure for each string or symbol argument of a call node.
     fn each_string_or_symbol_arg<F>(node: &ruby_prism::CallNode, mut f: F)
     where
         F: FnMut(String, ruby_prism::Location),
     {
-        let receiver = node.receiver();
-
-        if (receiver.is_none() || receiver.unwrap().as_self_node().is_some())
-            && let Some(arguments) = node.arguments()
-        {
```
alexcrocha, line 374: Moved the receiver guard out of `each_string_or_symbol_arg` and into `index_attr` as it was blocking `Foo.alias_method(:bar, :baz)` from being indexed. Now constant receivers produce a `ConstantReceiver` on the definition

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1653,6 +1653,16 @@ impl Visit<'_> for RubyIndexer<'_> {
                 let new_name_str_id = self.local_graph.intern_string(format!("{new_name}()"));
                 let old_name_str_id = self.local_graph.intern_string(format!("{old_name}()"));
 
+                // Only constant receivers (e.g. Foo.alias_method) are stored. `self.alias_method`
+                // is semantically identical to bare `alias_method`, so self maps to None.
+                // method_receiver below handles the constant reference, so we skip it here.
+                let receiver = node.receiver().as_ref().and_then(|recv_node| match recv_node {
+                    ruby_prism::Node::ConstantPathNode { .. } | ruby_prism::Node::ConstantReadNode { .. } => self
+                        .index_constant_reference(recv_node, false)
```
alexcrocha: `false` here avoids duplicate indexing and a `ReferenceId` collision. `method_receiver` below creates the reference.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1653,6 +1653,16 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                        .map(Receiver::ConstantReceiver),
+                    _ => None,
+                });
+
                 let method_receiver = self.method_receiver(node.receiver().as_ref(), node.location());
```
st0012: @alexcrocha and I found that `method_receiver` does not return consistent values:

```rb
class Foo
  bar # receiver Foo, should be <Foo>
end
Foo.bar # receiver <Foo>
```

And we decided that @alexcrocha will address this in a separate PR, which will potentially simplify the implementation here.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4909,6 +4921,136 @@ mod tests {
...
+        assert_definition_at!(&context, "2:1-2:28", MethodAlias, |def| {
+            assert_string_eq!(&context, def.new_name_str_id(), "bar()");
+            assert_string_eq!(&context, def.old_name_str_id(), "baz()");
+
+            let Receiver::ConstantReceiver(name_id) = def.receiver().as_ref().unwrap() else {
+                panic!("Expected ConstantReceiver");
+            };
+            let name = context.graph().names().get(name_id).unwrap();
+            let actual_name = context.graph().strings().get(name.str()).unwrap().as_str();
+            assert_eq!("Foo", actual_name);
```
st0012: Can we have a macro like `assert_method_alias_has_receiver(&context, def, "Foo")`

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4909,6 +4921,136 @@ mod tests {
...
+
+        assert_definition_at!(&context, "2:3-2:26", MethodAlias, |def| {
+            assert!(def.receiver().is_none());
+        });
+    }
+
+    #[test]
+    fn index_alias_method_in_singleton_class_has_no_receiver() {
+        // Ruby aliases class methods via `class << self`, which produces `receiver: None`.
+        // Valid Ruby aliasing never produces a SelfReceiver on MethodAliasDefinition.
```
st0012: What does this mean?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5009,6 +5033,247 @@ mod tests {
...
+
+            A::B.alias_method :new_name, :original
+            "
+        });
+
+        assert_no_local_diagnostics!(&context);
+
+        assert_definition_at!(&context, "7:1-7:39", MethodAlias, |def| {
+            assert_string_eq!(&context, def.new_name_str_id(), "new_name()");
+            assert!(matches!(def.receiver(), Some(Receiver::ConstantReceiver(_))));
```
st0012: Let's use that new assertion macro here too and assert the full name of the receiver const.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5009,6 +5033,247 @@ mod tests {
...
+            .graph()
+            .definitions()
+            .values()
+            .filter(|def| matches!(def, Definition::MethodAlias(_)))
+            .count();
+        assert_eq!(0, alias_count);
+    }
+
+    #[test]
+    fn index_alias_method_constant_receiver_inside_singleton_class() {
```
st0012: I feel this one is more like testing constant resolution inside singleton class context, than actually testing the method alias receiver.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5009,6 +5033,247 @@ mod tests {
...
+
+        assert_definition_at!(&context, "2:1-2:24", MethodAlias, |def| {
+            assert_string_eq!(&context, def.new_name_str_id(), "new_name()");
+            assert!(def.receiver().is_none());
+            assert!(def.lexical_nesting_id().is_none());
+        });
+    }
+
+    #[test]
+    fn index_alias_keyword_inside_method_body() {
```
st0012: Already covered in `index_alias_ignores_method_nesting`

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5009,6 +5033,247 @@ mod tests {
...
+        assert_no_local_diagnostics!(&context);
+
+        assert_definition_at!(&context, "7:5-7:42", MethodAlias, |def| {
+            assert_string_eq!(&context, def.new_name_str_id(), "new_name()");
+            assert!(matches!(def.receiver(), Some(Receiver::ConstantReceiver(_))));
+        });
+    }
+
+    #[test]
+    fn index_alias_keyword_at_top_level() {
```
st0012: Already covered in `index_alias_methods_top_level`

rust/rubydex/src/resolution.rs
```diff
@@ -509,8 +509,27 @@ impl<'a> Resolver<'a> {
...
                 }
                 Definition::MethodAlias(alias) => {
-                    let Some(owner_id) = self.resolve_lexical_owner(*alias.lexical_nesting_id()) else {
-                        continue;
+                    // Method aliases operate on instance methods. The SelfReceiver arm is for
+                    // RBS `alias self.x self.y`.
+                    let owner_id = match alias.receiver() {
+                        Some(Receiver::SelfReceiver(def_id)) => {
+                            let Some(&owner_decl_id) = self.graph.definition_id_to_declaration_id(*def_id) else {
+                                continue;
```
st0012: If you look at how we resolve `SelfReceiver` for `MethodDefinition` (around line 385), you can see we'll raise in these case. And such panic helped us identified https://github.com/Shopify/rubydex/pull/656. So let's also do the same here: panic instead of silently continue.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: `assert_method_has_receiver!` already exists, I'll use that one 👍

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Not a helpful but a confusing comment. Removed, the test name is explicit enough.

rust/rubydex/src/resolution.rs
alexcrocha: Updated 👍

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: That's fair, I removed.

rust/rubydex/src/resolution.rs
```diff
@@ -2540,6 +2557,209 @@ mod tests {
...
+        });
+        context.resolve();
+
+        assert_no_diagnostics!(&context);
+
+        assert_members_eq!(context, "Foo", ["new_name()", "original()"]);
+    }
+
+    #[test]
+    fn resolving_method_alias_cross_file_constant_receiver() {
```
vinistock: These two tests are probably not necessary. The file has no influence in the resolution logic.

rust/rubydex/src/resolution.rs
```diff
@@ -2540,6 +2557,209 @@ mod tests {
...
+            end
+            "
+        });
+        context.resolve();
+
+        assert_no_diagnostics!(&context);
+
+        // Constant receiver takes precedence over singleton class context
+        assert_members_eq!(context, "Bar", ["new_name()", "original()"]);
+    }
```
vinistock: I believe all of these examples are testing the same thing: that we resolve the constant reference to determine the alias owner.

We probably don't need all of them since we already have all of these scenarios covered for constant resolution tests. I'd keep a single one.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5003,12 +5027,146 @@ mod tests {
...
+            "
+            class Foo
+              def original; end
+            end
+
+            foo.alias_method :new_name, :original
+            "
+        });
+
+        assert_no_local_diagnostics!(&context);
```
vinistock, line 5186: Might be worth a todo for adding a diagnostic later.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1645,6 +1645,19 @@ impl Visit<'_> for RubyIndexer<'_> {
                 }
             }
             "alias_method" => {
+                let recv_node = node.receiver();
```
vinistock, line 1648: It seems that we're using `recv_node.as_ref()` in a couple of places. Would it be possible to make it a reference here already?


```suggestion
                let recv_node = node.receiver().as_ref();
```

alexcrocha: We can do
```rust
let recv_node = node.receiver();
let recv_ref = recv_node.as_ref();
```
otherwise `as_ref()` would be borrowing a temporary `node.receiver()`. 
Maybe this is what you had in mind?

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Added a todo in the handler.

rust/rubydex/src/resolution.rs
alexcrocha: Kept the first one

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: I applied this change

vinistock: Yeah, that works!

rust/rubydex/src/resolution.rs
```diff
@@ -509,8 +509,25 @@ impl<'a> Resolver<'a> {
                     }
                 }
                 Definition::MethodAlias(alias) => {
-                    let Some(owner_id) = self.resolve_lexical_owner(*alias.lexical_nesting_id()) else {
-                        continue;
+                    // Method aliases operate on instance methods. The SelfReceiver arm is for
+                    // RBS `alias self.x self.y`.
+                    let owner_id = match alias.receiver() {
+                        Some(Receiver::SelfReceiver(def_id)) => *self
```
Morriar, line 515: I'm wondering if the `SelfReceiver` arm here is actually reachable? From what I can trace, the Ruby indexer maps `self.alias_method` to `receiver = None` (the `_` arm at line 1687 of ruby_indexer.rs), and the RBS indexer doesn't create `MethodAlias` definitions at all. So this arm appears to be dead code.

The test `resolving_method_alias_with_self_receiver` seems to confirm this — it uses `self.alias_method` but that goes through the `None` arm, not this one.

Is this intended as a forward-looking placeholder for RBS support? If so, a comment clarifying that would help. Otherwise we could simplify to just two arms.

alexcrocha: That was indeed forward-looking for RBS support and [#631](https://github.com/Shopify/rubydex/issues/631) landed right after as part of the stack, making the `SelfReceiver` arm reachable via `AliasKind::Singleton`.

My comment above was aiming to clarify that. I'll be more explicit next time.

PR #646, opened 2026-03-05
Add `name_dependents` reverse index for incremental invalidation

st0012:
Build a reverse index during indexing that tracks which definitions, references, and names depend on each `NameId`. Name-to-name edges encode the dependency type at registration time:

- `ChildName`: `parent_scope` relationship (structural dependency)
- `NestedName`: `nesting` relationship (reference-only dependency)
- `Definition/Reference`: direct dependents of the name

This index will be consumed by the incremental invalidation engine to efficiently cascade changes without scanning the full graph.

### DIFF
--- a/rust/rubydex/src/indexing/local_graph.rs
+++ b/rust/rubydex/src/indexing/local_graph.rs
@@ -3,6 +3,7 @@ use std::collections::hash_map::Entry;
 use crate::diagnostic::{Diagnostic, Rule};
 use crate::model::definitions::Definition;
 use crate::model::document::Document;
+use crate::model::graph::NameDependent;
 use crate::model::identity_maps::IdentityHashMap;
 use crate::model::ids::{DefinitionId, NameId, ReferenceId, StringId, UriId};
 use crate::model::name::{Name, NameRef};
@@ -18,6 +19,7 @@ type LocalGraphParts = (
     IdentityHashMap<NameId, NameRef>,
     IdentityHashMap<ReferenceId, ConstantReference>,
     IdentityHashMap<ReferenceId, MethodRef>,
+    IdentityHashMap<NameId, Vec<NameDependent>>,
 );
 
 #[derive(Debug)]
@@ -29,6 +31,7 @@ pub struct LocalGraph {
     names: IdentityHashMap<NameId, NameRef>,
     constant_references: IdentityHashMap<ReferenceId, ConstantReference>,
     method_references: IdentityHashMap<ReferenceId, MethodRef>,
+    name_dependents: IdentityHashMap<NameId, Vec<NameDependent>>,
 }
 
 impl LocalGraph {
@@ -42,6 +45,7 @@ impl LocalGraph {
             names: IdentityHashMap::default(),
             constant_references: IdentityHashMap::default(),
             method_references: IdentityHashMap::default(),
+            name_dependents: IdentityHashMap::default(),
         }
     }
 
@@ -69,6 +73,13 @@ impl LocalGraph {
 
     pub fn add_definition(&mut self, definition: Definition) -> DefinitionId {
         let definition_id = definition.id();
+
+        if let Some(name_id) = definition.name_id() {
+            self.name_dependents
+                .entry(*name_id)
+                .or_default()
+                .push(NameDependent::Definition(definition_id));
+        }
 
         if self.definitions.insert(definition_id, definition).is_some() {
             debug_assert!(false, "DefinitionId collision in local graph");
@@ -117,6 +128,18 @@ impl LocalGraph {
                 entry.get_mut().increment_ref_count(1);
             }
             Entry::Vacant(entry) => {
+                if let Some(&parent_scope) = name.parent_scope().as_ref() {
+                    self.name_dependents
+                        .entry(parent_scope)
+                        .or_default()
+                        .push(NameDependent::ChildName(name_id));
+                }
+                if let Some(&nesting_id) = name.nesting().as_ref() {
+                    self.name_dependents
+                        .entry(nesting_id)
+                        .or_default()
+                        .push(NameDependent::NestedName(name_id));
+                }
                 entry.insert(NameRef::Unresolved(Box::new(name)));
             }
         }
@@ -133,6 +156,10 @@ impl LocalGraph {
 
     pub fn add_constant_reference(&mut self, reference: ConstantReference) -> ReferenceId {
         let reference_id = reference.id();
+        self.name_dependents
+            .entry(*reference.name_id())
+            .or_default()
+            .push(NameDependent::Reference(reference_id));
 
         if self.constant_references.insert(reference_id, reference).is_some() {
             debug_assert!(false, "ReferenceId collision in local graph");
@@ -172,6 +199,13 @@ impl LocalGraph {
         self.document.add_diagnostic(diagnostic);
     }
 
+    // Name dependents
+
+    #[must_use]
+    pub fn name_dependents(&self) -> &IdentityHashMap<NameId, Vec<NameDependent>> {
+        &self.name_dependents
+    }
+
     // Into parts
 
     #[must_use]
@@ -184,6 +218,7 @@ impl LocalGraph {
             self.names,
             self.constant_references,
             self.method_references,
+            self.name_dependents,
         )
     }
 }
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -5917,3 +5917,82 @@ mod tests {
         });
     }
 }
+
+#[cfg(test)]
+mod name_dependent_tests {
+    use crate::assert_dependents;
+    use crate::test_utils::LocalGraphTest;
+
+    fn index_source(source: &str) -> LocalGraphTest {
+        LocalGraphTest::new("file:///foo.rb", source)
+    }
+
+    #[test]
+    fn track_dependency_chain() {
+        let context = index_source(
+            "
+            module Bar; end
+            CONST = 1
+            CONST2 = CONST
+
+            module Foo
+              class Bar::Baz
+                CONST
+              end
+
+              CONST2
+            end
+            ",
+        );
+
+        assert_dependents!(&context, "Bar", [ChildName("Baz")]);
+        assert_dependents!(&context, "Foo", [NestedName("Baz"), NestedName("CONST2")]);
+        assert_dependents!(&context, "Bar::Baz", [Definition("Baz"), NestedName("CONST")]);
+    }
+
+    #[test]
+    fn multi_level_chain() {
+        let context = index_source(
+            "
+            module Foo
+              module Bar
+                module Baz
+                end
+              end
+            end
+            ",
+        );
+
+        assert_dependents!(&context, "Foo", [NestedName("Bar")]);
+        assert_dependents!(&context, "Bar", [NestedName("Baz")]);
+    }
+
+    #[test]
+    fn singleton_class() {
+        let context = index_source(
+            "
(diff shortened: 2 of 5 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -14,6 +14,18 @@ use crate::model::references::{ConstantReference, MethodRef};
...
+/// An entity whose validity depends on a particular `NameId`.
+/// Used as the value type in the `name_dependents` reverse index.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NameDependent {
+    Definition(DefinitionId),
+    Reference(ReferenceId),
+    /// This name's `parent_scope` is the key name — structural dependency.
+    ChildName(NameId),
+    /// This name's `nesting` is the key name — reference-only dependency.
+    NestedName(NameId),
```
vinistock, line 26: Is there a logic need for this distinction? Or could we just merge them into `Name` for dependent names?

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -117,6 +128,21 @@ impl LocalGraph {
                 entry.get_mut().increment_ref_count(1);
             }
             Entry::Vacant(entry) => {
+                let parent_scope = name.parent_scope().as_ref().copied();
+                if let Some(ps_id) = parent_scope {
+                    self.name_dependents
+                        .entry(ps_id)
+                        .or_default()
+                        .push(NameDependent::ChildName(name_id));
+                }
```
vinistock: Nit


```suggestion
                if let Some(parent_scope) = name.parent_scope() {
                    self.name_dependents
                        .entry(*parent_scope)
                        .or_default()
                        .push(NameDependent::ChildName(name_id));
                }
```

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -117,6 +128,21 @@ impl LocalGraph {
...
+                if let Some(nesting_id) = name.nesting().as_ref().copied()
+                    && Some(nesting_id) != parent_scope
```
vinistock: In what situation can be nesting ID be the same as the parent scope? Can this ever happen?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5917,3 +5917,201 @@ mod tests {
         });
     }
 }
+
+#[cfg(test)]
+mod name_dependent_tests {
```
vinistock, line 5922: Not opposed to it, but is it common in Rust to split tests groups in different modules?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5917,3 +5917,201 @@ mod tests {
...
+                "expected {}({}) in {}'s dependents",
+                stringify!($variant),
+                $dep,
+                $owner
+            );
+        }};
+    }
+
+    /// Asserts that no name matching `owner` has a dependent of the given variant pointing to `dep`.
+    macro_rules! assert_no_dependent {
```
vinistock: We will need to assert these for RBS indexing too, so I suggest we already start by putting this `local_graph_test`.

rust/rubydex/src/model/graph.rs
st0012: In the next PR (#641), we will have something like:

```rs
NameDependent::ChildName(id) => queue.push(InvalidationItem::UnresolveName(*id)),
NameDependent::NestedName(id) => queue.push(InvalidationItem::UnresolveReferences(*id)),
```

The main difference is that `UnresolveReferences` will only unresolve constant references:

```rb
class Foo; end

class Bar
  Foo
  class Baz; end
  # Bar has [NestedName(Foo), ChildName(Baz)]
end
```

When `Bar`'s ancestors changes:
- `ChildName(Baz)` will trigger a total invalidation on `Baz`
- `NestedName(Foo)` will only invalide the `Foo` reference

rust/rubydex/src/indexing/ruby_indexer.rs
st0012: Yes. ZJIT [does this](https://github.com/ruby/ruby/blob/master/zjit/src/hir.rs#L8672-L8674) as well. IMO it's a nice way to scope test helpers.

rust/rubydex/src/indexing/local_graph.rs
st0012: I think this does:

```rb
class Object
  class Object::Foo
    def bar
      puts "bar"
    end
  end
end

Foo.new.bar # bar
```

st0012: Actually, no. This code is dead

rust/rubydex/src/indexing/ruby_indexer.rs
st0012: Moved

rust/rubydex/src/model/graph.rs
vinistock: If the ancestors of `Bar` change, all constant references inside of the namespace have to be invalidated. I suspect that we can merge these two because the information of what needs to be invalidated is already encoded in the hashmap.

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: It's too bad we didn't start out like that. We should've created modules for indexing each individual type of thing. Same for resolution, all ancestors tests could be separate.

Anyway, not worth the investment to refactor immediately, but something we may want later.

rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -117,6 +128,19 @@ impl LocalGraph {
                 entry.get_mut().increment_ref_count(1);
             }
             Entry::Vacant(entry) => {
+                let parent_scope = name.parent_scope().as_ref();
+                if let Some(&ps_id) = parent_scope {
```
vinistock:
```suggestion
                if let Some(&parent_scope) = name.parent_scope().as_ref() {
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -5917,3 +5917,90 @@ mod tests {
...
+              class Bar::Baz
+                CONST
+              end
+
+              CONST2
+            end
+            ",
+        );
+
+        assert_has_dependent!(&context, "Bar", ChildName, "Baz");
```
vinistock: Not a big deal, but would it be possible to make the assertion look like this?

```rust
assert_has_dependent!(&context, "Bar", ChildName("Baz"));
```

or maybe this would be clearer?

```rust
assert_dependents!(&context, "Bar, [...])
```

rust/rubydex/src/model/graph.rs
```diff
@@ -800,8 +855,10 @@ impl Graph {
                 }
             }
 
-            if let Some(name_id) = self.definitions.get(def_id).unwrap().name_id() {
-                self.untrack_name(*name_id);
+            let def_name_id = self.definitions.get(def_id).unwrap().name_id().copied();
+            if let Some(name_id) = def_name_id {
```
vinistock:
```suggestion
            if let Some(name_id) = self.definitions.get(def_id).unwrap().name_id().copied() {
```

rust/rubydex/src/model/graph.rs
```diff
@@ -1021,6 +1078,58 @@ mod tests {
...
+            .find(|(_, n)| *n.str() == StringId::from("Foo"))
+            .unwrap()
+            .0;
+
+        let count = |deps: &[NameDependent], f: fn(&NameDependent) -> bool| deps.iter().filter(|d| f(d)).count();
+
+        let foo_deps = context.graph().name_dependents().get(&foo_name_id).unwrap();
+        assert_eq!(count(foo_deps, |d| matches!(d, NameDependent::Definition(_))), 2); // `module Foo` in both foo.rb and bar.rb
+        assert_eq!(count(foo_deps, |d| matches!(d, NameDependent::NestedName(_))), 2); // Bar, CONST
+        assert_eq!(count(foo_deps, |d| matches!(d, NameDependent::Reference(_))), 0);
```
vinistock: It would be nice to use a custom assertion here to verify exactly what dependents remain.

rust/rubydex/src/indexing/ruby_indexer.rs
st0012: I opened https://github.com/Shopify/rubydex/issues/649 in case anyway wants to give it a try.

rust/rubydex/src/indexing/ruby_indexer.rs
st0012: Updated 👍

rust/rubydex/src/model/graph.rs

rust/rubydex/src/model/graph.rs
st0012: If it's ok, I'd like to merge this as it is, and see if the invalidation algo looks better/worse without the 2nd enum after some rounds of reviews.

rust/rubydex/src/model/graph.rs
```diff
@@ -735,6 +780,15 @@ impl Graph {
                 debug_assert!(false, "Method ReferenceId collision in global graph");
             }
         }
+
+        for (name_id, deps) in name_dependents {
+            let global_deps = self.name_dependents.entry(name_id).or_default();
+            for dep in deps {
+                if !global_deps.contains(&dep) {
```
Morriar, line 787: The dedup check here is O(n) per dep, making the whole merge loop O(n²) for any name with many dependents. For the initial indexing pass where we're merging many local graphs, this could be costly on large codebases.

I wonder if we should use a `HashSet` during the merge and then convert back to `Vec`, or just rely on the fact that duplicates shouldn't occur at all if the local graph was built correctly (given that a local graph only covers one file)?

If duplicates are theoretically impossible, a `debug_assert` here would be more appropriate.

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -42,6 +43,79 @@ impl GraphTest {
         resolver.resolve_all();
     }
 
+    // Name dependents helpers (shared with LocalGraphTest for assert_dependents! macro)
+
+    /// # Panics
+    ///
+    /// Panics if no names match the given path.
+    #[must_use]
+    pub fn find_name_ids(&self, path: &str) -> Vec<NameId> {
```
Morriar, line 52: Both `GraphTest` and `LocalGraphTest` have identical implementations of `find_name_ids`, `name_dependents_for`, `name_str`, and `dependent_name_str`. Any way we could avoid the duplication here? A shared trait or a helper module would keep these in sync automatically.

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -42,6 +43,79 @@ impl GraphTest {
...
+            .graph()
+            .names()
+            .iter()
+            .filter(|(_, name_ref)| {
+                if *name_ref.str() != target_str_id {
+                    return false;
+                }
+                match parent {
+                    None => name_ref.parent_scope().as_ref().is_none(),
+                    Some(p) => name_ref.parent_scope().as_ref().is_some_and(|ps_id| {
```
Morriar, line 68: Two thoughts:

1. `find_name_ids` with a path like `"Bar::Baz"` only checks one level up — it verifies the immediate parent's string is `"Bar"`, but not that `Bar` is itself a top-level name. So `Foo::Bar::Baz` would also match `"Bar::Baz"` since the parent's str is `"Bar"` regardless of where `Bar` lives. Is that intentional?

2. Same question for `LocalGraphTest::find_name_ids` (same code).

PR #651, opened 2026-03-06
Attempt to use receiver as parent class when it's a call node

Morriar:
## Summary

* When `class Foo < Bar` is indexed and `Bar` is never defined, `Foo`'s ancestor chain was silently falling back to `[Foo, Object, BasicObject]` (partial) — `Bar` was completely absent. This is inconsistent with how unresolved mixins are handled, which push `Ancestor::Partial(name_id)` as a placeholder.
* Fixed by applying the same approach for superclasses: when no resolved parent is found but an unresolved reference exists, `linearize_parent_class` now returns `Ancestors::Partial([Ancestor::Partial(name_id)])` instead of falling back to Object's chain.
* Updated `assert_ancestors_eq` to compare by name string, so it works uniformly for both `Complete` and `Partial` ancestor entries (previously panicked on `Ancestors::Partial`).
* Added `Class#superclass_name` Ruby API that returns the unresolved superclass name as written in source, or `nil` when resolved or absent.

## Test plan

* [ ] New Rust test: `unresolved_superclass_is_represented_as_partial_ancestor` verifies `class Foo < Bar` yields `["Foo", "Bar"]`
* [ ] New Ruby tests: `test_superclass_name_returns_nil_when_superclass_is_resolved` and `test_superclass_name_returns_name_when_superclass_is_unresolved`
* [ ] `cargo test` — all 380 tests pass
* [ ] `bundle exec ruby -Itest test/declaration_test.rb` — all 23 tests pass
* [ ] `cargo clippy --all-targets -- -D warnings` — clean

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -644,13 +644,30 @@ impl<'a> RubyIndexer<'a> {
         let (comments, flags) = self.find_comments_for(offset.start());
         let lexical_nesting_id = self.parent_lexical_scope_id();
         let superclass = superclass_node.as_ref().and_then(|n| {
-            self.index_constant_reference(n, false).map(|id| {
-                self.local_graph.add_constant_reference(ConstantReference::new(
+            // Try direct constant reference first
+            if let Some(id) = self.index_constant_reference(n, false) {
+                return Some(self.local_graph.add_constant_reference(ConstantReference::new(
                     id,
                     self.uri_id,
                     Offset::from_prism_location(&n.location()),
-                ))
-            })
+                )));
+            }
+
+            // For call nodes (e.g. `ActiveRecord::Migration[7.0]`), try the receiver constant
+            if let ruby_prism::Node::CallNode { .. } = n {
+                let call = n.as_call_node().unwrap();
+                if let Some(receiver) = call.receiver()
+                    && let Some(id) = self.index_constant_reference(&receiver, false)
+                {
+                    return Some(self.local_graph.add_constant_reference(ConstantReference::new(
+                        id,
+                        self.uri_id,
+                        Offset::from_prism_location(&receiver.location()),
+                    )));
+                }
+            }
+
+            None
         });
 
         if let Some(superclass_node) = superclass_node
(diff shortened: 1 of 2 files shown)


ext/rubydex/declaration.c
```diff
@@ -272,6 +272,26 @@ static VALUE rdxr_declaration_ancestors(VALUE self) {
     return self;
 }
 
+// Class#superclass_name: () -> String?
+// Returns the unresolved superclass name as written in source, or nil if the superclass is
+// resolved (i.e. it exists in the indexed workspace) or there is no superclass.
+static VALUE rdxr_class_superclass_name(VALUE self) {
```
vinistock: I wonder if we should instead return a constant reference, where the concept of it being resolved or not is and what declaration it points to is encapsulated in it.

In fact, this is related to the reference related refactors I'm working on. The type hierarchy request needs to know direct ancestors (mixins and parent classes) and I wanted to make sure we were returning proper references.

What do you think?

We can always remove this part of the PR and ship just the partial ancestors part while we refactor references.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4631,7 +4645,7 @@ mod tests {
         });
 
         assert_definition_at!(&context, "3:1-3:54", Class, |def| {
-            assert!(def.superclass_ref().is_none(),);
+            assert!(def.superclass_ref().is_some(),);
```
vinistock: There's a weird trailing comma.
```suggestion
            assert!(def.superclass_ref().is_some());
```

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -372,36 +372,28 @@ macro_rules! assert_ancestors_eq {
             .get(&$crate::model::ids::DeclarationId::from($name))
             .unwrap();
 
-        match declaration.as_namespace().unwrap().ancestors() {
+        let ancestors = match declaration.as_namespace().unwrap().ancestors() {
```
vinistock: Is this change to allow asserting partial ancestors? I think Stan made this possible in one of his recent PRs.

vinistock (commented):
I think the resolution changes can be shipped, but I do think the `superclass` API will be nicer returning constant reference objects

vinistock (approved):

PR #654, opened 2026-03-10
Memoize name_depth to speed up resolution sorting by 3x

st0012:
## Summary

- Pre-compute `name_depth` for all names into an `IdentityHashMap<NameId, u32>` cache before sorting, eliminating redundant recursive walks during O(n log n) comparisons
- Switch `sort_by` to `sort_unstable_by` since the full `(depth, uri, offset)` key provides deterministic ordering without needing stability

## Problem

Profiling with `samply` revealed that **88% of sampled resolution time** was spent inside `name_depth` closures during `prepare_units` sorting. The function recursively walks `parent_scope` and `nesting` chains to compute depth, and was called from the sort comparator on every comparison — recomputing the same depths millions of times with no memoization.

Note: the depth sort is a **correctness requirement**, not just an optimization. Removing it entirely causes 13 test failures — the resolution loop's `made_progress` check gives up when children are processed before parents.

## Fix

Compute depths once for all names in a single memoized pass, then use O(1) lookups in the sort comparators.

## Benchmark

| **Resolution** | 50.2s (79.4%) | **16.7s (57.7%)** | **-66.7%** |
| **Total** | 63.2s | **29.0s** | **-54.1%** |
| Listing | 0.76s | 0.80s | ~same |
| Indexing | 11.5s | 10.8s | ~same |
| Querying | 0.72s | 0.66s | ~same |
| Memory (RSS) | 4863 MB | 4833 MB | ~same |
| Declarations | 879,648 | 879,648 | identical |
| Definitions | 1,043,725 | 1,043,725 | identical |

Resolution went from **50.2s → 16.7s** (3x speedup). Total indexing time cut in half. Output is identical (same counts, same orphan rate), confirming correctness.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -11,7 +11,7 @@ use crate::model::{
     },
     definitions::{Definition, Mixin, Receiver},
     graph::{CLASS_ID, Graph, MODULE_ID, OBJECT_ID},
-    identity_maps::{IdentityHashMap, IdentityHashSet},
+    identity_maps::{IdentityHashBuilder, IdentityHashMap, IdentityHashSet},
     ids::{DeclarationId, DefinitionId, NameId, ReferenceId, StringId},
     name::{Name, NameRef, ParentScope},
 };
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1583,8 +1597,8 @@ impl<'a> Resolver<'a> {
 
         // Sort namespaces based on their name complexity so that simpler names are always first
         // When the depth is the same, sort by URI and offset to maintain determinism
-        definitions.sort_by(|(_, (name_a, uri_a, offset_a)), (_, (name_b, uri_b, offset_b))| {
-            (Self::name_depth(name_a, names), uri_a, offset_a).cmp(&(Self::name_depth(name_b, names), uri_b, offset_b))
+        definitions.sort_unstable_by(|(_, (name_a, uri_a, offset_a)), (_, (name_b, uri_b, offset_b))| {
```
st0012, line 1600: This change is not required for the rest of the PR, but since we can't have 2 identical `depth, uri, offset` group, we can use unstable sort for a bit of speedup.

rust/rubydex/src/resolution.rs
```diff
@@ -1596,14 +1610,14 @@ impl<'a> Resolver<'a> {
...
-                    (names.get(constant_ref.name_id()).unwrap(), uri, constant_ref.offset()),
+                    (*constant_ref.name_id(), uri, constant_ref.offset()),
                 )
             })
             .collect::<Vec<_>>();
 
         // Sort constant references based on their name complexity so that simpler names are always first
-        const_refs.sort_by(|(_, (name_a, uri_a, offset_a)), (_, (name_b, uri_b, offset_b))| {
-            (Self::name_depth(name_a, names), uri_a, offset_a).cmp(&(Self::name_depth(name_b, names), uri_b, offset_b))
+        const_refs.sort_unstable_by(|(_, (name_a, uri_a, offset_a)), (_, (name_b, uri_b, offset_b))| {
```
st0012, line 1619: Same as the above

PR #656, opened 2026-03-11
Create Todo declarations for multi-level compact namespaces

### LINKED ISSUE #663: MCP panic: SelfReceiver definition should have a declaration
I was testing the MCP on a clone of https://github.com/steveclarke/real-world-rails.git and it reported the error:

```
thread '<unnamed>' (4202448) panicked at rubydex/src/resolution.rs:386:42:
SelfReceiver definition should have a declaration
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Rubydex indexing failed: SelfReceiver definition should have a declaration
```

I narrowed it down to be this file https://github.com/instructure/canvas-lms/blob/d5dc9ac2069c8287f45423494fa953401014b415/app/services/differentiation_tag/jobs/workers/tag_override_converter_worker.rb and specifically the methods `perform` and `convert_tags` that triggers the error.

st0012:
## Problem

`class A::B::C` where `A` is unknown panics when the class body contains `def self.foo` with instance variables:

```
thread panicked at rubydex/src/resolution.rs:386:42:
SelfReceiver definition should have a declaration
```

The panic occurs in `handle_remaining_definitions` because `C`'s definition never gets a declaration — the resolution loop retries indefinitely and then gives up.

Fixes #663

## Root cause

The Todo mechanism from #529 only handles **one level** of unknown parent in compact notation. For `class Foo::Bar` where `Foo` is unknown, `resolve_constant_internal(Foo)` returns `Unresolved(None)`, which triggers Todo creation in `name_owner_id`.

For multi-level paths like `class A::B::C`, the flow differs:

1. `name_owner_id(C)` calls `resolve_constant_internal(B)`
2. `B` has parent scope `A` — checks if `A`'s name is resolved
3. `A` is unresolved → returns `Retry(None)`
4. `name_owner_id` passes `Retry(None)` through unchanged — **never reaches the Todo creation branch**
5. The unit is re-enqueued, but nothing ever resolves `A`, so the loop stalls

## Fix

Two changes:

### 1. `name_owner_id` returns `Unresolved(None)` for unknown parent scopes

In the parent scope branch, both `Retry(None)` and `Unresolved(None)` from `resolve_constant_internal` now return `Unresolved(None)`. This distinguishes "parent is genuinely unknown" from other failure modes (circular aliases return `Retry(None)` via `resolve_to_primary_namespace`, nesting issues return `Retry(None)` from the nesting branch).

### 2. `handle_constant_declaration` creates Todos when it receives `Unresolved(None)`

The **reason** to create Todos comes from the caller — it has a definition that needs an owner. Todo creation is extracted into `create_todo_for_parent`, which recursively walks the parent chain creating placeholders. If real definitions appear later, the existing promotion mechanism upgrades Todos in place.

### 3. Todo tests moved to dedicated module

All tests asserting `<TODO>` behavior are now in `resolution::todo_tests`.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashSet, VecDeque},
+    collections::{HashSet, VecDeque, hash_map::Entry},
     hash::BuildHasher,
 };
 
@@ -1131,14 +1131,13 @@ impl<'a> Resolver<'a> {
 
         let parent_name = self.graph.names().get(&parent_scope).unwrap();
         let parent_str_id = *parent_name.str();
-        let parent_has_explicit_prefix = parent_name.parent_scope().as_ref().is_some();
-        // NLL: borrow of parent_name ends here
+        let parent_has_parent_scope = parent_name.parent_scope().as_ref().is_some();
+        // Non-Lexical Lifetimes: borrow of parent_name ends here
 
-        // For bare names (no explicit `::` prefix), always use OBJECT_ID as the owner.
-        // Using nesting here would create "Nesting::Bar" instead of "Bar" for a bare `Bar`
-        // reference, which is incorrect: if `Bar` can't be found anywhere, the placeholder
-        // should live at the top level so it can be promoted when `module Bar` appears later.
-        let parent_owner_id = if parent_has_explicit_prefix {
+        // For `class A::B::C` where `A` is bare (no `::` prefix), place the Todo under
+        // Object so it becomes top-level `A`. This way `module A; end` appearing later
+        // promotes it correctly. Using nesting would incorrectly create `SomeModule::A`.
+        let parent_owner_id = if parent_has_parent_scope {
             match self.name_owner_id(parent_scope) {
                 Outcome::Resolved(id, _) => id,
                 Outcome::Unresolved(None) => self.create_todo_for_parent(parent_scope),
@@ -1160,7 +1159,7 @@ impl<'a> Resolver<'a> {
 
         let declaration_id = DeclarationId::from(&fully_qualified_name);
 
-        if let std::collections::hash_map::Entry::Vacant(e) = self.graph.declarations_mut().entry(declaration_id) {
+        if let Entry::Vacant(e) = self.graph.declarations_mut().entry(declaration_id) {
             e.insert(Declaration::Namespace(Namespace::Todo(Box::new(TodoDeclaration::new(
                 fully_qualified_name,
                 parent_owner_id,
@@ -5824,5 +5823,39 @@ mod todo_tests {
         assert_declaration_kind_eq!(context, "Bar::Baz::Qux", "Class");
         assert_members_eq!(context, "Bar", vec!["Baz"]);
         assert_members_eq!(context, "Bar::Baz", vec!["Qux"]);
+    }
+
+    #[test]
+    fn no_todo_when_parent_is_reachable_through_include() {
+        // Baz::Qux inside Foo, where Baz comes from included Bar module.
+        // Baz::Qux should resolve through inheritance to Bar::Baz::Qux, not create
+        // a top-level Baz Todo.
+        let mut context = GraphTest::new();
+        context.index_uri("file:///file1.rb", {
+            r"
+            module Foo
+              include Bar
+
+              class Baz::Qux; end
+            end
+            "
+        });
+        context.index_uri("file:///file2.rb", {
+            r"
+            module Bar
+              module Baz; end
+            end
+            "
+        });
+        context.resolve();
+
+        assert_declaration_exists!(context, "Bar::Baz");
+        assert_declaration_exists!(context, "Bar::Baz::Qux");
+        assert_members_eq!(context, "Bar::Baz", vec!["Qux"]);
+        assert_declaration_does_not_exist!(context, "Foo::Baz");
+        // No spurious top-level Baz Todo should be created
+        assert_declaration_does_not_exist!(context, "Baz");
+        // Baz::Qux should NOT exist at top level
+        assert_declaration_does_not_exist!(context, "Baz::Qux");
     }
 }


st0012:
@Morriar There are different ways to resolve the panic mentioned in the description. Looking at #529, I don't think 1-level-only todo declaration creation is intentional, so I opened this PR. If that's the intend, I think we should document it somewhere (architecture.md?), and I have another way to address the panic.

rust/rubydex/src/resolution.rs
```diff
@@ -1060,15 +1060,16 @@ impl<'a> Resolver<'a> {
...
             // If we have `A::B`, the owner of `B` is whatever `A` resolves to.
             // If `A` is an alias, resolve through to get the actual namespace.
-            // On `Retry`, we don't create a Todo: the parent may still resolve through inheritance once ancestors are
-            // linearized. We only create Todos for `Unresolved` outcomes where the parent is genuinely unknown.
             match self.resolve_constant_internal(parent_scope) {
                 Outcome::Resolved(id, linearization) => self.resolve_to_primary_namespace(id, linearization),
-                // Retry or Unresolved(Some(_)) means we might find it later through ancestor linearization
-                Outcome::Retry(id) => Outcome::Retry(id),
+                // Retry(Some) or Unresolved(Some) means we might find it later through ancestor linearization
+                Outcome::Retry(Some(id)) => Outcome::Retry(Some(id)),
```
vinistock: Why do these lines need to change?

rust/rubydex/src/resolution.rs
```diff
@@ -1060,15 +1060,16 @@ impl<'a> Resolver<'a> {
...
                 Outcome::Unresolved(Some(id)) => Outcome::Unresolved(Some(id)),
-                // Only create a Todo when genuinely unresolvable (no pending linearizations)
-                Outcome::Unresolved(None) => {
+                // Retry(None) means the parent's own parent scope is unresolved (e.g., B in `A::B::C`
```
vinistock: If we're going to do this for all retries, then there might not be a reason for both `Retry` and `Unresolved` to exist.

rust/rubydex/src/resolution.rs
```diff
@@ -5463,4 +5464,166 @@ mod tests {
...
+        assert_declaration_kind_eq!(context, "A", "Module");
+        assert_declaration_kind_eq!(context, "A::B", "Module");
+        assert_declaration_kind_eq!(context, "A::B::C", "Class");
+        assert_members_eq!(context, "A", vec!["B"]);
+        assert_members_eq!(context, "A::B", vec!["C"]);
+        assert_members_eq!(context, "A::B::C", vec!["foo()"]);
+    }
+
+    #[test]
+    fn todo_chain_with_self_method_and_ivar() {
```
vinistock: Can we add a test demonstrating what happens in this scenario?

```ruby
module Foo
  class Bar::Baz::Qux
  end
end

module Bar; end
```

The two declarations should be `Foo` and `Bar::Baz::Qux`.

rust/rubydex/src/resolution.rs
```diff
@@ -1158,6 +1122,55 @@ impl<'a> Resolver<'a> {
...
+    /// so `B::C` can still be placed. Recurses for multi-level cases. Todos get promoted
+    /// when real definitions appear later.
+    fn create_todo_for_parent(&mut self, name_id: NameId) -> DeclarationId {
+        let name_ref = self.graph.names().get(&name_id).unwrap();
+        let parent_scope = *name_ref.parent_scope().as_ref().unwrap();
+
+        let parent_name = self.graph.names().get(&parent_scope).unwrap();
+        let parent_str_id = *parent_name.str();
+        let parent_has_explicit_prefix = parent_name.parent_scope().as_ref().is_some();
+        // NLL: borrow of parent_name ends here
```
vinistock: What is `NLL`?

rust/rubydex/src/resolution.rs
```diff
@@ -1158,6 +1122,55 @@ impl<'a> Resolver<'a> {
...
+    /// For `class A::B::C` where `A` can't be resolved, creates a Todo declaration for `A`
+    /// so `B::C` can still be placed. Recurses for multi-level cases. Todos get promoted
+    /// when real definitions appear later.
+    fn create_todo_for_parent(&mut self, name_id: NameId) -> DeclarationId {
+        let name_ref = self.graph.names().get(&name_id).unwrap();
+        let parent_scope = *name_ref.parent_scope().as_ref().unwrap();
+
+        let parent_name = self.graph.names().get(&parent_scope).unwrap();
+        let parent_str_id = *parent_name.str();
+        let parent_has_explicit_prefix = parent_name.parent_scope().as_ref().is_some();
```
vinistock: Let's keep the consistent naming of `parent_scope` instead of prefix.

rust/rubydex/src/resolution.rs
```diff
@@ -1158,6 +1122,55 @@ impl<'a> Resolver<'a> {
...
+            format!(
+                "{}::{}",
+                self.graph.declarations().get(&parent_owner_id).unwrap().name(),
+                self.graph.strings().get(&parent_str_id).unwrap().as_str()
+            )
+        };
+
+        let declaration_id = DeclarationId::from(&fully_qualified_name);
+
+        if let std::collections::hash_map::Entry::Vacant(e) = self.graph.declarations_mut().entry(declaration_id) {
```
vinistock: Let's import `Entry` instead of using the long fully qualified path.

rust/rubydex/src/resolution.rs
```diff
@@ -1158,6 +1122,55 @@ impl<'a> Resolver<'a> {
...
+
+        let parent_name = self.graph.names().get(&parent_scope).unwrap();
+        let parent_str_id = *parent_name.str();
+        let parent_has_explicit_prefix = parent_name.parent_scope().as_ref().is_some();
+        // NLL: borrow of parent_name ends here
+
+        // For bare names (no explicit `::` prefix), always use OBJECT_ID as the owner.
+        // Using nesting here would create "Nesting::Bar" instead of "Bar" for a bare `Bar`
+        // reference, which is incorrect: if `Bar` can't be found anywhere, the placeholder
+        // should live at the top level so it can be promoted when `module Bar` appears later.
```
vinistock: I'm not 100% sure that this statement is correct. For example, let's say you have this scenario:

```ruby
# Process this first

# file1.rb
module Foo
  include Bar
  
  class Baz::Qux; end
end

# file2.rb
module Bar
  module Baz; end
end
```

If we ended up processing `Baz::Qux` first, we would create `Baz` at the top level, but in reality this refers to `Bar::Baz::Qux`.

Is the code accounting for this?

rust/rubydex/src/resolution.rs
st0012: It means `Non-Lexical Lifetimes`. It's from the implementation in main.

rust/rubydex/src/resolution.rs
st0012: Yes this case passes:

```rs
    #[test]
    fn compact_namespace_inside_module_with_included_ancestor() {
        // Baz::Qux inside Foo, where Baz comes from included Bar module.
        // Baz::Qux should resolve through inheritance to Bar::Baz::Qux, not create
        // a top-level Baz Todo.
        context.index_uri("file:///file1.rb", {
            r"
            module Foo
              include Bar

              class Baz::Qux; end
            end
            "
        });
        context.index_uri("file:///file2.rb", {
            r"
            module Bar
              module Baz; end
            end
            "
        });
        context.resolve();

        assert_declaration_exists!(context, "Bar::Baz");
        assert_declaration_exists!(context, "Bar::Baz::Qux");
        assert_members_eq!(context, "Bar::Baz", vec!["Qux"]);
        assert_declaration_does_not_exist!(context, "Foo::Baz");
        // No spurious top-level Baz Todo should be created
        assert_declaration_does_not_exist!(context, "Baz");
        // Baz::Qux should NOT exist at top level
        assert_declaration_does_not_exist!(context, "Baz::Qux");
    }
```

rust/rubydex/src/resolution.rs
```diff
@@ -5574,83 +5639,223 @@ mod tests {
...
+        assert_declaration_kind_eq!(context, "A::B::D", "Class");
+        assert_members_eq!(context, "Object", vec!["A"]);
+        assert_members_eq!(context, "A", vec!["B"]);
+        assert_members_eq!(context, "A::B", vec!["C", "D"]);
+        assert_members_eq!(context, "A::B::C", vec!["c_method()"]);
+        assert_members_eq!(context, "A::B::D", vec!["d_method()"]);
+    }
+
+    #[test]
+    fn promoted_incrementally() {
```
vinistock, line 5739: This test will be particularly important for incremental edits. Right now, we erase all declarations, so there's nothing fancy going on to support this scenario.

PR #659, opened 2026-03-12
Use NonZeroU64 for Id<T> to enable niche optimization on Option<Id>

st0012:
## Summary

- Change `Id<T>` from `u64` to `NonZeroU64` so that `Option<Id<T>>` is 8 bytes instead of 16 bytes
- Hash value 0 is remapped to `u64::MAX` (collision probability is 2^-64, same as any other hash collision)
- Propagates savings through every struct with an `Option<XxxId>` field:
  - MethodRef: 40 → 32 bytes, Name: 48 → 40 bytes, ClassDefinition: 144 → 128 bytes
  - All other definition structs: -8 bytes each, namespace declarations: 224 → 216 bytes

## Benchmark (96k files)

| RSS | 3505 MB | 3424 MB | **-2.3%** |
| Indexing | 7.87s | 6.68s | -15% |
| Resolution | 56.34s | 55.90s | ~0% |
| Total | 65.79s | 64.11s | -3% |
| Declarations | 945,412 | 945,412 | identical |
| Definitions | 1,063,171 | 1,063,171 | identical |

**Note:** The original benchmark showed -11.2% memory (4480 → 3978 MB), but that was measured before #666 landed. Pre-#666, all 96K `LocalGraph`s were buffered in the channel at peak (verified: 96,240 buffered, RSS 4481 MB). Post-#666, merging overlaps with indexing so at most ~19K are buffered at once (RSS 3307 MB). The struct size savings (~18 MB theoretical) now apply to a much smaller peak working set, explaining the reduced delta.

### DIFF
--- a/rust/rubydex/src/model/id.rs
+++ b/rust/rubydex/src/model/id.rs
@@ -1,14 +1,24 @@
-use std::{
-    hash::{Hash, Hasher},
-    marker::PhantomData,
-    ops::Deref,
-};
+use std::{marker::PhantomData, num::NonZeroU64, ops::Deref};
 use xxhash_rust::xxh3;
 
-/// A deterministic type-safe ID representation
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Maps a u64 hash to a `NonZeroU64` by replacing 0 with `u64::MAX`.
+/// The probability of a 64-bit hash being exactly 0 is 2^-64 (~5.4e-20),
+/// and remapping 0 → MAX just means those two inputs collide — the same
+/// risk as any other hash collision, which the system already handles.
+const MAX_NONZERO: NonZeroU64 = NonZeroU64::new(u64::MAX).unwrap();
+
+#[inline]
+fn to_non_zero(value: u64) -> NonZeroU64 {
+    NonZeroU64::new(value).unwrap_or(MAX_NONZERO)
+}
+
+/// A deterministic type-safe ID representation.
+///
+/// Uses `NonZeroU64` internally so that `Option<Id<T>>` is 8 bytes (same as `Id<T>`)
+/// via niche optimization, instead of 16 bytes with a plain `u64`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Id<T> {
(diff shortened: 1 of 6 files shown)


rust/rubydex/src/model/id.rs
```diff
@@ -1,22 +1,43 @@
...
     _marker: PhantomData<T>,
 }
 
+impl<T> PartialEq for Id<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.value == other.value
+    }
+}
+
+impl<T> Eq for Id<T> {}
```
vinistock: Could we derive these? Or do we need these manual implementations?

rust/rubydex/src/model/id.rs
```diff
@@ -36,9 +59,21 @@ impl<T> std::fmt::Display for Id<T> {
...
+        Some(self.cmp(other))
+    }
+}
+
+impl<T> Ord for Id<T> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.value.cmp(&other.value)
+    }
+}
+
```
vinistock: Similar question: can we derive these?

rust/rubydex/src/model/id.rs
```diff
@@ -81,4 +116,16 @@ mod tests {
         let id = TestId::new(123);
         assert_eq!(*id, 123);
     }
+
+    #[test]
+    fn option_id_is_8_bytes() {
```
vinistock: Nit
```suggestion
    fn optional_id_is_still_8_bytes() {
```

rust/rubydex/src/model/id.rs
```diff
@@ -26,7 +47,9 @@ impl<T> Deref for Id<T> {
     type Target = u64;
 
     fn deref(&self) -> &Self::Target {
-        &self.value
+        // SAFETY: NonZeroU64 has the same layout as u64, and we only need the numeric value.
+        // We return a reference to the inner u64 representation.
+        unsafe { &*(std::ptr::from_ref(&self.value).cast::<u64>()) }
```
vinistock: I'm surprised it requires all of this to get the value out. Are there any simpler alternatives?

PR #661, opened 2026-03-12
Merge LocalGraphs concurrently with indexing workers

st0012:
## Summary

- Previously, all indexing workers completed before any LocalGraphs were merged into the global Graph
- With an unbounded channel, this buffered all ~96K LocalGraphs simultaneously — hundreds of MB of temporary data
- Now a dedicated merge thread consumes LocalGraphs as workers produce them via `std::thread::scope`, keeping only a small number buffered at any time
- Also improves indexing throughput by overlapping merge work with parsing

## Benchmark

| Memory (peak footprint) | 4480 MB | 3497 MB | **-21.9%** |
| Declarations | 895,844 | 895,844 | identical |
| Definitions | 1,063,171 | 1,063,171 | identical |


st0012:
Superseded by #666

PR #662, opened 2026-03-12
Fix method_receiver for self at class body level

### LINKED ISSUE #657: method_receiver returns inconsistent receiver for self.method
`self.bar` inside `class Foo` and `Foo.bar` both call a method on the `Foo` class object, but `method_receiver` returns different receivers:

```ruby 
  class Foo
    self.bar  # receiver: Foo
  end

  Foo.bar     # receiver: <Foo>
```

I would expect these to be equivalent, as both target a method on the singleton class of `Foo`.

Found while pairing on #645 and investigating the behaviour of `method_receiver`.

alexcrocha:
Fixes #657 

At class body level, `self` is the class object. Method calls on it (`self.bar`, `bar`, `Foo.bar`) all resolve through the singleton class. `method_receiver` was not setting `is_singleton_name` in the `SelfNode | None` arm for `Class`, `Module`, and `SingletonClass` definitions, so `self.bar` inside `class Foo` returned receiver `Foo` while `Foo.bar` returned `<Foo>`.

Now all three produce `<Foo>` consistently.

### Tests

| Test | Ruby source | Expected receiver |
|------|------------|-------------------|
| `self_at_class_level_matches_constant` | `self.bar` inside `class Foo` vs `Foo.baz` outside | Both `<Foo>` |
| `implicit_at_class_level` | `bar` inside `class Foo` | `<Foo>` |
| `self_at_module_level` | `self.bar` inside `module Foo` | `<Foo>` |
| `self_inside_singleton_class` | `self.bar` inside `class << self` | `<<Foo>>` |
| `implicit_inside_singleton_class` | `bar` inside `class << self` | `<<Foo>>` |
| `at_top_level` | `self.bar` at top level | `Object` |

Count assertions in `indexing_the_same_document_twice` updated: `alias_method` calls at class body level now create singleton names and constant references.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -2168,6 +2168,34 @@ mod tests {
         }};
     }
 
+    /// Asserts that a method reference has the expected receiver.
+    ///
+    /// Finds a `MethodRef` by name and checks its receiver's string representation.
+    ///
+    /// Usage:
+    /// - `assert_method_ref_receiver!(context, "bar", "<Foo>")`
+    macro_rules! assert_method_ref_receiver {
+        ($context:expr, $method_name:expr, $expected_receiver:expr) => {{
+            let method_ref = $context
+                .graph()
+                .method_references()
+                .values()
+                .find(|method_ref| *method_ref.str() == StringId::from($method_name))
+                .unwrap_or_else(|| panic!("should have a method reference for {}", $method_name));
+
+            let receiver = $context.graph().names().get(&method_ref.receiver().unwrap()).unwrap();
+
+            assert_eq!(
+                StringId::from($expected_receiver),
+                *receiver.str(),
+                "receiver mismatch for `{}`: expected `{}`, got `{}`",
+                $method_name,
+                $expected_receiver,
+                $context.graph().strings().get(receiver.str()).unwrap().as_str()
+            );
+        }};
+    }
+
     macro_rules! assert_promotable {
         ($def:expr) => {{
             assert!(
@@ -4142,20 +4170,9 @@ mod tests {
 
         assert_no_local_diagnostics!(&context);
 
-        let find_ref = |name: &str| {
-            context
-                .graph()
-                .method_references()
-                .values()
-                .find(|method_ref| *method_ref.str() == StringId::from(name))
-                .unwrap_or_else(|| panic!("should have a method reference for {name}"))
-        };
-
-        for name in ["bar", "baz", "qux"] {
-            let method_ref = find_ref(name);
-            let receiver = context.graph().names().get(&method_ref.receiver().unwrap()).unwrap();
-            assert_eq!(StringId::from("<Foo>"), *receiver.str(), "receiver mismatch for {name}");
-        }
+        assert_method_ref_receiver!(context, "bar", "<Foo>");
+        assert_method_ref_receiver!(context, "baz", "<Foo>");
+        assert_method_ref_receiver!(context, "qux", "<Foo>");
     }
 
     #[test]
@@ -4169,10 +4186,7 @@ mod tests {
         });
 
         assert_no_local_diagnostics!(&context);
-
-        let bar_ref = context.graph().method_references().values().next().unwrap();
-        let receiver = context.graph().names().get(&bar_ref.receiver().unwrap()).unwrap();
-        assert_eq!(StringId::from("<Foo>"), *receiver.str());
+        assert_method_ref_receiver!(context, "bar", "<Foo>");
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4119,6 +4128,126 @@ mod tests {
...
+        let constant_receiver = context.graph().names().get(&constant_ref.receiver().unwrap()).unwrap();
+        assert_eq!(StringId::from("<Foo>"), *constant_receiver.str());
+    }
+
+    #[test]
+    fn index_method_receiver_implicit_at_class_level() {
+        let context = index_source({
+            "
+            class Foo
+              bar
```
st0012: Let's merge this case with the above one?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4119,6 +4128,126 @@ mod tests {
...
+
+        assert_no_local_diagnostics!(&context);
+
+        let bar_ref = context.graph().method_references().values().next().unwrap();
+        let receiver = context.graph().names().get(&bar_ref.receiver().unwrap()).unwrap();
+        assert_eq!(StringId::from("<<Foo>>"), *receiver.str());
+    }
+
+    #[test]
+    fn index_method_receiver_implicit_inside_singleton_class() {
```
st0012: Merge with the above one

rust/rubydex/src/model/graph.rs
```diff
@@ -1833,18 +1833,18 @@ mod tests {
 
         context.index_uri("file:///foo.rb", source);
         assert_eq!(44, context.graph().definitions.len());
-        assert_eq!(7, context.graph().constant_references.len());
+        assert_eq!(9, context.graph().constant_references.len());
```
st0012, line 1836: Should we create singleton class for `attr_reader` calls too? If not, should we stop creating one for `alias_method` calls? Can be a follow up PR

st0012 (approved):

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4119,6 +4128,126 @@ mod tests {
...
+            .method_references()
+            .values()
+            .find(|method_ref| *method_ref.str() == StringId::from("bar"))
+            .expect("should have a method reference for bar");
+        let constant_ref = context
+            .graph()
+            .method_references()
+            .values()
+            .find(|method_ref| *method_ref.str() == StringId::from("baz"))
+            .expect("should have a method reference for baz");
```
st0012: Let's also create a macro/helper to simplify these.

alexcrocha: Good call

rust/rubydex/src/model/graph.rs
alexcrocha: That's a fair question. Let's do a follow up PR for this. I don't know the answer yet but I'll leave it for tomorrow

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -4119,6 +4156,69 @@ mod tests {
         assert!(parent_scope.parent_scope().is_none());
     }
 
+    #[test]
```
vinistock, line 4159: Do we have a test for when `self.bar` is inside a method definition?

alexcrocha: ~~It's missing. I'll add one as a follow-up~~ We have a [test](https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/indexing/ruby_indexer.rs#L4249) for this case already

vinistock: Nice!

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -2159,6 +2168,34 @@ mod tests {
...
+    macro_rules! assert_method_ref_receiver {
+        ($context:expr, $method_name:expr, $expected_receiver:expr) => {{
+            let method_ref = $context
+                .graph()
+                .method_references()
+                .values()
+                .find(|method_ref| *method_ref.str() == StringId::from($method_name))
+                .unwrap_or_else(|| panic!("should have a method reference for {}", $method_name));
+
+            let receiver = $context.graph().names().get(&method_ref.receiver().unwrap()).unwrap();
```
Morriar, line 2186: The `.unwrap()` on `receiver()` will panic with a generic message if the method ref has no receiver. I wonder if we should add a descriptive panic here?

```suggestion
            let receiver_id = method_ref.receiver().unwrap_or_else(|| panic!("method reference for `{}` has no receiver", $method_name));
            let receiver = $context.graph().names().get(&receiver_id).unwrap();
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -2159,6 +2168,34 @@ mod tests {
...
+    ///
+    /// Usage:
+    /// - `assert_method_ref_receiver!(context, "bar", "<Foo>")`
+    macro_rules! assert_method_ref_receiver {
+        ($context:expr, $method_name:expr, $expected_receiver:expr) => {{
+            let method_ref = $context
+                .graph()
+                .method_references()
+                .values()
+                .find(|method_ref| *method_ref.str() == StringId::from($method_name))
```
Morriar, line 2183: The macro uses `.find()` which returns the first match — if there are ever two method references with the same name in a test (e.g., `foo.bar` and `bar` in the same source), this could silently test the wrong one. Might be worth a comment noting this assumption, or asserting uniqueness.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Yes, we totally should. Addressed in #696

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Addressed in #696 as well. The helper now asserts uniqueness instead of taking the first match. Also updated the doc comment

Issue #649, opened 2026-03-06
Use test modules to group tests & their helpers


st0012:
[Example](https://github.com/Shopify/rubydex/pull/646#discussion_r2892678015)

We should be able to split our big indexing & resolution test suites and their helpers into different modules. Each module will have different group them in document symbol so it'll also help editor navigation.

alexcrocha:
Indexing & resolution tests and helpers have been extracted into their own test files and grouped into submodules.

PR #664, opened 2026-03-14
Group class and module indexing tests into dedicated test module

alexcrocha:
This PR is the second in a stack that addresses #649.

By grouping class and module indexing tests into a dedicated test module, we improve discoverability for both editor navigation and code search by giving class/module definition tests a named boundary.

Follows the existing `name_dependent_tests` pattern.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer_tests.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer_tests.rs
@@ -206,222 +206,6 @@ fn index_source_with_warnings() {
 }
 
 #[test]
-fn index_class_node() {
-    let context = index_source({
-        "
-        class Foo
-          class Bar
-            class Baz; end
-          end
-        end
-        "
-    });
-
-    assert_no_local_diagnostics!(&context);
-    assert_eq!(context.graph().definitions().len(), 3);
-
-    assert_definition_at!(&context, "1:1-5:4", Class, |def| {
-        assert_def_name_eq!(&context, def, "Foo");
-        assert_def_name_offset_eq!(&context, def, "1:7-1:10");
-        assert!(def.superclass_ref().is_none());
-        assert_eq!(1, def.members().len());
-        assert!(def.lexical_nesting_id().is_none());
-    });
-
-    assert_definition_at!(&context, "2:3-4:6", Class, |def| {
-        assert_def_name_eq!(&context, def, "Bar");
-        assert_def_name_offset_eq!(&context, def, "2:9-2:12");
-        assert!(def.superclass_ref().is_none());
-        assert_eq!(1, def.members().len());
-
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6018,6 +5802,234 @@ mod tests {
...
+#[cfg(test)]
+mod class_and_module_tests {
+    use crate::{
+        assert_def_name_eq, assert_def_name_offset_eq, assert_definition_at, assert_local_diagnostics_eq,
+        assert_no_local_diagnostics, test_utils::LocalGraphTest,
+    };
+
+    fn index_source(source: &str) -> LocalGraphTest {
+        LocalGraphTest::new("file:///foo.rb", source)
+    }
```
vinistock: Does it make sense to use nested modules so that all imports and shared helpers are at the top? It also avoids repeating the `#cfg[(test)]`.

For example:

```rust
#[cfg(test)]
mod tests {
  use crate::...

  fn index_source(...) {}

  mod classes_and_modules {}

  mod methods {}

  mod instance_variables {}
}
```

alexcrocha: That makes sense. The argument for sibling modules is that they would be fully self contained and ready to be extracted out of this file when we make that call. But to be honest the effort to doing so is the same now or later.

We can go with sub-modules for now.

st0012: If we were doing this, will `tests` have any test cases in it? Or it'll simply be a container of sub modules and helpers. I'd want to avoid having tests at both levels (imagine having test cases between `classes_and_modules` and `method` mods).

@alexcrocha Do you know which is the community preference? Can you find some examples for either cases?

alexcrocha: > Or it'll simply be a container of sub modules and helpers

That's the idea, and I think we can do that with our current tests. I wouldn't want to force tests into unrelated sub modules though, so I am not sure that will always be true in the future. 

> Do you know which is the community preference? Can you find some examples for either cases?

I can do an audit to get a better sense of the most idiomatic pattern in Rust

alexcrocha:
> Considering the size of this file, should we start extracting these tests in external test files?

Documenting the outcome of this question here. After a brief chat async, the team decided it was indeed time to split the tests into their own file. 

https://github.com/Shopify/rubydex/pull/705 is the new base of this stack

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: We decided to extract the tests into their own file. See https://github.com/Shopify/rubydex/pull/664#issuecomment-4139128764

PR #665, opened 2026-03-16
Support overloaded signature in method definition

soutaro:
To support RBS method definitions, this PR introduces explicitly overloaded method signatures.

At this stage, this PR only uses the `Simple` signatures. Follow-up work in #633 will use the `Overloaded` signatures for RBS method definitions that actually include overloads.

Context: This change is based on [the discussion](https://github.com/Shopify/rubydex/pull/633#discussion_r2896404043).

## Memory impact

I did a quick benchmark to estimate the memory impact on Shopify’s core codebase.

The results were inconsistent, ranging from approximately **+120MB** to **-65MB**, so I don’t think we can draw a reliable conclusion from that benchmark yet. Based on rough expectations, this change may increase memory usage by around **3MB**, but that number should be treated as directional rather than measured.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1489,7 +1489,7 @@ impl Visit<'_> for RubyIndexer<'_> {
                 comments.clone(),
                 flags.clone(),
                 parent_nesting_id,
-                Signatures::Simple(parameters.clone()),
+                Signatures::Simple(parameters.clone().into_boxed_slice()),
                 Visibility::Public,
                 self.current_nesting_definition_id().map(Receiver::SelfReceiver),
             )));
@@ -1505,7 +1505,7 @@ impl Visit<'_> for RubyIndexer<'_> {
                 comments,
                 flags,
                 parent_nesting_id,
-                Signatures::Simple(parameters),
+                Signatures::Simple(parameters.into_boxed_slice()),
                 Visibility::Private,
                 receiver,
             )));
@@ -1522,7 +1522,7 @@ impl Visit<'_> for RubyIndexer<'_> {
                 comments,
                 flags,
                 parent_nesting_id,
-                Signatures::Simple(parameters),
+                Signatures::Simple(parameters.into_boxed_slice()),
                 visibility,
                 receiver,
             )));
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/query.rs
```diff
@@ -419,12 +419,14 @@ fn method_argument_completion<'a>(
     // Find the first Method definition to extract keyword parameters
     for def_id in method_decl.definitions() {
         if let Some(Definition::Method(method_def)) = graph.definitions().get(def_id) {
-            for param in method_def.parameters() {
-                match param {
-                    Parameter::RequiredKeyword(p) | Parameter::OptionalKeyword(p) => {
-                        candidates.push(CompletionCandidate::KeywordArgument(*p.str()));
+            for signature in method_def.signatures().as_slice() {
```
vinistock, line 422: Not a problem to address in this PR, but something to consider in the future. For overloads, it would be nice to automatically filter out completion candidates that don't apply. For example:

```rbs
def foo: (up: Integer, down: Integer) -> void |
         (step: Integer) -> void
```

We wouldn't want to suggest the keyword `step` if the user already typed in `up:` or `down:` because `step` is not a part of that overload.

Maybe this filtering is a responsibility of the Rubydex client? I don't know, just thinking out loud.

soutaro: Good point! I think leaving this to the client would make more sense. The client has the editing context — like cursor position and which arguments have already been typed — so it's naturally positioned to determine which overload is active and filter candidates accordingly. This would also apply to positional arguments, not just keywords.

Morriar, line 422: I wonder if iterating all signatures here could produce duplicate keyword candidates when overloaded signatures share the same keyword parameter?

For example, `def foo: (String, name: String) -> void | (Integer, name: Integer) -> void` would emit `name:` twice. Since `KeywordArgument` isn't deduplicated through the `CompletionContext` (only `Declaration` members go through `dedup`), the caller would see duplicates.

Might be worth adding a test for an overloaded method with shared keyword params to confirm the expected behavior.

vinistock: I think after #668, we can return the signature objects instead.

soutaro: Sounds good. Will dedup the keywords after #668.

PR #666, opened 2026-03-16
Overlap graph merging with indexing to reduce memory usage

st0012:
## Summary

- Add `JobQueue::run_without_waiting()` that returns thread handles instead of blocking
- `index_files` now merges `LocalGraph`s on the main thread as they arrive while worker threads are still indexing
- This avoids buffering all `LocalGraph`s in the channel, reducing peak memory

## Benchmark (96k files)

|              | main    | branch  | delta |
|--------------|---------|---------|-------|
| Indexing     | 10.12s  | 7.11s   | -30%  |
| Resolution   | 51.43s  | 51.12s  | ~0%   |
| Total        | 63.07s  | 59.76s  | -5%   |
| RSS          | 5531 MB | 3447 MB | -38%  |

### DIFF
--- a/rust/rubydex/src/job_queue.rs
+++ b/rust/rubydex/src/job_queue.rs
@@ -53,10 +53,32 @@ impl JobQueue {
         Self::run_with_workers(queue, worker_count);
     }
 
+    /// Spin up worker threads and return their handles without waiting for completion.
+    ///
+    /// The caller is responsible for joining the returned handles. This allows
+    /// overlapping other work (e.g. merging results) with job processing.
+    pub fn run_without_waiting(queue: &Arc<JobQueue>) -> Vec<thread::JoinHandle<()>> {
+        let worker_count = thread::available_parallelism()
+            .map(std::num::NonZeroUsize::get)
+            .unwrap_or(4);
+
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/job_queue.rs
```diff
@@ -53,10 +53,32 @@ impl JobQueue {
         Self::run_with_workers(queue, worker_count);
     }
 
+    /// Spin up worker threads and return their handles without waiting for completion.
+    ///
+    /// The caller is responsible for joining the returned handles. This allows
+    /// overlapping other work (e.g. merging results) with job processing.
+    pub fn run_without_waiting(queue: &Arc<JobQueue>) -> Vec<thread::JoinHandle<()>> {
+        let worker_count = thread::available_parallelism()
```
Morriar, line 61: The `worker_count` detection is duplicated from `run` (lines 49–51). Any way we could extract this and add a small private helper:

```rust
    fn default_worker_count() -> usize {
        thread::available_parallelism()
            .map(std::num::NonZeroUsize::get)
            .unwrap_or(4)
    }
```

Then `run` can use it too.

PR #667, opened 2026-03-16
Turn comments into boxed slices

vinistock:
The comments vector doesn't change after being created. We always create and discard when necessary, but we never push more items or shrink it.

We can save a bit of memory by using boxed slices instead.


vinistock:
> Can we have some before/after stats?

There's no noticeable different in the benchmarks, but we're reducing the size of the structs so it should impact the live memory usage. Similar to my comment in #659.

Issue #668, opened 2026-03-16
Surface method signatures to the Ruby API


vinistock:
We need to be able to read method signatures to provide signature help in the Ruby LSP.

There's one interesting caveat to investigate. We don't eagerly resolve method aliases and their signatures are attached to the method they target. For example:

```ruby
class Parent
  def foo(a, b); end
end

class Child < Parent
  alias_method :new_foo, :foo
end
```

The signature for `new_foo` should be `[PositionalRequired(a), PositionalRequired(b)]` coming from the original `foo`.

Investigate if we can resolve method aliases lazily to return signatures (and consider splitting the work into separate PRs if necessary).

soutaro:
* [x] Want to do something for RBS untyped method `(?) -> T` (but how?)
* [x] Method aliases to inherited methods should work -- should return signatures from the inherited methods.

soutaro:
The untyped method type `(?) -> T` simply returns an empty parameters. This is inconsistent and should be fixed later.

PR #686, opened 2026-03-24
Expose method signature to Ruby

soutaro:
Towards https://github.com/Shopify/rubydex/issues/668

* RBS untyped method `(?) -> ` returns empty parameters, which should be fixed later.
* Method aliases to inherited methods doesn't work. It stops and returns empty signatures.

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -4,6 +4,7 @@
 #include "handle.h"
 #include "reference.h"
 #include "rustbindings.h"
+#include "signature.h"
 #include "utils.h"
 
 VALUE cDeclaration;
@@ -331,6 +332,18 @@ static VALUE rdxr_declaration_references(VALUE self) {
     return self;
 }
 
+// Method#signatures -> [Rubydex::Signature]
+static VALUE rdxr_method_declaration_signatures(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    SignatureArray *arr = rdx_declaration_method_signatures(graph, data->id);
+    return rdxi_signatures_to_ruby(arr, data->graph_obj, Qnil);
+}
+
 void rdxi_initialize_declaration(VALUE mRubydex) {
     cDeclaration = rb_define_class_under(mRubydex, "Declaration", rb_cObject);
     cNamespace = rb_define_class_under(mRubydex, "Namespace", cDeclaration);
@@ -341,6 +354,7 @@ void rdxi_initialize_declaration(VALUE mRubydex) {
     cConstant = rb_define_class_under(mRubydex, "Constant", cDeclaration);
     cConstantAlias = rb_define_class_under(mRubydex, "ConstantAlias", cDeclaration);
     cMethod = rb_define_class_under(mRubydex, "Method", cDeclaration);
+    rb_define_method(cMethod, "signatures", rdxr_method_declaration_signatures, 0);
     cGlobalVariable = rb_define_class_under(mRubydex, "GlobalVariable", cDeclaration);
     cInstanceVariable = rb_define_class_under(mRubydex, "InstanceVariable", cDeclaration);
     cClassVariable = rb_define_class_under(mRubydex, "ClassVariable", cDeclaration);
--- a/lib/rubydex/signature.rb
+++ b/lib/rubydex/signature.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Rubydex
+  # Represents a single signature of a method definition.
+  #
+  # A method definition may have multiple signatures when defined with overloads in RBS.
+  class Signature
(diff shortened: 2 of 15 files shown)


rust/rubydex-sys/src/declaration_api.rs
```diff
@@ -439,3 +441,42 @@ pub unsafe extern "C" fn rdx_declaration_references_iter_new(
...
+        let decl_id = DeclarationId::new(declaration_id);
+        let method_defs = rubydex::query::method_definitions(graph, decl_id);
+
+        if method_defs.is_empty() {
+            return ptr::null_mut();
+        }
+
+        let mut sig_entries: Vec<SignatureEntry> = Vec::new();
+        for (def_id, method_def) in &method_defs {
+            collect_method_signatures(graph, method_def, **def_id, &mut sig_entries);
```
Morriar:
```suggestion
            collect_method_signatures(graph, method_def, def_id.get(), &mut sig_entries);
```

rust/rubydex/src/query.rs
```diff
@@ -207,6 +207,78 @@ macro_rules! collect_candidates {
...
+/// For alias declarations, follows the alias to the original method and returns its definitions.
+///
+/// # Panics
+///
+/// Panics if:
+/// - the `declaration_id` does not exist in the graph
+/// - the declaration is not a method declaration
+/// - any definition or owner declaration referenced by the method is missing from the graph
+#[must_use]
+pub fn method_definitions(graph: &Graph, declaration_id: DeclarationId) -> Vec<(DefinitionId, &MethodDefinition)> {
```
Morriar: The name may be confusing, what do you think of something like `dealias_method_declaration_to_definitions`

Morriar: Do we need to return both the definition id and the reference to the definition?

soutaro: Agree that `method_definitions` is too ambiguous. I feel `dealias_` is a bit too implementation-specific though — the intent is "get the underlying `MethodDefinition`s from a declaration", not specifically about alias resolution.

How about `resolve_method_declaration_to_definitions`, or without the `resolve_` prefix?

soutaro: You're right — `MethodDefinition` has an `id()` method that computes the `DefinitionId`, so returning it separately is redundant. I'll simplify the return type to `Vec<&MethodDefinition>`.

Morriar: ~I like `resolve_method_declaration_to_definitions`~

Actually, we generally use `resolve` to go from a name to a declaration, it may be confusing to "resolve a declaration".

Let's go with `method_declaration_to_definitions`

ext/rubydex/declaration.c
```diff
@@ -331,6 +332,18 @@ static VALUE rdxr_declaration_references(VALUE self) {
     return self;
 }
 
+// Method#signatures -> [Rubydex::Signature]
+static VALUE rdxr_method_declaration_signatures(VALUE self) {
```
vinistock, line 336: Do we need to expose `Declaration#signature`? Signatures are associated to the specific `Definition` that creates them, so I'd expect us to access it like this:

```ruby
method_decl = graph["Foo#bar()"]
signatures = method_decl.definitions.map(&:signature)
```

I think we can remove this one.

lib/rubydex/signature.rb
```diff
@@ -0,0 +1,37 @@
...
+    # - `:req` — required positional parameter (`a`) or post-rest positional parameter (`d` in `def foo(*c, d)`)
+    # - `:opt` — optional positional parameter (`b = 1`)
+    # - `:rest` — rest positional parameter (`*c`)
+    # - `:keyreq` — required keyword parameter (`e:`)
+    # - `:key` — optional keyword parameter (`f: 1`)
+    # - `:keyrest` — rest keyword parameter (`**g`)
+    # - `:block` — block parameter (`&h`)
+    # - `:forward` — forward parameter (`...`), Rubydex-specific (Ruby expands this to `:rest`, `:keyrest`, `:block`)
+    #
+    #: Array[[Symbol, Symbol, Location]]
```
vinistock, line 23: Let's please create Ruby objects to represent each parameter type instead of using symbols.

I suspect you only need a parent class and then a bunch of empty definitions:

```ruby
class Parameter
  def initialize(name, location)
    # ...
  end
end

class PositionalParameter < Parameter; end
class OptionalPositionalParameter < Parameter; end
class KeywordParameter < Parameter; end
class OptionalKeywordParameter < Parameter; end
# ...
```

vinistock (commented):
This PR will likely be quicker to ship if we split it into smaller steps.

For example, the creation of the `Signature` class and all parameters can probably be shipped on its own

ext/rubydex/declaration.c
soutaro: I'd like to keep `#signatures` as a utility API. `#definitions` can contain method aliases, so `definitions.flat_map(&:signatures)` wouldn't work transparently — users would need to handle alias resolution themselves.

We could add a separate API for alias normalization, but if we're adding an API anyway, I think `#signatures` that does it all in one shot is the better choice.

PR #695, opened 2026-03-26
Index retroactive method visibility changes

alexcrocha:
One more step towards https://github.com/Shopify/rubydex/issues/89

This PR extends the definition-based visibility path from constants to methods by indexing retroactive `private/protected/public` calls as `MethodVisibilityDefinition`s.

### Why

Ruby supports retroactive visibility calls like `private :foo`, but the current codebase only models visibility from flag mode (`private` followed by later definitions) and scoped definitions (`private def foo` or `protected attr_reader :bar`). We already use a definition-based approach for constant visibility, so this applies the same model to methods and preserves those visibility changes in the graph.

This PR is intentionally limited to indexing. A follow-up PR will make `MethodVisibilityDefinition`s affect the resolved declaration graph.

### What

- add `Definition::MethodVisibility` and `MethodVisibilityDefinition`
- expose the new definition kind through the Rust/C/Ruby definition APIs
- index local retroactive `private/protected/public` calls with literal symbol and string targets
- store method targets using existing member names like `foo()`
- treat method visibility definitions as method-like in graph utilities
- diagnose unsupported dynamic retroactive visibility calls
- preserve existing behaviour for:
  - flag mode (`private`)
  - scoped definitions (`private def foo`)
  - scoped attr helpers (`protected attr_reader :bar`)


rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1252,6 +1252,71 @@ impl<'a> RubyIndexer<'a> {
...
+        };
+
+        let only_literals = args.iter().all(|arg| is_literal(&arg));
+        if !only_literals {
+            let has_any_literal = args.iter().any(|arg| is_literal(&arg));
+            let message = if has_any_literal {
+                "Method visibility called with mixed literal and non-literal arguments"
+            } else {
+                "Method visibility called with non-literal arguments"
+            };
```
vinistock: Could we use `each_string_or_symbol_arg` here?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1864,27 +1929,46 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                                || call.receiver().as_ref().is_some_and(|r| r.as_self_node().is_some());
+                            has_no_explicit_receiver
+                                && matches!(
+                                    call.name().as_slice(),
+                                    b"attr" | b"attr_reader" | b"attr_writer" | b"attr_accessor"
+                                )
+                        });
+
+                    let is_scoped_definition =
+                        message.as_str() == "module_function" || single_arg_is_def || single_arg_is_attr_helper;
```
vinistock: It seems that `module_function` works like the other cases. It accepts retroactive definitions and multiple names. Could we not handle it the same? Does it require any different logic?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1864,27 +1929,46 @@ impl Visit<'_> for RubyIndexer<'_> {
...
+                    let single_arg_is_attr_helper = has_single_arg
+                        && args.iter().next().unwrap().as_call_node().is_some_and(|call| {
+                            let has_no_explicit_receiver = call.receiver().is_none()
+                                || call.receiver().as_ref().is_some_and(|r| r.as_self_node().is_some());
+                            has_no_explicit_receiver
+                                && matches!(
+                                    call.name().as_slice(),
+                                    b"attr" | b"attr_reader" | b"attr_writer" | b"attr_accessor"
+                                )
+                        });
```
vinistock: There are a lot of checks in this section. Would it make it a bit easier to read if we extracted a helper method?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6658,6 +6742,260 @@ mod tests {
...
+        for def in context.graph().definitions().values() {
+            assert!(
+                !matches!(def, Definition::MethodVisibility(_)),
+                "should not create MethodVisibility for dynamic-only args"
+            );
+        }
+
+        // Existing method defs still indexed
+        assert_definition_at!(&context, "2:3-2:15", Method, |def| {
+            assert_def_str_eq!(&context, def, "foo()");
```
st0012: nit: this feels a bit redundant

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6658,6 +6742,260 @@ mod tests {
...
+        // No MethodVisibilityDefinition created
+        for def in context.graph().definitions().values() {
+            assert!(
+                !matches!(def, Definition::MethodVisibility(_)),
+                "should not create MethodVisibility for mixed args"
+            );
+        }
+
+        // Nested references still indexed
+        assert_constant_references_eq!(&context, ["SOME_CONST"]);
```
st0012: nit: this feels a bit redundant

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6658,6 +6742,260 @@ mod tests {
...
+            class Foo
+              VISIBILITY = :private
+
+              private VISIBILITY
+            end
+            ",
+        );
+
+        // Dynamic arg is visited — constant reference for VISIBILITY is indexed
+        assert_constant_references_eq!(&context, ["VISIBILITY"]);
```
st0012: I feel this can be part of `index_retroactive_method_visibility_mixed_args_diagnostic` if we just also define the constant in that case and assert its reference?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6658,6 +6742,260 @@ mod tests {
 
         assert_eq!(context.graph().definitions().len(), 3); // Foo, Foo::Qux, Foo#foo
     }
+
+    #[test]
+    fn index_retroactive_method_visibility() {
```
st0012: Can we create a dedicated test mod for these tests?

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: You'd be scared to see how it looked like before 😱 But I agree, there's still room for improvement

I added a new commit, let me know if you find it better

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Agree. Removed

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Since I folded the old `index_retroactive_method_visibility_dynamic_visits_args` test into this one as suggested below, I'll keep this and update the comment to make the intent more explicit that we are still indexing the constant reference despite the diagnostic

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Well spotted. I'll fold this test into `index_retroactive_method_visibility_mixed_args_diagnostic`

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: https://github.com/Shopify/rubydex/pull/672 is introducing a visibility tests mod. Depending on which one merges first I'll update the other

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: I added a new commit to handle retroactive `module_function` as well. It goes through the same logic with a extra guard to emit a diagnostic when used outside a module

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Refactored to use `each_string_or_symbol_arg` 👍

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1252,6 +1261,78 @@ impl<'a> RubyIndexer<'a> {
             self.add_member_to_current_owner(definition_id);
         }
     }
+
+    fn should_apply_inline_visibility(args: &ruby_prism::NodeList) -> bool {
+        if args.len() != 1 {
+            return false;
+        }
+
+        let arg = args.iter().next().unwrap();
```
vinistock: Instead of checking the length and then unwrapping the first, we can do a single operation:


```suggestion
        let Some(arg) = args.iter().next() else {
          return false;
        }
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1252,6 +1261,78 @@ impl<'a> RubyIndexer<'a> {
...
+            )
+        };
+
+        if !args.iter().all(&is_literal) {
+            let has_any_literal = args.iter().any(is_literal);
+            let message = if has_any_literal {
+                format!("`{call_name}` called with mixed literal and non-literal arguments")
+            } else {
+                format!("`{call_name}` called with non-literal arguments")
+            };
```
vinistock: Instead of adding a single diagnostic for the entire visibility call, we should add one diagnostic per incorrect argument.

That makes a bit clearer to the user where the problem is.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -6671,6 +6763,280 @@ mod tests {
 
         assert_eq!(context.graph().definitions().len(), 3); // Foo, Foo::Qux, Foo#foo
     }
+
```
vinistock: Let's add some examples for changing the visibility of inherited methods. We're not going to support it immediately, but we should have it documented and maybe include a `TODO` comment.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: I ended up removing that inline/retroactive routing entirely. Visibility args are now handled independently, so this no longer needs the `len()` guard,. Mixed cases like `private def foo; end, :bar` now work correctly, and mixed `attr_*` calls like `private attr_reader(:foo), :bar` now get diagnosed instead of being treated as inline private

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Fixed. With the change to visibility args being handled independently, each incorrect argument now gets its own diagnostic

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Good point. That reminded me that we already covered inherited method visibility in https://github.com/Shopify/rubydex/pull/691

I am leaning towards rebasing that draft and opening it for review so we can get the broader Ruby visibility behaviour coverage into the codebase, instead of cherry-picking only the cases relevant to this PR.

Does that sound like a good plan to you?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -378,27 +378,33 @@ impl<'a> RubyIndexer<'a> {
...
+                Some((name, Offset::from_prism_location(&arg.location())))
+            }
+            _ => None,
+        }
+    }
+
     fn each_string_or_symbol_arg<F>(node: &ruby_prism::CallNode, mut f: F)
     where
-        F: FnMut(String, ruby_prism::Location),
+        F: FnMut(String, Offset),
```
vinistock: Why are we changing this for an `Offset`? This means that we always eagerly instantiate the `Offset` for each argument, but we may not necessarily always need it.

I'd probably undo this whole refactor and continue passing a location, which the caller can then turn into an offset if desired.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1204,33 +1219,17 @@ impl<'a> RubyIndexer<'a> {
...
-                _ => {
-                    self.local_graph.add_diagnostic(
-                        Rule::InvalidPrivateConstant,
-                        Offset::from_prism_location(&argument.location()),
-                        "Private constant called with non-symbol argument".to_string(),
-                    );
+            let Some((name, offset)) = Self::extract_string_or_symbol(&argument) else {
+                self.local_graph.add_diagnostic(
+                    Rule::InvalidPrivateConstant,
+                    Offset::from_prism_location(&argument.location()),
```
vinistock: We're not using the offset that we instantiated and instead are creating a new one. If I understand correctly, we made this refactor to be able to associate the diagnostics with the specific arguments that aren't strings or symbols.

Could we instead make `each_string_or_symbol_arg` use a closure of `Option(String), Location`?

That way, if the name is `None`, we know it's a dynamic argument and we can use that as the signal to attach a diagnostic.

It will also prevent the fact that we're reimplementing `each_string_or_symbol_arg` inline here.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1250,6 +1249,91 @@ impl<'a> RubyIndexer<'a> {
...
+            if matches!(arg, ruby_prism::Node::DefNode { .. }) {
+                if apply_visibility {
+                    self.visibility_stack
+                        .push(VisibilityModifier::new(visibility, true, call_offset.clone()));
+                }
+                self.visit(&arg);
+                if apply_visibility {
+                    self.visibility_stack.pop();
+                }
+            } else if arg_count == 1 && Self::is_attr_call(&arg) {
```
vinistock: I think this branch could be merged with the previous one, but I'm not 100% sure I understand what `apply_visibility` is accounting for. Which scenario does it handle?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1250,6 +1249,91 @@ impl<'a> RubyIndexer<'a> {
...
+                // Sole attr_* arg: `private attr_reader :foo`
+                self.visibility_stack
+                    .push(VisibilityModifier::new(visibility, true, call_offset.clone()));
+                self.visit(&arg);
+                self.visibility_stack.pop();
+            } else if matches!(
+                arg,
+                ruby_prism::Node::SymbolNode { .. } | ruby_prism::Node::StringNode { .. }
```
vinistock, line 1298: Here we check if the argument is a string or a symbol and then invoke `create_method_visibility_definition`, which in turn invokes `extract_string_or_symbol`, checking if the argument is a symbol or string a second time.

Maybe we can call `extract_string_or_symbol` and use the return of that as the check?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1250,6 +1249,91 @@ impl<'a> RubyIndexer<'a> {
...
+        for arg in &args {
+            if matches!(arg, ruby_prism::Node::DefNode { .. }) {
+                if apply_visibility {
+                    self.visibility_stack
+                        .push(VisibilityModifier::new(visibility, true, call_offset.clone()));
+                }
+                self.visit(&arg);
+                if apply_visibility {
+                    self.visibility_stack.pop();
+                }
```
Morriar: I'm wondering about one edge case: if an unsupported arg appears *before* an inline `def`, the def is still visited but without the visibility modifier (since `apply_visibility` is false). For example:

```ruby
private SOME_CONST, def foo; end
```

In Ruby, `foo` would still be made private regardless of what `SOME_CONST` evaluates to. But the indexer would index `foo` as public here because `apply_visibility` was set to false by the earlier unsupported arg.

Should `DefNode` always get the scoped visibility pushed regardless of `apply_visibility`? The `DefNode` case is unambiguous (we know it's defining a method), unlike symbol/string args that could reference non-existent methods.

Not blocking since this is a pretty unusual pattern, just something to think about.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: This separate branch was only for the solo `attr_*` case, like `private attr_reader(:foo)` where the visibility changes always apply. You are right that we can combine it with the previous branch, I have merged them.

I also renamed `apply_visibility` to `visibility_still_applies` as I agree the intent was not very clear. It handles mixed visibility calls where an earlier arg is unsupported. In those cases we still want to visit the later args and index them, but visibility no longer applies to those args. For example, in `private attr_reader(:foo), def bar; end`, `bar` is still indexed but it is not made private.

Let me know if that still feels unclear

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Good point, this is a real edge case, but it does depend on what `SOME_CONST` evaluates to at runtime. Consider these 3 scenarios:

```
SOME_CONST = :foo
no error
{private: [:foo]}
{public: []}

SOME_CONST = 1
TypeError: 1 is not a symbol nor a string
{private: []}
{public: [:foo]}

SOME_CONST undefined
NameError: uninitialized constant SOME_CONST
{private: []}
{public: []}
```

Since we can't know that statically here, I didn't try to model it precisely in this PR (although, attempts were made, but things got messy fast). I suggest we revisit this at some point if we want to support it more accurately

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: You're right, the eager Offset change isn't worth it here. I decided to drop the refactor of the string / symbol name helper from this PR. It was only addressing code duplication, so I think it's better to keep this PR scoped and we can tackle the refactor as a separate follow up

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: No longer applicable after dropping the string/symbol helper refactor from this PR. See https://github.com/Shopify/rubydex/pull/695#discussion_r3025022197

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: I think this cleanup is still worth doing, but since I dropped the string/symbol helper refactor from this PR to keep the PR scoped, I'd rather tackle it in the separate follow up I mentioned in https://github.com/Shopify/rubydex/pull/695#discussion_r3025022197

rust/rubydex/src/indexing/ruby_indexer.rs
Morriar: Can't we just delete the `visibility_still_applies` checks for now?

btw do we have a test case for this?

```ruby
private SOME_CONST, def foo; end
```

rust/rubydex/src/indexing/ruby_indexer_tests.rs
```diff
@@ -3298,6 +3298,392 @@ mod constant_tests {
...
+}
+
+#[test]
+fn index_retroactive_method_visibility_receiver_ignored() {
+    let context = index_source(
+        "
+        class Foo
+          def foo; end
+
+          Foo.private :foo
```
st0012: `private` is a private method so we should have a diagnostics here:

```rb
class Foo
  def foo; end

  Foo.private :foo
end

Foo.new.foo

main.rb:4:in '<class:Foo>': private method 'private' called for class Foo (NoMethodError)

  Foo.private :foo
     ^^^^^^^^
	from main.rb:1:in '<main>'
```

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1250,6 +1259,99 @@ impl<'a> RubyIndexer<'a> {
...
+            } else {
+                // Unsupported arg — diagnostic + visit for side effects, stop applying visibility.
+                let arg_offset = Offset::from_prism_location(&arg.location());
+                self.local_graph.add_diagnostic(
+                    Rule::InvalidMethodVisibility,
+                    arg_offset,
+                    format!("`{call_name}` called with a non-literal argument"),
+                );
+                self.visit(&arg);
+                visibility_still_applies = false;
```
Morriar: I see that after hitting an unsupported arg, `visibility_still_applies` becomes false and subsequent literal symbols (`:bar` in `private SOME_CONST, :bar`) are silently skipped with no `MethodVisibilityDefinition` and no diagnostic.

I think the current behavior is reasonable (we can't know what `SOME_CONST` evaluates to at runtime, so conservatively stopping makes sense), but should we add a test for this "unsupported-then-literal" ordering to document the behavior? Right now we only test literal-then-unsupported (`private :foo, SOME_CONST`).

rust/rubydex/src/indexing/ruby_indexer_tests.rs
alexcrocha: Good catch! I amended this in.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Removed the `visibility_still_applies` checks for now. We are just emitting diagnostics for the unsupported cases.

Also added that case to our `index_inline_visibility_mixed_with_unsupported` test.

vinistock (approved):

PR #697, opened 2026-03-26
Add MethodDefinition#signatures to Ruby API

soutaro:
## Summary
- Expose method signature information through `MethodDefinition#signatures`, returning an array of `Rubydex::Signature` objects
- Each `Signature` holds an array of typed `Parameter` subclasses (`PositionalParameter`, `KeywordParameter`, etc.) with `name` and `location` attributes
- Add Rust C API (`ParameterKind`, `SignatureEntry`, `SignatureArray`) and C extension glue for conversion

Extracted from #686.

## Test plan
- Tests cover all 8 parameter kinds (positional, optional, rest, keyword, optional keyword, rest keyword, block, forward)
- Tests cover RBS-defined signatures, overloads, untyped parameters, and no-parameter methods

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ext/rubydex/signature.c
```diff
@@ -0,0 +1,89 @@
...
+
+    cParameter = rb_define_class_under(cSignature, "Parameter", rb_cObject);
+    cPositionalParameter = rb_define_class_under(cSignature, "PositionalParameter", cParameter);
+    cOptionalPositionalParameter = rb_define_class_under(cSignature, "OptionalPositionalParameter", cParameter);
+    cRestPositionalParameter = rb_define_class_under(cSignature, "RestPositionalParameter", cParameter);
+    cKeywordParameter = rb_define_class_under(cSignature, "KeywordParameter", cParameter);
+    cOptionalKeywordParameter = rb_define_class_under(cSignature, "OptionalKeywordParameter", cParameter);
+    cRestKeywordParameter = rb_define_class_under(cSignature, "RestKeywordParameter", cParameter);
+    cBlockParameter = rb_define_class_under(cSignature, "BlockParameter", cParameter);
+    cForwardParameter = rb_define_class_under(cSignature, "ForwardParameter", cParameter);
```
vinistock, line 70: Could we define these in Ruby? It seems like we may not even need new C files. We could just put the `rdxi_signatures_to_ruby` helper in `definition.c`.

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,170 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
...
+/// C-compatible enum representing the kind of a parameter.
+#[repr(C)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ParameterKind {
+    Req = 0,
+    Opt = 1,
+    Rest = 2,
+    Keyreq = 3,
+    Key = 4,
+    Keyrest = 5,
```
vinistock: I vote for using full names here


```suggestion
    RequiredPositional = 0,
    OptionalPositional = 1,
    Rest = 2,
    RequiredKeyword = 3,
    OptionalKeyword = 4,
    Keyrest = 5,
```

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,170 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
...
+pub unsafe extern "C" fn rdx_definition_signatures(pointer: GraphPointer, definition_id: u64) -> *mut SignatureArray {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        let Some(defn) = graph.definitions().get(&def_id) else {
+            panic!("Definition not found: {definition_id:?}");
+        };
+
+        let Definition::Method(method_def) = defn else {
+            return ptr::null_mut();
+        };
```
vinistock: Let's merge these two

```suggestion
        let Some(Definition::Method(defn)) = graph.definitions().get(&def_id) else {
            panic!("Expected a method definition to be found");
        };
```

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,170 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
...
+        let len = boxed.len();
+        let items_ptr = boxed.as_mut_ptr();
+        std::mem::forget(boxed);
+
+        Box::into_raw(Box::new(SignatureArray { items: items_ptr, len }))
+    })
+}
+
+/// Helper: build signature entries from a MethodDefinition and append them to the output vector.
+pub(crate) fn collect_method_signatures(
```
vinistock: Does this need to be public? I think it's only used in this file.

lib/rubydex/signature.rb
```diff
@@ -0,0 +1,31 @@
...
+    #: Array[Parameter]
+    attr_reader :parameters
+
+    #: MethodDefinition
+    attr_reader :method_definition
+
+    #: (parameters: Array[Parameter], method_definition: MethodDefinition) -> void
+    def initialize(parameters:, method_definition:)
+      @parameters = parameters
+      @method_definition = method_definition
```
vinistock: Do we need this back link? Considering that consumers will interact with the API like this:

```ruby
declaration.definitions.first.signatures
```

I think we can remove this.

soutaro: This is for `MethodDeclaration#signatures`. It will return an array of `Signature`, and then we should need the backlink to retrieve docs and something associated to the method definition.

We can drop the backlink and add a new class or use tuple for `#signatures` method though?

lib/rubydex/signature.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    #: Array[Parameter]
+    attr_reader :parameters
+
+    #: MethodDefinition
+    attr_reader :method_definition
+
+    #: (parameters: Array[Parameter], method_definition: MethodDefinition) -> void
+    def initialize(parameters:, method_definition:)
```
soutaro: Oops. I don't want keyword args here.

ext/rubydex/signature.c
soutaro: This is also for `MethodDeclaration#signatures`. That will be in `declaration.c` and then the `rdxi_signatures_to_ruby` will be used from two files.

Another question is if `Rubydex::Signature` is a good place to define the class. Does `Rubydex::MethodDefinition::Signature` make more sense?

vinistock: So far, we haven't really nested any of the classes in sub-modules. I think `Rubydex::Signature` is fine for now.

lib/rubydex/signature.rb
vinistock: Let's discuss in the team sync, but I think exposing signatures at the declaration level has the potential to be a bit confusing.

Maybe we can handle the alias case in some other way.

ext/rubydex/signature.c
```diff
@@ -0,0 +1,87 @@
...
+    case ParameterKind_OptionalKeyword:    return cOptionalKeywordParameter;
+    case ParameterKind_RestKeyword:        return cRestKeywordParameter;
+    case ParameterKind_Block:              return cBlockParameter;
+    case ParameterKind_Forward:            return cForwardParameter;
+    default: rb_raise(rb_eRuntimeError, "Unknown ParameterKind: %d", kind);
+    }
+}
+
+VALUE rdxi_signatures_to_ruby(SignatureArray *arr, VALUE graph_obj, VALUE default_method_def) {
+    if (arr == NULL || arr->len == 0) {
```
vinistock: How about we raise in `rdxr_method_definition_signatures` and `rdxr_method_alias_definition_signatures` if the pointer is NULL?

This should never really happen, so I'm trying to avoid having it fail silently.

ext/rubydex/signature.c
```diff
@@ -0,0 +1,87 @@
...
+    for (size_t i = 0; i < arr->len; i++) {
+        SignatureEntry sig_entry = arr->items[i];
+
+        VALUE method_def;
+        if (default_method_def != Qnil) {
+            method_def = default_method_def;
+        } else {
+            VALUE def_argv[] = {graph_obj, ULL2NUM(sig_entry.definition_id)};
+            method_def = rb_class_new_instance(2, def_argv, cMethodDefinition);
+        }
```
vinistock: I don't fully understand why we need these. Isn't `self` already the method or alias definition?

We shouldn't need to create a new one or even pass the `default_method_def`.

lib/rubydex/signature.rb
vinistock: Based on what we discussed, let's remove `method_definition` from the initializer and only accept the parameters. It will also simplify `rdxi_signatures_to_ruby` a bit.

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,226 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
...
+    RequiredKeyword = 3,
+    OptionalKeyword = 4,
+    RestKeyword = 5,
+    Block = 6,
+    Forward = 7,
+}
+
+fn map_parameter_kind(param: &Parameter) -> ParameterKind {
+    match param {
+        Parameter::RequiredPositional(_) | Parameter::Post(_) => ParameterKind::RequiredPositional,
```
vinistock: Let's create `Post` as a `ParameterKind` as well. It's similar to a `RequiredPositional`, but it does carry its own meaning.

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,226 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
...
+            return ptr::null_mut();
+        };
+
+        let mut sig_entries: Vec<SignatureEntry> = Vec::new();
+        collect_method_signatures(graph, method_def, definition_id, &mut sig_entries);
+
+        let mut boxed = sig_entries.into_boxed_slice();
+        let len = boxed.len();
+        let items_ptr = boxed.as_mut_ptr();
+        std::mem::forget(boxed);
```
vinistock: Do we need to forget here if we're ultimately passing this to `Box::into_raw`?

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -350,3 +350,226 @@ pub unsafe extern "C" fn rdx_definition_name_location(pointer: GraphPointer, def
         create_location_for_uri_and_offset(graph, document, name_offset)
     })
 }
+
```
vinistock, line 324: From this point on, everything is about signatures. What do you think about creating a `signature_api.rs` file to house these?

rust/rubydex/src/query.rs
```diff
@@ -458,6 +458,123 @@ fn method_argument_completion<'a>(
...
+        if let Ancestor::Complete(ancestor_id) = ancestor {
+            let ancestor_ns = graph.declarations().get(ancestor_id)?.as_namespace()?;
+            if let Some(&decl_id) = ancestor_ns.member(&str_id) {
+                return Some(decl_id);
+            }
+        }
+    }
+
+    None
+}
```
vinistock, line 608: We have an extremely similar implementation in `declaration_api::rdx_declaration_find_member`. Maybe we can move that one to `query.rs` and reuse it for both implementations?

rust/rubydex/src/query.rs
```diff
@@ -458,6 +458,123 @@ fn method_argument_completion<'a>(
...
+
+/// Result of dealiasing a single level of method alias.
+pub enum DealiasMethodResult {
+    /// The alias target is a concrete method definition.
+    Method(DefinitionId),
+    /// The alias target is another alias.
+    Alias(DefinitionId),
+}
+
+/// Dereferences a `MethodAliasDefinition` by one level, returning the definitions
```
vinistock, line 481: Does this mean that if someone creates an alias of an alias, we don't find the original signature?

E.g.:

```ruby
class Foo
  def bar(a, b, c); end

  alias_method :new_bar, :bar

  # We should follow the alias chain to be able to return the signatures
  # that apply to the alias
  alias_method :newest_bar, :new_bar
end
```

lib/rubydex/signature.rb
soutaro: Sure. Will remove the `method_definition`.

ext/rubydex/signature.c
soutaro: We don't need it because we remove the backlink from the signature object.

rust/rubydex-sys/src/definition_api.rs
soutaro: If we don't use `forget`, the contents of the `Box` will be deallocated by the destructor.
We could use `Box::into_raw` instead, but I don't think it's worth the rewrite since the `mem::forget` pattern is already used elsewhere in the codebase.

```rust
let items_ptr = Box::into_raw(boxed) as *mut SignatureEntry;
```

ext/rubydex/signature.c
soutaro: Yeah, I think the design is weird. Will take a look at this tonight.

rust/rubydex/src/query.rs
soutaro: Refactored to share the `query.rs` implementation.

rust/rubydex/src/query.rs
soutaro: It works. The nested dealiasing is implemented in `rdx_method_alias_definition_signatures`. (But I don't think this is good.)

soutaro: The nested dealiasing is now implemented in `query::deep_dealias_method`.

ext/rubydex/signature.c
soutaro: Fixed the implementation. `rdx_method_definition_signatures` and `rdx_method_alias_definition_signatures` always return an array. `rdxi_signatures_to_ruby` can safely skips checking if `arr` is `NULL`.

rust/rubydex/src/query.rs
```diff
@@ -458,6 +458,243 @@ fn method_argument_completion<'a>(
...
+            Some(Definition::Method(_)) => Some(DealiasMethodResult::Method(*def_id)),
+            Some(Definition::MethodAlias(_)) => Some(DealiasMethodResult::Alias(*def_id)),
+            _ => None,
+        })
+        .collect())
+}
+
+/// Result of following a `MethodAliasDefinition` chain to completion.
+#[derive(Debug)]
+pub struct DeepDealiasMethodResult {
```
soutaro: I'm not sure if we really need the details of the failure. It keeps every details the implementation has in this layer, and C API layer ignores them.

The implementation would be a bit simpler if we can ignore the error details.

ext/rubydex/signature.c
```diff
@@ -0,0 +1,72 @@
...
+
+    for (size_t i = 0; i < arr->len; i++) {
+        SignatureEntry sig_entry = arr->items[i];
+
+        VALUE parameters = rb_ary_new_capa((long)sig_entry.parameters_len);
+        for (size_t j = 0; j < sig_entry.parameters_len; j++) {
+            ParameterEntry param_entry = sig_entry.parameters[j];
+
+            VALUE param_class = parameter_class_for_kind(param_entry.kind);
+            VALUE name_sym = ID2SYM(rb_intern(param_entry.name));
```
vinistock: Does `rb_intern` make a copy of the string we pass? I'm asking because we free the signature array object further below. If Ruby interning just keeps a pointer, it could lead to a seg fault.

test/definition_test.rb
```diff
@@ -204,6 +204,286 @@ class NotDeprecated; end
...
+
+      method_def = graph["Object#baz()"].definitions.first
+      refute_nil(method_def)
+
+      path = context.absolute_path_to("file1.rb")
+      params = method_def.signatures.first.parameters
+      assert_equal(1, params.length)
+      assert_instance_of(Rubydex::Signature::ForwardParameter, params[0])
+      assert_equal(:"...", params[0].name)
+      assert_equal("#{path}:1:9-1:12", params[0].location.to_display.to_s) # ...
```
vinistock: It's the only parameter in this example, so I think we can remove the comment.
```suggestion
      assert_equal("#{path}:1:9-1:12", params[0].location.to_display.to_s)
```

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,214 @@
...
+        std::mem::forget(boxed);
+
+        Box::into_raw(Box::new(SignatureArray { items: items_ptr, len }))
+    })
+}
+
+/// Helper: build signature entries from a `MethodDefinition` and append them to the output vector.
+fn collect_method_signatures(
+    graph: &rubydex::model::graph::Graph,
+    method_def: &rubydex::model::definitions::MethodDefinition,
```
vinistock: Nit: let's import `Graph` and `MethodDefinition` instead of using the fully qualified path.

rust/rubydex/src/query.rs
```diff
@@ -458,6 +458,241 @@ fn method_argument_completion<'a>(
...
+/// found under the aliased name.
+///
+/// # Errors
+///
+/// Returns a `DealiasMethodError` if the alias target cannot be resolved.
+///
+/// # Panics
+///
+/// Panics if a `SelfReceiver` definition cannot be resolved to a namespace with a singleton class.
+pub fn dealias_method(
```
vinistock: This seems a bit more complicated than what I would've expected. It's a requirement that the graph must be resolved before performing queries like this one. To dealias a method, I would've expected something like this:

```rust
fn dealias(graph: &Graph, alias: &MethodAliasDefinition) {
  let decl_id = graph.definition_id_to_declaration_id(*def_id).unwrap();
  let alias_decl = graph.declarations().get(decl_id).unwrap();
  
  // Get the already resolved alias owner. No need to worry about the receiver
  // that HAS to have been resolved during resolution already
  let Some(method_decl_id) = find_member_in_ancestors(graph, alias_decl.owner_id(), *alias.old_name_str_id(), false) else {
        return Err(DealiasMethodError::MemberNotFound);
    };
}
```

I don't think we need the `EnclosingNamespace` either.

ext/rubydex/signature.c
soutaro: I believe `rb_intern` copies the string, but I couldn't find any documentation that explicitly says so. Changed to `rb_str_intern` with `rb_utf8_str_new_cstr`, which [creates *dynamic* symbols](https://github.com/ruby/ruby/blob/ruby_4_0/include/ruby/internal/intern/string.h#L1151-L1164) (subject to GC) from a [copied C string](https://github.com/ruby/ruby/blob/ruby_4_0/include/ruby/internal/intern/string.h#L1573-L1587).

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -31,12 +38,42 @@ impl GraphTest {
...
+        self.sources.get(uri).expect("source not found for URI")
+    }
+
+    /// Returns the source text for a definition, sliced by its offset.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition or its document does not exist.
+    #[must_use]
+    pub fn source_at(&self, definition_id: &DefinitionId) -> &str {
```
Morriar, line 66: I think we can move this to `Offset` and pass the document as we do with https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/offset.rs#L90-L103

rust/rubydex/src/query.rs
vinistock: I think differentiating the error message is probably worth it, but I think this should be an enum instead of a struct.

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// C-compatible array of signatures.
+#[repr(C)]
+pub struct SignatureArray {
+    pub items: *mut SignatureEntry,
+    pub len: usize,
+}
+
+/// Returns a newly allocated array of signatures for the given method definition id.
+/// Returns NULL if the definition is not a method definition.
```
st0012: I think it panics in this case?

rust/skills.json
st0012, line 1: Should this be included?

test/definition_test.rb
```diff
@@ -364,6 +364,286 @@ class << self
     end
   end
 
+  def test_method_definition_signatures_with_various_parameter_kinds
```
st0012, line 403: Can we include test cases for singleton methods?

rust/skills.json
soutaro: Oh no. I run the skill in the directory... 🙇‍♂️

rust/rubydex/src/query.rs
soutaro: Ignore errors and return only the successful method ids, for now.

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct ParameterEntry {
+    pub name: *const c_char,
+    pub kind: ParameterKind,
+    pub location: *mut Location,
+}
+
+/// C-compatible struct representing a single method signature (a list of parameters).
+#[repr(C)]
+pub struct SignatureEntry {
+    pub definition_id: u64,
```
st0012, line 52: Looks like this is never used. Do we keep it for future updates?

If that's the case, please add a comment for that.
If not, let's remove it and simplify related API.

ext/rubydex/signature.c
```diff
@@ -0,0 +1,72 @@
...
+VALUE cRestKeywordParameter;
+VALUE cForwardParameter;
+VALUE cBlockParameter;
+
+static VALUE parameter_class_for_kind(ParameterKind kind) {
+    switch (kind) {
+    case ParameterKind_RequiredPositional: return cPositionalParameter;
+    case ParameterKind_OptionalPositional: return cOptionalPositionalParameter;
+    case ParameterKind_RestPositional:     return cRestPositionalParameter;
+    case ParameterKind_Post:              return cPostParameter;
```
amomchilov, line 21:
```suggestion
    case ParameterKind_Post:               return cPostParameter;
```

ext/rubydex/signature.c
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+VALUE rdxi_signatures_to_ruby(SignatureArray *arr) {
+    VALUE signatures = rb_ary_new_capa((long)arr->len);
+
+    for (size_t i = 0; i < arr->len; i++) {
+        SignatureEntry sig_entry = arr->items[i];
+
+        VALUE parameters = rb_ary_new_capa((long)sig_entry.parameters_len);
```
amomchilov, line 37: For a follow-up PR, potential performance optimization:

Parameter-less functions are pretty common. We might see some nice memory gains  if we reuse the same empty array between them all.

lib/rubydex/signature.rb
```diff
@@ -0,0 +1,27 @@
...
+        @name = name
+        @location = location
+      end
+    end
+
+    #: Array[Parameter]
+    attr_reader :parameters
+
+    #: (Array[Parameter]) -> void
+    def initialize(parameters)
```
amomchilov, line 23: No return value?

lib/rubydex/signature.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Rubydex
+  class Signature
```
amomchilov, line 4: (Not for this PR, but in a follow-up)

In my experience, having all the parameters in a single array ends up making callers need to do lots of filtering (e.g. to find just positional args, just keyword args, pull out the block param, etc.).

As the need arises, I would encourage adding convenience methods like `positional_parameters`, `keyword_parameters`, `block`, etc.

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+    RestPositional = 2,
+    Post = 3,
+    RequiredKeyword = 4,
+    OptionalKeyword = 5,
+    RestKeyword = 6,
+    Forward = 7,
+    Block = 8,
+}
+
+fn map_parameter_kind(param: &Parameter) -> ParameterKind {
```
amomchilov, line 27: I think this would be nicer to have in the `impl Parameter` block, so you can just call like `p.kind`

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+        Parameter::OptionalKeyword(_) => ParameterKind::OptionalKeyword,
+        Parameter::RestKeyword(_) => ParameterKind::RestKeyword,
+        Parameter::Block(_) => ParameterKind::Block,
+        Parameter::Forward(_) => ParameterKind::Forward,
+    }
+}
+
+/// C-compatible struct representing a single parameter with its name, kind, and location.
+#[repr(C)]
+pub struct ParameterEntry {
```
amomchilov, line 43: Do we care about optimizing field layouts? There's some padding between the end of `kind` and the next pointer-aligned position where `location` starts.

If we add new fields in the future and put them at the end, we would be wasting the padding. How much do we care about optimizing for this case in Rubydex?

In this case, ordering `name`, `location` then `kind`?

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+use crate::location_api::{Location, create_location_for_uri_and_offset};
+use libc::c_char;
+use rubydex::model::definitions::{Definition, MethodDefinition, Parameter};
+use rubydex::model::graph::Graph;
+use rubydex::model::ids::DefinitionId;
+use std::ffi::CString;
+use std::ptr;
+
+/// C-compatible enum representing the kind of a parameter.
+#[repr(C)]
```
amomchilov, line 13: We can be explicit about the size, too


```suggestion
#[repr(C, u8)]
```

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let mut sig_entries: Vec<SignatureEntry> = Vec::new();
+        collect_method_signatures(graph, method_def, definition_id, &mut sig_entries);
+
+        let mut boxed = sig_entries.into_boxed_slice();
+        let len = boxed.len();
+        let items_ptr = boxed.as_mut_ptr();
+        std::mem::forget(boxed);
+
+        Box::into_raw(Box::new(SignatureArray { items: items_ptr, len }))
```
amomchilov, line 89: We're already using `Box::into_raw`, we can apply it here to remove the need for this `std::mem::forget(boxed)`.

```suggestion
        let len = sig_entries.len();
        let items_ptr = Box::into_raw(sig_entries.into_boxed_slice()) as *mut _;

        Box::into_raw(Box::new(SignatureArray { items: items_ptr, len }))
```

amomchilov: (or we can use `cast_const()`, depending on the convention this repo follows

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+                    kind: map_parameter_kind(param),
+                    location: create_location_for_uri_and_offset(graph, document, param_struct.offset()),
+                }
+            })
+            .collect::<Vec<_>>()
+            .into_boxed_slice();
+
+        let parameters_len = param_entries.len();
+        let parameters_ptr = param_entries.as_mut_ptr();
+        std::mem::forget(param_entries);
```
amomchilov, line 125: likewise

rust/rubydex-sys/src/signature_api.rs
vinistock: Yeah, I think we can remove this one.

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut sig_slice: Box<[SignatureEntry]> = unsafe { Box::from_raw(slice_ptr) };
+
+        for sig_entry in &mut sig_slice {
+            if !sig_entry.parameters.is_null() && sig_entry.parameters_len > 0 {
+                let param_slice_ptr = ptr::slice_from_raw_parts_mut(sig_entry.parameters, sig_entry.parameters_len);
+                let mut param_slice: Box<[ParameterEntry]> = unsafe { Box::from_raw(param_slice_ptr) };
+
+                for param_entry in &mut param_slice {
+                    if !param_entry.name.is_null() {
+                        let _ = unsafe { CString::from_raw(param_entry.name.cast_mut()) };
```
amomchilov, line 198: Equivalent, but more clear intent IMO

```suggestion
                        drop(unsafe { CString::from_raw(param_entry.name.cast_mut()) });
```

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -31,12 +37,36 @@ impl GraphTest {
...
+    ///
+    /// Panics if the URI has not been indexed.
+    #[must_use]
+    pub fn source(&self, uri: &str) -> &str {
+        self.sources.get(uri).expect("source not found for URI")
+    }
+
+    /// Returns the source text for a definition, sliced by its offset.
+    ///
+    /// # Panics
```
vinistock, line 62: Personally, I wouldn't include these changes as part of this PR. It's already pretty large and we can ship test improvements separately.

rust/rubydex-sys/src/signature_api.rs
```diff
@@ -0,0 +1,211 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_definition_signatures_free(ptr: *mut SignatureArray) {
+    if ptr.is_null() {
+        return;
+    }
+
+    // Take ownership of the SignatureArray
+    let arr = unsafe { Box::from_raw(ptr) };
+
+    if !arr.items.is_null() && arr.len > 0 {
```
amomchilov, line 186: Can this be inverted into an early-return, similar to the `ptr.is_null()` check above?

rust/rubydex/src/query.rs
```diff
@@ -469,6 +469,144 @@ fn method_argument_completion<'a>(
...
+        let pos = ancestors
+            .iter()
+            .position(|a| matches!(a, Ancestor::Complete(id) if *id == namespace_id))
+            .expect("namespace_id should be present in its own ancestor chain");
+        pos + 1
+    } else {
+        0
+    };
+
+    for ancestor in &ancestors[search_start..] {
```
vinistock, line 597: This refactor is switching 1 loop for 3. Before, we were iterating on ancestors and performing the checks inline.

Now, we're eagerly collecting ancestors using `iter().collect()` (which I think is not necessary? `ancestors` already returns a collection). Then we iterate to find the main namespace position and iterate again to search the ancestors.

I'd vote for us to keep the current implementation. At least for now.

rust/rubydex/src/query.rs
```diff
@@ -469,6 +469,144 @@ fn method_argument_completion<'a>(
...
+/// Recursively follows alias chains starting from `alias_id`, collecting all
+/// resolved `MethodDefinition` IDs. Unlike `dealias_method` which resolves one
+/// level, this function keeps following aliases until all chains are fully resolved.
+/// Circular aliases and missing targets are silently ignored.
+///
+/// # Panics
+///
+/// Panics if any alias definition in the chain has no corresponding declaration.
+#[must_use]
+pub fn deep_dealias_method(graph: &Graph, alias_id: DefinitionId) -> Vec<DefinitionId> {
```
vinistock, line 525: I think part of the test related changes are a result of this method returning definitions.

Should this return a `DeclarationId` instead? Then we return the declaration that an alias points to and, if desired, the consumer of the API can fetch the related definitions.

rust/rubydex-sys/src/signature_api.rs
soutaro: `Parameter` is a struct defined in the rubydex crate. I'd like to avoid having the function from the FFI crate.

rust/rubydex-sys/src/signature_api.rs
soutaro: Looks like it's not allowed. https://github.com/rust-lang/rust/issues/68585

ext/rubydex/signature.c
soutaro: Implemented in https://github.com/Shopify/rubydex/pull/788.

lib/rubydex/signature.rb
soutaro: Implemented in https://github.com/Shopify/rubydex/pull/788.

soutaro:
Implemented in #788, #779, and #792

PR #728, opened 2026-04-07
Fix default parent classes

vinistock:
We had originally considered `Object` to be the final possible class in ancestor chains, which was a temporary measure to get resolution working. This PR changes this to ensure it's actually `BasicObject`, which is the correct final ancestor.

The PR looks large, but there are basically only 3 actual fixes outside of test updates:

1. I created a `built_in` file with information about the basic core modules and classes. It adds definitions and declarations. I documented in a comment why it is necessary
2. Changed the final class checks from `OBJECT_ID` to `BASIC_OBJECT_ID`
3. Started finalizing the linearization context when we are done with a declaration. This is important because otherwise the `seen_ids` information can leak across recursive calls and that results in incorrect results

### DIFF
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -1115,13 +1115,13 @@ mod tests {
         let s = server_with_sources(&[(&a, "class Animal; end"), (&b, "module Greetable; end")]);
         let res = parse(&s.codebase_stats());
 
-        assert_eq!(res["files"], 2);
-        assert_json_int!(res, "declarations", 5);
-        assert_json_int!(res, "definitions", 2);
+        assert_eq!(res["files"], 3);
+        assert_json_int!(res, "declarations", 7);
+        assert_json_int!(res, "definitions", 7);
 
         let breakdown = &res["breakdown_by_kind"];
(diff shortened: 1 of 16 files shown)


alexcrocha (approved):
The synthetic `rubydex:built-in` document is now part of `graph.documents`. Should we filter it out from our public APIs? Not blocking, just want to understand if there's a plan for this, or if there's a reason it shouldn't be filtered out

st0012 (approved):
Can we have tests for:
- `class Foo < BasicObject`
- `class Foo < BasicObject; include Kernel`

vinistock:
@st0012  Added the tests.



@alexcrocha this is an interesting question. My initial impression is that we shouldn't automatically filter it, in case consumers want to understand where these base definitions and declarations are coming from. So I would leave the filtering to them. If we find that this makes the API too weird, we can always revisit.

vinistock:
### Merge activity

* **Apr 15, 2:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/728).
* **Apr 15, 2:55 PM UTC**: @vinistock merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/728).

PR #729, opened 2026-04-07
Fix top level constant resolution

vinistock:
We got the constant resolution algorithm slightly wrong in terms of the top level fallback. This PR fixes it.

1. First, there's no top level fallback when inside a class. All constants end up being available through inheritance. It looks like there's a top level fallback because not inheriting from `Object` is pretty uncommon, but you can see that this is the case because this crashes:

```ruby
# This crashes because `Foo` doesn't have `Object` in its ancestor chain,
# which is the owner of the `String` constant
class Foo < BasicObject
  String
end
```

And when using constants at the top level of a script, it works because `<main>` is an instance of `Object`.

```ruby
# This doesn't crash, but not because there's a top level fallback. It's simply
# because <main> is `Object`
String
```

2. The fallback **does exist for modules**. However, it's not a simple direct member access. Ruby will actually search the ancestors of `Object`. For example, this works:

```ruby
module Kernel
  FOO = 1
end

module Bar
  puts FOO
end
```

If Ruby can't find the constant in lexical scopes or ancestors of the surrounding nesting, it searches the ancestors of `Object`. The same exact behaviour occurs for top level references like `::FOO` - even inside classes!

```ruby
module Kernel
  FOO = 1
end

class Bar < BasicObject
  ::FOO # The top level reference will make Ruby search in the ancestors of Object
end
```

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1143,11 +1143,16 @@ impl<'a> Resolver<'a> {
         let parent_owner_id = if parent_has_parent_scope {
             match self.name_owner_id(parent_scope) {
                 Outcome::Resolved(id, _) => id,
-                Outcome::Unresolved(None) => self.create_todo_for_parent(parent_scope),
-                _ => *OBJECT_ID,
+                _ => self.create_todo_for_parent(parent_scope),
             }
         } else {
             *OBJECT_ID
+        };
+
+        // Ensure we follow constant aliases if that's the parent
+        let parent_owner_id = match self.resolve_to_primary_namespace(parent_owner_id, None) {
+            Outcome::Resolved(id, _) => id,
+            _ => *OBJECT_ID,
         };
 
         let fully_qualified_name = if parent_owner_id == *OBJECT_ID {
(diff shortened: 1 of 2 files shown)


vinistock:
* **#729** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/729?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#730](https://github.com/Shopify/rubydex/pull/730) <a href="https://app.graphite.com/github/pr/Shopify/rubydex/730?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#731](https://github.com/Shopify/rubydex/pull/731) <a href="https://app.graphite.com/github/pr/Shopify/rubydex/731?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/729?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#728** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/728?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

st0012 (approved):
Can you also update ruby-behaviours doc with this info?

vinistock:
### Merge activity

* **Apr 15, 2:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/729).
* **Apr 15, 2:57 PM UTC**: [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/729) rebased this pull request as part of a merge.
* **Apr 15, 3:02 PM UTC**: @vinistock merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/729).

PR #735, opened 2026-04-13
Ensure singleton class ancestors are linearized when created out of resolution loop

vinistock:
In `get_or_create_singleton_class`, we were lazily creating singleton classes and enqueueing their ancestors for linearization. The problem is that we invoke this method for handling definitions too and that happens _after the resolution loop finished_, which means that the enqueue work was always ignored.

There are a few ways to solve this:

1. Drain the queue again after handling definitions
2. Move handling all definitions inside of the same unified loop
3. Create an eager/lazy mode for linearizing the ancestors of lazily created singleton classes, so that any calls to `get_or_create_singleton_class` inside of the resolution loop enqueue work and all others eagerly linearize

I believe 1 is not a good option as we'll need to loop through everything again, including unresolved references. Option 2 seems like the most elegant of all, but it might be a considerably large refactor, so I went with option 3 for now.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -185,7 +185,7 @@ impl<'a> Resolver<'a> {
                 };
                 let str_id = *method.str_id();
                 match self.graph.definition_id_to_declaration_id(*def_id) {
-                    Some(&owner_decl_id) => match self.get_or_create_singleton_class(owner_decl_id) {
+                    Some(&owner_decl_id) => match self.get_or_create_singleton_class(owner_decl_id, false) {
                         Some(singleton_id) => {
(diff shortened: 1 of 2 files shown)


vinistock:
* **#757** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/757?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#735** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/735?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/735?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexcrocha (approved):

PR #737, opened 2026-04-14
Move resolution tests to separate file

alexcrocha:
Part of https://github.com/Shopify/rubydex/issues/649

This PR moves the tests from `resolution.rs` into `resolution_tests.rs`, separating test code from the indexer implementation as the file was getting too large at over 6k lines.

These relocated tests will be grouped into submodules in follow-up PRs.

PR #738, opened 2026-04-14
Resolve retroactive method visibility changes

alexcrocha:
This PR continues the work in #89 and follows #695, which added indexing for `MethodVisibilityDefinition` and left the `Definition::MethodVisibility` arm in resolution as a TODO.

This change resolves retroactive `private`, `protected`, and `public` visibility changes by deferring visibility definitions to a second pass inside `handle_remaining_definitions`, so methods and attrs are always declared before visibility is applied.

For inherited targets such as `private :foo` where `foo` comes from a parent class or included module, the resolver creates a child-owned `MethodDeclaration`, matching CRuby's behaviour where `Child.instance_method(:foo).owner` is `Child`. When ancestor resolution is partial, unresolved visibility changes are requeued for a later resolve pass.

We also add diagnostics for undefined visibility targets and introduce `Graph::visibility()` as a helper for querying a method declaration's resolved visibility.

Note: This PR intentionally covers retroactive `private`, `protected`, and `public` resolution only. `module_function` will come in a follow-up PR because its dual instance/singleton behaviour needs separate handling.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -14,6 +14,7 @@ use crate::model::ids::{ConstantReferenceId, DeclarationId, DefinitionId, Method
 use crate::model::name::{Name, NameRef, ParentScope, ResolvedName};
 use crate::model::references::{ConstantReference, MethodRef};
 use crate::model::string_ref::StringRef;
+use crate::model::visibility::Visibility;
 use crate::stats;
 
 /// An entity whose validity depends on a particular `NameId`.
@@ -599,6 +600,44 @@ impl Graph {
     #[must_use]
     pub fn name_dependents(&self) -> &IdentityHashMap<NameId, Vec<NameDependent>> {
         &self.name_dependents
+    }
+
+    /// Returns the visibility for a method declaration.
+    ///
+    /// Scans the declaration's backing definitions from the end:
+    /// - First `MethodVisibilityDefinition` wins
+    /// - Otherwise, falls back to the latest method-like definition's indexed visibility
+    /// - Returns `None` if the declaration has no definitions with visibility
+    #[must_use]
+    pub fn visibility(&self, declaration_id: &DeclarationId) -> Option<Visibility> {
+        let declaration = self.declarations.get(declaration_id)?;
+        let definitions = declaration.definitions();
+
+        // Scan from the end: last visibility directive wins
+        for def_id in definitions.iter().rev() {
+            if let Some(definition) = self.definitions.get(def_id) {
+                match definition {
+                    Definition::MethodVisibility(vis) => return Some(*vis.visibility()),
+                    Definition::Method(method) => return Some(*method.visibility()),
+                    Definition::AttrAccessor(attr) => return Some(*attr.visibility()),
+                    Definition::AttrReader(attr) => return Some(*attr.visibility()),
+                    Definition::AttrWriter(attr) => return Some(*attr.visibility()),
+                    Definition::Class(_)
+                    | Definition::SingletonClass(_)
+                    | Definition::Module(_)
+                    | Definition::Constant(_)
+                    | Definition::ConstantAlias(_)
+                    | Definition::ConstantVisibility(_)
+                    | Definition::GlobalVariable(_)
+                    | Definition::InstanceVariable(_)
+                    | Definition::ClassVariable(_)
+                    | Definition::MethodAlias(_)
+                    | Definition::GlobalVariableAlias(_) => {}
+                }
+            }
+        }
+
+        None
     }
 
     /// Drains the accumulated work items, returning them for use by the resolver.
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -3,6 +3,7 @@ use std::{
     hash::BuildHasher,
 };
 
+use crate::diagnostic::{Diagnostic, Rule};
 use crate::model::{
     built_in::{BASIC_OBJECT_ID, CLASS_ID, KERNEL_ID, MODULE_ID, OBJECT_ID},
     declaration::{
@@ -272,6 +273,8 @@ impl<'a> Resolver<'a> {
     /// Handle other definitions that don't require resolution, but need to have their declarations and membership created
     #[allow(clippy::too_many_lines)]
     fn handle_remaining_definitions(&mut self, other_ids: Vec<DefinitionId>) {
+        let mut method_visibility_ids = Vec::new();
+
         for id in other_ids {
             match self.graph.definitions().get(&id).unwrap() {
                 Definition::Method(method_definition) => {
@@ -538,8 +541,8 @@ impl<'a> Resolver<'a> {
                 Definition::ConstantVisibility(_constant_visibility) => {
                     // TODO
                 }
-                Definition::MethodVisibility(_method_visibility) => {
-                    // TODO
+                Definition::MethodVisibility(_) => {
+                    method_visibility_ids.push(id);
                 }
                 Definition::Class(_)
                 | Definition::SingletonClass(_)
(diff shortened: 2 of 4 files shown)


alexcrocha:
* **#738** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#737** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/737?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/rubydex/src/model/graph.rs
```diff
@@ -628,6 +629,34 @@ impl Graph {
...
+        // Scan from the end: last visibility directive wins
+        for def_id in definitions.iter().rev() {
+            if let Some(definition) = self.definitions.get(def_id) {
+                match definition {
+                    Definition::MethodVisibility(vis) => return Some(*vis.visibility()),
+                    Definition::Method(method) => return Some(*method.visibility()),
+                    Definition::AttrAccessor(attr) => return Some(*attr.visibility()),
+                    Definition::AttrReader(attr) => return Some(*attr.visibility()),
+                    Definition::AttrWriter(attr) => return Some(*attr.visibility()),
+                    _ => {}
```
vinistock: To force us to think about whether visibility applies to new definitions, how about we list all match arms here instead of using a catchall?

rust/rubydex/src/resolution.rs
```diff
@@ -17,17 +18,17 @@ use crate::model::{
 };
 
 enum Outcome {
-    /// The constant was successfully resolved to the given declaration ID. The second optional tuple element is a
-    /// declaration that still needs to have its ancestors linearized
+    /// The item was successfully resolved to the given declaration ID. The second optional element is a
```
vinistock: Are these comment updates related to the PR?

rust/rubydex/src/resolution.rs
```diff
@@ -541,6 +544,84 @@ impl<'a> Resolver<'a> {
...
+    /// regardless of whether `def bar` appeared earlier or later in the unordered definition list.
+    fn resolve_method_visibilities(&mut self, visibility_ids: Vec<DefinitionId>) {
+        for id in visibility_ids {
+            let Definition::MethodVisibility(method_visibility) = self.graph.definitions().get(&id).unwrap() else {
+                unreachable!()
+            };
+
+            let str_id = *method_visibility.str_id();
+            let uri_id = *method_visibility.uri_id();
+            let offset = method_visibility.offset().clone();
```
vinistock, line 574: Do we need this clone?

rust/rubydex/src/resolution.rs
```diff
@@ -541,6 +544,84 @@ impl<'a> Resolver<'a> {
...
+                            .add_diagnostic(diagnostic);
+                    }
+                    Outcome::Retry(_) => {
+                        // Ancestors not fully resolved — requeue for retry on the next resolve() call.
+                        // The inner Option<DeclarationId> is the owner itself, not a specific unlinearized
+                        // ancestor. The blocker is an unresolved constant (e.g., `include M` where M hasn't
+                        // resolved), which the convergence loop handles on the next pass.
+                        self.graph.extend_work(std::iter::once(Unit::Definition(id)));
+                    }
+                }
```
vinistock: We should not use `search_ancestors` here because we're already past the point of the resolution loop. Enqueueing retries here does nothing because we're no longer consuming from the queue.

Here's how I think we can go about this:

```rust
// Only namespaces can own declarations, so we can just unwrap
let owner = self.graph.declarations().get(&owner_id).unwrap().as_namespace().unwrap();

for ancestor in owner.ancestors() {
  match ancestor {
    Ancestor::Complete(id) => {
      let ancestor_decl = self.graph.declarations().get(&id).unwrap().as_namespace().unwrap();

      let Some(member) = ancestor_decl.member(&str_id) else {
        continue;
      };

      if id == owner_id {
        // method found and defined in the same namespace
        // change visibility
      } else {
        // method found in a different namespace (inherited)
        // create copy
      }
    }
    Ancestor::Partial(_) => {
      // Can't do anything in this case
    }
  }
}
```

rust/rubydex/src/resolution.rs
alexcrocha: No longer relevant as we're no longer using Outcome in this PR  as per https://github.com/Shopify/rubydex/pull/738#discussion_r3093945629 so I've reverted these comment updates

rust/rubydex/src/resolution.rs
alexcrocha: I Derived Copy on Offset, as a new commit since it touches other areas of the codebase, and we no longer need to clone

rust/rubydex/src/resolution.rs
alexcrocha: Thanks, this helped me understand the resolver's two-phase structure better. I amended to walking `owner.ancestors()` directly as suggested. 

Also, `Ancestor::Partial` entries still requeue via `extend_work`, so an unresolved ancestor (like an unindexed module) doesn't silently drop. It retries on the next `resolve()` call when the blocking constant resolves. Added a test for this as well.

rust/rubydex/src/resolution.rs
vinistock: Sorry, I was meaning to understand why we need to clone the offset in this case. I don't think we should derive `Copy` in `Offset` because that makes Rust implicitly create new copies of the structure when needed, which may end up creating unnecessary offset copies.

I'm okay with the explicit clone, I just wanted to understand why we needed it.

rust/rubydex/src/resolution.rs
```diff
@@ -550,6 +553,99 @@ impl<'a> Resolver<'a> {
...
+            };
+
+            let Some(namespace) = self
+                .graph
+                .declarations()
+                .get(&owner_id)
+                .and_then(|decl| decl.as_namespace())
+            else {
+                continue;
+            };
```
vinistock: Nit
```suggestion
            let Some(Declaration::Namespace(namespace)) = self
                .graph
                .declarations()
                .get(&owner_id)
            else {
                continue;
            };
```

rust/rubydex/src/resolution.rs
```diff
@@ -550,6 +553,99 @@ impl<'a> Resolver<'a> {
...
+                }
+            }
+
+            if visibility_applied {
+                continue;
+            }
+
+            if has_partial {
+                // Method might exist on an unresolved ancestor — requeue for retry.
+                self.graph.extend_work(std::iter::once(Unit::Definition(id)));
```
vinistock: This is not necessary. If the resolution loop failed to linearize ancestors or resolve constants, that will already be recorded in the remaining work. This is effectively duplicating it.

We can just ignore partial ancestors.

rust/rubydex/src/resolution.rs
alexcrocha: Ah, fair... I jumped the gun too quickly. Reverted the derive `Copy`.

We still need to clone here due to `method_visibility` being a reference borrowed from `self.graph.definitions()` which holds a shared borrow on `self.graph`, and keeping it as a reference would extend the borrow past the `&mut self` calls downstream, effectively blocking them

rust/rubydex/src/resolution.rs
alexcrocha: Paired with Vini on this to understand the resolution flow better. 
Documenting the outcome here for other reviewers.

Due to incremental changes, we still need to call `extend_work` when we have partial ancestor resolution as they can become complete on a later resolution pass. We have a [test](https://github.com/Shopify/rubydex/pull/738/changes#diff-75b8747bfffd417be07df84d0f4dc9ac4f6a650ca25d8e2f9b3484aaac35b849R4893-R4923) demonstrating this case

rust/rubydex/src/resolution.rs
vinistock: I see. However, we only need the offset in the failed case: when there's a diagnostic to create and we end up cloning the offset every time, even when there's no diagnostic.

Would it be possible to move it down and avoid cloning if there's no diagnostic?

rust/rubydex/src/resolution.rs
```diff
@@ -550,6 +553,90 @@ impl<'a> Resolver<'a> {
...
+                        let has_member = self
+                            .graph
+                            .declarations()
+                            .get(ancestor_id)
+                            .and_then(|decl| decl.as_namespace())
+                            .and_then(|ns| ns.member(&str_id))
+                            .is_some();
+
+                        if has_member {
+                            // Direct member: `create_declaration`'s FQN dedup attaches this
```
vinistock: LLMs are always eager to call fully qualified name FQN, but let's try to avoid that.

```suggestion
                            // Direct member: `create_declaration`'s fully qualified name dedup attaches this
```

rust/rubydex/src/resolution.rs
```diff
@@ -550,6 +553,90 @@ impl<'a> Resolver<'a> {
...
+                );
+                self.graph
+                    .declarations_mut()
+                    .get_mut(&owner_id)
+                    .unwrap()
+                    .add_diagnostic(diagnostic);
+            }
+        }
+
+        self.graph.extend_work(pending_work);
```
vinistock, line 640: Let's add a brief comment mentioning that we must extend work here to ensure that incremental resolution can resolve a previously unresolved visibility operation.

vinistock (approved):

rust/rubydex/src/resolution.rs
alexcrocha: Moving the clone down would keep `method_visibility` borrowed past the `&mut self` calls. We could refetch `self.graph.definitions().get(&id)` in the else branch, but I am not sure that's better than cloning here

PR #740, opened 2026-04-14
Use BUNDLE_GEMFILE to determine if we need a debug build

vinistock:
We were using the `.git` directory to determine if someone was working on Rubydex, so that we could decide between debug and release builds. Basically, there are two scenarios:

- When pre-compiling binaries on CI: we set the `RELEASE` flag
- When installing the gem for a platform we don't yet provide pre-compiled binaries: we fallback to this check

The issue is basically not confusing local development with the gem getting compiled in someone's machine. The `.git` directory choice is not a good one because if you install the gem from the git source, the directory will be there and then you get a slow debug build.

I think we can instead use `BUNDLE_GEMFILE`. Whenever we're working on Rubydex, Bundler is going to set the variable to point to the Rubydex Gemfile. If that is not set or if it points to something other than Rubydex's own Gemfile, then we can make release builds.

### DIFF
--- a/ext/rubydex/extconf.rb
+++ b/ext/rubydex/extconf.rb
@@ -8,7 +8,18 @@ unless system("cargo", "--version", out: File::NULL, err: File::NULL)
 end
 
 gem_dir = Pathname.new("../..").expand_path(__dir__)
-release = ENV["RELEASE"] || !gem_dir.join(".git").exist?
+
+# Use release mode for the compilation if:
+# - The RELEASE environment variable is set
+# - We're not working on Rubydex (BUNDLE_GEMFILE doesn't point to Rubydex's own Gemfile)
+#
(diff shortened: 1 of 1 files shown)


vinistock:
* **#740** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

paracycle (commented):
~~This is good, but will probably not solve the problem of people using `gem install rubydex` who have to build from source for one reason or another.~~

I was wrong about that, this changes the default compilation mode to `RELEASE` so it does the right thing by default.

paracycle (approved):

PR #744, opened 2026-04-15
Add abstract `Declaration#references`

vinistock:
For typing convenience, this PR adds an abstract `Declaration#references`. All declarations have the `references` method, so we don't want to be forced to narrow types when invoking `declaration.references`.


vinistock:
* **#744** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rbi/rubydex.rbi
alexcrocha, line 63: Should we narrow the return type here to `T::Enumerable[Rubydex::Reference]` now? Same for `InstanceVariable` and `ClassVariable` as well

vinistock: It makes no difference at the moment because we haven't actually implemented instance, class or global variable references. So all of these are currently hard coded to return an empty array.

Once we add support for those, these should reflect the specific reference classes (`InstanceVariableReference`, `ClassVariableReference`...).

alexcrocha (approved):

PR #754, opened 2026-04-17
Expose method reference receivers

vinistock:
This PR exposes the resolved declaration for method receivers. When there's no receiver or it's unresolved, we return `nil`. This helps us provide more accurate find references in the Ruby LSP.

Note: exposing the receiver as `Declaration?` is probably not 100% correct, but since we cannot know the type of method receivers until there's type inference, it's a decent mid-point.

I also prototyped changing the receiver to be a constant reference ID (instead of a name ID), but it leads to a lot of other issues. For example:

```ruby
class Foo
  def bar
    # In this method reference, the receiver is Foo. However, is the `self` considered
    # a constant reference to Foo? Not really...
    #
    # It would be pretty surprising to have this should up in find references and
    # much worse in a rename request.
    self.baz
  end
end
```

Maybe, in the future, the `self` would be something like a `TypeRef`? Anyway, it's clearly too early to model this so I went with a compromise.

### DIFF
--- a/ext/rubydex/reference.c
+++ b/ext/rubydex/reference.c
@@ -75,6 +75,28 @@ static VALUE rdxr_method_reference_location(VALUE self) {
     return location;
 }
 
+// MethodReference#receiver -> Rubydex::Declaration?
+// Returns the resolved declaration for the receiver of the method call. Returns nil when the receiver is not a
+// tracked constant or cannot be resolved.
+static VALUE rdxr_method_reference_receiver(VALUE self) {
+    HandleData *data;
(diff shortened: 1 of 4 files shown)


vinistock:
* **#754** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/754?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/754?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Morriar (approved):

PR #757, opened 2026-04-20
Index instance variable references

### LINKED ISSUE #517: Index instance variable references
The Ruby LSP provides find references and rename for instance variables. Indexing instance variables references is quite similar to constant references in the sense that we can resolve them during resolution with no other type checking steps.

Even the structure would be quite similar as I expect we would need to remember: uri_id, offsets and the name_id of that represents the `self` context for that variable.

Let's index instance variables and make sure that they are being resolved.

vinistock:
Closes #517

This PR starts indexing and handling instance variable references, which is required to match the Ruby LSP's find references for ivars.

The approach is mostly the same as any other reference. We have a new reference struct that gets collected during indexing and then merged into the main graph.

Resolution is running as a separate step at the end, which reuses the same logic as instance variable definitions.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer_tests.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer_tests.rs
@@ -89,6 +89,34 @@ macro_rules! assert_method_references_eq {
     }};
 }
 
+macro_rules! assert_instance_variable_references_eq {
+    ($context:expr, $expected_names:expr) => {{
+        let mut actual_references = $context
+            .graph()
+            .instance_variable_references()
+            .values()
+            .map(|r| {
+                (
+                    r.offset().start(),
+                    $context.graph().strings().get(r.str_id()).unwrap().as_str(),
+                )
+            })
+            .collect::<Vec<_>>();
+
+        actual_references.sort();
+
+        let actual_names = actual_references.iter().map(|(_, name)| *name).collect::<Vec<_>>();
+
+        assert_eq!(
+            $expected_names,
+            actual_names.as_slice(),
+            "instance variable references mismatch: expected `{:?}`, got `{:?}`",
+            $expected_names,
+            actual_names
+        );
+    }};
+}
+
 fn index_source(source: &str) -> LocalGraphTest {
     LocalGraphTest::new_with_backend("file:///foo.rb", source, super::backend())
 }
@@ -504,6 +532,38 @@ mod variable_tests {
     }
 
     #[test]
+    fn instance_variable_reads_are_indexed_as_references() {
+        let context = index_source({
+            "
+        class Foo
+          def bar
+            @baz
+          end
+        end
+        "
+        });
+
+        assert_no_local_diagnostics!(&context);
+        assert_instance_variable_references_eq!(context, ["@baz"]);
+    }
+
+    #[test]
+    fn instance_variable_operator_writes_are_references() {
+        let context = index_source({
+            "
+        class Foo
+          def bar
+            @baz += 1
+          end
+        end
+        "
+        });
+
+        assert_no_local_diagnostics!(&context);
+        assert_instance_variable_references_eq!(context, ["@baz"]);
+    }
+
+    #[test]
     fn index_instance_variable_definition() {
         let context = index_source({
             "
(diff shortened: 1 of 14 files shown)


vinistock:
* **#757** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/757?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/757?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#735** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/rubydex/src/model/graph.rs
```diff
@@ -986,6 +1033,23 @@ impl Graph {
...
+                // Detach from target declaration. Scan InstanceVariable declarations
+                // to find which one holds this reference.
+                for decl in self.declarations.values_mut() {
+                    if let Some(refs) = decl.instance_variable_references()
+                        && refs.contains(ref_id)
+                    {
+                        decl.remove_instance_variable_reference(ref_id);
+                        break;
+                    }
+                }
```
Morriar, line 1173: The scan over all declarations to find which one holds the ivar reference is O(N) in declarations. For constant references, removal is O(1) because we can look up the target declaration through the name system.

If we stored the resolved `DeclarationId` on `InstanceVariableRef` when resolving (similar to how constant refs have a `name_id` that resolves to a declaration), we could do a direct lookup here instead of scanning.

rust/rubydex/src/indexing/ruby_indexer_tests.rs
```diff
@@ -2075,14 +2135,12 @@ mod variable_tests {
               @baz, @qux = 3, 4
             end
 
-            @bar &= 5
```
Morriar, line 517: Should we keep this and make sure it creates a reference?

Morriar (commented):
Are the lint errors related?

rust/rubydex/src/indexing/ruby_indexer_tests.rs
vinistock: Yup.

rust/rubydex/src/model/graph.rs
vinistock: Good point. I followed the same approach we used for names, so that we can record resolved instance variables.

We might be able to refactor the code in the future to have the Unresolved/Resolved reference by generic since we always want to store a base type + the declaration_id it points to.

rust/rubydex/src/resolution.rs
```diff
@@ -722,7 +604,114 @@ impl<'a> Resolver<'a> {
...
-    /// Gets or creates a singleton class declaration for a given class/module declaration.  For class `Foo`, this
+    /// Resolves all instance variable references by linking them to their target declarations.
+    /// Runs as a post-pass after `handle_remaining_definitions` since ivar declarations must exist first.
+    fn resolve_instance_variable_references(&mut self) {
+        let ivar_ref_ids: Vec<InstanceVariableReferenceId> =
+            self.graph.instance_variable_references().keys().copied().collect();
+
+        for ref_id in ivar_ref_ids {
+            let ivar_ref = self.graph.instance_variable_references().get(&ref_id).unwrap();
+            let str_id = *ivar_ref.str_id();
```
st0012, line 735: Do we want to skip resolved ivar references here? Or we do want to iterate through ALL ivar references at every resolution?

vinistock: I added a new commit improving the incremental resolution of ivar references.

rust/rubydex/src/resolution.rs
```diff
@@ -748,27 +631,179 @@ impl<'a> Resolver<'a> {
...
+        None
+    }
+
+    /// Resolve an instance variable owner, creating missing singleton classes and re-queueing
+    /// the definition when its receiver or surrounding namespace is temporarily unresolved.
+    ///
+    /// Re-queueing is essential during incremental delete-then-readd of a receiver: without it,
+    /// ivars whose owner depends on a freshly-deleted receiver would be permanently dropped or
+    /// attached to `Object` (regression of #775).
+    fn resolve_ivar_owner(
```
st0012, line 773: How does this work with weird cases like:

```rb
class Foo::Bar # Foo should be a TODO
  def Foo.baz
    @baz = 1
  end
end
```

rust/rubydex/src/resolution_tests.rs
```diff
@@ -3692,6 +3692,181 @@ mod fqn_and_naming_tests {
     }
 }
 
+#[test]
```
st0012, line 3704: 2 additional cases we can cover:
- ivar reference in a singleton class body: `class Foo; class << self; @bar; end; end`
- ivar reference that can't be resolved from a parent: `class Bar < Foo; def bar; @bar; end; end` but `Foo` doesn't define `@bar`

vinistock: Added those two.

rust/rubydex/src/resolution.rs
vinistock: We look at the nesting definition for the ivar reference, which is the method and then check if it has a receiver.

At this point, the resolution loop ended, so if the receiver is already resolved, we know what is the `self` type inside of the method. Otherwise, we can't resolve the reference right now, so we enqueue as pending work for the next round of incremental resolution.

st0012:
I think we need to make sure ancestor change also updates the ivar references too:

```rs
    #[test]
    fn ivar_reference_retargets_when_surviving_namespace_superclass_changes() {
        context.index_uri(
            "file:///foo.rb",
            r"
            class Foo
              def initialize
                @var = 1
              end
            end
            ",
        );
        context.index_uri(
            "file:///baz.rb",
            r"
            class Baz
              def initialize
                @var = 1
              end
            end
            ",
        );
        context.index_uri("file:///bar_parent.rb", "class Bar < Foo; end");
        context.index_uri(
            "file:///bar_read.rb",
            r"
            class Bar
              def read
                @var
              end
            end
            ",
        );

        context.resolve();
        assert_declaration_references_count_eq!(context, "Foo#@var", 1);
        assert_declaration_references_count_eq!(context, "Baz#@var", 0);

        context.index_uri("file:///bar_parent.rb", "class Bar < Baz; end");
        context.resolve();

        assert_declaration_references_count_eq!(context, "Foo#@var", 0);
        assert_declaration_references_count_eq!(context, "Baz#@var", 1);
    }
```

vinistock:
This solution doesn't work because it doesn't fit the resolution algorithm in a way that accounts for incremental cases. I'll re-work this later.

PR #758, opened 2026-04-21
Expose constant alias targets in the Ruby API

vinistock:
Closes #689

This PR exposes constant alias targets in the Ruby API, which is necessary for correct completion in the Ruby LSP.

The implementation is pretty straight forward, we are just using existing Rust APIs.

### DIFF
--- a/test/declaration_test.rb
+++ b/test/declaration_test.rb
@@ -608,4 +608,114 @@ class DeclarationTest < Minitest::Test
       assert_equal("Foo#initialize()", decl.name)
     end
   end
+
+  def test_following_constant_alias_targets
+    with_context do |context|
+      context.write!("file1.rb", <<~RUBY)
+        module Foo
+          class Bar
+          end
+        end
+
+        module Baz
+          Qux = Foo
+        end
+
+        ALIAS = Baz
+        # This is the same as Foo::Bar. We need to be able to follow the alias step by step
+        ALIAS::Qux::Bar
+      RUBY
+
+      graph = Rubydex::Graph.new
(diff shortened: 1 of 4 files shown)


vinistock:
* **#758** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/758?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/758?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/declaration_test.rb
```diff
@@ -608,4 +608,111 @@ def initialize
...
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      alias_decl = graph["ALIAS"]
+      assert_instance_of(Rubydex::ConstantAlias, alias_decl)
+
+      target = alias_decl.target
+      assert_instance_of(Rubydex::Module, target)
+      assert_includes(["Foo", "Bar"], target.name)
```
st0012, line 718: Why does this module have 2 names?

vinistock: It doesn't. There's only one name, but there are two separate constant alias definitions with different targets.

Since we always pick the first one, there's no way to know exactly which one it will be. To avoid having the test be flaky, we can assert that the target name will always be either `Foo` or `Bar`.

st0012: Ah ok. This is not super intuitive tho, can we add a comment?

st0012 (approved):

PR #759, opened 2026-04-22
Unlink Rust shared shared object before copying

vinistock:
I believe this fixes the SIGKILL issues we've been getting when developing. My understanding of the problem is:

1. The Ruby LSP launches using the development version of Rubydex in the repo, which loads the `librubydex_sys.dylib`
2. As we're working on Rubydex, we re-compile and the dylib gets overridden
3. The problem is that we're not unlinking (rm) and then creating the new one, but just copying straight away, which macOS flags as insecure and interrupts with a SIGKILL

I tested this a bit and it seems that the issue indeed goes away.

### DIFF
--- a/ext/rubydex/extconf.rb
+++ b/ext/rubydex/extconf.rb
@@ -86,11 +86,15 @@ copy_dylib_commands = if Gem.win_platform?
   ""
 elsif RUBY_PLATFORM.include?("darwin")
   src_dylib = target_dir.join("librubydex_sys.dylib")
-  "\t$(COPY) #{src_dylib} #{lib_dir}"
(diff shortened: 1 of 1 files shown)


vinistock:
* **#759** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/759?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/759?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexcrocha (approved):

PR #764, opened 2026-04-24
Resolve constants against enclosing lexical scope in `class << self`

splantio:
## Summary

Fixes a constant-resolution bug in `class << self` bodies. Constants referenced from inside a singleton class block (or from methods defined inside one) were not resolving against the enclosing lexical scope. For example:

```ruby
module A
  module B
    class Sibling; end

    class Main
      class << self
        def m
          Sibling   # was left unresolved
        end
      end
    end
  end
end
```

## Root cause

The resolution algorithm already walks lexical scopes correctly; it follows the `nesting` linked list on each `Name`. The bug was in the indexer: when the singleton class `Name` was created, its `nesting` was set to `None`, so the walk terminated at the singleton class frame and never reached `A::B` or `A`.

## Fix

One-line change in `RubyIndexer::visit_singleton_class_node`: populate the singleton class `Name`'s `nesting` with the current lexical scope at the point of the `class << self` block, matching how regular classes and modules are indexed.

Two existing tests are updated to reflect the new `Name` identity: `<Foo>` with a nesting interns to a different `Name` than `<Foo>` without one.

## Tests

Added 5 tests covering:

- `Sibling` used inside a method in `class << self`
- `Sibling` used inside `class << self` nested under an outer class
- `Sibling` used directly in the `class << self` body (not inside a method)
- Non-regression: sibling still resolves from instance method / class body / top level
- Does not over-resolve unknown constants

All 528 rubydex tests pass; clippy and fmt are clean.

### DIFF
--- a/rust/rubydex-sys/src/name_api.rs
+++ b/rust/rubydex-sys/src/name_api.rs
@@ -44,9 +44,10 @@ pub fn nesting_stack_to_name_id(
 }
 
 /// Processes a qualified name (e.g., `"Foo::Bar"` or `"<Foo>"`) by splitting on `"::"` and registering each part in the
-/// graph. Singleton class names (starting with `<`) use `ParentScope::Attached` and `nesting=None`, matching how the
-/// indexer creates them. When a singleton is the first part (i.e., `current_name` has no parent), `current_nesting` is
-/// used as the attachment point.
+/// graph. Singleton class names (starting with `<`) use `ParentScope::Attached` and a `nesting` equal to the attached
+/// target, matching how the indexer creates them (`class << self` always sits lexically inside its attached class).
+/// When a singleton is the first part (i.e., `current_name` has no parent), `current_nesting` is used as the attachment
+/// point.
 fn process_qualified_name(
     graph: &mut Graph,
     qualified_name: &str,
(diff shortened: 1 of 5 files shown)


vinistock:
Oh, wow I can't believe we forgot to add a test for that. Thanks for the PR! The scenarios are all great, but the fix is actually significantly simpler. The resolution algorithm already walks lexical scopes correctly, but we're mistakenly not associating the nesting with the name objects for singleton.

Here's the fix (needs some test updates too, but logic changes it's just this):

```
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1507,9 +1507,10 @@ impl Visit<'_> for RubyIndexer<'_> {
         };
 
         let string_id = self.local_graph.intern_string(singleton_class_name);
+        let nesting = self.current_lexical_scope_name_id();
         let name_id = self
             .local_graph
-            .add_name(Name::new(string_id, ParentScope::Attached(attached_target), None));
+            .add_name(Name::new(string_id, ParentScope::Attached(attached_target), nesting));
```

vinistock (approved):
The failure for lint is fixed on main, so a rebase will probably fix it. Thanks for the contribution!

PR #766, opened 2026-04-29
Resolve retroactive constant visibility changes

alexcrocha:
Continues #89 and follows #506, which indexed `ConstantVisibilityDefinition` and left the `Definition::ConstantVisibility` arm in resolution as a TODO.

This PR resolves retroactive `private_constant` and `public_constant` calls by reusing the second-pass mechanism from #738: visibility definitions are processed in `handle_remaining_definitions` after the ancestors linearization loop terminates, so we can resolve the owner and look up the named constant reliably before we apply the visibility.

Visibility is applied only to direct members of the owner. Inherited constants are rejected because both `private_constant` and `public_constant` raise `NameError` on inherited names in Ruby. When the target is missing, an `UndefinedConstantVisibilityTarget` diagnostic is emitted.

`Graph::visibility()` was changed to treat constant visibility as sticky once set: reassigning a private constant in a reopened class doesn't flip it back to public.

Note: we don't handle the case when an owner isn't yet in the graph or is indexed in a later resolve(). A follow-up PR will handle incremental resolution properly

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -288,6 +288,16 @@ impl Graph {
         &self.documents
     }
 
+    /// Attaches a diagnostic to the document with the given `uri_id`. The diagnostic clears
+    /// automatically when the document is deleted or re-indexed.
+    ///
+    /// # Panics
+    ///
+    /// Panics if no document is registered for `uri_id`.
+    pub fn add_document_diagnostic(&mut self, uri_id: UriId, diagnostic: Diagnostic) {
+        self.documents.get_mut(&uri_id).unwrap().add_diagnostic(diagnostic);
+    }
+
     /// # Panics
     ///
     /// Panics if the definition is not found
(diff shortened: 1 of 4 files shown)


alexcrocha:
* **#778** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#766** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#769** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/769?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rust/rubydex/src/resolution.rs
```diff
@@ -538,8 +538,88 @@ impl<'a> Resolver<'a> {
                         Declaration::GlobalVariable(Box::new(GlobalVariableDeclaration::new(name, *OBJECT_ID)))
                     });
                 }
-                Definition::ConstantVisibility(_constant_visibility) => {
-                    // TODO
+                Definition::ConstantVisibility(constant_visibility) => {
+                    // Both `private_constant` and `public_constant` raise `NameError` on inherited
+                    // names in Ruby, so we attach only to direct members.
```
vinistock: Another way to phrase it

```suggestion
                    // Both `private_constant` and `public_constant` can only target direct members. Inheritance or surrounding lexical scopes are not taken into account
```

rust/rubydex/src/model/graph.rs
```diff
@@ -622,12 +635,12 @@ impl Graph {
...
-                    | Definition::SingletonClass(_)
+                    Definition::ConstantVisibility(vis) => return Some(*vis.visibility()),
+                    Definition::Constant(_)
+                    | Definition::Class(_)
                     | Definition::Module(_)
-                    | Definition::Constant(_)
-                    | Definition::ConstantAlias(_)
-                    | Definition::ConstantVisibility(_)
+                    | Definition::ConstantAlias(_) => has_constant_def = true,
+                    Definition::SingletonClass(_)
```
vinistock: It seems that what the logic is revealing now is we have different rules for visibility depending on the type of entity.

I think the code would be easier to understand, if we matched the type of the `declaration` first and then defined the specific logic of what to search in definitions for that type.

```rust
match declaration {
  Declaration(Namespace(_)) => {
    definitions.iter().rev().find(|def| {
      // are any of the definitions a visibility change to private?
    }).unwrap_or(Visibility::Public)
  }
  Declaration(InstanceVariable(_)) => {
    None
  }
}
```

rust/rubydex/src/resolution.rs
```diff
@@ -538,8 +538,88 @@ impl<'a> Resolver<'a> {
...
+                    let name_id = *constant_visibility.name_id();
```
vinistock: This modelling is actually wrong. The target of private_constant and public_constant is just a simple StringId since the target is not resolved by Ruby. It's always just a direct member access, which means we are wasting memory storing a Name that will never have any relevant information other than the StringId.

rust/rubydex/src/resolution.rs
```diff
@@ -538,8 +538,88 @@ impl<'a> Resolver<'a> {
...
+                            let parent_str_id = self.graph.names().get(parent_id).unwrap().str();
+                            let receiver_name = self.graph.strings().get(parent_str_id).unwrap().as_str().to_string();
+                            let diagnostic = Diagnostic::new(
+                                Rule::UndefinedConstantVisibilityTarget,
+                                uri_id,
+                                offset,
+                                format!(
+                                    "unresolved namespace `{receiver_name}` for visibility change of `{constant_name}`"
+                                ),
+                            );
```
vinistock: All of this should disappear once `name_id` is converted to `str_id` because all you have to do is actually this:

```rust
declaration_for_immediate_lexical_scope.member(str_id)
```

rust/rubydex/src/model/graph.rs
alexcrocha: Refactored. We are now matching by declaration type first and it does feel simpler to reason about

rust/rubydex/src/resolution.rs
alexcrocha: Rebased this PR on top of #769, which adds a `target: StringId` to `ConstantVisibilityDefinition`. We no longer need to walk through `name_id` for the target string

rust/rubydex/src/resolution.rs
alexcrocha: We had talked about not supporting receivers for now, but since we are indeed already handling them, I kept the logic in `name_owner_id` while refactoring to use the new target `StringId`

rust/rubydex/src/resolution_tests.rs
```diff
@@ -5128,4 +5128,201 @@ mod visibility_resolution_tests {
...
+            r"
+            Foo.private_constant :BAR
+            ",
+        );
+        context.resolve();
+
+        assert_diagnostics_eq!(
+            context,
+            &[
+                "undefined-constant-visibility-target: unresolved namespace `Foo` for visibility change of `BAR` (1:23-1:26)"
```
vinistock: I think we shouldn't push a diagnostic in this scenario. The rationale is avoiding over-notification of errors for the user.

We already report the receiver as undefined. If we don't know what the receiver is, then there's no expectation that we would be able to process the visibility operation. The root cause is the undefined receiver.

After the receiver gets defined, then that may change of course. But I would avoid the double notification for better DX.

rust/rubydex/src/resolution.rs
```diff
@@ -538,8 +538,83 @@ impl<'a> Resolver<'a> {
...
+                    // Inheritance or surrounding lexical scopes are not taken into account.
+                    let receiver = *constant_visibility.receiver();
+                    let target = *constant_visibility.target();
+                    let uri_id = *constant_visibility.uri_id();
+                    let offset = constant_visibility.offset().clone();
+                    let lexical_nesting_id = *constant_visibility.lexical_nesting_id();
+                    let constant_name = self.graph.strings().get(&target).unwrap().as_str().to_string();
+
+                    let owner_id = if let Some(receiver_name_id) = receiver {
+                        let receiver_decl = match self.resolve_constant_internal(receiver_name_id) {
```
vinistock: The visibility handling happens _after_ the resolution loop is already complete. We cannot invoke mutating methods at this point, such as triggering constant resolution.

At this point, if the receiver is not yet resolved, then it's because we cannot resolve it (or it's in fact undefined). This should instead be:

```rust
let NameRef(Resolved(name)) = self.graph.names().get(receiver_name_id).unwrap() else {
  // If the name is not resolved, we just skip
  // There's no chance that it will get resolved at this stage and we should have already emitted a diagnostic for unresolved constant
  continue;
}
```

vinistock (approved):

alexcrocha:
### Merge activity

* **May 1, 9:29 PM UTC**: @alexcrocha merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/766).

PR #767, opened 2026-04-29
Take self type into account for completion

vinistock:
I noticed that completion for expression-like scenarios was actually wrong under a few scenarios.

The first is that we weren't taking into the correct resolution rules for the top level that we fixed in #729. For completion, we need to match the same idea. For example, a class inheriting from `BasicObject` should not see any top level constants because `Object` is not in its inheritance chain. Also, modules need the `Object` inheritance fallback to show all top level constants and the ones defined in `Kernel`, `BasicObject` and so on.

The second part is that we weren't correctly separating two concepts: the lexical scope from the type of `self`. The lexical scope _might be the type of self_, but that's not always the case. For example:

```ruby
class Foo
end

class Bar
  def Foo.baz
    # Inside this method, the lexical scope is Bar. However, the type of self
    # is Foo::<Foo> (singleton of Foo).
    #
    # Constants and class variables follow the lexical scope
    # Methods and instance variables follow the type of self
    # Globals are accessible anywhere
  end
end
```

This PR fixes expression completion to apply these rules and adds a bunch of new tests to prevent regressions.

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -13,6 +13,9 @@ static VALUE mRubydex;
 static VALUE cKeyword;
 static VALUE cKeywordParameter;
 
+// Interned once in `rdxi_initialize_graph` to avoid repeated symbol-table lookups on hot completion paths.
+static ID id_self_receiver;
+
 // Free function for the custom Graph allocator. We always have to call into Rust to free data allocated by it
 static void graph_free(void *ptr) {
     if (ptr) {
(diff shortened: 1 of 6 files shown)


vinistock:
* **#767** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/767?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Morriar (approved):

PR #769, opened 2026-04-29
Store the constant visibility target as a StringId

alexcrocha:
Constant visibility resolution (in #766) needs the target constant's name to look up the member on the resolved owner. Currently, the only way to get that name from a `ConstantVisibilityDefinition` is to walk the `name_id` through the graph (`graph.names().get(*def.name_id()).str()`), even though the indexer already had the `StringId` in hand when it built the definition.

By storing the target as a `StringId` on the definition, the resolver gets a direct handle (`owner.member(*def.target())`), and consumers that still want the full name (receiver and lexical scope) are unaffected.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1258,11 +1258,8 @@ impl<'a> RubyIndexer<'a> {
             let str_id = self.local_graph.intern_string(name);
             let offset = Offset::from_prism_location(&location);
             let definition = Definition::ConstantVisibility(Box::new(ConstantVisibilityDefinition::new(
-                self.local_graph.add_name(Name::new(
-                    str_id,
-                    receiver_name_id.map_or(ParentScope::None, ParentScope::Some),
-                    self.current_lexical_scope_name_id(),
-                )),
+                receiver_name_id,
+                str_id,
                 visibility,
                 self.uri_id,
                 offset,
--- a/rust/rubydex/src/model/definitions.rs
+++ b/rust/rubydex/src/model/definitions.rs
@@ -155,8 +155,8 @@ impl Definition {
             Definition::Module(d) => Some(d.name_id()),
             Definition::Constant(d) => Some(d.name_id()),
             Definition::ConstantAlias(d) => Some(d.name_id()),
-            Definition::ConstantVisibility(d) => Some(d.name_id()),
-            Definition::MethodVisibility(_)
+            Definition::ConstantVisibility(_)
+            | Definition::MethodVisibility(_)
             | Definition::GlobalVariable(_)
             | Definition::InstanceVariable(_)
             | Definition::ClassVariable(_)
@@ -711,7 +711,8 @@ impl ConstantAliasDefinition {
 
 #[derive(Debug)]
 pub struct ConstantVisibilityDefinition {
-    name_id: NameId,
+    receiver: Option<NameId>,
+    target: StringId,
     visibility: Visibility,
     uri_id: UriId,
     offset: Offset,
@@ -722,8 +723,10 @@ pub struct ConstantVisibilityDefinition {
 
 impl ConstantVisibilityDefinition {
     #[must_use]
+    #[allow(clippy::too_many_arguments)]
     pub const fn new(
-        name_id: NameId,
+        receiver: Option<NameId>,
+        target: StringId,
         visibility: Visibility,
         uri_id: UriId,
         offset: Offset,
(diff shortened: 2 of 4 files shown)


alexcrocha:
* **#778** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#766** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/766?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#769** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1263,6 +1263,7 @@ impl<'a> RubyIndexer<'a> {
                     receiver_name_id.map_or(ParentScope::None, ParentScope::Some),
                     self.current_lexical_scope_name_id(),
                 )),
+                str_id,
```
vinistock, line 1262: It seems that we're still using the `str_id` as part of the `add_name` operation above, which mixes the target with the receiver.

rust/rubydex/src/model/definitions.rs
```diff
@@ -712,6 +712,7 @@ impl ConstantAliasDefinition {
 #[derive(Debug)]
 pub struct ConstantVisibilityDefinition {
     name_id: NameId,
```
Morriar: Do we still need the `NameId` then?

Morriar: Actually I'm confused as to why we need both a `lexical_nesting_id` and `name_id`, can't the nesting be already represented using the `Name#nesting` field?

Morriar: Also I'm wondering if we could use a `ConstantReference` to hold the visibility's target?

alexcrocha: We still need the `lexical_nesting_id` when we don't have an explicit receiver because at that point `name_id` is `None`.

Using `ConstantReference` instead could actually work. In that case `:BAR` (in `private_constant :BAR`) becomes a real reference and it would show up when finding all references for BAR. Is that behaviour something we'd expect?

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Updated 👍

rust/rubydex/src/model/graph.rs
```diff
@@ -320,9 +317,13 @@ impl Graph {
             Definition::ConstantAlias(it) => {
                 return self.name_id_to_declaration_id(*it.name_id());
             }
-            Definition::ConstantVisibility(it) => {
-                return self.name_id_to_declaration_id(*it.name_id());
-            }
+            // TODO: alias receivers (`ALIAS.private_constant :BAR`) need resolver-side handling
```
vinistock: I'm not sure I understand this todo. Why does this method need to care about the resolve side of things?

rust/rubydex/src/model/graph.rs
```diff
@@ -320,9 +317,13 @@ impl Graph {
...
+            Definition::ConstantVisibility(it) => (
+                it.receiver()
+                    .as_ref()
+                    .or_else(|| self.find_enclosing_namespace_name_id(it.lexical_nesting_id().as_ref())),
+                it.target(),
+            ),
```
vinistock, line 325: Here we have two cases to be handled: when there's a receiver, we need to get the declaration based on its receiver name ID.

When there's none, then we need to use the lexical nesting ID (because the owner is whatever `self` is).

This is pseudo code, but I think it should be something like this:

```rust
it
  .receiver()
  .unwrap_or_else(|| {

    self.find_enclosing_namespace_name_id(it.lexical_nesting_id().as_ref()))
  })
  .and_then(|name_id| {
    self.name_id_to_declaration_id(name_id)
  })
```

rust/rubydex/src/model/graph.rs
alexcrocha: Thanks for catching. It should not have been pushed. Removed

rust/rubydex/src/model/graph.rs
vinistock: Actually, I misunderstood. The code is correct.

vinistock (approved):

alexcrocha:
### Merge activity

* **May 1, 8:30 PM UTC**: @alexcrocha merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/769).

PR #770, opened 2026-04-29
Index constant references in parameter default values

connie-ho:
### Related

- [PR #764](https://github.com/Shopify/rubydex/pull/764) — sibling fix for constant resolution inside `class << self`; same shape: data-layer fix in the indexer, not the resolver.
- Checkout Classic cleanup graph — downstream tool that surfaced the bug by mistakenly flagging `MoneyBagContainer` as unused.

### Context

Constants referenced only in a method-parameter default value are silently dropped from `graph.constant_references`. Any tool relying on `constant_references` for reachability or rename analysis treats those constants as unused.

```ruby
class Foo
  def for(price, container: MoneyBagContainer.new)
    container.presentment_money(price)
  end
end
```

Today: `PresentmentMoneyResponseFormat` and friends from the body are indexed; `MoneyBagContainer` is missing entirely.

### Solution

`visit_def_node` overrides the default ruby-prism visitor and only descends into `node.body()`. Parameter sub-nodes go to `collect_parameters`, which extracts names but never walks default-value expressions. Visiting those sub-trees in `visit_def_node` — before pushing `Nesting::Method` — feeds the missing references to the existing resolution algorithm. Visiting *outside* `Nesting::Method` matches Ruby semantics: defaults resolve against the method's enclosing lexical scope, not its local scope.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -234,6 +234,7 @@ impl<'a> RubyIndexer<'a> {
                     Offset::from_prism_location(&name_loc),
                     str_id,
                 )));
+                self.visit(&opt_param.value());
             }
 
             if let Some(rest) = parameters_list.rest() {
@@ -276,6 +277,7 @@ impl<'a> RubyIndexer<'a> {
                         let str_id = self.local_graph.intern_string(self.offset_to_string(&offset));
 
                         parameters.push(Parameter::OptionalKeyword(ParameterStruct::new(offset, str_id)));
+                        self.visit(&optional.value());
                     }
                     _ => {}
                 }
--- a/rust/rubydex/src/indexing/ruby_indexer_tests.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer_tests.rs
@@ -4061,6 +4061,19 @@ mod constant_reference_tests {
     }
 
     #[test]
+    fn index_unresolved_constant_references_in_default_values() {
+        let context = index_source({
+            "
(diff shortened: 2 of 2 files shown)


rust/rubydex/src/resolution_tests.rs
```diff
@@ -141,6 +141,102 @@ mod constant_resolution_tests {
         assert_no_members!(context, "Bar::Baz");
         assert_owner_eq!(context, "Bar::Baz", "Bar");
     }
+
+    #[test]
```
vinistock: From a resolution perspective, there's no difference between a constant reference in a default parameter value vs a constant reference anywhere else.

Instead of having resolution tests for this, we should add indexer tests to verify that we're collecting the reference initially. There's no logic related to resolution here.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1760,6 +1760,22 @@ impl Visit<'_> for RubyIndexer<'_> {
             definition_id
         };
 
+        // Default-value expressions live in the method's signature, not its body. They
+        // resolve constants against the enclosing lexical scope, so visit them here —
+        // before `Nesting::Method` is pushed for the body.
+        if let Some(parameters_list) = node.parameters() {
```
vinistock: Could we trigger this inside `collect_parameters`? That way, we avoid looping through the same set of parameters multiple times.

For a large codebase, that makes a difference in performance.

connie-ho: done and done!

vinistock (approved):

PR #771, opened 2026-04-30
Add interactive CLI mode

### LINKED ISSUE #6: Add a rake task that outputs performance stats for generating an index
This task should behave similarly to --time-index from Ruby LSP.

Stats should include:
- Time taken to generate index
- Number of each type of entry in the index
- Total size of the index in bytes

It would also be nice to have a test that we can run where many entries are generated and then accessed quickly so that our memory sharing approach across layers can be stress tested.

vinistock:
Closes #6

We weren't really doing much with our Ruby executable. I propose that we add an interactive CLI mode through IRB.

The idea is that we always index and resolve the current workspace. By default, we just print the time it takes to index and resolve. If you pass the interactive flag, then we start a custom IRB session where you can query the populated graph. It's quite useful for a quick exploration.

Usage:

```sh
bundle exec rdx -i

# Indexing workspace... finished in xxms
# Resolving graph... finished in xxms
# rubydex> graph
```

### DIFF
--- a/exe/rdx
+++ b/exe/rdx
@@ -2,11 +2,12 @@
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+
 require "optparse"
 
-OptionParser.new do |parser|
-  parser.banner = "Usage: [path1, path2]"
+options = {}
 
+OptionParser.new do |parser|
   parser.on("--version", "Print the gem's version") do
     require "rubydex/version"
     puts "v#{Rubydex::VERSION}"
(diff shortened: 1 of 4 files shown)


vinistock:
* **#771** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

exe/rdx
```diff
@@ -17,31 +18,34 @@ OptionParser.new do |parser|
...
+  print(message)
+  start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
+  block.call
+  duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
+  puts " finished in #{duration.round(2)}ms"
 end
 
-graph = Rubydex::Graph.new(workspace_path: workspace_path)
-graph.index_workspace
+graph = Rubydex::Graph.new
```
Morriar, line 37: We don't care about `workspace_path` anymore?

Morriar (approved):
It's a nice addition but I wonder if a `--stats` option would be a more straightforward way to solve #6?

exe/rdx
vinistock: It actually doesn't work well for the same reasons that make the Ruby LSP need its own composed bundle. If Rubydex itself is not in the bundle of the given workspace, then its own dependencies will not be available as part of `Bundler.setup`.

So I figured we just always defaulted to the current workspace considering people add rubydex to the bundle.

vinistock:
> It's a nice addition but I wonder if a --stats option would be a more straightforward way to solve https://github.com/Shopify/rubydex/issues/6?

We can follow up with stats. I just wanted the executable to show something useful by default when not used with the interactive flag.

The question is: what stats would a user of Rubydex want to see exactly? We may need to revisit what we include.

PR #775, opened 2026-05-01
Re-queue definitions with missing receivers

st0012:
## Issue

Incremental resolution can temporarily lose or misplace definitions whose owner depends on resolving a receiver name.

Two cases were broken:

1. A receiver is deleted and later re-added. Definitions that depend on that receiver can be processed while the receiver is missing, then skipped instead of retried.
2. A receiver remains present, but lexical lookup changes because a closer namespace is added. Incremental resolution can keep the definition on the old receiver even though a fresh index would move it.

Both make the incremental graph diverge from a fresh index: declarations can be missing after re-add, stale declarations can remain while the receiver is deleted, or singleton methods can stay attached to the wrong receiver.

Examples affected by this:

```ruby
# foo.rb
class Foo; end

# singleton.rb
class << Foo
  def bar; end
end
```

```ruby
# singleton.rb
def Foo.bar
  @x = 1
end
```

```ruby
# alias.rb
Foo.alias_method :new_name, :old_name
```

```ruby
# file1.rb
module Foo; end

module Bar
  class Qux
    def Foo.baz; end # initially Foo
  end
end

# file2.rb added later
module Bar
  module Foo; end # now Bar::Foo should win
end
```

## Fix

Re-queue definitions whose receiver or lexical owner is temporarily unresolved, so a later resolve pass can place them on the correct owner.

Also track constant receivers as receiver-specific name dependents. When lexical lookup changes, incremental invalidation now detaches and requeues receiver-dependent definitions without confusing them with definitions that define the name itself.

This also avoids falling back from a missing singleton-class lexical owner to `Object`, which could attach singleton-body declarations to the wrong namespace.

## Tests

Added regression coverage for:

- `class << Foo; def bar; end; end`
- `def Foo.bar; end`
- instance variables in explicit singleton methods
- `Foo.alias_method :new_name, :old_name`
- aliases inside singleton class bodies
- instance variables directly inside singleton class bodies
- receiver re-resolution when a closer lexical namespace is added

Verified locally:

```sh
cargo test -p rubydex
```

The full `rubydex` crate test suite passes locally.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -645,7 +645,7 @@ impl Graph {
         std::mem::take(&mut self.pending_work)
     }
 
-    fn push_work(&mut self, unit: Unit) {
+    pub(crate) fn push_work(&mut self, unit: Unit) {
         self.pending_work.push(unit);
     }
 
@@ -2398,6 +2398,34 @@ mod incremental_resolution_tests {
         }
     }
 
+    /// Compares incremental resolution against a fresh index at the declaration-ID level.
+    ///
+    /// This is a broad consistency check: it catches both stale declarations left
+    /// behind by incremental invalidation and declarations that incremental
+    /// resolution failed to recreate.
+    fn assert_declaration_ids_match(incremental: &GraphTest, fresh: &GraphTest) {
+        let extras: Vec<_> = incremental
+            .graph()
+            .declarations()
+            .iter()
+            .filter(|(id, _)| !fresh.graph().declarations().contains_key(id))
+            .map(|(_, d)| format!("{} ({})", d.name(), d.kind()))
+            .collect();
+
+        let missing: Vec<_> = fresh
+            .graph()
+            .declarations()
+            .iter()
+            .filter(|(id, _)| !incremental.graph().declarations().contains_key(id))
+            .map(|(_, d)| format!("{} ({})", d.name(), d.kind()))
+            .collect();
+
+        assert!(
+            extras.is_empty() && missing.is_empty(),
+            "Declaration mismatch:\n  Extra: {extras:?}\n  Missing: {missing:?}"
+        );
+    }
+
     #[test]
     fn new_namespace_shadowing_include_target_invalidates_references() {
         let mut context = GraphTest::new();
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -2398,6 +2398,34 @@ mod incremental_resolution_tests {
         }
     }
 
+    /// Compares incremental resolution against a fresh index at the declaration-ID level.
+    ///
+    /// This is a broad consistency check: it catches both stale declarations left
+    /// behind by incremental invalidation and declarations that incremental
+    /// resolution failed to recreate.
+    fn assert_declaration_ids_match(incremental: &GraphTest, fresh: &GraphTest) {
```
vinistock, line 2406: Nit: should this be a macro instead?

rust/rubydex/src/resolution.rs
```diff
@@ -688,6 +711,21 @@ impl<'a> Resolver<'a> {
         }
     }
 
+    /// Wraps `resolve_lexical_owner` for `handle_remaining_definitions` callers:
+    /// when the owner can't be resolved, re-queue the definition so the next
+    /// resolve can place it on the right owner instead of dropping the work.
+    fn resolve_lexical_owner_or_requeue(
```
vinistock: I agree with the solution: receivers we failed to process have to be re-enqueued so that in the next incremental iteration they may get resolved properly.

However, shouldn't that just be the default behaviour? What I mean is: do we need to create separate `_or_requeue` methods? I think we should just add the logic to re-enqueue directly on the methods.

rust/rubydex/src/model/graph.rs
```diff
@@ -3772,6 +3800,162 @@ mod incremental_resolution_tests {
         assert_declaration_exists!(context, "Parent::Target::<Target>");
     }
 
+    #[test]
```
vinistock, line 3803: Scenario we may have missed: do we already have a test for a receiver that gets invalidated due to resolution changes?

For example:

```ruby
# file1.rb
module Foo
end

module Bar
  class Qux
    # Initial state: Foo resolves to top
    # level Foo
    def Foo.baz
    end
  end
end

# add file2.rb
# Now `Foo` should resolve to `Bar::Foo`
module Bar
  module Foo; end
end
```

rust/rubydex/src/model/graph.rs
st0012: Why is a macro better for this case?

rust/rubydex/src/resolution.rs
st0012: Updated 👍

rust/rubydex/src/model/graph.rs
st0012: We don't cover this case and it's not working indeed. But this IMO should be addressed in a separate PR as it's not really a missing receiver case. I'll open an issue to track it.

st0012: https://github.com/Shopify/rubydex/issues/777

rust/rubydex/src/model/graph.rs
vinistock: It's not necessarily better, it's just that most `assert_` things are always macros (e.g.: `assert!`, `assert_eq!`...).

rust/rubydex/src/resolution.rs
```diff
@@ -640,6 +648,19 @@ impl<'a> Resolver<'a> {
...
     }
 
+    /// Resolves a constant receiver for `handle_remaining_definitions`.
+    /// If the receiver name is unresolved, preserve the definition for a later
+    /// resolve cycle instead of dropping work during an incremental delete/re-add gap.
+    fn resolve_constant_receiver(&mut self, name_id: NameId, id: DefinitionId) -> Option<DeclarationId> {
+        match self.graph.names().get(&name_id).unwrap() {
+            NameRef::Resolved(resolved) => Some(*resolved.declaration_id()),
+            NameRef::Unresolved(_) => {
+                self.graph.push_work(Unit::Definition(id));
```
vinistock, line 658: One thing to consider, is asserting at the end of incremental resolution tests if  `pending_work` has any duplicates. It might become hard to guarantee that we're not duplicating work without assertions.

vinistock (approved):

PR #776, opened 2026-05-01
Fix singleton class lifecycle cleanup

st0012:
## Summary
Fixes singleton class lifecycle cleanup during document deletion and re-indexing. Singleton classes are now pruned only when they are no longer anchored by definitions, members, constant references, or surviving `extend` mixins on their attached object.

The cleanup path is split into focused steps for method references, constant references, unmapped definitions, raw definition entries, and declaration pruning. Removability is checked through the graph so singleton classes can account for external anchors, while TODO parents are not kept alive by unresolved references alone.

The invalidation worklist now keeps declaration lifecycle modes under `InvalidationItem::Declaration { id, reason }`, so declaration-specific invalidation reasons stay localized instead of widening the top-level work item enum.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -29,14 +29,51 @@ pub enum NameDependent {
     NestedName(NameId),
 }
 
+/// Declaration-specific modes processed by the unified invalidation worklist.
+enum DeclarationInvalidationReason {
+    /// Definitions or structural ownership changed. The declaration may be
+    /// removed, or it may survive and need ancestor relinearization.
+    Changed,
+    /// Ancestor chain is stale, but the declaration should not be pruned just
+    /// because it is currently waiting for new document data to be merged.
+    RecomputeAncestors,
+    /// A local anchor was detached. Remove only if no graph-level anchors remain.
+    PruneIfUnanchored,
+}
+
 /// Items processed by the unified invalidation worklist.
 enum InvalidationItem {
-    /// Ancestor chain is stale, or declaration has become empty and needs removal.
-    Declaration(DeclarationId),
+    Declaration {
+        id: DeclarationId,
+        reason: DeclarationInvalidationReason,
+    },
     /// Structural dependency broken — unresolve the name and cascade to all dependents.
     Name(NameId),
     /// Ancestor context changed — unresolve references under this name but keep the name resolved.
     References(NameId),
+}
+
+impl InvalidationItem {
+    fn declaration_changed(id: DeclarationId) -> Self {
+        Self::Declaration {
+            id,
+            reason: DeclarationInvalidationReason::Changed,
+        }
+    }
+
+    fn declaration_recompute_ancestors(id: DeclarationId) -> Self {
+        Self::Declaration {
+            id,
+            reason: DeclarationInvalidationReason::RecomputeAncestors,
+        }
+    }
+
+    fn declaration_prune_if_unanchored(id: DeclarationId) -> Self {
+        Self::Declaration {
+            id,
+            reason: DeclarationInvalidationReason::PruneIfUnanchored,
+        }
+    }
 }
 
 /// A work item produced by graph mutations (update/delete) that needs resolution.
@@ -984,7 +1021,7 @@ impl Graph {
     /// Identifies declarations affected by old/new documents and feeds them into `invalidate_graph`.
     ///
     /// Does NOT mutate declarations or remove raw data — definition detachment is deferred to
-    /// `invalidate_declaration`, and raw data cleanup to `remove_document_data`.
+    /// `handle_declaration_change`, and raw data cleanup to `remove_document_data`.
     fn invalidate(&mut self, old_document: Option<&Document>, new_local_graph: Option<&LocalGraph>) {
         let capacity = old_document.map_or(0, |d| d.definitions().len())
             + new_local_graph.map_or(0, |lg| lg.definitions().len() + lg.constant_references().len());
@@ -993,13 +1030,29 @@ impl Graph {
 
         // Identify declarations affected by removed definitions
         if let Some(document) = old_document {
+            let mut attached_singletons = Vec::new();
+
             for def_id in document.definitions() {
                 if let Some(declaration_id) = self.definition_id_to_declaration_id(*def_id).copied() {
                     pending_detachments.entry(declaration_id).or_default().push(*def_id);
+
+                    if let Some(singleton_id) = self
+                        .declarations
+                        .get(&declaration_id)
+                        .and_then(Declaration::as_namespace)
+                        .and_then(Namespace::singleton_class)
+                    {
+                        attached_singletons.push(*singleton_id);
+                    }
                 }
             }
+
+            for singleton_id in attached_singletons {
+                items.push(InvalidationItem::declaration_changed(singleton_id));
+            }
+
             for decl_id in pending_detachments.keys() {
-                items.push(InvalidationItem::Declaration(*decl_id));
+                items.push(InvalidationItem::declaration_changed(*decl_id));
             }
         }
 
@@ -1009,7 +1062,9 @@ impl Graph {
                 if let Some(name_id) = def.name_id()
                     && let Some(NameRef::Resolved(resolved)) = self.names.get(name_id)
                 {
-                    items.push(InvalidationItem::Declaration(*resolved.declaration_id()));
+                    items.push(InvalidationItem::declaration_recompute_ancestors(
+                        *resolved.declaration_id(),
+                    ));
                 }
             }
 
@@ -1024,7 +1079,9 @@ impl Graph {
                     && let Some(nesting_id) = name_ref.nesting()
                     && let Some(NameRef::Resolved(resolved)) = self.names.get(nesting_id)
                 {
-                    items.push(InvalidationItem::Declaration(*resolved.declaration_id()));
+                    items.push(InvalidationItem::declaration_recompute_ancestors(
+                        *resolved.declaration_id(),
+                    ));
                 }
             }
         }
@@ -1035,22 +1092,40 @@ impl Graph {
     }
 
     /// Removes raw document data (refs, defs, names, strings) from maps.
-    /// Does not touch declarations or perform invalidation -- that is handled by `invalidate`.
+    ///
+    /// `invalidate` has already handled dependency invalidation and the normal
+    /// definition detachment path. This cleanup removes the raw map entries and
+    /// prunes declarations that become removable only after those entries are detached.
     fn remove_document_data(&mut self, document: &Document) {
+        self.remove_document_method_references(document);
+
+        let mut declarations_to_prune = self.remove_document_constant_references(document);
+        declarations_to_prune.extend(self.detach_unmapped_document_definitions(document));
+        self.prune_declarations_if_unanchored(declarations_to_prune);
+
+        self.remove_document_definitions(document);
+    }
+
+    fn remove_document_method_references(&mut self, document: &Document) {
         for ref_id in document.method_references() {
             if let Some(method_ref) = self.method_references.remove(ref_id) {
                 self.untrack_string(*method_ref.str());
             }
         }
+    }
 
+    fn remove_document_constant_references(&mut self, document: &Document) -> Vec<DeclarationId> {
+        let mut declarations_to_prune = Vec::new();
         for ref_id in document.constant_references() {
             if let Some(constant_ref) = self.constant_references.remove(ref_id) {
                 // Detach from target declaration. References unresolved during invalidation
                 // were already detached; this catches the rest.
-                if let NameRef::Resolved(resolved) = self.names.get(constant_ref.name_id()).unwrap()
-                    && let Some(declaration) = self.declarations.get_mut(resolved.declaration_id())
+                let target_id = self.name_id_to_declaration_id(*constant_ref.name_id()).copied();
+                if let Some(target_id) = target_id
+                    && let Some(declaration) = self.declarations.get_mut(&target_id)
                 {
                     declaration.remove_constant_reference(ref_id);
+                    declarations_to_prune.push(target_id);
                 }
 
                 self.remove_name_dependent(*constant_ref.name_id(), NameDependent::Reference(*ref_id));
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -1035,52 +1035,115 @@ impl Graph {
...
     /// Removes raw document data (refs, defs, names, strings) from maps.
-    /// Does not touch declarations or perform invalidation -- that is handled by `invalidate`.
+    ///
+    /// `invalidate` has already handled dependency invalidation and the normal
+    /// definition detachment path. This cleanup removes the raw map entries and
+    /// prunes declarations that become removable only after those entries are detached.
     fn remove_document_data(&mut self, document: &Document) {
+        self.remove_document_method_references(document);
+
+        let mut declarations_to_prune = self.remove_document_constant_references(document);
```
st0012, line 1102: 🐛 **Bug**: Consider reviewing this cleanup path for surviving namespaces whose deleted or reindexed definition contributed an `extend` mixin. `remove_document_data` currently builds prune candidates from removed constant references and unmapped definitions, but mapped class/module definitions are detached through the normal invalidation path. If another file still defines `Foo` and the removed file only contributed `class Foo; extend Bar; end`, the attached singleton `Foo::<Foo>` may no longer be needed, yet it is never queued for pruning. The same gap can leave stale singleton ancestors when an existing `extend Bar` changes to `extend Baz`, because the owner is relinearized but its singleton may not be reconsidered.

**Suggestion:** When cleaning up a document, also collect existing singleton class declarations attached to mapped namespace definitions that contributed an `extend` mixin, and add those singleton declaration IDs to `declarations_to_prune` before calling `prune_removable_declarations`. A focused regression would be: `base.rb` contains `class Foo; end`, `bar.rb` contains `module Bar; end`, and `ext.rb` contains `class Foo; extend Bar; end`; after deleting or replacing `ext.rb` with no `extend`, `Foo::<Foo>` should be removed if it has no other anchors and `Foo.singleton_class_id` should be cleared.

rust/rubydex/src/model/graph.rs
```diff
@@ -1035,52 +1035,115 @@ impl Graph {
...
+    }
+
+    fn detach_constant_reference_from_resolved_target(
+        &mut self,
+        name_id: NameId,
+        ref_id: ConstantReferenceId,
+    ) -> Option<DeclarationId> {
+        let target_id = match self.names.get(&name_id) {
+            Some(NameRef::Resolved(resolved)) => *resolved.declaration_id(),
+            Some(NameRef::Unresolved(_)) | None => return None,
```
st0012: 🐛 **Bug**: Worth checking the case where invalidation has already detached a removed document's constant reference before raw document cleanup runs. This cleanup only returns a pruning candidate when the removed reference name is still `Resolved`, but `consume_document_changes` and `delete_document` run invalidation first. If invalidation structurally unresolves the attached singleton reference used by a call like `Foo.new`, this line can return `None`, leaving the now-empty singleton out of `prune_removable_declarations` and allowing `Foo.singleton_class_id` to diverge from a fresh graph.

**Suggestion:** Preserve pruning candidates for references detached during invalidation, not only references whose `NameRef` remains resolved during raw cleanup. One option is to have the unresolve paths that remove constant references return or enqueue the old declaration ID for `prune_removable_declarations` when the reference belongs to the document being removed. Add a regression with a surviving `class Foo; end` file and a deleted file containing both a `Foo` reopener and `Foo.new`, then assert `Foo::<Foo>` is removed and `Foo.singleton_class_id` is cleared after deletion.

rust/rubydex/src/model/graph.rs
```diff
@@ -1280,7 +1393,7 @@ impl Graph {
                         if self
                             .declarations
                             .get(&old_decl_id)
-                            .is_some_and(Declaration::has_no_definitions)
+                            .is_some_and(Declaration::is_removable)
```
st0012: 🐛 **Bug**: This path still calls the declaration-local `is_removable` predicate even though the PR introduces `declaration_is_removable` to account for graph-level external anchors. For singleton classes, the local predicate can disagree with the graph-level rule because a singleton may need to survive due to an `extend` mixin on its attached object. If a dependent definition is detached through this structural invalidation path, the singleton can be queued for removal even though the graph-level predicate would keep it alive.

**Suggestion:** Consider using the graph-level predicate here so all pruning decisions apply the same singleton keepalive rule:

```suggestion
                            .is_some_and(|_| self.declaration_is_removable(old_decl_id))
```

rust/rubydex/src/model/graph.rs
```diff
@@ -1213,6 +1346,10 @@ impl Graph {
                         ns.remove_member(&unqualified_str_id);
                     }
                 }
+
+                if self.declaration_is_removable(owner_id) {
+                    queue.push(InvalidationItem::Declaration(owner_id));
+                }
             }
 
             self.declarations.remove(&decl_id);
```
st0012: 🐛 **Bug**: Worth checking the remove branch for namespace declarations because it deletes the declaration without detaching it from its old ancestors' `descendants` sets. The update branch already performs this reverse-index cleanup before clearing ancestors, which suggests removal needs the same treatment. With the new singleton cleanup behavior, replacing `extend Bar` with `extend Baz` could remove and recreate `Foo::<Foo>` while `Bar.descendants()` still contains the old singleton id, causing later invalidation from `Bar` to traverse a stale descendant relationship.

**Suggestion:** Before `self.declarations.remove(&decl_id)`, mirror the update-path cleanup for namespace declarations: walk the declaration's current ancestors and remove `decl_id` from each ancestor namespace's descendants set. Add coverage asserting that the old extend target no longer lists the singleton class as a descendant after the replacement.

rust/rubydex/src/model/graph.rs
```diff
@@ -1092,6 +1179,52 @@ impl Graph {
...
     }
 
+    fn prune_removable_declarations(&mut self, candidate_declaration_ids: Vec<DeclarationId>) {
+        if candidate_declaration_ids.is_empty() {
+            return;
+        }
+
+        let prune_items: Vec<_> = candidate_declaration_ids
+            .into_iter()
+            .filter(|id| self.declaration_is_removable(*id))
```
st0012: 🐛 **Bug**: This filtering step appears to drop singleton candidates unless they are currently removable. That is risky because a singleton class can survive due to methods or references while still needing its old `extend`-derived ancestors detached and relinearized after the attached object's definitions change. If it is filtered out here, it may never reach the update path that clears old ancestor relationships.

**Suggestion:** Consider separating prune-only candidates from singleton invalidation candidates. Singleton IDs collected because their attached object's definitions changed should still reach `invalidate_graph` even when they are not removable, so the update branch can detach old ancestors, clear descendant links, and enqueue ancestor relinearization.

rust/rubydex/src/model/graph.rs
```diff
@@ -1213,8 +1346,13 @@ impl Graph {
                         ns.remove_member(&unqualified_str_id);
                     }
                 }
+
+                if self.declaration_is_removable(owner_id) {
```
st0012: 🐛 **Bug**: Worth reviewing this owner-pruning path carefully: it reuses the same removability predicate for every owner, but the graph can contain synthetic namespace declarations that intentionally have no definitions. A Todo namespace can still be live because it owns members from surviving documents, so pruning it after one child is removed can cascade through its members and delete declarations that should remain indexed.

**Suggestion:** Avoid using the singleton cleanup predicate as a general owner-removal predicate. Either scope this owner-prune to singleton namespaces only, or make the graph-level removability check require namespace anchors such as definitions, members, references, and singleton links to be empty before removing synthetic namespaces. An incremental test with sibling Todo-owned declarations would make this concrete, for example deleting one of `A::B::C` / `A::B::D` should not remove the shared `A::B` Todo owner or the surviving sibling.

rust/rubydex/src/model/graph.rs
```diff
@@ -1255,6 +1390,34 @@ impl Graph {
...
+            })
+            .unwrap_or_default();
+
+        for ancestor_id in ancestor_ids {
+            if let Some(ancestor_namespace) = self
+                .declarations
+                .get_mut(&ancestor_id)
+                .and_then(Declaration::as_namespace_mut)
+            {
+                ancestor_namespace.remove_descendant(&decl_id);
```
st0012, line 1520: 🐛 **Bug**: This helper removes the current declaration from each ancestor's descendant set, but the detached ancestor is not returned or queued for a follow-up removability check. That can leave an ancestor singleton attached even after the last descendant relationship that made it useful has been removed.

**Suggestion:** Consider having `detach_declaration_from_ancestors` return the detached ancestor IDs, or give it access to the invalidation queue. After `remove_descendant`, enqueue any detached ancestor that now satisfies the appropriate removability check so singleton lifecycle cleanup remains complete when ancestry changes remove the last live relationship.

rust/rubydex/src/model/graph.rs
```diff
@@ -29,16 +29,53 @@ pub enum NameDependent {
...
 
+/// Declaration-specific modes processed by the unified invalidation worklist.
+enum DeclarationInvalidationReason {
+    /// Definitions or structural ownership changed. The declaration may be
+    /// removed, or it may survive and need ancestor relinearization.
+    Changed,
+    /// Ancestor chain is stale, but the declaration should not be pruned just
+    /// because it is currently waiting for new document data to be merged.
+    RecomputeAncestors,
+    /// A local anchor was detached. Remove only if no graph-level anchors remain.
```
vinistock, line 40: I think it's worth expanding on what this means. I don't know what a local anchor is supposed to be.

rust/rubydex/src/model/graph.rs
```diff
@@ -29,16 +29,53 @@ pub enum NameDependent {
     NestedName(NameId),
 }
 
+/// Declaration-specific modes processed by the unified invalidation worklist.
+enum DeclarationInvalidationReason {
+    /// Definitions or structural ownership changed. The declaration may be
+    /// removed, or it may survive and need ancestor relinearization.
```
vinistock, line 35: Does the `need ancestor relinearization` belong here? If it needs re-linearization, wouldn't the reason be `RecomputeAncestors`?

More generally, I think `DeclarationInvalidationReason` can be better documented with examples as it's not immediately obvious how this is used and what triggers each scenario.

rust/rubydex/src/model/graph.rs
```diff
@@ -993,13 +1030,29 @@ impl Graph {
...
+                        .and_then(Namespace::singleton_class)
+                    {
+                        attached_singletons.push(*singleton_id);
+                    }
                 }
             }
+
+            for singleton_id in attached_singletons {
+                items.push(InvalidationItem::declaration_changed(singleton_id));
+            }
```
vinistock, line 1052: Can't we push to `items` directly above? Why do we need to accumulate in a vector and then loop to push into `items`?

Also, `items.extend(attached_singletons)` is more efficient.

rust/rubydex/src/model/graph.rs
```diff
@@ -1035,52 +1092,98 @@ impl Graph {
...
                 // Detach from target declaration. References unresolved during invalidation
                 // were already detached; this catches the rest.
-                if let NameRef::Resolved(resolved) = self.names.get(constant_ref.name_id()).unwrap()
-                    && let Some(declaration) = self.declarations.get_mut(resolved.declaration_id())
+                let target_id = self.name_id_to_declaration_id(*constant_ref.name_id()).copied();
+                if let Some(target_id) = target_id
+                    && let Some(declaration) = self.declarations.get_mut(&target_id)
                 {
                     declaration.remove_constant_reference(ref_id);
+                    declarations_to_prune.push(target_id);
```
vinistock, line 1128: If I'm following the logic correctly, this means that we're enqueueing all declarations for possible pruning. Is that understanding correct?

This is a inefficient because it only really applies to singleton classes.

Consider this for example:

```ruby
class Foo; end

Foo
```

For `Foo`, it does not matter if the reference is removed. The declaration can never be pruned unless the definition is removed. Only `Foo::<Foo>` might possibly get pruned by the removal of constant references.

rust/rubydex/src/model/graph.rs
```diff
@@ -1035,52 +1092,98 @@ impl Graph {
...
-            .collect();
+            .collect()
+    }
 
-        if !missed_def_ids.is_empty() {
-            for declaration in self.declarations.values_mut() {
-                for def_id in &missed_def_ids {
-                    declaration.remove_definition(def_id);
-                }
+    fn detach_definitions_from_any_declaration(
```
vinistock, line 1163: Just to confirm my understanding here: it seems that we're taking orphan definitions and looping through them to try to remove their ID from being registered in any declaration. In what scenario can this happen?

If a definition ID got registered in a declaration at any point, it means we managed to connect them correctly. If an incremental change comes it that would make the definition orphan, then I would expect that iteration to follow `definition_id_to_declaration_id` and successfully remove itself from the declaration.

So I'm not sure what conditions can lead to:

1. We correctly register a definition ID in a declaration
2. The definition becomes orphan, but we haven't removed the ID from the declaration
3. Now we need to loop through all of them to cleanup

This feels like a scenario that shouldn't ever happen. It might be a bug elsewhere in the invalidation algorithm.

rust/rubydex/src/model/graph.rs
```diff
@@ -1092,6 +1195,67 @@ impl Graph {
...
+            // an attached singleton class, and — for singleton classes — an
+            // `extend` on the attached object (handled by the early return above).
+            // TODO parents are not kept alive by references alone.
+            return namespace.definitions().is_empty()
+                && namespace.members().is_empty()
+                && !has_reference_anchor
+                && namespace.singleton_class().is_none();
+        }
+
+        declaration.has_no_definitions()
```
vinistock, line 1242: This might be a better fit for a method in `Declaration`. Something like `declaration.is_unanchored()`.

rust/rubydex/src/model/graph.rs
```diff
@@ -1161,98 +1334,194 @@ impl Graph {
...
+            for def_id in def_ids {
+                self.push_work(Unit::Definition(def_id));
             }
+        }
+    }
 
-            let Some(namespace) = self.declarations.get_mut(&decl_id).and_then(|d| d.as_namespace_mut()) else {
-                return;
-            };
+    fn detach_edges_and_queue_prune(&mut self, decl_id: DeclarationId, queue: &mut Vec<InvalidationItem>) {
```
vinistock, line 1433: What is an `edge` here?

PR #779, opened 2026-05-05
Add `follow_method_alias` query

vinistock:
This PR extracts a portion of #697

The PR implements `follow_method_alias`, which accepts a `definition_id` for an alias and follows the chain until reaching the first concrete method definition. This is required to expose `MethodDefinition#signatures` in the Ruby API (as the definition may be an alias).

This PR also moves the core logic of finding members in the ancestor chain to `query.rs`, so that we can reuse it for this implementation and for the FFI.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -380,13 +380,15 @@ impl Graph {
                 )
             }
             Definition::MethodAlias(it) => {
-                if let Some(Receiver::SelfReceiver(def_id)) = it.receiver() {
-                    return self.find_self_receiver_declaration(*def_id, *it.new_name_str_id());
-                }
-                (
-                    self.find_enclosing_namespace_name_id(it.lexical_nesting_id().as_ref()),
-                    it.new_name_str_id(),
-                )
+                let nesting_name_id = match it.receiver() {
+                    Some(Receiver::SelfReceiver(def_id)) => {
+                        return self.find_self_receiver_declaration(*def_id, *it.new_name_str_id());
+                    }
+                    Some(Receiver::ConstantReceiver(name_id)) => Some(name_id),
+                    None => self.find_enclosing_namespace_name_id(it.lexical_nesting_id().as_ref()),
+                };
+
+                (nesting_name_id, it.new_name_str_id())
             }
         };
 
--- a/rust/rubydex/src/query.rs
+++ b/rust/rubydex/src/query.rs
@@ -10,7 +10,7 @@ use crate::model::declaration::{Ancestor, Declaration, Namespace};
 use crate::model::definitions::{Definition, Parameter};
 use crate::model::graph::Graph;
 use crate::model::identity_maps::IdentityHashSet;
-use crate::model::ids::{DeclarationId, NameId, StringId, UriId};
+use crate::model::ids::{DeclarationId, DefinitionId, NameId, StringId, UriId};
 use crate::model::keywords::{self, Keyword};
 use crate::model::name::NameRef;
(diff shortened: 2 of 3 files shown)


vinistock:
* **#779** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/779?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/779?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/model/graph.rs
```diff
@@ -375,13 +375,15 @@ impl Graph {
...
-                }
-                (
-                    self.find_enclosing_namespace_name_id(it.lexical_nesting_id().as_ref()),
-                    it.new_name_str_id(),
-                )
+                let nesting_name_id = match it.receiver() {
+                    Some(Receiver::SelfReceiver(def_id)) => {
+                        return self.find_self_receiver_declaration(*def_id, *it.new_name_str_id());
+                    }
+                    Some(Receiver::ConstantReceiver(name_id)) => Some(name_id),
```
vinistock, line 387: We were not handling constant receivers (e.g.: `Foo.alias_method :bar, :baz`), which ended up being surfaced in the new tests.

rust/rubydex/src/query.rs
```diff
@@ -610,6 +610,123 @@ fn method_argument_completion<'a>(
...
+            if *ancestor_id == declaration_id {
+                found_main_namespace = true;
+            }
+            continue;
+        }
+
+        if let Some(member_id) = graph
+            .declarations()
+            .get(ancestor_id)?
+            .as_namespace()?
```
Morriar: The `?` operators here would exit the entire function if an ancestor doesn't exist in the graph or isn't a namespace. The old code in the FFI used `.unwrap()` (would panic), but the more important difference is that `?` aborts the search entirely rather than skipping that ancestor and continuing to the next one.

In practice this probably never fires (ancestors should always be valid namespaces), but if it did, a member that exists on a later ancestor would be missed. Should this use `and_then` instead to skip over bad ancestors?

```suggestion
        let member = graph
            .declarations()
            .get(ancestor_id)
            .and_then(|decl| decl.as_namespace())
            .and_then(|ns| ns.member(&member_str_id));

        if let Some(member_id) = member {
            return Some(*member_id);
        }
```

rust/rubydex/src/query.rs
```diff
@@ -610,6 +610,123 @@ fn method_argument_completion<'a>(
     Ok(candidates)
 }
 
+/// Finds the given member on the ancestor chain of the declaration. Use `only_inherited` to skip all ancestors until
+/// the main namespace and start from its parent.
+#[must_use]
+pub fn find_member_in_ancestors(
```
soutaro, line 699: How about introducing a new error enum for this function?

It would be nice to distinguish the valid “member not found” case from other errors, which shouldn’t happen unless there is something wrong with the graph.

vinistock: Yes, I agree. We've not been really good at doing this (returning better error reasons and using `Result` objects). I switched to return a `Result<Declaration, Error>` with multiple reasons.

For complete graph inconsistencies, like an ancestor ID not pointing to a module or class, we are choosing to panic so that the bug is exposed as quickly as possible. Corrupt data can lead to very difficult to trace bugs, so it's better to just outright crash and fix than to attempt a soft fail.

soutaro (approved):
👍

PR #780, opened 2026-05-05
Index private_class_method / public_class_method calls

alexcrocha:
Part of #89 

This PR adds indexer support for `private_class_method` and `public_class_method` calls. Each call becomes a `MethodVisibilityDefinition` with the `SINGLETON_METHOD_VISIBILITY` flag set, which the resolver applies to the singleton method declaration in a follow-up PR. This is the base of a Graphite stack.

We need a definition because the indexer can't yet attach the visibility change to the singleton method:

```ruby
class Foo
  def self.bar; end
  private_class_method :bar
end
```

The method declaration for `Foo::<Foo>#bar()` only exists after resolution.

### Why a bit flag

I added a `DefinitionFlags::SINGLETON_METHOD_VISIBILITY` bit on the existing `MethodVisibilityDefinition` rather than introducing a parallel `SingletonMethodVisibilityDefinition` type. The two would have carried the same fields and the indexer treats instance and singleton visibility calls almost identically. A bit flag captures the singleton-vs-instance distinction without doubling the definition variants and the FFI plumbing that comes with them.

### Why no explicit receiver support

`Foo.private_class_method :bar` is legal Ruby but rare in practice. Supporting it would require a new definition type to carry the receiver. Ignoring it lets us fold the singleton case into the existing `MethodVisibilityDefinition` and get resolution-time differentiation from a single bit flag, which keeps the blast radius of the change small. Calls with an explicit receiver bypass the visibility logic, though the call is still visited so nested references stay indexed.

### In this PR

- add `DefinitionFlags::SINGLETON_METHOD_VISIBILITY`
- index `private_class_method` / `public_class_method` calls with literal symbol or string targets, with an implicit receiver only
- skip singleton-flagged definitions from `resolve_method_visibilities` so the new flag stays inert until the follow-up PR wires up resolution
- diagnose top-level calls and non-literal arguments

### Not in this PR

A few things are intentionally deferred to follow-up PRs in the stack:

- applying the visibility to the singleton method declaration
- extracting the literal-extraction helper we duplicate from `handle_constant_visibility`
- supporting array arguments (`private_class_method [:a, :b]`)
- supporting inline-def arguments (`private_class_method def self.foo; end`)

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1200,9 +1200,7 @@ impl<'a> RubyIndexer<'a> {
             }
             Some(ruby_prism::Node::SelfNode { .. }) | None => match self.nesting_stack.last() {
                 Some(Nesting::Method(_)) => {
-                    // Dynamic private constant (called from a method), we ignore it but don't report an error since it's valid Ruby
-                    // if being called from a singleton method.
-
+                    self.visit_call_node_parts(node);
                     return;
                 }
                 None => {
@@ -1211,7 +1209,7 @@ impl<'a> RubyIndexer<'a> {
                         Offset::from_prism_location(&node.location()),
                         "Private constant called at top level".to_string(),
                     );
-
+                    self.visit_call_node_parts(node);
                     return;
                 }
                 _ => None,
@@ -1222,7 +1220,7 @@ impl<'a> RubyIndexer<'a> {
                     Offset::from_prism_location(&node.location()),
                     "Dynamic receiver for private constant".to_string(),
                 );
-
+                self.visit_call_node_parts(node);
                 return;
             }
         };
@@ -1252,8 +1250,8 @@ impl<'a> RubyIndexer<'a> {
                         Offset::from_prism_location(&argument.location()),
                         "Private constant called with non-symbol argument".to_string(),
                     );
-
-                    return;
+                    self.visit(&argument);
+                    continue;
                 }
             };
 
@@ -1273,6 +1271,60 @@ impl<'a> RubyIndexer<'a> {
             let definition_id = self.local_graph.add_definition(definition);
 
             self.add_member_to_current_owner(definition_id);
+        }
+    }
+
+    fn handle_singleton_method_visibility(
+        &mut self,
+        node: &ruby_prism::CallNode,
+        visibility: Visibility,
+        call_name: &str,
+    ) {
+        match node.receiver() {
+            Some(ruby_prism::Node::SelfNode { .. }) | None => match self.nesting_stack.last() {
+                Some(Nesting::Method(_)) => {
+                    self.visit_call_node_parts(node);
+                    return;
+                }
+                None => {
+                    self.local_graph.add_diagnostic(
+                        Rule::InvalidMethodVisibility,
+                        Offset::from_prism_location(&node.location()),
+                        format!("`{call_name}` called at top level"),
+                    );
+                    self.visit_call_node_parts(node);
+                    return;
+                }
+                _ => {}
+            },
+            _ => {
+                self.visit_call_node_parts(node);
+                return;
+            }
+        }
+
+        let Some(arguments) = node.arguments() else {
+            return;
+        };
+
+        for argument in &arguments.arguments() {
+            match argument {
+                ruby_prism::Node::SymbolNode { .. } | ruby_prism::Node::StringNode { .. } => {
+                    self.create_method_visibility_definition(
+                        &argument,
+                        visibility,
+                        DefinitionFlags::SINGLETON_METHOD_VISIBILITY,
+                    );
+                }
+                _ => {
+                    self.local_graph.add_diagnostic(
+                        Rule::InvalidMethodVisibility,
+                        Offset::from_prism_location(&argument.location()),
+                        format!("`{call_name}` called with a non-literal argument"),
+                    );
+                    self.visit(&argument);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


alexcrocha:
* **#780** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/780?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1276,6 +1276,57 @@ impl<'a> RubyIndexer<'a> {
...
     }
 
+    fn handle_singleton_method_visibility(
+        &mut self,
+        node: &ruby_prism::CallNode,
+        visibility: Visibility,
+        call_name: &str,
+    ) {
+        match node.receiver() {
+            Some(ruby_prism::Node::SelfNode { .. }) | None => match self.nesting_stack.last() {
```
st0012, line 1284: Should the branches under this condition call `visit_call_node_parts` before returning/continuing?

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1276,6 +1276,57 @@ impl<'a> RubyIndexer<'a> {
...
+                    );
+                }
+                _ => {
+                    self.local_graph.add_diagnostic(
+                        Rule::InvalidMethodVisibility,
+                        Offset::from_prism_location(&argument.location()),
+                        format!("`{call_name}` called with a non-literal argument"),
+                    );
+                    self.visit(&argument);
+                    return;
```
st0012: Since we're iterating through arguments here, I think we should `continue` instead of `return` here.

rust/rubydex/src/indexing/ruby_indexer_tests.rs
```diff
@@ -3735,6 +3735,89 @@ mod visibility_tests {
...
+            assert_string_eq!(&context, def.str_id(), "bar()");
+            assert_eq!(def.visibility(), &Visibility::Public);
+        });
+    }
+
+    #[test]
+    fn index_private_class_method_calls_diagnostics() {
+        let context = index_source(
+            r"
+            private_class_method :NOT_INDEXED
```
st0012, line 3799: What happens when we have `private_class_method :NOT_INDEXED, :indexed` and `private_class_method :indexed_1, ignored, :indexed_2`?
Will `:indexed` and `:indexed_2` be dropped? Would that be expected?

rust/rubydex/src/model/definitions.rs
```diff
@@ -40,6 +40,7 @@ bitflags! {
     pub struct DefinitionFlags: u8 {
         const DEPRECATED = 0b0001;
         const PROMOTABLE = 0b0010;
+        const SINGLETON_METHOD_VISIBILITY = 0b0100;
```
vinistock, line 43: Note for the future: we're using `DefinitionFlags` for all definitions, but the reality is that most flags are only applicable to certain types. For example, deprecated doesn't make much sense for these visibility related definitions, promotable is only applicable to constants and this new one is only applicable to visibility operations.

Right now, this is not a problem because we didn't run out of space yet, but it would be nice to specialize a bunch of different flags.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1276,6 +1276,57 @@ impl<'a> RubyIndexer<'a> {
...
+                _ => {}
+            },
+            _ => {
+                self.visit_call_node_parts(node);
+                return;
+            }
+        }
+
+        let Some(arguments) = node.arguments() else {
+            return;
```
vinistock, line 1307: For a future PR, we can add a diagnostic here. Invoking `private|public_class_method` with no arguments does nothing.

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1276,6 +1276,57 @@ impl<'a> RubyIndexer<'a> {
...
 
+    fn handle_singleton_method_visibility(
+        &mut self,
+        node: &ruby_prism::CallNode,
+        visibility: Visibility,
+        call_name: &str,
+    ) {
+        match node.receiver() {
+            Some(ruby_prism::Node::SelfNode { .. }) | None => match self.nesting_stack.last() {
+                Some(Nesting::Method(_)) => return,
```
vinistock: This means that if we find `private_class_method` inside of a method, we're ignoring it. I think this is okay, but let's document it in a test.

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: I think that's already happening. When we process the arguments below, if it's a string or a symbol, we consider it as a valid call, otherwise we visit the argument and add a diagnostic.

rust/rubydex/src/indexing/ruby_indexer.rs
vinistock: Yeah, this should be `continue` and we're probably missing a test for multiple arguments.

vinistock (approved):
General implementation is good, just a few fixes

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Switched to `continue` and added `index_private_class_method_continues_past_invalid_arg` test

rust/rubydex/src/indexing/ruby_indexer_tests.rs
alexcrocha: They were, but with the switch to `continue` instead of `return` they are no longer dropped.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Added `index_private_class_method_inside_method_body_emits_no_diagnostic` test

alexcrocha:
### Merge activity

* **May 7, 7:57 PM UTC**: @alexcrocha merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/780).

PR #781, opened 2026-05-05
Resolve retroactive singleton method visibility changes

alexcrocha:
Part of #89 and follows #780, which added the `SINGLETON_METHOD_VISIBILITY` bit flag and indexed singleton-flagged `MethodVisibilityDefinition`s.

This PR routes singleton-flagged defs through the singleton class and folds the singleton path into the existing `resolve_method_visibilities` pass. Visibility definitions are processed after the main convergence loop, so the target method declaration is guaranteed to exist by the time we attach visibility.

For inherited targets like `private_class_method :foo` where `foo` comes from a parent class's singleton, we create a child-owned `MethodDeclaration` on the child's singleton class. This matches what #738 did for instance methods, and what Ruby reports when you ask `Child.singleton_class.instance_method(:foo).owner`:

```ruby
class Parent
  def self.foo; end
end

class Child < Parent
  private_class_method :foo
end

Child.singleton_class.instance_method(:foo).owner
# => #<Class:Child>   (Ruby copies the method to Child when visibility changes)

Parent.singleton_class.instance_method(:foo).owner
# => #<Class:Parent>  (untouched on the parent)
```

### Why document-scoped diagnostics for the singleton path

When a singleton target doesn't resolve, the diagnostic attaches to the document, not to the singleton declaration. Walking ancestors via `get_or_create_singleton_class(Foo, ...)` can synthesize `Foo::<Foo>` as a side effect, even when it has no real members. For `class Foo; private_class_method :missing; end` where `Foo` has no other singleton methods, that synthetic singleton class only exists to host the diagnostic. Attaching to it would leak the declaration on file delete. Document-scoped clears with the file. Instance-method diagnostics keep their existing declaration scope: their owner can't be synthetic.

### Source-order processing of visibility defs

Ruby applies visibility statements in the order they appear in source, so the later one wins:

```ruby
class Foo
  def bar; end
  private :bar
  public :bar          # Ruby: bar is Public
end
```

This was already broken on `main` for instance methods; surfaced while wiring up the singleton path, fixed here for both. `prepare_units` sorted `definitions` and `const_refs` by `(uri_id, offset)` for determinism but left `others` (which holds the visibility defs queued by `handle_remaining_definitions`) in `IdentityHashMap` iteration order, which is bucket-hash order. So `Foo#bar` could end up `Private` because the resolver happened to apply `public` first and `private` second. Same sort applied to `others` so the override-order invariant holds for both instance and singleton paths.

### In this PR

- branch on `is_singleton_method_visibility()` inside `resolve_method_visibilities`: resolve through `get_or_create_singleton_class` for the singleton path, render `"class method"` vs `"method"` in the diagnostic, attach document-scoped (singleton) vs declaration-scoped (instance)
- add `Graph::find_singleton_method_visibility_declaration` and route through it from the existing `Definition::MethodVisibility` arm in `definition_to_declaration_id` when the flag is set
- sort `others` by `(uri_id, offset)` in `prepare_units` so visibility overrides apply in source order

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -604,10 +604,8 @@ impl<'a> Resolver<'a> {
                         self.graph.add_document_diagnostic(uri_id, diagnostic);
                     }
                 }
-                Definition::MethodVisibility(visibility_def) => {
-                    if !visibility_def.flags().is_singleton_method_visibility() {
-                        method_visibility_ids.push(id);
-                    }
+                Definition::MethodVisibility(_) => {
+                    method_visibility_ids.push(id);
                 }
                 Definition::Class(_)
                 | Definition::SingletonClass(_)
@@ -622,7 +620,8 @@ impl<'a> Resolver<'a> {
         self.resolve_method_visibilities(method_visibility_ids);
     }
 
-    /// Resolves retroactive method visibility changes (`private :foo`, `protected :foo`, `public :foo`).
+    /// Resolves retroactive method visibility changes (`private :foo`, `protected :foo`, `public :foo`,
+    /// `private_class_method :foo`, `public_class_method :foo`).
     ///
     /// Runs as a second pass after all methods/attrs are declared, so `private :bar` works
     /// regardless of whether `def bar` appeared before or after it in source.
@@ -638,9 +637,19 @@ impl<'a> Resolver<'a> {
             let uri_id = *method_visibility.uri_id();
             let offset = method_visibility.offset().clone();
             let lexical_nesting_id = *method_visibility.lexical_nesting_id();
+            let is_singleton = method_visibility.flags().is_singleton_method_visibility();
 
-            let Some(owner_id) = self.resolve_lexical_owner(lexical_nesting_id, id) else {
+            let Some(lexical_owner_id) = self.resolve_lexical_owner(lexical_nesting_id, id) else {
                 continue;
+            };
+
+            let owner_id = if is_singleton {
+                let Some(singleton_id) = self.get_or_create_singleton_class(lexical_owner_id, true) else {
+                    continue;
+                };
+                singleton_id
+            } else {
+                lexical_owner_id
             };
 
             let Some(Declaration::Namespace(namespace)) = self.graph.declarations().get(&owner_id) else {
@@ -691,13 +700,20 @@ impl<'a> Resolver<'a> {
                     Rule::UndefinedMethodVisibilityTarget,
                     uri_id,
                     offset,
-                    format!("undefined method `{method_name}` for visibility change in `{owner_name}`"),
+                    format!("undefined method `{owner_name}#{method_name}` for visibility change"),
                 );
-                self.graph
-                    .declarations_mut()
-                    .get_mut(&owner_id)
-                    .unwrap()
-                    .add_diagnostic(diagnostic);
+                if is_singleton {
+                    // Document-scoped: the singleton class may be synthetic (created by this
+                    // visibility resolution) and won't be cleaned up on file delete, so attaching
+                    // the diagnostic to the declaration would leave it orphaned.
+                    self.graph.add_document_diagnostic(uri_id, diagnostic);
+                } else {
+                    self.graph
+                        .declarations_mut()
+                        .get_mut(&owner_id)
+                        .unwrap()
+                        .add_diagnostic(diagnostic);
+                }
             }
         }
 
@@ -1882,7 +1898,7 @@ impl<'a> Resolver<'a> {
                             singleton_methods.push(Unit::Definition(id));
                         }
                         _ => {
-                            others.push(id);
+                            others.push((id, (*definition.uri_id(), definition.offset())));
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


alexcrocha:
* **#781** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/resolution_tests.rs
```diff
@@ -5306,4 +5325,104 @@ mod visibility_resolution_tests {
...
+              private_class_method :nonexistent
+            end
+            ",
+        );
+        context.resolve();
+
+        assert_diagnostics_eq!(
+            context,
+            &[
+                "undefined-method-visibility-target: undefined class method `nonexistent()` for visibility change in `Foo::<Foo>` (2:25-2:36)"
```
st0012: Should `nonexistent` be a class method of `Foo`, or a class method of `Foo::<Foo>`, as the message suggests? I think it's the former?

rust/rubydex/src/resolution_tests.rs
```diff
@@ -5306,4 +5325,104 @@ mod visibility_resolution_tests {
...
+            ]
+        );
+    }
+
+    #[test]
+    fn retroactive_singleton_method_visibility_undefined_target_diagnostic_clears_when_file_deleted() {
```
st0012, line 5397: Let's add a similar test but from the other direction: the diagnostics should also disappear when we update `foo.rb` to define the missing method.

rust/rubydex/src/resolution.rs
```diff
@@ -687,17 +696,25 @@ impl<'a> Resolver<'a> {
                 // Ancestors are fully resolved — method definitively doesn't exist.
                 let method_name = self.graph.strings().get(&str_id).unwrap().as_str().to_string();
                 let owner_name = self.graph.declarations().get(&owner_id).unwrap().name().to_string();
+                let kind = if is_singleton { "class method" } else { "method" };
                 let diagnostic = Diagnostic::new(
                     Rule::UndefinedMethodVisibilityTarget,
                     uri_id,
                     offset,
-                    format!("undefined method `{method_name}` for visibility change in `{owner_name}`"),
+                    format!("undefined {kind} `{method_name}` for visibility change in `{owner_name}`"),
```
vinistock: Not necessary to change, but an alternative idea: we can use the fully qualified name of the method to avoid the need to check if it's singleton. Like this:

```rust
format!("undefined method {owner_name}#{method_name} for visibility change")
```

It will look like `Foo#bar()` or `Foo::<Foo>#bar()`.

rust/rubydex/src/resolution.rs
```diff
@@ -1922,12 +1939,15 @@ impl<'a> Resolver<'a> {
...
 
+        others.sort_unstable_by_key(|(_, key)| *key);
+
         // Definitions first, then constant refs, then singleton methods, then ancestors
         self.unit_queue.extend(definitions.into_iter().map(|(id, _)| id));
         self.unit_queue.extend(const_refs.into_iter().map(|(id, _)| id));
         self.unit_queue.extend(singleton_methods);
         self.unit_queue.extend(ancestors.into_iter().map(Unit::Ancestors));
 
+        let mut others: Vec<DefinitionId> = others.into_iter().map(|(id, _)| id).collect();
```
vinistock: If we create a brand new vector, do we still need to `shrink_to_fit`? I think not because the new vector will not have the same capacity as the other one, which will be garbage collected.

We may be able to swap these 3 final lines with

```rust
others.into_iter().map(|(id, _)| id).collect()
```

rust/rubydex/src/resolution_tests.rs
vinistock: Yes, this can be confusing because of all of the ways Ruby allows you to define methods. For example:

```ruby
class Foo
  def self.qux; end

  class << self
    def bar; end
  end
end
```

You could argue that `qux` is a class method defined on `Foo` and `bar` is an instance method defined on `Foo::<Foo>`. Or you could argue that there's no such thing as a "class method", everything is just methods defined on different classes or modules.

Maybe my previous suggestion on the diagnostic message will make it clearer?

vinistock (approved):

rust/rubydex/src/resolution_tests.rs
alexcrocha: Right, that makes sense. I'll go with your suggestion in the other comment as it is indeed clearer

rust/rubydex/src/resolution.rs
alexcrocha: This is better, thanks for the suggestion

rust/rubydex/src/resolution.rs
alexcrocha: You are right, we don't need it at all! Revised

alexcrocha:
### Merge activity

* **May 7, 8:07 PM UTC**: @alexcrocha merged this pull request with [Graphite](https://app.graphite.com/github/pr/Shopify/rubydex/781).

PR #782, opened 2026-05-05
Support array and inline-def arguments to private_class_method

alexcrocha:
Part of #89 and follows #781. With the basic call shape and resolution wired up, this PR extends the indexer to recognize the two other argument shapes Ruby accepts for `private_class_method` / `public_class_method`:

```ruby
class Foo
  def self.a; end
  def self.b; end

  private_class_method [:a, :b]                  # array literal
  private_class_method def self.foo; end         # inline singleton def
end
```

Both are common in Ruby code and already supported by ruby-lsp. Parity matters because ruby-lsp's indexing is migrating to Rubydex. Without this PR, these forms silently report the wrong visibility downstream.

### Inline def synthesizes a visibility def by name

For `private_class_method def self.foo; end` the indexer extracts the def's name, builds a `MethodVisibilityDefinition` for it with the singleton flag, then visits the `DefNode` so the `MethodDefinition` is indexed normally. At resolution time the visibility def attaches to the resulting declaration through the existing fully-qualified-name lookup, so no new "inline" definition kind and no new resolution arm are needed. The same path handles `def self.x` nested inside the array form.

A receiverless `def` (`private_class_method [def x; end]` or `private_class_method def x; end`) is a coding error: the def isn't a singleton method, so the call doesn't apply to it. We diagnose specifically (`` `private_class_method` requires a singleton method definition ``) and still visit the `DefNode` so it's indexed.

### Sole-arg guard for the array form

`private_class_method [:a]` works at runtime because Ruby unwraps a sole array argument; `private_class_method [:a], :b` doesn't. The match arm guards on `arg_count == 1` to mirror that. A multi-arg call with an array falls through to a separate arm with a tailored diagnostic (`` `private_class_method` array argument must be the only argument ``).

### Why `attr_*` arguments get their own diagnostic

`private attr_reader(:foo)` works for instance methods because `attr_reader` returns the names array and `private` consumes it. `private_class_method attr_reader(:foo)` doesn't: `attr_reader` defines instance methods, so the names don't exist on the singleton, and applying class-method visibility to them is meaningless. Common enough confusion to call out specifically rather than collapse into a generic "non-literal argument" message.

### Continue past invalid args, in both loops

Inherited from #780 and extended here to array elements: a bad arg or element produces a diagnostic but the loop keeps going. `private_class_method [:a, SOME_CONST, :c]` indexes `:a` and `:c` and diagnoses `SOME_CONST`. `private_class_method [:a], :b` diagnoses the array-not-sole-arg case and still indexes `:b`. IDEs see partial-valid code constantly during editing; halting on the first bad arg would discard intelligence the user can still benefit from.

### In this PR

- accept `ArrayNode` (sole-arg) with `SymbolNode`, `StringNode`, and `DefNode` elements
- accept `DefNode` as a top-level argument; reject receiverless `def` with a specific diagnostic
- diagnose `attr_*` arguments and array-not-sole-arg with tailored messages
- continue past invalid args in both the outer arg loop and the inner array-element loop, indexing whatever's still valid
- factor a `create_method_visibility_definition_from_name` helper so DefNode and Symbol/String paths share definition construction

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -1274,6 +1274,7 @@ impl<'a> RubyIndexer<'a> {
         }
     }
 
+    #[allow(clippy::too_many_lines)]
     fn handle_singleton_method_visibility(
         &mut self,
         node: &ruby_prism::CallNode,
@@ -1307,7 +1308,10 @@ impl<'a> RubyIndexer<'a> {
             return;
         };
 
-        for argument in &arguments.arguments() {
+        let args = arguments.arguments();
+        let arg_count = args.len();
+
+        for argument in &args {
             match argument {
                 ruby_prism::Node::SymbolNode { .. } | ruby_prism::Node::StringNode { .. } => {
                     self.create_method_visibility_definition(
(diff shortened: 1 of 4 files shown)


alexcrocha:
* **#782** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/782?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

st0012 (approved):

vinistock (approved):

PR #783, opened 2026-05-05
Extract shared helpers for visibility-call processing

alexcrocha:
The preceding stack (#780, #781, #782) built up the visibility-call repeated the same symbol and string argument parsing across the visibility-call handlers and `each_string_or_symbol_arg`. This PR extracts that parsing into `extract_literal_name`. The visibility handlers call it directly when they need custom handling for non-literal arguments, while `each_string_or_symbol_arg` now builds on it. No behaviour change.

Both indexer backends (`RubyIndexer` and `RubyOperationBuilder`) carry the duplication and get their own copy of the helper.

### DIFF
--- a/rust/rubydex/src/indexing/ruby_indexer.rs
+++ b/rust/rubydex/src/indexing/ruby_indexer.rs
@@ -386,21 +386,8 @@ impl<'a> RubyIndexer<'a> {
     {
         if let Some(arguments) = node.arguments() {
             for argument in &arguments.arguments() {
-                match argument {
-                    ruby_prism::Node::SymbolNode { .. } => {
-                        let symbol = argument.as_symbol_node().unwrap();
-
-                        if let Some(value_loc) = symbol.value_loc() {
-                            let name = Self::location_to_string(&value_loc);
-                            f(name, value_loc);
-                        }
-                    }
-                    ruby_prism::Node::StringNode { .. } => {
-                        let string = argument.as_string_node().unwrap();
-                        let name = String::from_utf8_lossy(string.unescaped()).to_string();
-                        f(name, argument.location());
-                    }
-                    _ => {}
+                if let Some((name, location)) = Self::extract_literal_name(&argument) {
+                    f(name, location);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


alexcrocha:
* **#783** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/783?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/indexing/ruby_indexer.rs
```diff
@@ -1414,6 +1386,22 @@ impl<'a> RubyIndexer<'a> {
         }
     }
 
+    fn extract_literal_name<'b>(arg: &ruby_prism::Node<'b>) -> Option<(String, ruby_prism::Location<'b>)> {
```
vinistock, line 1389: This look like it's doing the same thing as `each_string_or_symbol_arg`, but without the loop. Can we use this method there to avoid having two versions of the logic?

rust/rubydex/src/operation/ruby_builder.rs
```diff
@@ -1067,6 +1040,22 @@ impl<'a> RubyOperationBuilder<'a> {
         Some(())
     }
 
+    fn extract_literal_name<'b>(arg: &ruby_prism::Node<'b>) -> Option<(String, ruby_prism::Location<'b>)> {
```
vinistock, line 1043: Since we need this in both indexers, it might be a good candidate for extracting an `Indexer` trait that can share helpers between the two.

rust/rubydex/src/indexing/ruby_indexer.rs
alexcrocha: Yes, `each_string_or_symbol_arg` uses `extract_literal_name` in this diff. I updated the PR description to make that explicit as it was easy to miss.

rust/rubydex/src/operation/ruby_builder.rs
alexcrocha: Sounds like a good idea. There are a couple other helpers that would be good candidates as well so I suggest we introduce an `Indexer` trait as a follow up.

vinistock (approved):

PR #784, opened 2026-05-06
Handle visibility in completion

vinistock:
This PR starts handling visibility of methods and constants in completion. A considerable part of the complexity unfortunately comes from protected methods since both the receiver and lexical scope types must be in the same ancestor chain for valid invocations.

The idea is to switch from our original macro to a `collect_members` method which accepts some filtering mechanisms, so that we can filter the candidates according to all of the different scenarios.

### DIFF
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -340,6 +340,54 @@ impl Declaration {
     }
 
     #[must_use]
+    pub fn as_constant(&self) -> Option<&ConstantDeclaration> {
+        match self {
+            Declaration::Constant(constant) => Some(constant),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_constant_alias(&self) -> Option<&ConstantAliasDeclaration> {
+        match self {
+            Declaration::ConstantAlias(alias) => Some(alias),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_method(&self) -> Option<&MethodDeclaration> {
+        match self {
+            Declaration::Method(method) => Some(method),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_global_variable(&self) -> Option<&GlobalVariableDeclaration> {
+        match self {
+            Declaration::GlobalVariable(global) => Some(global),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_class_variable(&self) -> Option<&ClassVariableDeclaration> {
+        match self {
+            Declaration::ClassVariable(cvar) => Some(cvar),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_instance_variable(&self) -> Option<&InstanceVariableDeclaration> {
+        match self {
+            Declaration::InstanceVariable(ivar) => Some(ivar),
+            _ => None,
+        }
+    }
+
+    #[must_use]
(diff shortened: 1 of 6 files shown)


vinistock:
* **#784** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/784?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/784?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/graph_test.rb
```diff
@@ -1048,6 +1048,77 @@ def bar
...
+        PUBLIC_CONST = 1
+        SECRET = 2
+        private_constant :SECRET
+      end
+    RUBY
+    graph.resolve
+
+    candidates = graph.complete_namespace_access("Foo").map(&:name)
+    assert_includes(candidates, "Foo::PUBLIC_CONST")
+    refute_includes(candidates, "Foo::SECRET")
```
Morriar, line 1120: We have tests for `complete_method_call` with `self_receiver:` (private + protected), and a test for `complete_namespace_access` with private constants, but no test for `complete_namespace_access` with `self_receiver:` and private singleton methods (e.g., `private_class_method :foo` then `Foo::` completion).

The code handles it (visibility filter is wired in for singleton methods), but it's the one new code path without coverage. Should we add one here?

test/graph_test.rb
```diff
@@ -1048,6 +1048,77 @@ def bar
...
+      end
+    RUBY
+    graph.resolve
+
+    external = graph.complete_method_call("Foo").map(&:name)
+    assert_includes(external, "Foo#public_one()")
+    refute_includes(external, "Foo#secret()")
+
+    internal = graph.complete_method_call("Foo", self_receiver: "Foo").map(&:name)
+    assert_includes(internal, "Foo#secret()")
```
Morriar, line 1070:
```suggestion
    assert_includes(external, "Foo#public_one()")
    assert_includes(internal, "Foo#secret()")
```
?

rust/rubydex/src/model/declaration.rs
```diff
@@ -323,6 +323,34 @@ impl Declaration {
...
+        matches!(self, Declaration::Method(_))
+    }
+
+    #[must_use]
+    pub fn is_constant(&self) -> bool {
+        matches!(
+            self,
+            Declaration::Namespace(_) | Declaration::Constant(_) | Declaration::ConstantAlias(_)
+        )
+    }
```
Morriar: I think I prefered the old way where we were passing the list directly to `collect_candidates` instead of introducing `is_constant` as a public method. We already have `as_namespace` that returns a `Option<&Namespace>`, so we could imagine a `as_constant` to return a `Option<&Constant>` and it becomes weird with `is_constant` that checks for `Namespace | Constant | ConstantAlias`

Morriar (approved):

rust/rubydex/src/model/declaration.rs
vinistock: Yeah, agreed. I created the `as_*` methods and started using them directly in the filter.

test/graph_test.rb
vinistock: All of those are covered in the `query.rs` tests. I wanted to add some Ruby tests, but wanted to also avoid writing the exact same ones from `query.rs`, so I just picked a few.

PR #788, opened 2026-05-07
Add `MethodDefinition#signatures`

soutaro:
Towards https://github.com/Shopify/rubydex/issues/668.

This PR extracts `Signature` class definition and `MethodDefinition#signatures` implementation from https://github.com/Shopify/rubydex/pull/697.


ext/rubydex/signature.c
```diff
@@ -0,0 +1,83 @@
...
+                rb_ary_push(parameters, param);
+            }
+
+            signature = rb_class_new_instance(1, &parameters, cSignature);
+        }
+
+        rb_ary_push(signatures, signature);
+    }
+
+    rdx_definition_signatures_free(arr);
```
st0012, line 62: We need to make sure this still runs when `rb_class_new_instance` or other calls above it caused exception. I think the codebase already has some ensure convention.

st0012 (commented):
Is there a way we can distinguish between method taking `(?)` untyped parameters, and taking no parameters, as both seem to return empty arrays?

ext/rubydex/signature.c
```diff
@@ -0,0 +1,83 @@
...
+    cPostParameter = rb_define_class_under(cSignature, "PostParameter", cParameter);
+    cKeywordParameter = rb_define_class_under(cSignature, "KeywordParameter", cParameter);
+    cOptionalKeywordParameter = rb_define_class_under(cSignature, "OptionalKeywordParameter", cParameter);
+    cRestKeywordParameter = rb_define_class_under(cSignature, "RestKeywordParameter", cParameter);
+    cForwardParameter = rb_define_class_under(cSignature, "ForwardParameter", cParameter);
+    cBlockParameter = rb_define_class_under(cSignature, "BlockParameter", cParameter);
+
+    empty_params = rb_ary_new();
+    OBJ_FREEZE(empty_params);
+    rb_gc_register_mark_object(empty_params);
```
vinistock, line 82: Oh, nice.

ext/rubydex/signature.c
vinistock: I agree in general, but I think this code is pretty safe. We're creating new instances of our own objects and array push is probably safe enough.

vinistock (approved):

PR #789, opened 2026-05-07
Add source code and test helper in GraphTest

soutaro:
Add `GraphTest::source_of(definition_id)`, which the source code for the definition. The tests in #697 use these helpers.

Extracted from https://github.com/Shopify/rubydex/pull/697.

### DIFF
--- a/rust/rubydex/src/test_utils/graph_test.rs
+++ b/rust/rubydex/src/test_utils/graph_test.rs
@@ -1,20 +1,26 @@
+use std::collections::HashMap;
+
 use super::normalize_indentation;
 #[cfg(test)]
 use crate::diagnostic::Rule;
 use crate::indexing::{self, LanguageId};
 use crate::model::graph::{Graph, NameDependent};
-use crate::model::ids::{NameId, StringId};
+use crate::model::ids::{DefinitionId, NameId, StringId};
 use crate::resolution::Resolver;
 
 #[derive(Default)]
 pub struct GraphTest {
     graph: Graph,
+    sources: HashMap<String, String>,
 }
 
 impl GraphTest {
     #[must_use]
     pub fn new() -> Self {
-        Self { graph: Graph::new() }
+        Self {
+            graph: Graph::new(),
+            sources: HashMap::new(),
+        }
     }
 
     #[must_use]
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -31,12 +37,37 @@ impl GraphTest {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition or its document does not exist.
+    #[must_use]
+    pub fn source_of(&self, definition_id: &DefinitionId) -> &str {
+        let def = self.graph.definitions().get(definition_id).unwrap();
+        let uri = self.graph.documents().get(def.uri_id()).unwrap().uri();
+        let source = self.source(uri);
+        &source[def.offset().start() as usize..def.offset().end() as usize]
```
soutaro, line 70: @Morriar [suggested](https://github.com/Shopify/rubydex/pull/697/changes#r3028558055) to get the content through `Document`, but `Document` doesn't have the actual source code. (It has offset of each line, but doesn't have the source text.)

soutaro:
We won't need this test helper if the dealiase returns `DeclarationId`, not `DefinitionId` as in #779.

rust/rubydex/src/test_utils/graph_test.rs
```diff
@@ -31,12 +37,37 @@ impl GraphTest {
...
+
+    /// Returns the source text for a definition, sliced by its offset.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the definition or its document does not exist.
+    #[must_use]
+    pub fn source_of(&self, definition_id: &DefinitionId) -> &str {
+        let def = self.graph.definitions().get(definition_id).unwrap();
+        let uri = self.graph.documents().get(def.uri_id()).unwrap().uri();
```
st0012, line 68: Can this method work with definitions coming from `built_in.rs`? They have uri but don't have document AFAIK. So it'll probably panic on this line?

rust/rubydex/src/test_utils/graph_test.rs
st0012, line 74: Should we also update `sources` when we remove a uri?

rust/rubydex/src/test_utils/graph_test.rs
soutaro: Absolutely. But I think we can easily avoid testing with builtins...

st0012: I don't think we'll intentionally test against the built in definitions. What could happen is that we build other helpers on top of this method, which takes a declaration and get all the sources of it. In that case it'd be reasonable to pass `Object` etc to it and then it'd explode.
We can likely prevent it by checking if the definition is a built in before invoking this.

soutaro:
The test utility is not used. Closing this PR for this time. Reopen if we need it in future.

PR #793, opened 2026-05-07
Follow alias target for visibility

vinistock:
We were not following alias targets to get the inherited visibility of an alias. Additionally, we were returning `None` as the final fallback, which is reserved for declaration types where visibility simply doesn't apply (like ivars).

This PR makes sure we follow aliases to detect the inherited visibility and I changed the final fallback to `public` in case we fail to determine the visibility of a method. I also moved `Todo` together with `Class` since it's basically the same thing.


vinistock:
* **#793** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/793?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexcrocha (approved):

PR #794, opened 2026-05-08
Make self type required for completion

vinistock:
Completion is based on two things, depending on the scenario:

1. Either the receiver type or lexical scope
2. The type of `self` (which is not always the same as the lexical scope)

We were making the type of `self` optional before, but the reality is that we always need it. Consider this example:

```ruby
class Foo
  # completion here

  def bar
    # and completion here
  end
end
```

In both cases, the lexical scope is the same, but the type of `self` is different `Foo::<Foo>` vs `Foo`. We already provide the tools for consumers to figure out the type of `self` for these cases, so let's make it required in the API in order to provide correct completion.

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -16,17 +16,20 @@ static VALUE cKeywordParameter;
 // Interned once in `rdxi_initialize_graph` to avoid repeated symbol-table lookups on hot completion paths.
 static ID id_self_receiver;
 
-// Extracts the optional `self_receiver:` kwarg from `opts`. Returns NULL when the kwarg is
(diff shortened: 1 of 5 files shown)


vinistock:
* **#794** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexcrocha (approved):

PR #795, opened 2026-05-08
Add `Reference#location` for better typing

vinistock:
All reference types have a `location` method, but we never defined an abstract version of it in the parent. That makes typing weird because if you have a generic `Reference` type, then `location` doesn't exist on it.

Let's include an abstract implementation to get more intuitive typing.


vinistock:
* **#795** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/795?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexcrocha (approved):

PR #798, opened 2026-05-08
Add operation-based indexing pipeline

Morriar:
## What

Introduces an intermediate representation (IR) between Ruby parsing and graph construction. Instead of the indexer writing directly to a `LocalGraph`, a new pipeline produces an ordered `Vec<Operation>` (an explicit list of what each file contributes to the graph), then applies those operations to build the graph.

**Current pipeline:**

```mermaid
graph LR
    AST --> RubyIndexer --> LocalGraph --> Graph --> Resolution
```

**New pipeline:**

```mermaid
graph LR
    AST --> RubyOperationBuilder --> Operations["Vec&lt;Operation&gt;"] --> OperationApplier --> LocalGraph --> Graph --> Resolution
```

(This is an intermediate step to ensure parity with the current indexer / local graph and reduce the scope of this PR.)

**Target pipeline:**

```mermaid
graph LR
    AST --> Operations["Vec&lt;Operation&gt;"] --> Graph["Graph (resolution applies operations directly)"]
```

Once we commit to this direction, we can remove the `LocalGraph` and feed the operations directly to the resolver to let it create _both_ the definitions and the declarations.

## Why

Four motivations that reinforce each other:

**DSL plugin API**

Plugins need to contribute definitions to the graph. Without operations, plugins mutate `LocalGraph` internals: fragile, tightly coupled, and bad for parallelism (Ruby plugins calling into the graph would require synchronization).

With operations, a plugin just produces `Vec<Operation>`, a pure data structure that can be built in parallel and applied later. Operations become the **public contract** between Rubydex and plugins.

```ruby
class Post < ApplicationRecord
  has_many :comments
end
```

An ActiveRecord plugin sees `has_many :comments` and emits operations. It doesn't touch graph internals, it speaks the same operation protocol as the built-in indexer.

```
EnterMethod(comments)
ExitScope
EnterMethod(comments=)
ExitScope
EnterMethod(comment_ids)
ExitScope
EnterMethod(comment_ids=)
ExitScope
```

**Unloading**

Each file's contribution is its operation list. Negating the operations (walk in reverse, opposite effect) cleanly unloads the file without graph diffing.

```ruby
# user.rb produces:
# 1. EnterClass(User)
# 2.   EnterMethod(name)
# 3.   ExitScope
# 4.   EnterMethod(email)
# 5.   ExitScope
# 6. ExitScope

# To unload user.rb: negate in reverse order
# 6. (no-op)
# 5. (no-op)
# 4. RemoveMethod(email)
# 3. (no-op)
# 2. RemoveMethod(name)
# 1. RemoveClass(User)
```

Because DSL plugins produce the same IR primitives as the built-in indexer, we know how to undo their contributions for free: no plugin-specific cleanup logic, no tracking which definitions in the graph came from which plugin.

**Incremental re-indexing**

Because operations are a flat list, we can diff the old and new operation lists when a file changes. Only the delta gets applied: remove operations that disappeared, add new ones. Most edits change very few operations, so the diff is small.

```
# user.rb: rename `name` to `full_name`

Old operations:                    New operations:
  EnterClass(User)                   EnterClass(User)
    EnterMethod(name)          →       EnterMethod(full_name)
    ExitScope                          ExitScope
    EnterMethod(email)                 EnterMethod(email)
    ExitScope                          ExitScope
  ExitScope                          ExitScope

# Diff: remove EnterMethod(name), add EnterMethod(full_name)
```

No full rebuild, no negating the entire file, just the minimal set of changes.

**Order-dependent diagnostics**

Operations preserve source order, which the current graph doesn't. Consider:

```ruby
class Foo
  private_constant :BAR
  BAR = 42
end
```

Ruby raises `NameError` because `private_constant :BAR` runs before `BAR` exists. Today Rubydex can't catch this because both the constant and the visibility change become unordered definitions in the graph, and resolution wires them together regardless of order. With operations, the ordering is explicit (`SetConstantVisibility` before `DefineConstant`), and a diagnostic pass can detect the forward reference.

## How

**`Operation` enum** (`operation/mod.rs`): ~20 variants covering namespace scopes (`EnterClass`, `EnterModule`, `EnterSingletonClass`, `ExitScope`), method operations (`EnterMethod`, `DefineAttribute`, `AliasMethod`, `SetMethodVisibility`, `SetDefaultVisibility`), constants (`DefineConstant`, `AliasConstant`, `SetConstantVisibility`), variables (`DefineInstanceVariable`, `DefineClassVariable`, `DefineGlobalVariable`), mixins (`Mixin`), and references (`ReferenceConstant`, `ReferenceMethod`). Each variant is its own struct so consumers can take fields by value or by reference.

**`OperationPrinter`** (`operation/printer.rs`): Debug formatter that pretty-prints `Vec<Operation>` with indentation reflecting scope nesting. Used for debugging and testing the builder output.

**`RubyOperationBuilder`** (`operation/ruby_builder.rs`): Walks the Ruby AST and produces `Vec<Operation>`. Same visitor pattern as `RubyIndexer`, but emits operations instead of mutating a graph. ~3100 lines, 1:1 correspondence with every `RubyIndexer` behavior.

**`OperationApplier`** (`operation/applier.rs`): Converts `Vec<Operation>` → `LocalGraph`. Consumes operations by value (zero clones). This is scaffolding that exists only because the current merge pipeline expects a `LocalGraph`. In the target architecture, resolution consumes operations directly and both `OperationApplier` and `LocalGraph` go away.

**`IndexerBackend` enum** (`indexing.rs`, `main.rs`): `--indexer ruby_indexer|operation_builder` CLI flag to switch backends. Both backends run the same shared test suite (166 test functions × 2 backends).

## Parity verification

Both the indexer test suite (166 tests) and the resolution test suite (208 tests) run with both backends using a `#[path]` module trick — each parent module provides a `backend()` function via `super::backend()`, so the same test file exercises both `RubyIndexer` and `OperationBuilder`.

Both backends produce **identical output** on Shopify core:

|                    | RubyIndexer | OperationBuilder |
| ------------------ | ----------- | ---------------- |
| Total definitions  | 926,926     | 926,926          |
| Total declarations | 749,360     | 749,360          |

### Performance

|               | RubyIndexer | OperationBuilder | Delta  |
| ------------- | ----------- | ---------------- | ------ |
| Indexing time | 3.98s       | 4.02s            | +1.0%  |
| Total time    | 29.3s       | 29.9s            | +1.9%  |
| Memory (RSS)  | 3,063 MB    | 3,143 MB         | +2.6%  |

Essentially identical performance. The intermediate `Vec<Operation>` per file adds a very small overhead.

### DIFF
--- a/rust/rubydex/src/model/definitions.rs
+++ b/rust/rubydex/src/model/definitions.rs
@@ -29,7 +29,7 @@ use crate::{
     assert_mem_size,
     model::{
         comment::Comment,
-        ids::{ConstantReferenceId, DefinitionId, NameId, StringId, UriId},
+        ids::{self, ConstantReferenceId, DefinitionId, NameId, StringId, UriId},
         visibility::Visibility,
     },
     offset::Offset,
@@ -289,7 +289,7 @@ impl ClassDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.name_id))
+        ids::namespace_definition_id(self.uri_id, &self.offset, self.name_id)
     }
 
     #[must_use]
@@ -411,7 +411,7 @@ impl SingletonClassDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.name_id))
+        ids::namespace_definition_id(self.uri_id, &self.offset, self.name_id)
     }
 
     #[must_use]
@@ -515,7 +515,7 @@ impl ModuleDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.name_id))
+        ids::namespace_definition_id(self.uri_id, &self.offset, self.name_id)
     }
 
     #[must_use]
@@ -611,7 +611,7 @@ impl ConstantDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        DefinitionId::from(&format!("{}{}{}", *self.uri_id, self.offset.start(), *self.name_id))
+        ids::namespace_definition_id(self.uri_id, &self.offset, self.name_id)
     }
 
     #[must_use]
@@ -926,7 +926,7 @@ pub struct MethodDefinition {
 assert_mem_size!(MethodDefinition, 96);
 
 /// The receiver of a singleton method definition.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Receiver {
     /// `def self.foo` - receiver is the enclosing definition (class, module, singleton class or DSL)
     SelfReceiver(DefinitionId),
@@ -965,18 +965,7 @@ impl MethodDefinition {
 
     #[must_use]
     pub fn id(&self) -> DefinitionId {
-        let mut formatted_id = format!("{}{}{}", *self.uri_id, self.offset.start(), *self.str_id);
-
-        if let Some(receiver) = &self.receiver {
-            match receiver {
-                Receiver::SelfReceiver(def_id) => formatted_id.push_str(&def_id.to_string()),
-                Receiver::ConstantReceiver(name_id) => {
-                    formatted_id.push_str(&name_id.to_string());
-                }
-            }
-        }
-
-        DefinitionId::from(&formatted_id)
+        ids::method_definition_id(self.uri_id, &self.offset, self.str_id, self.receiver.as_ref())
     }
 
     #[must_use]
--- a/rust/rubydex/src/model/ids.rs
+++ b/rust/rubydex/src/model/ids.rs
@@ -1,4 +1,8 @@
-use crate::{assert_mem_size, model::id::Id};
+use crate::{
+    assert_mem_size,
+    model::{definitions::Receiver, id::Id},
+    offset::Offset,
+};
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
 pub struct DeclarationMarker;
@@ -11,6 +15,28 @@ pub struct DefinitionMarker;
 // DefinitionId represents the ID of a definition found in a specific file
 pub type DefinitionId = Id<DefinitionMarker>;
 assert_mem_size!(DefinitionId, 8);
+
+#[must_use]
+pub fn namespace_definition_id(uri_id: UriId, offset: &Offset, name_id: NameId) -> DefinitionId {
+    DefinitionId::from(&format!("{}{}{}", *uri_id, offset.start(), *name_id))
+}
+
+#[must_use]
+pub fn method_definition_id(
+    uri_id: UriId,
+    offset: &Offset,
+    str_id: StringId,
+    receiver: Option<&Receiver>,
+) -> DefinitionId {
+    let mut formatted_id = format!("{}{}{}", *uri_id, offset.start(), *str_id);
+    if let Some(receiver) = receiver {
+        match receiver {
+            Receiver::SelfReceiver(def_id) => formatted_id.push_str(&def_id.to_string()),
+            Receiver::ConstantReceiver(name_id) => formatted_id.push_str(&name_id.to_string()),
+        }
+    }
+    DefinitionId::from(&formatted_id)
+}
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
 pub struct UriMarker;
--- a/rust/rubydex/src/operation/mod.rs
+++ b/rust/rubydex/src/operation/mod.rs
@@ -1,0 +1,284 @@
+//! Operations represent the ordered actions extracted from Ruby/RBS source code.
+//!
+//! Unlike definitions (which are unordered and declarative), operations are ordered and imperative.
+//! They model what Ruby actually does at execution time: define a class, define a method, change
+//! visibility, include a module, etc.
+//!
+//! Each operation is self-contained: it carries enough context (`name_id`, etc.) to know what it
+//! defines, while scope context is provided by surrounding Enter/Exit scope operations.
+//!
+//! The builder produces a `Vec<Operation>` for each file. These operations are then applied in order
+//! by the applier to produce definitions and declarations in the graph.
+//!
+//! # Example
+//!
+//! ```ruby
+//! class Foo
+//!   def bar; end
+//!   private :bar
+//! end
+//! ```
+//!
+//! Produces the operations:
+//! 1. `EnterClass` (name: Foo)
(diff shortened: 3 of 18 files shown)


rust/rubydex/src/model/definitions.rs
```diff
@@ -926,7 +926,7 @@ pub struct MethodDefinition {
 assert_mem_size!(MethodDefinition, 96);
 
 /// The receiver of a singleton method definition.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
vinistock, line 929: Do we need the clone?

rust/rubydex/src/model/ids.rs
```diff
@@ -12,6 +16,28 @@ pub struct DefinitionMarker;
 pub type DefinitionId = Id<DefinitionMarker>;
 assert_mem_size!(DefinitionId, 8);
 
+#[must_use]
+pub fn namespace_definition_id(uri_id: UriId, offset: &Offset, name_id: NameId) -> DefinitionId {
+    DefinitionId::from(&format!("{}{}{}", *uri_id, offset.start(), *name_id))
+}
+
+#[must_use]
+pub fn method_definition_id(
```
vinistock, line 25: Can we define these as associated functions of `DefinitionId` instead? Basically, so that we can do this instead:

```rust
DefinitionId::for_namespace()
DefinitionId::for_method()
```

rust/rubydex/src/operation/mod.rs
vinistock, line 1: The `mod.rs` thing is the old Rust naming style. Can we use this instead?

```
rubydex/src/operation.rs # defines the module
           /operation/whatever.rs
```

Also, my vote goes to naming it `IR` instead of operation.

rust/rubydex/src/operation/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+/// The kind of mixin operation.
+#[derive(Debug, Clone, Copy)]
+pub enum MixinKind {
+    Include,
+    Prepend,
+    Extend,
+}
+
+#[derive(Debug)]
+pub struct EnterClass {
```
vinistock, line 118: Let's default all of these fields to private and expose with getters.

rust/rubydex/src/operation/printer.rs
```diff
@@ -0,0 +1,260 @@
...
+    visibility::Visibility,
+};
+use crate::operation::{
+    AliasConstant, AliasGlobalVariable, AliasMethod, AttrKind, DefineAttribute, DefineClassVariable, DefineConstant,
+    DefineGlobalVariable, DefineInstanceVariable, EnterClass, EnterMethod, EnterModule, EnterSingletonClass, Mixin,
+    MixinKind, Operation, ReferenceConstant, ReferenceMethod, SetConstantVisibility, SetDefaultVisibility,
+    SetMethodVisibility, Target,
+};
+
+struct OperationPrinter<'a> {
```
vinistock, line 17: Not necessary for this PR, but it might be useful to consider the implementation as an M x N matrix of IR compilers and VMs.

For the compilers, we already have a trait with default behaviours to guide the implementation (Prism's visit trait). We should create a virtual machine trait that defaults to doing nothing for all instructions. It's a small thing, but it makes it easier to discover which instructions you haven't handled yet and you get better LSP support.

rust/rubydex/src/operation/ruby_builder.rs
```diff
@@ -0,0 +1,2915 @@
+//! Visit the Ruby AST and produce an ordered list of operations.
+//!
+//! Walks the parsed AST and produces `Vec<Operation>` that can later be applied
```
vinistock, line 3: Idea: is it worth encapsulating a list of compiled instructions into a `Chunk`?

rust/rubydex/src/operation/ruby_builder.rs
```diff
@@ -0,0 +1,2915 @@
...
+    pub strings: IdentityHashMap<StringId, StringRef>,
+    pub names: IdentityHashMap<NameId, NameRef>,
+}
+
+#[derive(Clone, Copy)]
+enum Nesting {
+    /// A lexical scope (class/module keyword) that produces a new constant resolution scope.
+    LexicalScope { name_id: NameId, is_module: bool },
+    /// An owner that doesn't produce a lexical scope (Class.new/Module.new).
+    Owner { name_id: NameId, is_module: bool },
```
vinistock, line 40: On the other approach, we walk the nesting backwards to figure out if we're inside a module or not. Is storing the boolean here mostly a performance improvement? Or is there some other aspect?

rust/rubydex/src/operation/ruby_builder.rs
```diff
@@ -0,0 +1,2915 @@
...
+    /// A method entry, used for instance variable ownership.
+    Method { receiver: Option<NestingReceiver> },
+}
+
+/// Tracks receiver info for methods on the nesting stack, so `method_receiver` can work
+/// without looking up definitions. Distinct from `operation::Target` which represents
+/// the source-level receiver without resolved names.
+#[derive(Clone, Copy)]
+enum NestingReceiver {
```
vinistock, line 49: I think it would be nicer to extract this into `definitions.rs` since the case of having something where the receiver is either `self` or a constant is so common. We can probably reuse the same enum for method and method alias definitions.

rust/rubydex/src/operation/ruby_builder.rs
```diff
@@ -0,0 +1,2915 @@
...
+    #[test]
+    fn build_class_node() {
+        assert_operations(
+            "
+            class Foo
+              class Bar; end
+            end
+            ",
+            "
+            EnterClass(Foo)
```
vinistock, line 2289: I love this style of tests. Very easy to visually verify what the expected IR is.

vinistock (approved):
I really believe moving towards an IR will give us a much nicer way to express all of the richness in Ruby code.

It's challenging to reason about the implementation because we are supporting both mechanisms temporarily and without a VM to go along with the IR, but I would rather we pull the future forward and iterate fast on the refactor than holding this back.

Here are some questions/things to explore in case they prove useful:

1. It would be nice to test out more granular instructions to see what the code looks like. We don't have to do it for all definitions, but maybe experiment with classes by supporting `Class.new` and `Struct.new` to see what emerges
2. What do the reverse instructions look like for incremental resolution? Is it possible to generically reverse all operations (something like `[IR::Undo, IR::DefineClass]`)? Or do we need to create each reverse operation one by one (e.g.: `[IR::UndefClass]`)?
3. After we have a proper VM that processes the IR directly without the intermediate local graph, how do we account for ordering and parallelism in processing the chunks? For example, what happens if we start interpreting one chunk and hit a constant dependency? I'd assume there's no way to order the chunks correctly because you can have different constant dependencies per instruction (and not per chunk), so is it possible to pause the execution of a chunk and resume after? Or do we need full dependency tracking?
4. An interesting aspect is that the boundary between compiler and VM can be quite blurry. For example, if we find a `private` in the code, do we apply the visibility to the IR during compilation or does it become an instruction that only gets processed by the VM? For this particular example, I think the answer is the latter, but I'm curious about what other boundaries will change

PR #802, opened 2026-05-12
Resolve module_function visibility

### LINKED ISSUE #89: Start keeping track of visibility for applicable definitions

alexcrocha:
Closes #89 and finishes the singleton-method visibility resolution work started in #780/#781/#782. This PR resolves retroactive `module_function` visibility:

```ruby
module Foo
  def bar; end
  module_function :bar
end

Foo.bar           # public singleton method
Foo.new.bar       # NoMethodError (private instance method)
```

`Foo#bar` becomes private and `Foo.bar` is added as a public singleton method.

### One definition per side

The indexer emits two `MethodVisibility` definitions at the `module_function :bar` call site: one for `Foo#bar` and one for `Foo::<Foo>#bar`. The singleton-side definition uses the existing `SINGLETON_METHOD_VISIBILITY` flag; both carry `Visibility::ModuleFunction`.

The definitions must remain separate because reverse lookup maps each definition to one declaration. Giving each side its own definition lets invalidation detach the instance and singleton declarations independently when the source method or `module_function` call is removed. The flag bits are included in `MethodVisibilityDefinition::id()` so the two definitions at the same source location have distinct IDs.

### Resolution

The instance-side definition follows the existing method-visibility path. The singleton-side definition looks for the same instance method, then attaches a declaration to the singleton class. `Graph::visibility()` exposes these as `Private` and `Public`, respectively.

The singleton class is created only after the target method is found. For `module_function :missing`, the instance side emits the diagnostic while the singleton side stays silent, leaving no empty `Foo::<Foo>` namespace behind.

The existing inline form (`module_function def bar; end`) is unchanged. Regression coverage verifies that its singleton companion can still be changed by a later singleton visibility call.

### Known difference from Ruby

Rubydex resolves each argument independently. For `module_function :a, :missing`, it makes `Foo#a` private, adds the public `Foo.a`, and reports `:missing`. Ruby makes `Foo#a` private before raising, but creates no singleton copies for the call. Resolving independently preserves useful information while code is incomplete in an editor.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -16,6 +16,7 @@ use crate::model::{
     identity_maps::{IdentityHashBuilder, IdentityHashMap, IdentityHashSet},
     ids::{ConstantReferenceId, DeclarationId, DefinitionId, NameId, StringId, UriId},
     name::{Name, NameRef, ParentScope},
+    visibility::Visibility,
 };
 
 enum Outcome {
@@ -635,7 +636,7 @@ impl<'a> Resolver<'a> {
     }
 
     /// Resolves retroactive method visibility changes (`private :foo`, `protected :foo`, `public :foo`,
-    /// `private_class_method :foo`, `public_class_method :foo`).
+    /// `private_class_method :foo`, `public_class_method :foo`, `module_function :foo`).
     ///
     /// Runs as a second pass after all methods/attrs are declared, so `private :bar` works
     /// regardless of whether `def bar` appeared before or after it in source.
@@ -652,12 +653,14 @@ impl<'a> Resolver<'a> {
             let offset = method_visibility.offset().clone();
             let lexical_nesting_id = *method_visibility.lexical_nesting_id();
             let is_singleton = method_visibility.flags().is_singleton_method_visibility();
+            let is_module_function = *method_visibility.visibility() == Visibility::ModuleFunction;
+            let is_module_function_singleton = is_singleton && is_module_function;
 
             let Some(lexical_owner_id) = self.resolve_lexical_owner(lexical_nesting_id, id) else {
                 continue;
             };
 
-            let owner_id = if is_singleton {
+            let namespace_id = if is_singleton && !is_module_function {
                 let Some(singleton_id) =
                     self.get_or_create_singleton_class(lexical_owner_id, SingletonAncestors::Eager)
                 else {
@@ -668,12 +671,12 @@ impl<'a> Resolver<'a> {
                 lexical_owner_id
             };
 
-            let Some(Declaration::Namespace(namespace)) = self.graph.declarations().get(&owner_id) else {
+            let Some(Declaration::Namespace(namespace)) = self.graph.declarations().get(&namespace_id) else {
                 continue;
             };
 
-            let mut visibility_applied = false;
-            let mut has_partial = false;
+            let mut visibility_definition_attached = false;
+            let mut ancestor_chain_is_incomplete = false;
 
             for ancestor in namespace.ancestors() {
                 match ancestor {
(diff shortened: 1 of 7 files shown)


alexcrocha:
* **#802** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/802?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/802?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rust/rubydex/src/resolution.rs
```diff
@@ -671,49 +684,71 @@ impl<'a> Resolver<'a> {
...
 
                         if has_member {
-                            // Direct member: `create_declaration`'s fully qualified name dedup attaches
-                            // this visibility definition to the existing method declaration.
-                            // Inherited: a new child-owned declaration is created.
-                            self.create_declaration(str_id, id, owner_id, |name| {
-                                Declaration::Method(Box::new(MethodDeclaration::new(name, owner_id)))
-                            });
-                            visibility_applied = true;
+                            target_found = true;
```
soutaro: I was a bit confused for the case where:

1. The ancestor chain includes a partial ancestor
2. Another ancestor that is located after the partial ancestor has the member

and if resolution comes from the module it was partial in this loop later. (Should we stop target lookup if partial ancestor is detected?)

This is actually **ok** because `Declaration::Method` is created with `declaration_owner_id`, not something looked up from the ancestor chain. 👍

rust/rubydex/src/resolution.rs
```diff
@@ -652,12 +653,24 @@ impl<'a> Resolver<'a> {
...
+            // members we search for the target method) and `declaration_owner_id` (the
+            // namespace the resulting declaration belongs to). Three shapes:
+            //
+            // - `private`/`protected`/`public` (and `module_function :bar`'s instance-side):
+            //   both owners are the lexical owner.
+            // - `private_class_method`: both owners are the singleton class. Create it upfront.
+            // - `module_function :bar` singleton-side: target owner is the lexical (instance)
+            //   owner; declaration owner is the singleton class. Defer creating the singleton
+            //   class until the target is found, so a missing target doesn't leave an empty
+            //   `Foo::<Foo>`.
```
vinistock: I found these comments to be a bit confusing. I'm not really sure I understand why there are two owners.

rust/rubydex/src/resolution.rs
```diff
@@ -687,40 +700,63 @@ impl<'a> Resolver<'a> {
...
+                    target_owner_id
+                };
+
+                // `create_declaration` deduplicates by fully qualified name: if a declaration
+                // with this name already exists, the visibility def attaches to it; otherwise
+                // a new declaration is created.
+                self.create_declaration(str_id, id, declaration_owner_id, |name| {
+                    Declaration::Method(Box::new(MethodDeclaration::new(name, declaration_owner_id)))
+                });
                 continue;
```
vinistock: Why did we need to pull this down? Couldn't we create the declaration in the same previous spot?

rust/rubydex/src/resolution_tests.rs
vinistock, line 1: I have a good test case to add:

```ruby
module Foo
  module_function def bar; end
  
  class << self
    private :bar
  end
end
```

Both `Foo#bar()` and `Foo::<Foo>#bar()` should be private.

vinistock (commented):
Do we already support module function and the other visibility operations in the RBS indexer?

rust/rubydex/src/resolution_tests.rs
alexcrocha: Added!

alexcrocha:
I pushed a new commit simplifying the logic, which was a bit complex to grok. There's no more target/declaration owner split, and declaration creation is now next to the method match.

There's some duplication now, which is what I was trying to eliminate in the previous version. Let me know if this is more readable.

(Clippy complains that the function has too many lines. I'll fix it once we are happy with the shape of implementation)

rust/rubydex/src/resolution.rs
alexcrocha: See https://github.com/Shopify/rubydex/pull/802#issuecomment-4988098648

rust/rubydex/src/resolution.rs
alexcrocha: See https://github.com/Shopify/rubydex/pull/802#issuecomment-4988098648

rust/rubydex/src/resolution.rs
```diff
@@ -652,11 +653,57 @@ impl<'a> Resolver<'a> {
...
 
+            if is_module_function_singleton {
+                let Some(Declaration::Namespace(namespace)) = self.graph.declarations().get(&lexical_owner_id) else {
+                    continue;
+                };
+
+                let mut visibility_definition_attached = false;
+                let mut ancestor_chain_is_incomplete = false;
+
+                for ancestor in namespace.ancestors() {
```
vinistock: Would it be possible to merge this loop with the one below? They seem nearly identical, so I'm wondering if we can go from

```
if conditions
  loop
  end

  maybe continue to next iteration
end

if conditions
  loop
  end
end
```

to

```
loop
  if conditions
  elsif conditions
  end
end
```

vinistock (commented):
Looks good in general, I'm just wondering if we can unify the method visibility handling into a single loop

rust/rubydex/src/resolution.rs
alexcrocha: Yes we can! Updated

vinistock (approved):

PR #808, opened 2026-05-15
MCP: exclude the same default dirs as the ruby gem

bnferguson:
## Background 
When testing out the MCP Claude brought up that it was getting declarations appearing in `search_declarations` from my worktrees in the `.claude` dir. Digging a bit further in it seems that while the Ruby gem skips things via `Rubydex::Graph::IGNORED_DIRECTORIES` the MCP side had no such mechanism. 

It looks like PR #712 added the `excluded` parameter to `collect_file_paths`. Then PR #713 wired exclusions into the Ruby `Graph` class. But the MCP was never fully wired in. 

## Approach
So that's what this PR is doing! Wire'n it in. Currently it mirrors/duplicates the `Rubydex::Graph::IGNORED_DIRECTORIES` over in Rust. Considered having it respect `.gitignore` but since that was already explored in the closed PR #205 (perhaps more to avoid the bringing in a new crate than the concept) and there was the existing concept in the Ruby gem, I decided to match that. 

## Testing

There's an e2e test added that can be run with: 

`cargo test -p rubydex-mcp --test mcp mcp_server_skips_ignored_directories`

Another approach is tossing a `.rb` file in your `.claude` or `node_modules` and seeing if the declaration comes up when invoking `search_declarations` in the MCP. 

## Environment

- `rubydex-mcp`: `0.1.0`
- rubydex gem: `0.2.3`
- OS: macOS (Darwin 25.x, Apple Silicon)

### DIFF
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -20,6 +20,20 @@ use rubydex::model::{
 };
 use url::Url;
 
+/// Directory names that are never worth descending into while indexing a workspace.
+/// Mirrors `Rubydex::Graph::IGNORED_DIRECTORIES` in the Ruby gem.
+const IGNORED_DIRECTORIES: &[&str] = &[
+    ".bundle",
+    ".claude",
+    ".git",
+    ".github",
+    ".ruby-lsp",
+    ".vscode",
+    "log",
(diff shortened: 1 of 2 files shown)


bnferguson:
I have signed the CLA!

Issue #803, opened 2026-05-12
`index_workspace` fails with `FileError` on the `rails` meta-gem


andrewmcodes:
## Summary

Following the basic usage example in the README, `Rubydex::Graph#index_workspace` returns a `FileError` for the `rails` gem because the gem is a meta-gem and ships without a `lib/` directory. RubyGems still resolves a default require path of `lib` for it, and Rubydex treats the missing directory as a hard error.

## Reproduction

In a Rails 8 application's console:

```ruby
graph = Rubydex::Graph.new
# => #<Rubydex::Graph:0x000000013fcad8f8 @workspace_path="/path/to/app">

graph.encoding = "utf16"
# => "utf16"

graph.index_workspace
# => ["FileError: Path `/path/to/app/.bundle/ruby/4.0.0/gems/rails-8.0.2.1/lib` does not exist"]
```

## Expected behavior

`index_workspace` should successfully index the workspace and its dependencies. The `rails` meta-gem should be tolerated (skipped, or its missing `lib/` ignored) rather than producing a `FileError`.

## Root cause

The `rails-8.0.2.1` gem installs only:

```
.../gems/rails-8.0.2.1/
├── MIT-LICENSE
└── README.md
```

The source `rails.gemspec` does not explicitly set `require_paths`. RubyGems falls back to its built-in default of `["lib"]`, defined in `Gem::Specification`'s default-value table:

```ruby
# rubygems/specification.rb
require_paths: ["lib"],
```

The setter's RDoc states the same:

> The default value is `"lib"`.

You can confirm this with no network access:

```bash
$ ruby -e 'require "rubygems"; p Gem::Specification.default_value(:require_paths)'
# => ["lib"]

$ ruby -e 'require "rubygems"; p Gem::Specification.new.require_paths'
# => ["lib"]
```

As a result, the RubyGems-generated stub spec at `specifications/rails-8.0.2.1.gemspec` writes the default back out:

```ruby
s.require_paths = ["lib".freeze]
```

…even though no `lib/` directory exists in the gem. Rubydex appears to iterate over each dependency's resolved `require_paths` and raise `FileError` when the directory is missing. Any meta-gem in this shape will hit this, and `rails` itself is almost always a direct dependency of Rails apps, so this is broadly reproducible.

## Environment

- Rubydex: `0.2.3` (`arm64-darwin`)
- Ruby: `4.0.3`
[...]

Issue #608, opened 2026-02-26
MCP server follow-up improvements


st0012:
From PR #586 review — deferred suggestions and ideas for future iterations.

## New tools

- [ ] **`resolve_at_position`** — Given file + line + column, return the declaration being referenced. The most powerful navigation primitive for LLM debugging workflows. (Morriar)
- [ ] **`find_method_references`** — Expose `graph.method_references()` as a name-based search for method call sites (e.g., "find all calls to `validate`"). Not fully resolved, but very useful. (Morriar)
- [ ] **`get_diagnostics`** — Expose `graph.all_diagnostics()` to return parse/type errors. Lets an LLM understand what's broken before editing. (Morriar)

## Tool enhancements

- [ ] **Include method parameters in `get_declaration`** — `MethodDefinition#parameters()` data exists but isn't serialized. LLMs need signatures without reading source. (Morriar)
- [x] **~~Reverse index for `find_constant_references`~~** — ~~Build `declaration_id → Vec<const_ref_offset>` at startup for O(k) lookups instead of O(n) scans.~~ Addressed by using `declaration.references()` which already tracks resolved references per declaration. (Morriar, vinistock)
- [ ] **Audit `codebase_stats` tool usefulness** — Is aggregate codebase info useful to agents, or only to tool developers? (vinistock)
- [ ] **Type-safe `Location` to prevent double `to_presentation`** — Follow up similar to PR #607 to make it impossible to call `to_presentation()` multiple times and accidentally bump line/col values. (vinistock)

## Code organization

- [ ] **Split server.rs into separate files per tool handler** — Prevents the file from growing too large and enables unit testing each handler independently. (vinistock)
- [ ] **Add unit tests alongside integration tests** — Current tests all spawn the server and do real IO. Add unit tests for tool logic, keep a few integration tests. (vinistock)

vinistock:
We should give a Ruby executable a try so that we can hook up to dependencies with `index_workspace`. People can configure the MCP launching with flexibility, so that we can avoid handling every possible version manager:

```shell
rbenv exec bundle exec rdx --mcp
mise x -- bundle exec rdx --mcp
```

PR #810, opened 2026-05-15
MCP: Fix `get_declaration` leaking internal `Debug` representation for `Unresolved` ancestors

bnferguson:
## Background

While working with the MCP Claude was complaining about debug output in `get_declaration`, digging into it it seems that when you have a ancestor that is unindexed (for instance, something defined in a gem), rather than say what the ancestor is rubydex prints the debug info. For instance for a class like `class Dog < ActiveRecord::Base; end`

you get:

```
      "ancestors": [{
        "name": "Unresolved(Name { str: Id { value: 13424087348773472403, _marker: PhantomData<...> }, parent_scope: Some(Id { value: 13486303262390686460,
  _marker: PhantomData<...> }), nesting: None, ref_count: 1 })",
        "kind": "Unresolved"
      }]
```

instead of: 

```
      "ancestors": [{
        "name": "ActiveRecord::Base",
        "kind": "Unresolved"
      }]
```

The latter seems to make a bit more sense to me but I don't know if perhaps there's reasoning I'm missing for the former. So feel free to close this (obvs) if this is way off base and not in fact a bug or a fix that makes total sense. 

## Approach

Went with a very simple/minimal fix by reusing `Graph::build_concatenated_name_from_name`, which already existed in `stats/orphan_report.rs` for the orphan report. It walks the `parent_scope` chain off the `Name` and concatenates each segment into a qualified constant path.

Buuuut to call it from the MCP crate, its visibility moves from `pub(crate)` to `pub`. 

The architecture docs note that `model/graph.rs` is where all the `Graph`-y public bits are so it seems like this my make sense to be moved over there. But in the interest of keeping this PR for a bug that may not even be a bug I kept the fix narrow. Happy to move things around. 

There's a slight behavior change with this approach too. Previously, if `graph.names().get(name_id)` returned `None`, the `?` would short-circuit `filter_map` and the ancestor would silently vanish from the response. After this change, that case renders as `"<unknown>"` (the existing fallback inside `build_concatenated_name_from_name`) and the ancestor stays in the array. TLDR; a missing name is now visible rather than swallowed. That seems preferable to me at first glance but may not be desired. 

I do believe old swallowing behavior could be maintained pretty easily. 

## Testing

Added a regression test: 

`cargo test -p rubydex-mcp get_declaration_unresolved_ancestor_renders_qualified_name`

Also if you want to test manually, index a workspace whose classes inherit from gem-provided base classes (any `ActiveRecord` model should do), and call `get_declaration`. You should no longer see the debug output but instead see `ActiveRecord::Base`

## Notes
(For the cross linking context sickos out there [that's me])

- Distinct from #707 (resolver creating `TODODeclaration`s for unresolved references), that's about why something is unresolved; this is about how the `Unresolved` variant is serialized to the MCP client.
- Distinct from #708 (singleton class ancestor chain correctness), different failure mode.
- Not currently tracked in #608 (MCP server follow-up improvements).
- Related: #803 (`index_workspace` doesn't index gems) explains why these ancestors are unresolved

### DIFF
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -196,13 +196,10 @@ fn format_ancestors(graph: &Graph, ancestors: &Ancestors) -> Vec<serde_json::Val
                     "kind": ancestor_decl.kind(),
                 }))
             }
-            Ancestor::Partial(name_id) => {
-                let name_ref = graph.names().get(name_id)?;
-                Some(serde_json::json!({
-                    "name": format!("{name_ref:?}"),
-                    "kind": "Unresolved",
-                }))
-            }
+            Ancestor::Partial(name_id) => Some(serde_json::json!({
+                "name": graph.build_concatenated_name_from_name(*name_id),
+                "kind": "Unresolved",
+            })),
         })
         .collect()
 }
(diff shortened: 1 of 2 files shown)


bnferguson:
I have signed the CLA!

rafaelfranca:
Thank you for the pull request. Waiting to see what will happen with #863.

st0012:
@bnferguson Thanks for the PR. But recently we've migrated the MCP server implementation to Ruby and removed the Rust version.

PR #818, opened 2026-05-22
Obfuscate IDs with a per-process random mask

### LINKED ISSUE #816: Obfuscate IDs
The IDs of various model objects aren't exposed on the Ruby API, but they're [public in the C API](https://github.com/Shopify/rubydex/blob/cb809fa80e7400c43213fbde657578610a8a3ce3/ext/rubydex/handle.h#L8) and Rust API ([example](https://github.com/Shopify/rubydex/blob/cb809fa80e7400c43213fbde657578610a8a3ce3/rust/rubydex/src/model/name.rs#L142)).

C or Rust library users might access these fields, notice that they _happen_ to be deterministic, and start relying on that false assumption (an example of [Hyrum's law](https://www.hyrumslaw.com/)). E.g. they might serialize them in a cache, and compare IDs against a future Rubydex run to check for changes.

This would limit our ability to evolve our choice of IDs (e.g. changing the hash function from XXH3 to something else, switching to a database incrementing id, etc.).

We should obfuscate the ID, to intentionally make it non-deterministic between runs.

## Proposed solution

Randomly generate a mask (of equal size to the ID size, currently 64 bits) on startup, and XOR all hashes with the mask before storing into "id" fields. This preserves equality checks without needing to do anything special.

## Prior art

Ruby's hash codes are randomly seeded (for a different reason, but same idea):

```console
$ ruby -e 'p nil.hash'
-3851756969988544677

$ ruby -e 'p nil.hash'
-929755904964107744
```

amomchilov:
Slop exploration of Fixes Shopify/rubydex#816.

## Why

IDs in rubydex (`DeclarationId`, `NameId`, `UriId`, `DefinitionId`,
`ConstantReferenceId`, `MethodReferenceId`, …) are derived from a stable
XXH3 hash of a canonical string. Because the hash function is
deterministic, the resulting IDs are identical across runs, which
encourages callers (Ruby gem consumers, MCP clients, log readers) to
depend on specific numeric values as if they were a stable contract.
They are not: IDs are an internal implementation detail and the indexer
should be free to change how they're derived without breaking anyone.

## What

XOR every ID produced from a string with a process-wide random mask
seeded via `std::collections::hash_map::RandomState` — no new crate
dependency. This keeps IDs stable within a single run (so equality,
hashing, and lookups still work) while making them non-deterministic
across runs, applying Hyrum's law defensively.

The mask is only applied in `From<&str>` / `From<&String>`; `Id::new`
still takes a raw `u64` unchanged so FFI round-tripping (raw `u64`
across the C boundary, reconstruct via `Id::new`) keeps working.

## Internal sites that needed adjusting

A few internal sites previously relied on raw ID ordering for
determinism and had to be adapted:

- **`follow_method_alias`**: tie-break by source position
  `(uri_id, offset)` instead of by `min(DefinitionId)`. Source position
  is stable across runs and matches the user-visible "last alias wins"
  intuition.
- **`Resolver::prepare_units`**: sort the `others` queue by URI string
  and offset instead of `(UriId, offset)`, so the cross-file processing
  order (and the "first-indexed wins" semantics that depend on it)
  stays stable across runs.
- **MCP `find_constant_references`**: sort the result references by
  source location, so the JSON output order is deterministic.
  References live in an `IdentityHashSet` keyed by obfuscated
  `ConstantReferenceId`, so iterating them directly would shuffle every
  run.
- **`declaration_search` test helper / depth-order test**: compare
  results by name (as a set or with a secondary name sort) instead of
  by raw ID, since the underlying HashMap iteration order now depends
  on obfuscated IDs.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- `cargo test` — 30 consecutive runs green during development; 5
  additional runs green on the final commit.
- The mask is genuinely process-random: verified during development
  with a small probe binary that emits the same string's ID across
  process invocations and confirmed values differ each run.

## Ruby tests

Not run from this worktree (Nix gem env not materialised here and AE
guidance is to avoid `dev up` in sub-worktrees). CI will be
authoritative — only Rust files were touched, so the Ruby unit tests
should be unaffected modulo the FFI surface, which preserves the same
`(u64) ↔ Id<T>` round-trip semantics as before.

Co-authored-by: Claude <noreply@anthropic.com>
Orchestrated-by: ae <noreply@shopify.com>

### DIFF
--- a/rust/rubydex/src/model/id.rs
+++ b/rust/rubydex/src/model/id.rs
@@ -1,5 +1,26 @@
-use std::{marker::PhantomData, num::NonZeroU64, ops::Deref};
+use std::{
+    collections::hash_map::RandomState, hash::BuildHasher, marker::PhantomData, num::NonZeroU64, ops::Deref,
+    sync::LazyLock,
+};
 use xxhash_rust::xxh3;
+
+/// Process-wide random mask XOR-ed into every ID produced from a string hash.
+///
+/// IDs are an internal implementation detail; the indexer happens to derive them from a stable
+/// XXH3 hash of a canonical string, but C and Rust callers must not rely on that. Picking a fresh
+/// random mask each time the process starts makes IDs non-deterministic across runs (an example of
+/// applying [Hyrum's law](https://www.hyrumslaw.com/) defensively), while keeping them stable
+/// within a single run so equality, hashing, and lookups still work.
+///
+/// We use [`RandomState`], which is OS-seeded, to obtain entropy without taking on an extra crate
+/// dependency. Hashing a fixed input under a freshly-constructed `RandomState` yields a value
+/// that's effectively random for the lifetime of the process.
+static ID_OBFUSCATION_MASK: LazyLock<u64> = LazyLock::new(|| RandomState::new().hash_one(0_u64));
+
+#[inline]
+fn obfuscate(hash: u64) -> u64 {
+    hash ^ *ID_OBFUSCATION_MASK
+}
 
 /// Maps a u64 hash to a `NonZeroU64` by replacing 0 with `u64::MAX`.
 /// The probability of a 64-bit hash being exactly 0 is 2^-64 (~5.4e-20),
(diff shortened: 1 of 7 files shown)


rust/rubydex/src/model/id.rs
```diff
@@ -107,4 +128,21 @@ mod tests {
...
+        let id = TestId::from("obfuscation_check");
+        assert_ne!(id.get(), raw, "IDs should be obfuscated by the per-process mask");
+    }
+
+    #[test]
+    fn mask_is_nonzero() {
+        // A zero mask would defeat obfuscation. The probability of `RandomState` producing a 0
+        // here is 2^-64, so this is effectively a sanity assertion.
+        assert_ne!(*ID_OBFUSCATION_MASK, 0, "ID mask should be a random non-zero value");
+    }
```
amomchilov, line 147: AI does some funny things.

Morriar (changes requested):
It's on purpose to allow database caching.

PR #823, opened 2026-05-26
Add namespace ancestor predicate

Morriar:
## Summary

Adds `Rubydex::Namespace#has_ancestor?(ancestor_name)` to the Ruby API.

This provides a small convenience helper for checking whether a namespace’s ancestor chain contains a class or module by fully-qualified name, without requiring callers to manually enumerate and compare `#ancestors`.

Example:

```ruby
child.has_ancestor?("Namespace::Parent")
child.has_ancestor?("Mixins::Foo")
```

Unqualified names are not matched:

```ruby
child.has_ancestor?("Parent") # => false
child.has_ancestor?("Foo")    # => false
```

## Changes

* Adds `Namespace#has_ancestor?(ancestor_name)` in `lib/rubydex/declaration.rb`
* Updates the RBI signature
* Adds coverage showing that ancestor names are matched as fully-qualified names

### DIFF
--- a/test/declaration_test.rb
+++ b/test/declaration_test.rb
@@ -305,6 +305,49 @@ class DeclarationTest < Minitest::Test
     end
   end
 
+  def test_has_ancestor
+    with_context do |context|
+      context.write!("file1.rb", <<~RUBY)
+        module Mixins
+          module Foo; end
+          module Bar; end
+        end
+
+        module Namespace
+          class Parent
+            extend Mixins::Bar
+          end
+
+          class Child < Parent
+            include Mixins::Foo
+            prepend Mixins::Bar
+            extend Mixins::Foo
+          end
+        end
+      RUBY
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      child = graph["Namespace::Child"]
+      assert(child.has_ancestor?("Namespace::Child"))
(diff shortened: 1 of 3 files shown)


test/declaration_test.rb
```diff
@@ -305,6 +305,38 @@ class Child < Parent
...
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      child = graph["Namespace::Child"]
+      assert(child.has_ancestor?("Namespace::Child"))
+      assert(child.has_ancestor?("Namespace::Parent"))
+      assert(child.has_ancestor?("Mixins::Foo"))
+      assert(child.has_ancestor?("Mixins::Bar"))
```
st0012, line 337: Can we add a few assertions for the singleton class chain?

st0012 (approved):

test/declaration_test.rb
Morriar: Added assertions for the singleton class chain: the child singleton class, the extended module, the parent singleton class, and the parent extended module. Also kept a fully-qualified-name negative assertion.

st0012:
Do you think we can make this method take a splat arg? So `SUPERCLASSES.any? { |superclass| controller.has_ancestor?(superclass) }` can become `controller.has_ancestor?(*SUPERCLASSES)`

PR #827, opened 2026-05-27
Indexing `.rake` and `.ru` files

Morriar:
So Rubydex is aware of Rake and Rack files.

### DIFF
--- a/lib/rubydex/graph.rb
+++ b/lib/rubydex/graph.rb
@@ -17,6 +17,8 @@ module Rubydex
       "tmp",
     ].freeze
 
+    INDEXABLE_EXTENSIONS = [".rb", ".rake", ".rbs", ".ru"].freeze
+
     #: String
     attr_accessor :workspace_path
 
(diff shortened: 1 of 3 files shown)


lib/rubydex/graph.rb
```diff
@@ -17,6 +17,8 @@ class Graph
       "tmp",
     ].freeze
 
+    INDEXABLE_RUBY_EXTENSIONS = [".rake", ".rb", ".ru"].freeze
```
st0012: Why don't we allow `.rbs` on the Ruby API side?

amomchilov: microopt:


```suggestion
    INDEXABLE_RUBY_EXTENSIONS = [ ".rb", ".rake", ".ru"].freeze
```

amomchilov (approved):

lib/rubydex/graph.rb
Morriar: You're right, we should accept `rbs` here as well 👍

PR #837, opened 2026-05-29
Expose mixin locations

Morriar:
## Summary

* Add `Mixin#location` to the Ruby API.
* Track mixin offsets in Rust so the location covers the full `include` / `prepend` / `extend` expression, not just the referenced constant.
* Add Ruby API signatures and coverage for include, prepend, and extend locations.


amomchilov (approved):

vinistock:
As discussed on Slack, I think we don't need to pay the memory price to remember the full mixin's location. Especially considering that someone can do multiple operations in a single line, which diminishes the relevance of the mixin location vs the constant reference location.

For example:

```ruby
class Foo
  include Bar, Baz, Qux
end
```

PR #840, opened 2026-06-01
Make MCP executable recover on non-FHS Linux

st0012:
## Context

`rubydex_mcp` now ships as a precompiled Rust executable inside the gem and is launched through the Ruby wrapper at `exe/rubydex_mcp`. On normal platforms, the wrapper can `exec` that packaged binary directly.

Nix-style Linux environments can break that direct launch. The binary is present and executable, but the kernel may still raise `ENOENT` because the ELF interpreter path baked into the binary, for example `/lib64/ld-linux-x86-64.so.2`, does not exist in the host filesystem. From an MCP client, this looks like the server cannot start even though the gem installed correctly.

## Change

This keeps the normal path unchanged: the wrapper still tries to `exec` the packaged MCP binary first.

If that direct `exec` raises `Errno::ENOENT`, the wrapper now looks for the dynamic loader already mapped into the current Ruby process through `/proc/self/maps`. When it finds that loader, it re-execs the packaged MCP binary through it.

The launcher logic now lives in `Rubydex::MCPServerLauncher`, keeping `exe/rubydex_mcp` small and keeping these process-launch details out of the main Ruby API.

### DIFF
--- a/exe/rubydex_mcp
+++ b/exe/rubydex_mcp
@@ -1,17 +1,8 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require "rbconfig"
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 
-host_os = RbConfig::CONFIG.fetch("host_os")
-executable = host_os.match?(/mswin|mingw|cygwin/) ? "rubydex_mcp.exe" : "rubydex_mcp"
-binary = File.expand_path("../lib/rubydex/bin/#{executable}", __dir__)
+require "rubydex/mcp_server_launcher"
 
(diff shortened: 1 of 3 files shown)


st0012:
* **#840** <a href="https://app.graphite.com/github/pr/Shopify/rubydex/840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rubydex/840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rafaelfranca:
This can be closed

PR #841, opened 2026-06-02
Make Graph#index_source language_id optional

st0012:
## Problem

`Graph#index_source` currently requires callers to always pass a `language_id`, even though Rubydex can infer the language from the URI in the same way file indexing does. This makes callers repeat extension-based dispatch logic and adds friction for common in-memory indexing paths.

This follows from the API discussion in https://github.com/shop/world/pull/775250/changes#r3325323365.

## Changes

- Made the Ruby API accept an omitted or `nil` `language_id`.
- Moved language inference for omitted `language_id` into the Rust FFI boundary, defaulting `.rbs` URIs to RBS and everything else to Ruby.
- Updated the RBI signature and `index_source` tests to cover inference, explicit overrides, invalid language IDs, and the resulting graph declarations.

```ruby
# Before
graph.index_source(uri, source, language_id)

# After
graph.index_source(uri, source)
graph.index_source(uri, source, language_id) # explicit override still works
```

### DIFF
--- a/rust/rubydex-sys/src/graph_api.rs
+++ b/rust/rubydex-sys/src/graph_api.rs
@@ -562,13 +562,15 @@ pub enum IndexSourceResult {
     UnsupportedLanguageId = 4,
 }
 
-/// Indexes source code from memory using the specified language.  Returns `IndexSourceResult::Success` on success
-/// or a specific error variant if string conversion or language lookup fails.
+/// Indexes source code from memory using the specified language. If `language_id` is null, infers the language from
+/// the URI using the same default as file indexing. Returns `IndexSourceResult::Success` on success or a specific error
+/// variant if string conversion or language lookup fails.
 ///
 /// # Safety
 ///
 /// - `pointer` must be a valid `GraphPointer` previously returned by this crate.
-/// - `uri` and `language_id` must be valid, null-terminated UTF-8 strings.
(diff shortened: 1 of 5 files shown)


rust/rubydex-sys/src/graph_api.rs
```diff
@@ -562,13 +562,20 @@ pub enum IndexSourceResult {
     UnsupportedLanguageId = 4,
 }
 
-/// Indexes source code from memory using the specified language.  Returns `IndexSourceResult::Success` on success
-/// or a specific error variant if string conversion or language lookup fails.
+fn language_from_uri(uri: &str) -> LanguageId {
+    let extension = Path::new(uri).extension();
+    extension.map_or(LanguageId::Ruby, LanguageId::from)
+}
```
alexcrocha: Should we turn this into a `LanguageId` method? There's one other usage at https://github.com/Shopify/rubydex/blob/88a0d4659b01f1ad4ff9855005365d81f8294252/rust/rubydex/src/indexing.rs#L103

alexcrocha (approved):

rust/rubydex-sys/src/graph_api.rs
st0012: Updated 👍

PR #848, opened 2026-06-05
Prepare Rust crates for publication

st0012:
## Summary

- Add crates.io metadata and crate-local READMEs for the `rubydex`, `rubydex-sys`, and `rubydex-mcp` crates.
- Add `rust-version` to `rubydex-sys` for consistency with the rest of the workspace.
- Make internal `rubydex` dependencies publishable by pairing local `path` dependencies with `version = "0.2.5"`, matching the Ruby gem version.

## Notes

This keeps local workspace and Ruby gem builds using path dependencies, while making the manifests valid for crates.io publication. For the first publish, `rubydex` must be published first. After the crates.io index sees `rubydex 0.2.5`, `rubydex-sys` and `rubydex-mcp` can be published.

## Test plan

- `PATH="/Users/hung-wulo/.cargo/bin:$PATH" cargo check --workspace --locked`
- `PATH="/Users/hung-wulo/.cargo/bin:$PATH" cargo test --workspace --locked`
- `PATH="/Users/hung-wulo/.cargo/bin:$PATH" cargo publish --dry-run -p rubydex --locked --allow-dirty`

### DIFF
--- a/rust/rubydex-sys/Cargo.toml
+++ b/rust/rubydex-sys/Cargo.toml
@@ -1,14 +1,21 @@
 [package]
 name = "rubydex-sys"
-version = "0.1.0"
+version = "0.2.5"
 edition = "2024"
+rust-version = "1.89.0"
 license = "MIT"
+description = "C FFI bindings for the Rubydex Ruby code indexing engine."
+homepage = "https://github.com/Shopify/rubydex"
+repository = "https://github.com/Shopify/rubydex"
+readme = "README.md"
+keywords = ["ruby", "ffi", "static-analysis"]
+categories = ["development-tools", "external-ffi-bindings"]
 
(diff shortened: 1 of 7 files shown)


rust/rubydex-sys/Cargo.toml
```diff
@@ -2,13 +2,20 @@
 name = "rubydex-sys"
 version = "0.1.0"
```
rafaelfranca: Can we match the crates version with the gem version?

st0012: Updated. But we haven't decided how to bump versions for crates as we could be releasing Ruby API changes and not touching Rust at all 🤔

rafaelfranca (approved):

PR #859, opened 2026-06-12
Build Linux MCP executable with static musl

st0012:
## Summary

This prototypes building the packaged `rubydex_mcp` executable as a fully static musl binary for Linux precompiled gem builds.

The Ruby native extension keeps using the normal platform-specific build path. Only the standalone MCP executable is cross-built to musl during `RELEASE=true` Linux packaging, then verified before it is copied into the gem.

## Why

A standalone Linux executable can fail to start on environments that do not provide the dynamic loader path baked into the binary. For the MCP server, we can avoid that class of startup failure by packaging a fully static executable with no ELF interpreter segment.

This is intentionally narrower than full musl support for the Ruby native extension. The extension is loaded by a running Ruby process and has different platform constraints.

## What changed

- Install `musl-tools` and the matching Rust musl target in Linux cibuildgem compile jobs.
- Add `Rubydex::MCPExecutableBuild` to choose the MCP executable target and build command.
- Split the MCP Cargo build from the Ruby extension build only when the MCP target differs from the extension target.
- Verify the Linux musl MCP artifact has no `PT_INTERP` entry before packaging it.

## Testing

- `shadowenv exec -- ruby -Itest -Ilib test/mcp_executable_build_test.rb`
- `shadowenv exec -- bundle exec rubocop ext/rubydex/extconf.rb lib/rubydex/mcp_executable_build.rb test/mcp_executable_build_test.rb`
- `git diff --check -- .github/workflows/cibuildgem.yaml ext/rubydex/extconf.rb lib/rubydex/mcp_executable_build.rb test/mcp_executable_build_test.rb`

## Notes

This is a draft PR to let CI prove whether the Linux packaging job can produce the fully static MCP artifact. If CI passes, the next useful check is to test the CI-built gem or executable artifact in a target environment without publishing a release.

### DIFF
--- a/.github/workflows/cibuildgem.yaml
+++ b/.github/workflows/cibuildgem.yaml
@@ -26,6 +26,12 @@ jobs:
         with:
           ruby-version: "3.2.9"
           bundler-cache: true
+      - name: "Install musl build tools"
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y musl-tools
+          rustup target add ${{ runner.arch == 'ARM64' && 'aarch64-unknown-linux-musl' || 'x86_64-unknown-linux-musl' }}
       - name: "Install cargo-about"
         run: cargo install cargo-about --features cli
       - name: "Run cibuildgem"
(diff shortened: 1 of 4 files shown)


rafaelfranca:
This can be closed, we wrote the mcp in Ruby now

Issue #858, opened 2026-06-12
Diagnose superclass set in one definition, after another definition left it implicit


amomchilov:
Hard to name this, but let me give an example.

**foo.rb**:

```ruby
class Foo < Bar; end
```

**foo/version.rb**

```ruby
class Foo
  VERSION = "1.2.3"
end
```

Ruby will set the superclass to whatever it loads first at runtime. That depends on the load order, which we don't want Rubydex to do. So we deviate from the runtime behaviour, picking `< Bar` even if we see `foo/version.rb` before `foo.rb`. This compromise makes makes sense in the multiple file case, to ignore the load order race.

But within a single file, this will **always** be wrong:

```ruby
class Foo; end
class Foo < String; end # ❌ TypeError: superclass mismatch for `class Foo` 
```

Notably, the inverse is fine:

```ruby
class Foo < String; end
class Foo; end # No `< String`, but that's ok
```

## Proposal

Add a diagnostic that checks for contradictory superclass definitions with each file.

Within the scope of a single file, we can eagerly assume that no superclass clause implies `< Object`, and then check that any further definition doesn't contradict that. Notably, this shouldn't change the interfile behaviour we havem where a missing superclass clause does *not* imply `< Object`.

Issue #857, opened 2026-06-12
Diagnose for contradictory superclasses on the same class


amomchilov:
Should error:

```ruby
class Foo < Bar; end
class Foo < NotBar; end
```

vinistock:
Duplicate of #354

amomchilov:
Spoke offline, chose to keep an issue to track this independently, but linked to the other issue.

PR #860, opened 2026-06-12
Add more tests for superclass resolution behaviour

amomchilov:
Documents existing behaviour as-is.

Adds some pending tests that will need to be fixed in the future for:
1. https://github.com/Shopify/rubydex/issues/857
2. https://github.com/Shopify/rubydex/issues/858

### DIFF
--- a/rust/rubydex/src/resolution_tests.rs
+++ b/rust/rubydex/src/resolution_tests.rs
@@ -983,6 +983,22 @@ mod superclass_tests {
     use super::*;
 
     #[test]
+    fn missing_superclass_implies_object() {
+        let mut context = graph_test();
+        context.index_uri("file:///foo.rb", {
+            r"
+            class Foo
+            end
+            "
+        });
+        context.resolve();
+
+        assert_no_diagnostics!(&context);
+
+        assert_ancestors_eq!(context, "Foo", ["Foo", "Object", "Kernel", "BasicObject"]);
+    }
+
+    #[test]
     fn linearizing_super_classes() {
         let mut context = graph_test();
         context.index_uri("file:///foo.rb", {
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution_tests.rs
```diff
@@ -982,6 +982,139 @@ mod constant_alias_tests {
...
+
+            assert_diagnostics_eq!(
+                context,
+                &[
+                    "contradictory-superclasses: Attempted to define `Foo` with contradictory superclasses (`Object` in file:///foo.rb:1:1-1:10 and `Bar` in file:///foo.rb:3:1-3:16)."
+                ]
+            );
+        }
+
+        // Contrast with this case, which Ruby allows::
```
vinistock, line 1101: I think this is already encoded in the other tests, so I would remove it from this one.

vinistock (approved):

rust/rubydex/src/resolution_tests.rs
amomchilov: Which one? I don't see it

vinistock: Is this not the same as C1 in `superclass_is_load_order_independent`? Also, I would split each thing into their own test rather than checking multiple things in each example.

amomchilov: 1. It's exercising the same code, but it's making a different claim about it. I tried to explain it on the docs of this test:

    > Unlike the `superclass_is_load_order_independent()` test which covers contradictions across files,
    > contradictions within the same file are deterministically an issue.

2. What I'm trying to emphasize here is the contrast that one order is deterministically always an issue, whereas the other order is deterministically always okay. And that it's not automatically an issue to not have the superclass clause repeated the same, but that the order is what matters (within the scope of a single file)

vinistock: Okay, that's fine. What do you think about splitting everything into separate examples though?

PR #863, opened 2026-06-16
Replace `rubydex_mcp` server with Ruby implementation

st0012:
## Summary

Adds a Ruby implementation of the MCP server, exposed through `rdx --mcp`, with a small internal MCP/JSON-RPC protocol layer.

The existing Rust MCP crate and build plumbing are left in place. If this PR is accepted, I'll remove it in a follow-up PR.

## Parity with the Rust MCP server

Matches the existing Rust MCP server for:

- tool names and schemas: `search_declarations`, `get_declaration`, `get_descendants`, `find_constant_references`, `get_file_declarations`, `codebase_stats`
- JSON payload shapes, pagination defaults and caps, kind filtering, exact/fuzzy search modes, relative path formatting, declaration details, descendants, file declarations, and stats
- constant-reference behavior for classes/modules/constants, with methods returning an empty reference list

Known differences:

- startup indexing follows the Ruby API workspace indexing behavior.
- unresolved partial ancestors are not exposed by the current Ruby API, so the Ruby MCP server only reports resolved ancestors.

## Implementation notes

- The server owns JSON-RPC dispatch, request workers, and serialized stdout writes.
- Tool execution is instance-based: tools are initialized with the indexed graph and derive workspace paths from `Graph#workspace_path`.
- Shared tool helpers live in `BaseTool`, keeping path formatting, pagination, declaration lookup, and response formatting out of server code.

## Follow-up work

- Remove the Rust MCP implementation in a separate PR.
- Open a follow-up PR for Bundler integration.


exe/rubydex_mcp
```diff
@@ -1,17 +1,52 @@
...
+  end
+
+  parser.on("-V", "--version", "Print version") do
+    puts "rubydex_mcp #{Rubydex::VERSION}"
+    exit
+  end
+end
+
+begin
+  option_parser.parse!(ARGV)
```
vinistock: FWIW, one of the things I regret in the LSP was using `parse!` instead of the non-raising version `parse`. It makes it more difficult to make backwards incompatible changes to accepted flags, even for development related stuff.

exe/rubydex_mcp
```diff
@@ -1,17 +1,52 @@
...
 
-exec(binary, *ARGV)
+if ARGV.length > 1
+  warn("error: unexpected argument '#{ARGV[1]}' found")
+  warn
+  warn(option_parser)
+  exit(2)
+end
+
+path = ARGV.fetch(0, ".")
```
vinistock: Better to use absolute paths when possible.

```suggestion
path = ARGV.fetch(0, Dir.pwd)
```

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+require "uri"
+
+require "rubydex"
+require "rubydex/mcp_server/protocol"
+require "rubydex/mcp_server/tools/codebase_stats_tool"
+require "rubydex/mcp_server/tools/find_constant_references_tool"
+require "rubydex/mcp_server/tools/get_declaration_tool"
+require "rubydex/mcp_server/tools/get_descendants_tool"
+require "rubydex/mcp_server/tools/get_file_declarations_tool"
+require "rubydex/mcp_server/tools/search_declarations_tool"
```
vinistock: So we don't have to keep adding to the list.
```suggestion
Dir.glob("rubydex/mcp_server/tools/**/*.rb").each { |f| require f }
```

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+
+module Rubydex
+  module MCPServer
+    SERVER_INSTRUCTIONS = <<~TEXT
+      Rubydex provides semantic Ruby code intelligence.
+
+      ONLY use these tools for Ruby files (.rb, .rbi, .rbs) -- never for Rust, JavaScript, or other languages.
```
vinistock: Now I'm wondering if the reference to other language names will make it better or worse at picking up the MCP 😂

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      Use these tools INSTEAD OF Grep when working with Ruby code structure.
+
+      Decision guide:
+      - Know a name? -> search_declarations (fuzzy search by name)
+      - Have an exact fully qualified name? -> get_declaration (full details with docs, ancestors, members)
+      - Need reverse hierarchy? -> get_descendants (what inherits from this class/module)
+      - Refactoring a class/module/constant? -> find_constant_references (all precise usages across codebase)
+      - Exploring a file? -> get_file_declarations (structural overview)
+      - Want general statistics? -> codebase_stats (size and composition)
```
vinistock: Is this maybe a better fit for the individual tool descriptions?

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      Typical workflow: search_declarations -> get_declaration -> find_constant_references.
```
vinistock: I'd remove this. I can see LLMs getting eager to satisfy this workflow even if it doesn't make sense for what they're trying to discover.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      Fully qualified name format: "Foo::Bar" for classes/modules/constants, "Foo::Bar#method_name" for instance methods.
+
+      Pagination: tools that may return a high number of results include `total` for pagination. When `total` exceeds the number of returned items, use `offset` to fetch the next page.
+
+      Use Grep instead for: literal string search, log messages, comments, non-Ruby files, or content search rather than structural queries.
+    TEXT
+    TOOLS = [
```
vinistock: Can we auto-track tools with a parent class instead? Something like:

```ruby
class Tool
  @tools = []

  def self.inherited(base)
    @tools << base
  end
end
```

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+        payload.to_json(*args)
+      end
+    end
+
+    class State
+      #: (String) -> void
+      def initialize(root_path)
+        @root_path = root_path
+        @mutex = Mutex.new
+        @graph = nil
```
vinistock: I'd instantiate the graph here already to avoid having it be nilable. It doesn't do any work until you tell it to index.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+        Thread.new do
+          graph = Graph.new(workspace_path: @root_path)
+          errors = graph.index_workspace
+          errors.each { |error| warn("Indexing error: #{error}") }
+          graph.resolve
+          warn("Rubydex indexed #{graph.documents.count} files, #{graph.declarations.count} declarations")
+
+          @mutex.synchronize do
+            @graph = graph
+          end
```
vinistock: Initializing the graph would also eliminate the need for this, I think.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+        rescue Exception => e # rubocop:disable Lint/RescueException
+          warn("Rubydex indexing failed: #{e.message}")
+          @mutex.synchronize do
+            @error = e.message
+          end
```
vinistock: We explicitly design indexing to never raise and instead return a list of errors. I don't think we would ever hit this code path.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+          file_path
+        else
+          File.join(root_path, file_path)
+        end
+        canonical_target = File.realpath(absolute_target)
+        graph.documents.find do |document|
+          path = file_path_for_uri(document.uri)
+          path && File.expand_path(path) == canonical_target
+        end
+      rescue SystemCallError
```
vinistock: Why are we rescuing `SystemCallError` here?

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,239 @@
...
+    TOOLS = [
+      SearchDeclarationsTool,
+      GetDeclarationTool,
+      GetDescendantsTool,
+      FindConstantReferencesTool,
+      GetFileDeclarationsTool,
+      CodebaseStatsTool,
+    ].freeze
+
+    class Error
```
vinistock: I think this belongs in `protocol`, like `Response`.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+        }
+      end
+    end
+
+    class StdioTransport
+      #: (Server, ?IO, ?IO) -> void
+      def initialize(server, input: $stdin, output: $stdout)
+        @server = server
+        @input = input
+        @output = output
```
vinistock: If this is using JSON RPC, we need to turn on sync and binmode on the pipes.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+          {
+            name: @tool_name,
+            description: @description,
+            inputSchema: input_schema,
+          }
+        end
+      end
+    end
+
+    class Server
```
vinistock: IMO, it would be easier to follow the architecture if we merged `Server` and `State` into a single object and kept things like error codes and convenience objects inside protocol.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+        @server_state = server_state
+      end
+
+      PARSE_ERROR = -32_700
+      INVALID_REQUEST = -32_600
+      METHOD_NOT_FOUND = -32_601
+      INVALID_PARAMS = -32_602
+      INTERNAL_ERROR = -32_603
+
+      #: (Hash | Array | untyped) -> Hash | Array[Hash]?
```
vinistock: Can the MCP send asynchronous messages to the client like an LSP can? If so, we're making the same mistake we originally made in the LSP here: implementing a `handle` method that returns the response to the sent to the client.

That approach makes it really difficult to have a single operation (like a tool call) both send an asynchronous notification and return a response.

That's why in the LSP we have an outgoing queue with a dispatcher thread. Any request/notification can enqueue as many messages as desired to be sent to the client.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+        end
+
+        unless request.is_a?(Hash)
+          return error_response(nil, INVALID_REQUEST, "Invalid Request", data: "Request must be a hash")
+        end
+
+        has_id = request.key?(:id) || request.key?("id")
+        id = request.key?(:id) ? request[:id] : request["id"]
+        method = request[:method] || request["method"]
+        params = request.key?(:params) ? request[:params] : request["params"]
```
vinistock: Do we need to check if it's string or symbol? If we control the JSON parsing, we should be able to use `symbolize_names: true` and standardize everything on symbols.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+      rescue StandardError => e
+        has_id ? error_response(id, INTERNAL_ERROR, "Internal error", data: e.message) : nil
+      end
+
+      private
+
+      #: (Hash) -> Hash
+      def call_tool(params)
+        tool_name = params[:name] || params["name"]
+        tool = TOOLS.find { |candidate| candidate.to_h.fetch(:name) == tool_name }
```
vinistock: Instead of an array, we should probably use a hash and make this a simple lookup.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,202 @@
...
+      #: (Server, ?IO, ?IO) -> void
+      def initialize(server, input: $stdin, output: $stdout)
+        @server = server
+        @input = input
+        @output = output
+      end
+
+      #: -> void
+      def open
+        @input.each_line do |line|
```
vinistock, line 136: Is this using JSONRPC? The separator should be `\r\n\r\n` and not a single line. So you need something like `gets("\r\n\r\n")`. And the first line you get is the content length, so after the gets you need to read the specified number of bytes.

For example, something like this:

```ruby
headers = io.gets("\r\n\r\n")
raise unless headers

length = headers[/Content-Length: (\d+)/i, 1]
raise unless length

raw_message = io.read(length)
message = JSON.parse(raw_message)
```

st0012: According to the latest [spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports), stdio messages do NOT include content-length. Each message takes exactly one line. You can see the official MCP SDK is [implemented this way](https://github.com/modelcontextprotocol/ruby-sdk/blob/5518e8db032517896e0025119a9073f4aa946ba4/lib/mcp/server/transports/stdio_transport.rb#L23-L25) too.

lib/rubydex/mcp_server/protocol.rb
st0012: In MCP, requests/notifications behave differently depending on the transport. In stdo, both response and notification are dispatched directly to the stdout, and there's no similar outgoing queue concept in the [official SDK implementation](https://github.com/modelcontextprotocol/ruby-sdk/blob/main/lib/mcp/server/transports/stdio_transport.rb).

exe/rubydex_mcp
vinistock, line 1: I'd vote to put this in the main executable. Something like `bundle exec rdx --mcp`.

It's easier to discover all of the included functionality if it's not spread around different executables.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+        end
+      end
+
+      class << self
+        attr_reader :tools
+
+        #: (Class) -> void
+        def inherited(tool)
+          tools << tool
+          @tools_by_name = nil
```
vinistock: Do we expect `tools_by_name` to be invoked before we're done loading the code? Or in other words, why do we need to bust the cache for each rule being defined?

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+
+        unless params.nil? || params.is_a?(Hash)
+          return error_response(id, INVALID_PARAMS, "Invalid params", data: "Method parameters must be an object or null")
+        end
+
+        result = case method
+        when "initialize"
+          {
+            # This is not the latest protocol version. But as of implementation,
+            # this is the default protocol version the official MCP Ruby SDK uses.
```
vinistock: Why is the version of the protocol being used by the SDK matter?

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+
+    class StdioTransport
+      #: (Server, ?IO, ?IO) -> void
+      def initialize(server, input: $stdin, output: $stdout)
+        @server = server
+        @input = input
+        @output = output
+        @input.binmode
+        @output.binmode
+        @output.sync = true
```
vinistock, line 131: I think we need sync for `@input` too, no?

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+      end
+
+      #: -> void
+      def open
+        @input.each_line do |line|
+          response = @server.handle(JSON.parse(line, symbolize_names: true))
+          next unless response
+
+          @output.puts(JSON.generate(response))
+          @output.flush
```
vinistock: AFAIK, there should not be a need to manually `flush` if `sync` is enabled.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,170 @@
...
+          @mutex.synchronize do
+            @index_finished = true
+          end
+          warn("Rubydex indexed #{@graph.documents.count} files, #{@graph.declarations.count} declarations")
+        end
+      end
+
+      #: -> Graph | Error
+      def graph_or_error
+        @mutex.synchronize do
```
vinistock: Do we need a mutex for this? We are synchronizing on setting a boolean instance variable, which nearly instantaneous. I'm not sure I understand what the mutex guards against.

lib/rubydex/mcp_server.rb
```diff
@@ -0,0 +1,170 @@
...
+      #: -> Thread
+      def spawn_indexer
+        Thread.new do
+          errors = @graph.index_workspace
+          errors.each { |error| warn("Indexing error: #{error}") }
+          @graph.resolve
+          @mutex.synchronize do
+            @index_finished = true
+          end
+          warn("Rubydex indexed #{@graph.documents.count} files, #{@graph.declarations.count} declarations")
```
vinistock: Another learning from the LSP: if the idea for this is to print log/debug messages, but this part of the code doesn't have a way to send proper messages to the client, then something is misplaced in the architecture.

Ideally, we should never print to stderr and should instead use proper log messages.

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+        end
+
+        #: -> Hash[String, Class]
+        def tools_by_name
+          @tools_by_name ||= tools.to_h { |tool| [tool.tool_name, tool] }
+        end
+
+        #: (?String) -> String
+        def tool_name(value = UNSET)
+          @tool_name = value unless value.equal?(UNSET)
```
vinistock: Do we need this `UNSET` mechanism? Or could we get away with an attr_accessor?

lib/rubydex/mcp_server/protocol.rb
```diff
@@ -0,0 +1,238 @@
...
+          {
+            name: tool_name,
+            description: description,
+            inputSchema: input_schema,
+          }
+        end
+      end
+    end
+
+    class Server
```
vinistock: Are we re-opening the `Server` class in two places? If so, can we move server specific things, like `handle`, `call_tool` into `mcp_server.rb` so that everything is centralized?

I think the protocol bits need to be server agnostic. For example, `StdioTransport#open` could accept a block instead of keeping a handle to the server, so that we keep the server in control of the message handling flow.

Something like

```ruby
class Server
  def main_loop
    @io.open do |message|
      # handling the message decides which tools to call
      # and then uses @io to send the responses back
      handle(message)
    end
  end
end
```

lib/rubydex/mcp_server/protocol.rb
vinistock: Sorry, I think there's something missing still. In an LSP, you can also print directly to stdout from requests and that would work, but it's not an ideal architecture as your output gets spread across the logic implementation.

Also, with a single thread, it means you'll be blocked on IO more often then you'd like.

The outgoing queue is not a protocol specific thing, it's an implementation technique to reduce the time blocked on IO and centralize messages.

exe/rubydex_mcp
rafaelfranca: Agree

rafaelfranca (approved):

Issue #811, opened 2026-05-18
SelfReceiver `.expect` panic via constant alias to re-opened namespace (follow-up to #663)


shiroinock:
After seeing the rubydex session at RubyKaigi 2026 I tried adopting it on a mid-sized Rails codebase, and ran into the panic below at the first `Graph#resolve`. Filing this in case the trigger is interesting — and to check whether a defensive fix is welcome before I send a PR.

## Summary

`Resolver::handle_remaining_definitions` still has two `.expect("SelfReceiver definition should have a declaration")` call sites that can be hit at fresh resolve time when a class definition exists but its enclosing namespace fails name resolution:

- `rust/rubydex/src/resolution.rs:381` — `Definition::InstanceVariable` / `Definition::Method` arm
- `rust/rubydex/src/resolution.rs:525` — `Definition::MethodAlias` arm

This is the same panic message as #663 but reached through a different input path that #656's multi-level Todo-declaration mechanism does not cover.

## Reproduction

Tested against `main` at `cb809fa` (`Rubydex::VERSION = "0.2.3"`).

`Gemfile`:

```ruby
source 'https://rubygems.org'
gem 'rubydex',  '0.2.3'
gem 'minitest', '5.25.4'
gem 'aasm',     '5.0.3'
```

`repro.rb`:

```ruby
require 'rubydex'

def gem_lib(name)
  Gem::Specification.find_by_name(name).full_gem_path + '/lib'
end

g = Rubydex::Graph.new
g.encoding = 'utf8'
[gem_lib('minitest'), gem_lib('aasm')].each do |dir|
  Dir.glob("#{dir}/**/*.rb").each do |path|
    g.index_source(path, File.read(path), 'ruby')
  end
end
g.resolve
```

```
$ RUST_BACKTRACE=1 bundle exec ruby repro.rb
thread '<unnamed>' panicked at rubydex/src/resolution.rs:381:42:
SelfReceiver definition should have a declaration
   5: rubydex::resolution::Resolver::handle_remaining_definitions
             at rubydex/src/resolution.rs:381:42
   6: rubydex::resolution::Resolver::resolve
             at rubydex/src/resolution.rs:136:14
```

Neither `minitest` nor `aasm` alone reproduces it — both gems are required. A delta-debug pass over a real workspace narrowed the trigger to exactly this pair.

## Mechanism

`aasm-5.0.3/lib/aasm/minitest.rb` is:

```ruby
Minitest = MiniTest unless defined? Minitest
[...]

PR #867, opened 2026-06-18
Fix panic resolving singleton methods on non-namespace owners

paracycle:
## Summary

`Graph::find_self_receiver_declaration` resolves the owner of a `def self.foo`-style singleton method and then `unwrap()`ed both the declaration lookup and `as_namespace()`. That hard-codes the assumption that the owner always exists and is always a namespace.

That assumption breaks for valid Ruby where a `class`/`module` reopens a same-named **constant** or **constant alias** that was not promoted to a namespace. In that case the owning declaration stays a non-namespace (`Constant`/`ConstantAlias`), `as_namespace()` returns `None`, and the `unwrap()` panics:

```
thread 'main' panicked at rubydex/src/model/graph.rs:457:14:
called `Option::unwrap()` on a `None` value
```

The fix resolves the owner and singleton class with `?` so the lookup returns `None` instead of panicking — consistent with the sibling `find_singleton_method_visibility_declaration`, which already handles the same situation gracefully. A method whose owner can't be resolved to a namespace simply has no owning declaration (an orphan), which is an already-expected outcome elsewhere in resolution.

## How to reproduce (without any new tooling)

This is a latent bug in core graph code, reachable through any consumer that resolves definitions to declarations via the public `Graph::definition_to_declaration_id` — for example the built-in DOT exporter (`--dot`), which calls it for every definition.

Minimal valid Ruby that triggers it:

```ruby
class Bar
end

Aliased = Bar      # non-promotable: produces a ConstantAlias declaration

class Aliased      # valid Ruby: reopens Bar
  def self.foo; end
end
```

`Aliased = Bar` assigns a constant to another constant, producing a `ConstantAlias` declaration (a constant read value is not promotable). `class Aliased` reopens `Bar` at runtime, but rubydex attaches the class definition to the existing `ConstantAlias` declaration without promoting it to a namespace. The `def self.foo` then has a `SelfReceiver` owner whose declaration is not a namespace.

To observe the crash on `main`, run any path that resolves definitions over that file, e.g.:

```
rubydex_cli /path/to/dir --dot
```

or iterate definitions in a test:

```rust
for definition in graph.definitions().values() {
    let _ = graph.definition_to_declaration_id(definition);
}
```

Before this change that panics; after it, it returns `None` for the singleton method and completes normally.

## Test

Adds `singleton_method_in_non_namespace_owner_does_not_panic`, which indexes the snippet above, asserts the declaration stays a `ConstantAlias`, and resolves every definition to its declaration (mirroring the DOT exporter). The test panics on the unpatched code and passes with the fix.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -447,20 +447,21 @@ impl Graph {
     }
 
     /// Looks up the declaration for a `SelfReceiver` method/alias through the singleton class.
+    ///
+    /// Returns `None` when the owner cannot be resolved to a namespace with a singleton class. This
+    /// can happen when the enclosing construct resolved to a non-namespace declaration (e.g. a
+    /// constant or constant alias that a same-named `class`/`module` reopened without promotion), in
+    /// which case the method has no owning declaration.
     fn find_self_receiver_declaration(&self, def_id: DefinitionId, member_str_id: StringId) -> Option<&DeclarationId> {
         let owner_decl_id = self.definition_id_to_declaration_id(def_id)?;
         let singleton_id = self
             .declarations
-            .get(owner_decl_id)
-            .unwrap()
-            .as_namespace()
-            .unwrap()
+            .get(owner_decl_id)?
+            .as_namespace()?
             .singleton_class()?;
         self.declarations
-            .get(singleton_id)
-            .unwrap()
-            .as_namespace()
-            .unwrap()
+            .get(singleton_id)?
+            .as_namespace()?
             .member(&member_str_id)
     }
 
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -1568,6 +1569,41 @@ mod tests {
...
+        context.resolve();
+
+        // The declaration stays a non-namespace constant alias.
+        assert_declaration_kind_eq!(context, "Aliased", "ConstantAlias");
+
+        // Mirrors what consumers like the DOT exporter do: resolve every definition to its
+        // declaration. This previously panicked on the singleton method's non-namespace owner.
+        for definition in context.graph().definitions().values() {
+            let _ = context.graph().definition_to_declaration_id(definition);
+        }
```
vinistock, line 1601: Do we need this part for the test? I think the panic is only during resolve.

In fact, I just realized we have a stale `# Panics` comment in the `definition_to_declaration_id`, which can no longer panic.

vinistock (approved):
For future reference, also related to #811. It might be possible to consider the `class` or `module` keywords as source of truths for constant aliases with unresolved targets, but the incremental scenarios are a bit complex.

Better to fix the panic right away.

rust/rubydex/src/model/graph.rs
paracycle: Yes, it is needed. Calling `resolve()` alone did not panic (against the old codebase) without this loop.

paracycle: I removed the `# Panics` comment as well.

PR #868, opened 2026-06-18
Query the in-memory graph with Cypher

paracycle:
## Goal

Give clients a flexible, future-proof way to query the in-memory graph using **Cypher** — the de facto standard graph query language. Instead of adding a bespoke method for every traversal, this exposes the graph through a query language clients already know, so new introspection needs become queries rather than new APIs. Queries are **read-only**, run directly against the existing in-memory `Graph` (no duplication, no embedded database).

### References
- openCypher specification: https://opencypher.org/resources/
- Cypher query language (Neo4j docs): https://neo4j.com/docs/cypher-manual/current/

## Architecture

The Cypher engine itself — lexer, recursive-descent parser, AST, the tree-walking executor, values, and result formatting — lives in a separate, published crate, [`cypher-parser`](https://crates.io/crates/cypher-parser). The executor is generic over a `GraphProvider` trait, so it has no dependency on rubydex.

`rubydex` depends on `cypher-parser` and provides only the rubydex-specific pieces:
- `query::cypher::schema` — `impl GraphProvider for Graph`, the property-graph mapping.
- `query::cypher::schema_info` — the static schema description.

## How it's exposed

- **`rubydex_cli`**: `--query "<CYPHER>"`, `--schema`, `--format table|json`.
- **Ruby API** — the whole query API lives on `Rubydex::Query`:
  - `Rubydex::Query.parse(str)` → an opaque, reusable parsed query (raises `ArgumentError` on syntax errors, needs no graph).
  - `Rubydex::Query#render(graph, format = :table)` → runs a parsed query against a graph and returns the formatted output (table or JSON).
  - `Rubydex::Query.schema(format = :table)` → describes the queryable schema.
- **`rdx` command CLI**:
  ```
  rdx query <CYPHER>  [--format table|json]
  rdx query --schema  [--format table|json]
  rdx console
  ```

### Parse first, then build the graph

Both CLIs parse the query into the opaque parsed object **before** indexing/resolution, so a malformed query fails fast (~0.1s) instead of after a full workspace index:

```
parse query  ->  build graph (index + resolve)  ->  run parsed query
```

A parsed `Rubydex::Query` is reusable: parse once, run against many graphs.

## Graph schema exposed to queries

`rdx query --schema` / `rubydex_cli --schema` / `Rubydex::Query.schema` print this model:

**Node labels:** `Document`, `Definition`, `Declaration`, the grouping label `Namespace`, and declaration kind sub-labels (`Class`, `Module`, `SingletonClass`, `Method`, `Constant`, `ConstantAlias`, `GlobalVariable`, `InstanceVariable`, `ClassVariable`).

**Relationship types:** `DEFINES` (Document→Definition), `DECLARES` (Definition→Declaration), `CONTAINS` (lexical nesting), `HAS_PARENT` (direct superclass; reverse-traverse for direct subclasses, `*` for the chain), `INCLUDES`/`PREPENDS`/`EXTENDS` (mixins), `OWNS` (members), `HAS_ANCESTOR` (linearized ancestor chain, incl. modules), `HAS_DESCENDANT` (reverse of `HAS_ANCESTOR`), `REFERENCES` (Document→Declaration).

**Properties:** Declaration: `name`, `unqualified_name`, `kind`, `visibility`, `definition_count`; Definition: `kind`, `name`, `file`, `line`; Document: `uri`, `path` (file-system path), `name` (base file name).

**Supported syntax:** `MATCH` (node patterns with label disjunction `:A|B` and inline properties; relationship patterns with direction, type lists, and variable length `*min..max`), `WHERE` (`=, <>, <, <=, >, >=, CONTAINS, STARTS WITH, ENDS WITH`, `AND/OR/NOT`), `RETURN` (`DISTINCT`, `AS`, aggregates `count/collect/min/max/sum/avg`), `ORDER BY`, `SKIP`, `LIMIT`. Read-only; write clauses are intentionally unsupported.

## Try it

```bash
# Discover the model
rdx query --schema

# All classes or modules
rdx query "MATCH (n:Class|Module) RETURN n.name ORDER BY n.name"

# All (transitive) subclasses of a base class, as JSON
rdx query "MATCH (c:Class)-[:HAS_PARENT*1..]->(p {name: 'ApplicationRecord'}) RETURN DISTINCT c.name" --format json

# Count definitions per file
rdx query "MATCH (d:Document)-[:DEFINES]->(def:Definition) RETURN d.path, count(def) AS defs ORDER BY defs DESC"
```

From Ruby:
```ruby
query = Rubydex::Query.parse("MATCH (n:Class|Module) RETURN n.name")  # fails fast on bad syntax
graph = Rubydex::Graph.new
graph.index_workspace
graph.resolve
puts query.render(graph, :json)
```

## What's here

Core work:
1. A read-only Cypher engine over the in-memory graph, extracted into the standalone [`cypher-parser`](https://crates.io/crates/cypher-parser) crate (generic over `GraphProvider`).
2. rubydex's `GraphProvider` mapping plus a single-source schema catalog (`RelType` + `NODE_LABELS`/`NODE_PROPERTIES`), rendered by `schema_info`.
3. `--query`/`--schema` on `rubydex_cli` and a `rdx query` / `console` command CLI, parsing the query up front so a syntax error fails fast before indexing.

Review refinements folded in: direction-correct `HAS_PARENT`/`HAS_ANCESTOR`/`HAS_DESCENDANT` edge naming, `rdx query --schema` as a flag, a single "build graph once, then dispatch" CLI structure, extraction of `Rubydex::Query` into `query.c` and a shared `utils` helper, robust `file://` path handling, and clap validation that rejects nonsensical flag combinations.

### DIFF
--- a/exe/rdx
+++ b/exe/rdx
@@ -5,75 +5,151 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 
 require "optparse"
 
-options = {}
+USAGE = <<~TEXT
+  Usage: rdx <command> [options]
 
-option_parser = OptionParser.new do |parser|
-  parser.on("--version", "Print the gem's version") do
-    require "rubydex/version"
-    puts "v#{Rubydex::VERSION}"
-    exit
-  end
+  Commands:
+    query <CYPHER>   Run a Cypher query against the workspace graph and print the result.
+                     Use `query --schema` to describe the queryable schema (labels,
+                     relationships, properties) without indexing the workspace.
+    console          Open an interactive session with a populated graph for the current workspace
+    mcp [PATH]       Run the MCP server for AI assistants (workspace defaults to the current dir)
+    help             Show this help message
 
-  parser.on("-h", "--help", "Prints this help") do
-    puts parser
-    exit
-  end
+  Run `rdx <command> --help` for command-specific options.
+TEXT
 
-  parser.on("-i", "--interactive", "Open an interactive session with a populated graph for the current workspace") do
-    options[:interactive] = true
-  end
-
-  parser.on("--mcp", "Run the MCP server for AI assistants") do
-    options[:mcp] = true
-  end
+def abort_with_usage(message)
+  warn(message)
+  warn("")
+  warn(USAGE)
+  exit(1)
 end
 
-begin
-  arguments = option_parser.parse(ARGV)
-rescue OptionParser::ParseError => e
-  warn("error: #{e.message}")
-  warn
-  warn(option_parser)
-  exit(2)
+# Top-level --version / --help / bare invocation, handled before command dispatch. Each branch
+# performs its action and reports whether it handled the invocation, so we exit exactly once here
+# rather than from inside every branch.
+handled =
+  case ARGV.first
+  when "--version", "version"
+    require "rubydex/version"
+    puts "v#{Rubydex::VERSION}"
+    true
+  when nil, "-h", "--help", "help"
+    puts USAGE
+    true
+  else
+    false
+  end
+
+exit if handled
+
+command = ARGV.shift
+
+def with_timer(io, message)
+  io.print(message)
+  start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
+  yield
+  duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
+  io.puts(" finished in #{duration.round(2)}ms")
+end
+
+# Builds the workspace graph, sending progress messages to `progress_io`.
+def build_graph(progress_io)
+  graph = Rubydex::Graph.new
+  graph.load_config
+  with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+  with_timer(progress_io, "Resolving graph...") { graph.resolve }
+  graph
 end
 
 require "rubydex"
 
-if options[:mcp]
-  if arguments.length > 1
-    warn("error: unexpected argument '#{arguments[1]}' found")
-    warn
-    warn(option_parser)
-    exit(2)
+# Resolve the command into an operation on a populated graph. Each command parses its own options
+# and does any graph-independent work here; a command that needs no graph (like `query --schema`)
+# handles itself and exits. Whatever falls through returns a lambda that runs against the graph.
+operation =
+  case command
+  when "query"
+    schema = false
+    format = "table"
+    OptionParser.new do |parser|
+      parser.banner = "Usage: rdx query <CYPHER> [options]"
+      parser.on("--schema", "Describe the queryable schema instead of running a query") { schema = true }
+      parser.on("--format FORMAT", ["table", "json"], "Output format (table or json)") { |value| format = value }
+      parser.on("-h", "--help", "Show this help") do
+        puts parser
+        exit
+      end
+    end.parse!
+
+    query = ARGV.shift
+
+    if schema
+      warn("warning: ignoring query argument because `--schema` was given") if query
+      # The schema is static, so describe it without building a graph, then exit.
+      print(Rubydex::Query.schema(format))
+      exit
+    end
+
+    abort_with_usage("`query` requires a Cypher query argument (or pass `--schema`)") if query.nil? || query.empty?
+
+    # Parse the query up front so a malformed query fails fast, before the expensive indexing below.
+    parsed = begin
+      Rubydex::Query.parse(query)
+    rescue ArgumentError => e
+      abort(e.message)
+    end
+
+    lambda do |graph|
+      print(parsed.render(graph, format))
+    rescue ArgumentError => e
+      abort(e.message)
+    end
+  when "console"
+    OptionParser.new do |parser|
+      parser.banner = "Usage: rdx console"
+      parser.on("-h", "--help", "Show this help") do
+        puts parser
+        exit
+      end
+    end.parse!
+
+    lambda do |graph|
+      require "irb"
+      IRB.setup(nil)
+      IRB.conf[:IRB_NAME] = "rubydex"
+      IRB::Irb.new(IRB::WorkSpace.new(binding)).run(IRB.conf)
+    rescue LoadError
+      abort("Interactive mode requires `irb` to be in the bundle")
+    end
+  when "mcp"
+    parser = OptionParser.new do |p|
+      p.banner = "Usage: rdx mcp [PATH]"
+      p.on("-h", "--help", "Show this help") do
+        puts p
+        exit
+      end
+    end
+    begin
+      parser.parse!
+    rescue OptionParser::ParseError => e
+      abort_with_usage(e.message)
+    end
+
+    path = ARGV.shift || Dir.pwd
+    abort_with_usage("unexpected argument: #{ARGV.first}") unless ARGV.empty?
+
+    # The MCP server manages its own graph lifecycle for the given workspace, so it runs here
+    # instead of falling through to the shared graph build below.
+    require "rubydex/mcp_server"
+    Rubydex::MCPServer.run(path)
+    exit
+  else
+    abort_with_usage("unknown command: #{command}")
   end
 
-  require "rubydex/mcp_server"
-  Rubydex::MCPServer.run(arguments.fetch(0, Dir.pwd))
-  exit
-end
-
-def __with_timer(message, &block)
-  print(message)
-  start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
-  block.call
-  duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
-  puts " finished in #{duration.round(2)}ms"
-end
-
-graph = Rubydex::Graph.new
-graph.load_config
-
-__with_timer("Indexing workspace...") { graph.index_workspace }
-__with_timer("Resolving graph...") { graph.resolve }
-
-if options[:interactive]
-  begin
-    require "irb"
-    IRB.setup(nil)
-    IRB.conf[:IRB_NAME] = "rubydex"
-    workspace = IRB::WorkSpace.new(binding)
-    IRB::Irb.new(workspace).run(IRB.conf)
-  rescue LoadError
-    abort("Interactive mode requires `irb` to be in the bundle")
-  end
-end
+# Everything that reaches here operates on a populated graph. Progress goes to stderr so stdout
+# carries only the command's output (e.g. for piping a query's JSON).
+graph = build_graph($stderr)
+operation.call(graph)
--- a/rust/rubydex/src/query/cypher.rs
+++ b/rust/rubydex/src/query/cypher.rs
@@ -1,0 +1,57 @@
+//! A small Cypher query engine that runs read-only queries directly against the in-memory
+//! [`Graph`](crate::model::graph::Graph).
+//!
+//! Supported subset:
+//! - `MATCH` with node patterns `(v:Label {prop: value})` — labels may be a disjunction
+//!   (`(v:Class|Module)` matches a node with **any** of the listed labels) — and relationship
+//!   patterns `-[:TYPE]->`, `<-[:TYPE]-`, `-[:TYPE]-`, including variable-length `-[:TYPE*min..max]->`.
+//! - `WHERE` with `=`, `<>`, `<`, `<=`, `>`, `>=`, `CONTAINS`, `STARTS WITH`, `ENDS WITH`,
+//!   combined with `AND`, `OR`, `NOT`.
+//! - `RETURN` with `DISTINCT`, `AS` aliases, and the aggregates `count`, `collect`, `min`, `max`,
+//!   `sum`, `avg`.
+//! - `ORDER BY`, `SKIP`, `LIMIT`.
+//!
+//! See [`schema`] for the node labels and relationship types exposed to queries.
+
+// The whole Cypher engine — lexer, parser, AST, executor, values, and formatting — lives in the
+// graph-independent `cypher-parser` crate. rubydex only provides the `GraphProvider` mapping for its
+// `Graph` (in `schema`) and the static schema description (in `schema_info`).
+//
+// `Query` is the opaque parsed-query object: callers can `parse` a query string once (failing fast
+// on syntax errors), then `run_parsed` it against a graph that was built afterwards.
+pub use cypher_parser::{CypherError, OutputFormat, Query, parse};
+
+pub mod schema;
+pub mod schema_info;
+
+use crate::model::graph::Graph;
+
+/// Parses and executes a Cypher query against the graph, returning the formatted output.
+///
+/// # Errors
+///
+/// Returns a [`CypherError`] if the query cannot be parsed or executed.
+pub fn run_query(graph: &Graph, query: &str, output_format: OutputFormat) -> Result<String, CypherError> {
+    cypher_parser::run_query(graph, query, output_format)
+}
+
+/// Executes an already-parsed [`Query`] against the graph and formats the result. Pair with
+/// [`parse`] to validate a query before building the graph.
+///
+/// # Errors
+///
+/// Returns a [`CypherError`] if the query cannot be executed.
+pub fn run_parsed(graph: &Graph, query: &Query, output_format: OutputFormat) -> Result<String, CypherError> {
+    let result = cypher_parser::execute(graph, query)?;
+    Ok(cypher_parser::format::format(&result, output_format))
+}
+
+/// Returns a description of the queryable schema (node labels, relationship types, and properties)
+/// in the requested format. The schema is static and does not require a graph.
+#[must_use]
+pub fn schema(output_format: OutputFormat) -> String {
+    schema_info::describe(output_format)
+}
+
+#[cfg(test)]
+mod tests;
--- a/rust/rubydex/src/query/cypher/schema.rs
+++ b/rust/rubydex/src/query/cypher/schema.rs
@@ -1,0 +1,790 @@
+//! Maps the rubydex [`Graph`] onto a property-graph schema for Cypher execution.
+//!
+//! Node labels:
+//! - `Document` — a source file.
+//! - `Definition` — a per-file occurrence of a Ruby construct.
+//! - `Declaration` — the global, merged concept of a named entity. Declarations also carry
+//!   kind sub-labels (`Class`, `Module`, `SingletonClass`, `Method`, `Constant`, `ConstantAlias`,
+//!   `GlobalVariable`, `InstanceVariable`, `ClassVariable`) plus the grouping label `Namespace`
+//!   (any of `Class`/`Module`/`SingletonClass`).
+//!
+//! Relationship types mirror `dot.rs`:
+//! - `DEFINES`: `Document` → `Definition`
+//! - `DECLARES`: `Definition` → `Declaration`
+//! - `CONTAINS`: `Definition` → `Definition` (lexical nesting in one file, e.g. a class written
+//!   inside a module; the source-level counterpart of declaration-level `OWNS`)
+//! - `HAS_PARENT`: `Declaration` → `Declaration` (direct superclass; reverse-traverse for direct
+//!   subclasses, `*` for the full chain)
+//! - `INCLUDES` / `PREPENDS` / `EXTENDS`: `Declaration` → `Declaration` (mixins)
+//! - `OWNS`: `Declaration` → `Declaration` (declaration-level membership, e.g. a namespace's methods
+//!   and nested constants, merged across all files)
+//! - `HAS_ANCESTOR`: `Declaration` → `Declaration` (linearized ancestor chain, incl. modules)
+//! - `HAS_DESCENDANT`: `Declaration` → `Declaration` (reverse of `HAS_ANCESTOR`)
+//! - `REFERENCES`: `Document` → `Declaration` (constant references)
+
+use std::collections::{HashSet, VecDeque};
+
+use crate::model::declaration::Declaration;
+use crate::model::definitions::{Definition, Mixin};
+use crate::model::graph::Graph;
+use crate::model::ids::{ConstantReferenceId, DeclarationId, DefinitionId, UriId};
+
+use cypher_parser::{CypherValue, GraphProvider};
+
+/// A handle to a node in the graph.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum NodeRef {
+    Declaration(DeclarationId),
+    Definition(DefinitionId),
+    Document(UriId),
+}
+
+/// A relationship type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum RelType {
+    /// `Document` → `Definition`: a file defines a construct occurrence.
+    Defines,
+    /// `Definition` → `Declaration`: a per-file occurrence declares the global merged entity.
+    Declares,
+    /// `Definition` → `Definition`: lexical nesting within a single file, e.g. a class written
+    /// inside a module. This is the source-level structure; the declaration-level (merged across
+    /// files) counterpart is [`RelType::Owns`].
+    Contains,
+    /// `Declaration` → `Declaration`: the direct superclass (a single hop). Reverse-traverse for
+    /// direct subclasses, or use `*` for the full class chain. For the transitive relation
+    /// (including modules) see [`RelType::HasAncestor`].
+    HasParent,
+    /// `Declaration` → `Declaration`: an included module mixin.
+    Includes,
+    /// `Declaration` → `Declaration`: a prepended module mixin.
+    Prepends,
+    /// `Declaration` → `Declaration`: an extended module mixin.
+    Extends,
+    /// `Declaration` → `Declaration`: declaration-level membership (a namespace's methods and
+    /// nested constants), merged across all files. The per-file source counterpart is
+    /// [`RelType::Contains`].
+    Owns,
+    /// `Declaration` → `Declaration`: an entry in the linearized ancestor chain (transitive
+    /// superclasses plus included/prepended modules).
+    HasAncestor,
+    /// `Declaration` → `Declaration`: the reverse of [`RelType::HasAncestor`].
+    HasDescendant,
+    /// `Document` → `Declaration`: a constant reference in the file resolves to a declaration.
+    References,
+}
+
+/// Catalog metadata for a relationship type: the type itself, its canonical name, endpoint labels,
+/// and a short description. Entries live in [`REL_SCHEMAS`], the single source of truth that
+/// `RelType`'s accessors and the `--schema` catalog both derive from.
+pub struct RelSchema {
+    pub rel: RelType,
+    pub name: &'static str,
+    pub from: &'static str,
+    pub to: &'static str,
+    pub description: &'static str,
+}
+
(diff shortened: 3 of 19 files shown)


exe/rdx
```diff
@@ -5,40 +5,108 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
...
   start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
-  block.call
+  yield
   duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
-  puts " finished in #{duration.round(2)}ms"
+  io.puts(" finished in #{duration.round(2)}ms")
+end
+
+# Builds the workspace graph, sending progress messages to `progress_io`.
+def build_graph(progress_io)
```
vinistock, line 58: Other than quick one off switches, like `--version`, I would assume everything in this executable always depends on a populated graph (like interactive mode or query).

What do you think of keeping the one off switches as early returns at the top, then we populate the graph and the different commands simply perform different operations on it?

rust/rubydex/src/query/cypher.rs
vinistock, line 1: As far as I understand, the `mod.rs` convention was superseded with just using the desired module as the file name (despite `mod.rs` still working for compatibility).

Can we rename `query/cypher/mod.rs` -> `query/cypher.rs`?

rust/rubydex/src/query/cypher.rs
```diff
@@ -0,0 +1,57 @@
...
+    cypher_parser::run_query(graph, query, output_format)
+}
+
+/// Executes an already-parsed [`Query`] against the graph and formats the result. Pair with
+/// [`parse`] to validate a query before building the graph.
+///
+/// # Errors
+///
+/// Returns a [`CypherError`] if the query cannot be executed.
+pub fn run_parsed(graph: &Graph, query: &Query, output_format: OutputFormat) -> Result<String, CypherError> {
```
vinistock, line 44: Is there a scenario where a consumer would use this? Is it for caching the parsed query somewhere and then skipping the parse step?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,541 @@
...
+//!
+//! Node labels:
+//! - `Document` — a source file.
+//! - `Definition` — a per-file occurrence of a Ruby construct.
+//! - `Declaration` — the global, merged concept of a named entity. Declarations also carry
+//!   kind sub-labels (`Class`, `Module`, `SingletonClass`, `Method`, `Constant`, `ConstantAlias`,
+//!   `GlobalVariable`, `InstanceVariable`, `ClassVariable`) plus the grouping label `Namespace`
+//!   (any of `Class`/`Module`/`SingletonClass`).
+//!
+//! Relationship types mirror `dot.rs`:
```
vinistock, line 11: Can cypher help us get rid of `dot.rs`? Is there a way to pretty print the graph or generate visualizations?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,541 @@
...
+    Defines,
+    Declares,
+    Contains,
+    Inherits,
+    Includes,
+    Prepends,
+    Extends,
+    Owns,
+    Ancestor,
+    Descendant,
```
vinistock: What's the difference between an `Inherits` and an `Ancestor` relationship? 

Also, we have `Descendant` (`Ancestor` equivalent), but no equivalent for `Inherits`. Is the assymetry expected?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,541 @@
...
+    Definition(DefinitionId),
+    Document(UriId),
+}
+
+/// A relationship type.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum RelType {
+    Defines,
+    Declares,
+    Contains,
```
vinistock, line 52: What does contains represent?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,541 @@
...
+    graph.name_id_to_declaration_id(*constant_ref.name_id()).copied()
+}
+
+/// Resolves a constant reference to a namespace declaration, following constant aliases.
+fn resolve_ref_to_namespace(graph: &Graph, ref_id: ConstantReferenceId) -> Option<DeclarationId> {
+    resolve_to_namespace(graph, resolve_ref(graph, ref_id)?)
+}
+
+/// Walks constant-alias chains until reaching a namespace declaration.
+fn resolve_to_namespace(graph: &Graph, declaration_id: DeclarationId) -> Option<DeclarationId> {
```
vinistock, line 771: This method already exists in `query.rs` (although it may not be handling the circular alias case).

Can we make that one public instead?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,541 @@
...
+}
+
+fn document_property(graph: &Graph, id: UriId, prop: &str) -> CypherValue {
+    let Some(document) = graph.documents().get(&id) else {
+        return CypherValue::Null;
+    };
+
+    match prop {
+        "uri" => CypherValue::Str(document.uri().to_string()),
+        "path" | "name" => {
```
vinistock: Does `prop` here mean the field to be returned from a given entity?

And if so, is `path` here URI path or file path? If it's file path, we should use the `Url` crate to make sure we're handling all types of URIs (like Windows).

Also, why is `name` also considered `path`?

vinistock (commented):
Still trying to wrap my head around the entire engine, but left some comments already. Excited to have a unified way of querying the graph.

I wonder if there's some IRB trick we can use to enter a "query" mode that accepts the Cypher queries directly (non-valid Ruby). Something like:

```shell
bundle exec rdx -i
Indexing...
Resolving...
> graph["Foo"]
=> <Declaration ...>
>
> query_mode!
> MATCH (n:Class|Module) RETURN n.name ORDER BY n.name
=> [Foo]
```

rust/rubydex/src/query/cypher.rs
paracycle: Good call. I'll move `cypher/mod.rs` to `cypher.rs`.

rust/rubydex/src/query/cypher.rs
paracycle: So, this is actually the main way that we are using the API for 2 main reasons:
1. It makes sure that if there are any query parsing errors, we can catch those early at query parse time, which we can do before we index the codebase. Then we pass the parsed query to this method to run it against a graph.
2. We can also use this to cache queries and run them multiple times against the same graph multiple times.

My motivation for the split was mainly for 1, but 2 comes as a nice by-product.

At this point, the complementary `run_query` (which takes a string query) is in the PR as a utility method and is only used in tests. We can remove that version, if you want.

rust/rubydex/src/query/cypher/schema.rs
paracycle: I don't think there is too much overlap. DOT rendering is for full-graph visualization, and Cypher is for selective querying. We could, however, supply the ability to render _query results_ as DOT, and we could potentially merge the relationship/edge definition parts of two. But, I'd rather do all that later, and not in this PR.

rust/rubydex/src/query/cypher/schema.rs
paracycle: `Inherits` represents a direct superclass relationship, where as `Ancestor` is the full ancestor chain membership. So for:
```ruby
class Base
end

module Bar
end

class Foo < Base
  include Bar
end
```
the query:
```cypher
MATCH (c:Class {name: 'Foo'})-[:INHERITS]->(p) RETURN p.name, p.kind
```
gives:
```
p.name | p.kind
-------+-------
Base   | Class
```
but the query:
```cypher
MATCH (c:Class {name: 'Foo'})-[:ANCESTOR]->(a) RETURN a.name, a.kind
```
gives:
```
a.name      | a.kind
------------+-------
Bar         | Module
Base        | Class
Object      | Class
Kernel      | Module
BasicObject | Class
```

Notice that `Bar` has an `ANCESTOR` relationship to `Foo`, but not an `INHERITS` one.

As for the asymmetry, that is due to the fact that we store descendants information, but not subclass information in the graph. We _could_ do more processing to extract `subclass` information, but it will be more expensive.

At the same time, you can already query the relationship going in the other direction to query direct subclassing relationship as well. So, for the above example:
```cypher
MATCH (p:Class {name:'Base'})<-[:INHERITS]-(c) RETURN c.name, c.kind
```
gives:
```
c.name | c.kind
-------+-------
Foo    | Class
```

rust/rubydex/src/query/cypher/schema.rs
paracycle: `CONTAINS` represents lexical nesting between definitions, so it is a `Definition` to `Definition` edge. For example, a class written textually inside a module in the same file. It's the source-level counterpart of `OWNS`, which is declaration-level membership merged across all files.

I will be documenting this directly on the type so that every `RelType` variant has a doc comment spelling out its source/target node type and meaning.

rust/rubydex/src/query/cypher/schema.rs
paracycle: Agree there's duplication worth removing, but they're not quite the same and I want to unify carefully rather than just making the existing one public.

`query.rs::resolve_to_namespace` returns `Result<Option<…>>` (it errors when a declaration is neither a namespace nor an alias-to-namespace) and does a single `resolve_alias` step. This method returns `Option<…>` and walks alias chains in a loop to handle cyclic aliases, which is the case you are also noting. 

We could extract a shared helper that keeps the cyclic-alias handling, and then have `query.rs` adopt it as well. But, in this PR, I want to stay away from code that is behaviour changing in the core.

rust/rubydex/src/query/cypher/schema.rs
paracycle: Yes, `prop` is the property name read off a node. For example, the `x` in `RETURN n.x`. 

You're right that collapsing `path` and `name` to both mean the basename was the wrong decision. I will be using the Uri methods that already exist in `document.rs` to make: 
- `uri`: full URI, e.g. `file:///app/models/user.rb`
- `path`: file-system path for `file://` URIs, e.g. `/app/models/user.rb`
- `name`: base file name for `file://` URIs, e.g. `user.rb`

I am planning that for `path` and `name`, non-`file://` URIs fall back to the raw URI, since we don't always have a good fallback for those URIs.

exe/rdx
paracycle: Generally, that is correct, except for the `schema` subcommand handling, which returns a rendering of the schema (with documentation) without needing to index anything.

I could turn `schema` subcommand into a `--schema` flag for the `query` subcommand, if you want, and that would then become special handling for that subcommand, and we can do what you are suggesting.

Let me know what you prefer.

paracycle:
> I wonder if there's some IRB trick we can use to enter a "query" mode that accepts the Cypher queries directly (non-valid Ruby). Something like:
> 
> ```shell
> bundle exec rdx -i
> Indexing...
> Resolving...
> > graph["Foo"]
> => <Declaration ...>
> >
> > query_mode!
> > MATCH (n:Class|Module) RETURN n.name ORDER BY n.name
> => [Foo]
> ```

Great idea. I am prototyping what's possible here, but obviously it won't be a part of this PR.

paracycle:
Ok, this PR has a prototype for the console mode extension: https://github.com/Shopify/rubydex/pull/883

paracycle:
@vinistock If it is helpful, I had this diagram generated from the codebase for how the query engine works at a high level:

image

Morriar (commented):
Why did we need our own implementation of the Cypher parser? None of the existing Rust implementations could be reused?

Could you clean the commit history to make this PR easier to read? First commit seems to introduce a whole parser that gets removed later.

paracycle:
> Why did we need our own implementation of the Cypher parser? None of the existing Rust implementations could be reused?

The existing implementations are either unmaintained, or were doing much more than what we needed, or were pulling in a lot of dependencies. I wanted to start with a parser under our control and with no dependencies, that support only the subset of the syntax that we want to support (e.g. we will probably never need `CREATE` queries for building the schema).

If needed, we can swap what we have with another library down the line, but this seems to work just fine.

> Could you clean the commit history to make this PR easier to read? First commit seems to introduce a whole parser that gets removed later.

Sure, I can do that tomorrow.

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,565 @@
...
+    Prepends,
+    /// `Declaration` → `Declaration`: an extended module mixin.
+    Extends,
+    /// `Declaration` → `Declaration`: declaration-level membership (a namespace's methods and
+    /// nested constants), merged across all files. The per-file source counterpart is
+    /// [`RelType::Contains`].
+    Owns,
+    /// `Declaration` → `Declaration`: an entry in the linearized ancestor chain (transitive
+    /// superclasses plus included/prepended modules).
+    Ancestor,
```
Morriar: With just `Ancestor`, the direction is unclear:

```cypher
MATCH (a:Class)-[:ANCESTOR]->(b) RETURN a.name, b.name
```

Is `a` the ancestor or the descendant?

What do you think of using `HAS_ANCESTOR`?

And same for `HAS_DESCENDANT`.

Morriar: Since in Cypher we can easily go through transitive edges, should we provide a `HAS_PARENT` and `HAS_CHILD` and let the user decide if they want transitivity?

```cypher
// Direct edges
(a)-[:HAS_PARENT]->(b)

// Transitive edges
(a)-[:HAS_PARENT*]->(b)

// Fixed depth of exactly 3
(a)-[:HAS_PARENT*3]->(b)

// Depth range (1 to 5)
(a)-[:HAS_PARENT*1..5]->(b)

// At least 2 deep, no upper bound
(a)-[:HAS_PARENT*2..]->(b)
```

rust/rubydex/src/query/cypher/schema_info.rs
```diff
@@ -0,0 +1,374 @@
...
+/// A property exposed on a node type.
+struct PropInfo {
+    node_type: &'static str,
+    property: &'static str,
+    description: &'static str,
+}
+
+const LABELS: &[LabelInfo] = &[
+    LabelInfo {
+        label: "Document",
```
Morriar: Could we avoid the dual source of truth between `schema` and `schema_info`?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,565 @@
...
+    fn name(&self, node: NodeRef) -> String {
+        node_name(self, node)
+    }
+}
+
+/// Returns all nodes matching the given labels. An empty slice matches every node; otherwise a node
+/// is returned if it matches **any** of the labels (label disjunction, e.g. `(:Class|Module)`).
+#[must_use]
+pub fn scan(graph: &Graph, labels: &[String]) -> Vec<NodeRef> {
+    if labels.is_empty() {
```
Morriar, line 397: Should we apply an implicit default `LIMIT` to queries to avoid loading the entire graph in memory?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -0,0 +1,565 @@
...
+    /// nested constants), merged across all files. The per-file source counterpart is
+    /// [`RelType::Contains`].
+    Owns,
+    /// `Declaration` → `Declaration`: an entry in the linearized ancestor chain (transitive
+    /// superclasses plus included/prepended modules).
+    Ancestor,
+    /// `Declaration` → `Declaration`: the reverse of [`RelType::Ancestor`].
+    Descendant,
+    /// `Document` → `Declaration`: a constant reference in the file resolves to a declaration.
+    References,
```
Morriar, line 73: To me it reads as `a` references (action verb) `b` and not the list of references contained in `a`.

exe/rdx
vinistock: I think a sub-command is a nice idea indeed 👍

rust/rubydex/src/query/cypher.rs
vinistock: Ah, perfect. Thanks for explaining! I think it's fine to still keep the string version.

rust/rubydex/src/query/cypher/schema.rs
vinistock: Ah, got it. Yeah, we've been meaning to explore moving ancestor/descendants to a poset, which would allow for richer queries against all of the different types of ancestor related operations.

rust/rubydex/src/query/cypher/schema.rs
vinistock: Perfect, yes please let's document each one as it will help keep in mind what the intended edge means.

rust/rubydex/src/query/cypher/schema.rs
vinistock: Fair, we can follow up. I think the circular thing could be an existing bug, so it's worth at least checking if we can unify.

ext/rubydex/graph.c
```diff
@@ -9,13 +9,17 @@
...
+static VALUE cQuery;
 static VALUE mRubydex;
 static VALUE cKeyword;
 static VALUE cKeywordParameter;
 
 // Interned once in `rdxi_initialize_graph` to avoid repeated symbol-table lookups on hot completion paths.
 static ID id_self_receiver;
 
+// Coerces an optional format argument (String, Symbol, or nil) to a C string; defined below.
+static const char *cypher_format_cstr(VALUE format);
```
vinistock: It might be worth putting this in utils as it can probably be useful in other places too.

ext/rubydex/graph.c
```diff
@@ -750,6 +754,99 @@ static VALUE rdxr_graph_keyword(VALUE self, VALUE name) {
...
+        .dfree = query_free,
+        .dsize = NULL,
+        .dcompact = NULL,
+    },
+    .parent = NULL,
+    .data = NULL,
+    .flags = RUBY_TYPED_FREE_IMMEDIATELY,
+};
+
+// Rubydex::Query.parse(query) -> Rubydex::Query
```
vinistock: This can probably be done in a follow up (and likely in one go with an LLM), but we moved all of the docs to call-seq to have it surfaced in the new static doc website.

ext/rubydex/graph.c
```diff
@@ -784,4 +881,10 @@ void rdxi_initialize_graph(VALUE moduleRubydex) {
     rb_define_method(cGraph, "exclude_paths", rdxr_graph_exclude_paths, 1);
     rb_define_method(cGraph, "excluded_paths", rdxr_graph_excluded_paths, 0);
     rb_define_method(cGraph, "keyword", rdxr_graph_keyword, 1);
+
+    cQuery = rb_define_class_under(mRubydex, "Query", rb_cObject);
+    rb_undef_alloc_func(cQuery);
+    rb_define_singleton_method(cQuery, "parse", rdxr_query_parse, 1);
+    rb_define_singleton_method(cQuery, "schema", rdxr_cypher_schema, -1);
+    rb_define_method(cQuery, "render", rdxr_query_render, -1);
```
vinistock: We've been trying to keep one entity per C file. It would be nice to house all of these in a `query.c`.

rust/rubydex/src/main.rs
```diff
@@ -52,6 +53,38 @@ struct Args {
...
+    )]
+    schema: bool,
+
+    #[arg(
+        long = "format",
+        value_enum,
+        default_value = "table",
+        help = "Output format for --query and --schema results"
+    )]
+    format: Format,
```
vinistock: Does clap have a way to tie this to `query` and `schema`? Basically, ensuring that something that makes no sense like `rubydex_cli --stats --format=table` gets rejected.

vinistock (commented):
Just some minor things, but I think we can ship and iterate on it.

Note: it seems that, like `query.rs`, cypher queries are going to be quite heavy on graph traversal (depending on what the user is searching for of course). If possible, it would be nice to scan the graph in parallel

rust/rubydex/src/main.rs
paracycle: Done

ext/rubydex/graph.c
paracycle: Done

ext/rubydex/graph.c
paracycle: Done

rust/rubydex/src/query/cypher/schema.rs
paracycle: I am not sure why we would need a `LIMIT` when we already have the whole graph in memory? If we find that we need it, maybe we can add it later on.

rust/rubydex/src/query/cypher/schema_info.rs
paracycle: I merged all sources of truth into the `schema.rs` file.

rust/rubydex/src/query/cypher/schema.rs
paracycle: Isn't that exactly what this relation is? `a -[:REFERENCES]-> b` means that the `Document a` has a reference to `Declaration b`.

rust/rubydex/src/query/cypher/schema.rs
paracycle: Good call, I renamed `ANCESTOR` to `HAS_ANCESTOR`, `DESCENDANT` to `HAS_DESCENDANT` and `INHERITS` to `HAS_PARENT`.

We don't define a `HAS_CHILD` since we don't have an explicit index on that, and you can always query that using `HAS_PARENT` with a reversed direction.

ext/rubydex/graph.c
paracycle: Done

exe/rdx
paracycle: I moved this to a flag for the `query` subcommand.

paracycle:
@vinistock and @Morriar: I think I've addressed all the comments and I have a clean commit history now. Can you both do a final review?

rust/rubydex/src/model/graph.rs
```diff
@@ -92,8 +93,14 @@ pub struct Graph {
 
     /// Project configuration
     config: Config,
+
+    /// Lazily-memoized decoded file paths, keyed by document URI. `Document::file_path` runs
+    /// `Url::parse`, which is hot during Cypher property reads; the URI is immutable, so we parse at
+    /// most once per document here. Kept on the graph (a single instance) rather than on `Document`
+    /// so the per-document struct stays small across a hyper-scale workspace.
+    file_path_cache: Mutex<IdentityHashMap<UriId, Option<PathBuf>>>,
```
vinistock: Can we delay introducing this? This means extra work to maintain the state of the graph up to date on changes since now we need to update not only `documents`, but also make sure that the cache doesn't have stale entries.

For example, right now, if documents are deleted or moved this cache is not having its state updated, so it goes out of sync with `documents`.

Unless this is truly necessary for performance, I'd vote for starting off without this.

Also, I think the comment on top doesn't hold. Putting this inside of `Document` means avoiding the cost of storing an extra `UriId` for each key in this hashmap, so it would probably be more memory efficient to put it inside of the struct.

exe/rdx
```diff
@@ -5,75 +5,149 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
...
-    abort("Interactive mode requires `irb` to be in the bundle")
+    lambda do |graph|
+      require "irb"
+      IRB.setup(nil)
+      IRB.conf[:IRB_NAME] = "rubydex"
+      IRB::Irb.new(IRB::WorkSpace.new(binding)).run(IRB.conf)
+    rescue LoadError
+      abort("Interactive mode requires `irb` to be in the bundle")
+    end
+  when "mcp"
```
st0012, line 126: Is this changing `rdx --mcp` to `rdx mcp`?

rust/rubydex/src/main.rs
```diff
@@ -76,6 +84,45 @@ struct Args {
         help = "Number of files to re-index per incremental cycle"
     )]
     incremental_files: usize,
+
+    #[arg(long = "query", value_name = "CYPHER", help = "Run a Cypher query against the graph")]
```
st0012: Do we need to expose this feature in both Ruby executable and Rust cli? If we're mostly using the former now, we can defer the latter until we actually need it?

vinistock (approved):
I believe the changes are fine, but I'd avoid the new state in the graph for now

exe/rdx
```diff
@@ -5,75 +5,149 @@ $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
...
+    end
 
-def __with_timer(message, &block)
-  print(message)
-  start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
-  block.call
-  duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
-  puts " finished in #{duration.round(2)}ms"
-end
+    query = ARGV.shift
```
st0012: Should we warn if `--schema` and query are both passed?

```suggestion
    query = ARGV.shift
    
    if schema
      # The schema is static, so describe it without building a graph, then exit.
      print(Rubydex::Query.schema(format))
      if query
        print("Warning: query string after --schema flag is ignored")
      end
      
      exit
    end
```

rust/rubydex/src/model/graph.rs
paracycle: Sure, I can hold this back. The original patch was adding it as a `OnceLock<Option<PathBuf>>` field to the `Document` struct making it increase in size by 32 bytes, which implied a memory increase per document.

This centralized cache performs the memoization only if the file path is used and only for the documents it has queried for it, so the optimization is that the cost (even if it is higher) is only borne when this operation is performed.

exe/rdx
paracycle: Yes, this PR is making all of the flags into subcommands, since subcommands make more sense.

rust/rubydex/src/main.rs
paracycle: I am fine either way. I wasn't sure if the team working on Rubydex mostly use the Rust executable or the Ruby one, so I added both. Happy to remove the Rust one.

st0012: Personally I only use the Rust cli when testing Rubydex itself. For actually indexing codebase and interact with it etc the Ruby layer is more useful as it has access to dependencies easily. So I'd vote for remove it for now and only add it when we actually need it.
@vinistock @Morriar WDYT?

exe/rdx
st0012: I'm fine with the change, but can you also update readme's MCP server section too?

I also think this feature is worth mentioning in the readme.

exe/rdx
paracycle: Done!

rust/rubydex/src/main.rs
paracycle: I removed the commit that was adding that, for now. Based on what other says, we can keep it out or add it back in in another PR.

exe/rdx
paracycle: Updated

rust/rubydex/src/model/graph.rs
paracycle: I removed that commit from this PR.

st0012:
@paracycle As it changes multiple flags into sub-command, next version should be `v0.3.0`.

PR #869, opened 2026-06-18
Add opt-in client/server mode to the rdx executable

paracycle:
## Goal

`rdx query` pays for indexing and resolving the workspace on every call. This adds a resident server
that pays it once, and a `--server` flag that sends the query to it. Without the flag nothing changes,
and a platform without `fork` or UNIX sockets keeps running inline.

`rdx server start|stop|restart|status` manages the server for a workspace.

This pull request is the infrastructure only: the runtime state, the wire protocol, the process
lifecycle and the CLI surface. Keeping the graph in step with a changing workspace is #979, stacked on
this one.

## Identity

A workspace maps to a runtime directory keyed by an app id, covering the workspace path, the protocol
version, the Ruby version, the rubydex version and a fingerprint of the loaded native extension. A gem
upgrade or a protocol change therefore addresses a different directory and forces a fresh server,
because a C extension cannot be reloaded in place.

The directory holds three files:

| File | Role |
|---|---|
| `lock` | Empty. Nobody reads or writes it. The server holds an exclusive lock for its whole life. Never deleted. |
| `state.json` | The pid, the token, the version and the start time. Replaced by a rename. |
| `socket` | The UNIX domain socket. |

The two files stay separate because of Windows: `flock` there is `LockFileEx` over the whole file, and
an exclusive lock denies every other process read access to that range, so a record inside the locked
file would be unreadable exactly while a server runs. The lock file is never unlinked either, because
a lock belongs to an inode and unlinking the path would let a second process lock a new file and start
a second server for the same workspace.

**The lock answers "does a server own this workspace".** The kernel releases it when the holder dies,
so a recycled pid cannot look alive. The recorded pid is a display value for a human, and the CLI never
signals it. An authenticated request over the socket is the only stop.

## Wire protocol

A frame is a decimal length line followed by exactly that many bytes of JSON. The two directions carry
opposite risks, so each has its own reader:

- The server bounds a whole request with one deadline, because it handles one client at a time and a
  client that drips bytes would otherwise hold the accept loop.
- The client bounds only the gap between chunks, and waits for the first byte without a limit, because
  its server sends nothing until the query finishes.

Reads happen in 64 KiB chunks, since `read_nonblock` allocates its `maxlen` before it reads: a peer
declaring 300 MiB and sending one byte cost 300 MiB of resident memory before that change. The module
raises two error types, and a body that is not JSON is one of them, so no caller has to rescue
`JSON::ParserError`.

## Failure

A resident server belongs to every client, so one bad request must not take it away from the rest. Each
connection is rescued on its own: the fault is logged with its backtrace, the client gets status 1, and
the accept loop lives on.

A server that holds its lock but does not answer is reported rather than waited on. `status` probes it
without starting, stopping or restarting anything, and prints the recorded pid and start time so a human
can find the process. `status`, `stop` and `restart` all exit 1 in that case instead of hanging.

The report claims no more than what the probe saw. `serve` takes one request at a time, so a server
working on a long query is indistinguishable from one that is stuck, and the message says so rather
than telling the reader to kill a process that may be doing exactly what they asked.

## Freshness

The graph is a snapshot taken at boot. A file edited afterwards is still answered from the state it had
at boot, so `rdx query --server` and `rdx query` disagree about a workspace that changed. #979 fixes
that, and it is stacked on this branch.

## Tests

`test/server/` covers the framing, the runtime state, the client lifecycle and the error handling;
`test/server/integration_test.rb` drives the real executable as a subprocess. Every test was checked
against the behaviour it guards.

### DIFF
--- a/lib/rubydex/cli.rb
+++ b/lib/rubydex/cli.rb
@@ -26,6 +26,14 @@ module Rubydex
         require "rubydex"
 
         dispatch(argv.shift, argv)
+      rescue StandardError => e
+        # A command loads `rubydex/server` only when it needs it, so the constant can be absent.
+        # A direct reference here would raise `NameError` and hide the real error.
+        raise unless defined?(Rubydex::Server::Error) && e.is_a?(Rubydex::Server::Error)
+
+        # A server that does not start or answer is a runtime condition, not a defect in rdx.
+        warn("rdx server: #{e.message}")
+        exit(1)
       end
 
       # Reports `message`, then the usage text, and exits non-zero. Public because the subcommands
--- a/lib/rubydex/server.rb
+++ b/lib/rubydex/server.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require "rubydex/progress"
+require "rubydex/version"
+
+module Rubydex
+  # Client/server mode for the `rdx` executable.
+  #
+  # A resident server process indexes and resolves the workspace once, and keeps the graph in
+  # memory. Later commands reach it over a UNIX domain socket and skip that work.
+  module Server
+    # Increase this after any incompatible change to the request or response shape.
+    PROTOCOL = 1
+
+    class Error < StandardError; end
+
+    class << self
+      # Without `O_NOFOLLOW`, a symlink can redirect the permissions this code sets on its runtime
+      # directory.
+      #: -> bool
+      def supported?
+        Process.respond_to?(:fork) && defined?(::UNIXSocket) && !Gem.win_platform? &&
+          !State::NOFOLLOW.nil?
+      end
+
+      #: (workspace_path: String, ?progress_io: IO?) -> Rubydex::Graph
+      def build_graph(workspace_path:, progress_io: nil)
+        # The server boot must build the same graph as the inline CLI path.
+        graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+        Progress.with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+        Progress.with_timer(progress_io, "Resolving graph...") { graph.resolve }
+        graph
+      end
+    end
+  end
+end
+
+require "rubydex/server/state"
+require "rubydex/server/frame"
+require "rubydex/server/core"
+require "rubydex/server/client"
+require "rubydex/server/commands"
--- a/lib/rubydex/server/client.rb
+++ b/lib/rubydex/server/client.rb
@@ -1,0 +1,287 @@
+# frozen_string_literal: true
+
+require "socket"
+
+module Rubydex
+  module Server
+    # The short-lived client side. It requires no native extension when a server answers, which
+    # is the source of the time saving.
+    module Client
+      HANDSHAKE_TIMEOUT = 10.0
+
+      # A new server indexes the whole workspace first, so this limit is far larger.
+      BOOT_TIMEOUT = 120.0
+
+      STOP_TIMEOUT = 5.0
+
+      class << self
+        #: (State state, Hash[Symbol, untyped] options, ?stdout: IO, ?stderr: IO) -> Integer
+        def query(state, options, stdout: $stdout, stderr: $stderr)
+          request(
+            state,
+            {
+              "command" => "query",
+              "query" => options[:query],
+              "query_format" => options[:query_format] || "table",
+            },
+            stdout: stdout,
+            stderr: stderr,
+          )
+        end
+
+        # The server answers only when the query completes, so this read has no overall deadline.
+        #: (State state, Hash[String, untyped] payload, ?stdout: IO, ?stderr: IO) -> Integer
+        def request(state, payload, stdout: $stdout, stderr: $stderr)
+          socket = connection(state)
+
+          begin
+            exchange(state, socket, payload, stdout: stdout, stderr: stderr)
+          rescue Frame::ReadTimeout, Frame::Malformed => e
+            # The server failed, so the client reports the failure and does not raise.
+            stderr.puts("rdx server: #{e.message}")
+            1
+          end
+        end
+
+        # Talks to a server that already exists, and changes nothing. `status` uses this to diagnose
+        # a server that does not answer, so one budget covers every step. Returns `nil` on no answer.
+        #: (State state, Hash[String, untyped] payload, ?stdout: IO, ?stderr: IO, ?timeout: Float) -> Integer?
+        def probe(state, payload, stdout: $stdout, stderr: $stderr, timeout: Frame::REQUEST_TIMEOUT)
+          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+
+          socket = connect(state, timeout: remaining(deadline))
+          return unless socket
+
+          exchange(state, socket, payload, stdout: stdout, stderr: stderr, total_timeout: remaining(deadline))
+        rescue Frame::ReadTimeout, Frame::Malformed, Errno::EPIPE, Errno::ECONNRESET
+          nil
+        end
+
+        #: (State state) -> UNIXSocket
+        def connection(state)
+          ensure_server(state)
+
+          socket = connect(state)
+          return socket if socket
+
+          # A restart repairs a stale socket, but not a live owner that does not answer: a new
+          # server cannot take the lock from it.
+          unless state.server_running?
+            restart(state)
+            socket = connect(state)
+            return socket if socket
+          end
+
+          if state.server_running?
+            raise Error, "the rdx server for #{state.workspace_path} is not answering; run `rdx server status`"
+          end
+
+          raise Error, "could not connect to the rdx server at #{state.socket_path}"
+        end
+
+        #: (State state) -> void
+        def ensure_server(state)
+          if state.server_running? && File.socket?(state.socket_path)
+            return if state.version_compatible?
+
+            stop(state)
+          end
+
+          state.clean!
+          start(state)
+        end
+
+        # A held lock means another process has already started a server here, so this waits for it.
+        #
+        # Take the lock before `fork`. `server_running?` acquires the lock to test it, so the
+        # parent's readiness check can otherwise acquire it before the child does.
+        #: (State state) -> void
+        def start(state)
+          previous = state.read
+
+          return wait_until_ready(state, previous: previous) if state.server_running?
+
+          state.ensure_dir!
+          lock = state.open_lock
+
+          # Another process acquired the lock after the check above.
+          unless lock.flock(File::LOCK_EX | File::LOCK_NB)
+            lock.close
+            return wait_until_ready(state, previous: previous)
+          end
+
+          begin
+            pid = fork { run_server(state, lock) }
+          rescue StandardError
+            lock.flock(File::LOCK_UN)
+            lock.close
+            raise
+          end
+
+          Process.detach(pid) if pid
+          # The child inherits the open file description, so the lock outlives this handle.
+          lock.close
+          wait_until_ready(state, previous: previous)
+        end
+
+        # Returns false when the old server keeps its lock, because a new one cannot start then.
+        #: (State state) -> bool
+        def restart(state)
+          return false unless stop(state)
+
+          state.clean!
+          start(state)
+          true
+        end
+
+        # An authenticated request is the only stop. `Process.daemon` forks a second time, so a
+        # recorded pid can belong to an unrelated process, and no caller may signal it.
+        #: (State state) -> bool
+        def stop(state)
+          return true unless state.server_running?
+
+          socket = connect(state)
+
+          if socket
+            begin
+              exchange(
+                state,
+                socket,
+                { "command" => "stop" },
+                stdout: $stdout,
+                stderr: $stderr,
+                total_timeout: Frame::REQUEST_TIMEOUT,
+              )
+            rescue Errno::EPIPE, Errno::ECONNRESET, Frame::ReadTimeout, Frame::Malformed
+              # A server that dies during its answer has stopped. The lock proves that, below.
+              nil
+            end
+          end
+
+          stopped = wait_until_stopped(state)
+          state.clean! if stopped
+          stopped
+        end
+
+        private
+
+        # One exchange with a server that already answers. The three callers differ only in how they
+        # reached it, and in what a failure means, so every failure raises and each caller decides.
+        #: (State state, UNIXSocket socket, Hash[String, untyped] payload, stdout: IO, stderr: IO, ?total_timeout: Float?) -> Integer
+        def exchange(state, socket, payload, stdout:, stderr:, total_timeout: nil)
+          Frame.write(socket, payload.merge(base_payload(state)))
+
+          response = Frame.read_response(socket, total_timeout: total_timeout)
+          raise Frame::Malformed, "the server closed the connection without answering" unless response
+
+          stdout.print(response["stdout"]) if response["stdout"]
+          stderr.print(response["stderr"]) if response["stderr"]
+
+          status = response["status"]
+          raise Frame::Malformed, "the response carried no status" unless status.is_a?(Integer)
+
+          status
+        ensure
+          socket.close unless socket.closed?
+        end
+
+        # The daemon inherits the held lock across both forks.
+        #: (State state, File lock) -> void
+        def run_server(state, lock)
+          Process.daemon(true)
+          Core.new(state, lock: lock).run
+        end
+
+        #: (State state, ?timeout: Float) -> UNIXSocket?
+        def connect(state, timeout: HANDSHAKE_TIMEOUT)
+          # A relative socket name keeps the path below the `sockaddr_un` limit.
+          socket = Dir.chdir(File.dirname(state.socket_path)) do
+            UNIXSocket.new(File.basename(state.socket_path))
+          end
+          version = Frame.read_line(socket, timeout: timeout).chomp
+
+          if version == state.expected_version
+            socket
+          else
+            socket.close
+            nil
+          end
+        rescue Errno::ENOENT,
+               Errno::ECONNREFUSED,
+               Errno::ECONNRESET,
+               Frame::ReadTimeout,
+               Frame::Malformed
+          begin
+            socket&.close
+          rescue IOError
+            nil
+          end
+          nil
+        end
+
+        # `previous` is the record from before the spawn. A held lock, or a changed record, proves
+        # that a boot began. A free lock after that proves the server died, so this stops early.
+        #: (State state, previous: Hash[String, untyped]?) -> void
+        def wait_until_ready(state, previous:)
+          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + BOOT_TIMEOUT
+          seen_lock = false
+
+          loop do
+            running = state.server_running?
+            seen_lock ||= running
+
+            # The lock is part of the test: a crashed server leaves a socket and a valid record.
+            break if running && File.socket?(state.socket_path) && state.version_compatible?
+
+            if (seen_lock || state.read != previous) && !running
+              raise Error, "the rdx server exited during startup; see #{server_log_hint(state)} for details"
+            end
+
+            raise Error, "timed out waiting for the rdx server to start (see #{server_log_hint(state)})" if Process.clock_gettime(Process::CLOCK_MONOTONIC) > deadline
+
+            sleep(0.05)
+          end
+        end
+
+        # The kernel frees the lock when the process exits, however it exited.
+        #: (State state) -> bool
+        def wait_until_stopped(state)
+          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + STOP_TIMEOUT
+
+          while state.server_running?
+            return false if Process.clock_gettime(Process::CLOCK_MONOTONIC) > deadline
+
+            sleep(0.02)
+          end
+
+          true
+        end
+
+        # Floored, so a read never gets a zero or negative timeout.
+        #: (Float deadline) -> Float
+        def remaining(deadline)
+          left = deadline - Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          [left, 0.05].max
+        end
+
+        # Where the detached server writes its output, so a user can read a boot failure.
+        #: (State state) -> String
+        def server_log_hint(state)
+          override = ENV["RDX_SERVER_LOG"]
+          override && !override.empty? ? override : state.log_path
+        end
+
+        #: (State state) -> Hash[String, untyped]
+        def base_payload(state)
+          {
+            "protocol" => PROTOCOL,
+            "token" => state.token,
+            "cwd" => Dir.pwd,
+            "argv" => ARGV,
+            "env" => {},
+          }
+        end
+      end
+    end
+  end
+end
--- a/lib/rubydex/server/commands.rb
+++ b/lib/rubydex/server/commands.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+module Rubydex
+  module Server
+    module Commands
+      class << self
+        #: (State state, ?stdout: IO) -> Integer
+        def start(state, stdout: $stdout)
+          if state.server_running? && state.version_compatible? && File.socket?(state.socket_path)
+            stdout.puts("rdx server already running")
+            return 0
+          end
+
+          Client.ensure_server(state)
+          stdout.puts("rdx server started")
+          0
+        end
+
+        #: (State state, ?stdout: IO) -> Integer
+        def stop(state, stdout: $stdout)
+          unless state.server_running?
+            stdout.puts("rdx server: no server running for #{state.workspace_path}")
+            state.clean!
+            return 0
+          end
+
+          if Client.stop(state)
+            stdout.puts("rdx server stopped")
+            return 0
+          end
+
+          stdout.puts(unresponsive_report(state, "it did not stop within #{Client::STOP_TIMEOUT}s"))
+          1
+        end
+
+        #: (State state, ?stdout: IO) -> Integer
+        def restart(state, stdout: $stdout)
+          unless Client.restart(state)
+            stdout.puts(unresponsive_report(state, "it did not stop within #{Client::STOP_TIMEOUT}s"))
+            return 1
+          end
+
+          stdout.puts("rdx server restarted")
+          0
+        end
+
+        # `timeout` exists so a test can exercise the unanswered path without the real wait.
+        #: (State state, ?stdout: IO, ?stderr: IO, ?timeout: Float) -> Integer
+        def status(state, stdout: $stdout, stderr: $stderr, timeout: Frame::REQUEST_TIMEOUT)
+          unless state.server_running?
+            stdout.puts("rdx server: not running for #{state.workspace_path}")
+            return 0
+          end
+
+          unless File.socket?(state.socket_path)
+            stdout.puts(unresponsive_report(state, "it has not created its socket"))
+            return 1
+          end
+
+          # `probe` must not change the server or spend a fresh timeout on the handshake and the
+          # answer, because it diagnoses a server that will not answer.
+          status = Client.probe(
+            state,
+            { "command" => "status" },
+            stdout: stdout,
+            stderr: stderr,
+            timeout: timeout,
+          )
+          return status if status
+
+          stdout.puts(unresponsive_report(state, "it did not answer within #{timeout}s"))
+          1
+        end
+
+        private
+
+        # The report states an observation, not a diagnosis: the server answers one request at a
+        # time, so a long query looks the same as a stuck process.
+        #: (State state, String reason) -> String
+        def unresponsive_report(state, reason)
+          <<~REPORT
+            rdx server: a server holds #{state.workspace_path}, but #{reason}
+              recorded pid:   #{state.server_pid || "unknown"}
+              recorded start: #{state.started_at || "unknown"}
+              socket:         #{state.socket_path}
+              log:            #{state.log_path}
+            The server answers one request at a time, so a long query looks the same from here. The
+            log records the boot and any failure, but not the request in flight, so it cannot tell
+            you which of the two this is. rdx never signals a recorded pid: check the log for errors,
+            and stop that process yourself only if it stays silent for longer than the work should
+            take.
+          REPORT
+        end
+      end
+    end
+  end
+end
--- a/lib/rubydex/server/core.rb
+++ b/lib/rubydex/server/core.rb
@@ -1,0 +1,217 @@
+# frozen_string_literal: true
+
+require "socket"
+
+module Rubydex
+  module Server
+    # The resident server process. It answers one client at a time, in process, over a UNIX socket.
+    #
+    # The graph is a snapshot from boot. A file edited after that answers with its boot content.
+    class Core
+      #: (State state, ?lock: File?) -> void
+      def initialize(state, lock: nil)
+        @state = state
+        @lock = lock
+        @running = true
+        @started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      end
+
+      # Blocks for the lifetime of the server.
+      #: -> void
+      def run
+        redirect_output
+        # Recorded before the slow index, so a client can separate a boot crash from a slow start.
+        # The socket is the readiness signal, and it appears only when the graph is ready.
+        @state.record!
+
+        require "rubydex"
+        @graph = Server.build_graph(workspace_path: @state.workspace_path)
+
+        server = open_socket
+        log("rdx server ready (pid=#{Process.pid}, workspace=#{@state.workspace_path})")
+        serve(server)
+      rescue StandardError => e
+        # A detached crash reaches nobody, so the log is the only record of it.
+        log("rdx server crashed: #{e.class}: #{e.message}")
+        e.backtrace&.each { |frame| log("  #{frame}") }
+        raise
+      ensure
+        cleanup(server)
+      end
+
+      private
+
+      #: -> UNIXServer
+      def open_socket
+        File.unlink(@state.socket_path) if File.exist?(@state.socket_path)
+        # `sockaddr_un` caps the path near 104 bytes, so the bind uses a short relative name.
+        server = Dir.chdir(File.dirname(@state.socket_path)) do
+          UNIXServer.new(File.basename(@state.socket_path))
+        end
+        File.chmod(0o600, @state.socket_path)
+        server
+      end
+
+      #: (UNIXServer server) -> void
+      def serve(server)
+        while @running
+          client = begin
+            server.accept
+          rescue IOError, Errno::EBADF
+            break
+          end
+
+          handle(client)
+        end
+      end
+
+      #: (UNIXSocket client) -> void
+      def handle(client)
+        # The first write lets the client detect an old server before it sends a request.
+        client.puts(@state.expected_version)
+
+        request = Frame.read_request(client)
+        return unless request
+
+        if authorized?(request["token"])
+          dispatch(request, client)
+        else
+          answer(client, response(stderr: "rdx server: unauthorized request\n", status: 1))
+        end
+      rescue Errno::EPIPE, Errno::ECONNRESET
+        nil
+      rescue Frame::Malformed, Frame::ReadTimeout => e
+        # A peer that breaks the protocol must not stop the server.
+        log("rdx server: rejected a request: #{e.class}: #{e.message}")
+        answer(client, response(stderr: "rdx server: #{e.message}\n", status: 1))
+      rescue StandardError => e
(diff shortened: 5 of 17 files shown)


lib/rubydex/cli.rb
```diff
@@ -0,0 +1,255 @@
+# frozen_string_literal: true
```
Morriar, line 1: Do we need to split the cli right now? I think if we were to split it I'd try to have each command in it's own file.

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+require "rubydex/version"
+
+module Rubydex
+  # Client/server mode for the `rdx` executable.
+  #
+  # The expensive indexing + resolution work is performed once by a resident server process that
+  # keeps the built `Rubydex::Graph` in memory. Subsequent commands (currently `--query`) run against
+  # the already-built graph over a UNIX domain socket, making follow-up queries effectively instant.
+  #
+  # See `tmp/rdx-server-plan/README.md` for the full design.
```
Morriar: Sure

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+  module Server
+    # Wire protocol version. Bump on any incompatible change to the request/response shape.
+    PROTOCOL = 1
+
+    # How long the client waits for the server's version line / handshake before giving up.
+    HANDSHAKE_TIMEOUT = 10.0
```
Morriar: Should this be in `Client`?

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    # How long the client waits for a freshly spawned server to become ready (socket to appear).
+    BOOT_TIMEOUT = 120.0
```
Morriar: Should this be in `Client`?

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    class Error < StandardError; end
+
+    # Raised when a read from the server exceeds its timeout.
+    class ServerReadTimeout < Error; end
```
Morriar: In `Client`?

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+      #: -> bool
+      def supported?
+        Process.respond_to?(:fork) && defined?(::UNIXSocket) && !Gem.win_platform?
+      end
+
+      # Whether the user has explicitly disabled the server via the environment.
+      #: -> bool
+      def disabled?
+        ENV.key?("DISABLE_RDX_SERVER")
+      end
```
Morriar: Do we need this?

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+        # Honor the workspace's `.rubydex` config (exclusions etc.) before indexing, matching the
+        # inline CLI path. A missing default config is ignored.
+        graph.load_config
+        with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+        with_timer(progress_io, "Resolving graph...") { graph.resolve }
+        graph
+      end
+
+      #: (IO? io, String message) { -> void } -> void
+      def with_timer(io, message)
```
Morriar: Could we share the one from `cli.rb`?

lib/rubydex/server.rb
```diff
@@ -0,0 +1,75 @@
...
+  # Client/server mode for the `rdx` executable.
+  #
+  # The expensive indexing + resolution work is performed once by a resident server process that
+  # keeps the built `Rubydex::Graph` in memory. Subsequent commands (currently `--query`) run against
+  # the already-built graph over a UNIX domain socket, making follow-up queries effectively instant.
+  #
+  # See `tmp/rdx-server-plan/README.md` for the full design.
+  module Server
+    # Wire protocol version. Bump on any incompatible change to the request/response shape.
+    PROTOCOL = 1
```
Morriar: Where do we ever enforce this?

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
...
+          socket.close
+
+          unless response
+            stderr.puts("rdx server: empty response")
+            return 1
+          end
+
+          stdout.print(response["stdout"]) if response["stdout"]
+          stderr.print(response["stderr"]) if response["stderr"]
+          response["status"] || 0
```
Morriar: Should we expect all responses to always have a status, should we make the client/server error if the response doesn't have a status?

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
+# frozen_string_literal: true
+
+require "socket"
+
+module Rubydex
+  module Server
+    # The short-lived client side. Ensures a compatible server is running for the workspace, then
+    # talks to it over the UNIX socket. Kept deliberately lightweight: it never `require`s the full
+    # `rubydex` extension on the happy path, so the server actually saves time.
```
Morriar:
```suggestion
    # `rubydex` extension on the happy path, so the client actually saves time.
```
?

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
...
+            return 1
+          end
+
+          stdout.print(response["stdout"]) if response["stdout"]
+          stderr.print(response["stderr"]) if response["stderr"]
+          response["status"] || 0
+        end
+
+        # Ensures a server is running, then connects with the version handshake. Restarts once on a
+        # failed connect.
```
Morriar: Do we have a rule to break comments at 80 chars? 🤔

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
...
+        def stop(cache)
+          pid = cache.server_pid
+
+          if pid && process_alive?(pid)
+            graceful_stop(cache, pid)
+
+            if process_alive?(pid)
+              terminate(pid)
+              wait_for_exit(pid)
+            end
```
Morriar: Beware, if the server crashed and the pid was reused you may be terminating the wrong process.

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
...
+
+        #: (Integer pid) -> void
+        def terminate(pid)
+          Process.kill("TERM", pid)
+        rescue Errno::ESRCH
+          nil
+        end
+
+        #: (Integer pid) -> bool
+        def process_alive?(pid)
```
Morriar: It may be safer to authenticate the process as well, like establish a hand check or ping a health endpoint so you know the process with this pid is actually a rdx server?

lib/rubydex/server/client.rb
```diff
@@ -0,0 +1,248 @@
...
+            "cwd" => Dir.pwd,
+            "argv" => ARGV,
+            "env" => {},
+          }
+        end
+
+        #: -> Float
+        def monotonic
+          Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        end
```
Morriar: Smelly extraction, other places use `Process.clock_gettime(Process::CLOCK_MONOTONIC)` directly?

lib/rubydex/server/request.rb
```diff
@@ -0,0 +1,51 @@
...
+          data = JSON.dump(payload)
+          socket.puts(data.bytesize)
+          socket.write(data)
+          socket.flush
+        end
+
+        # Reads a framed JSON payload from `socket`. Returns `nil` on EOF / malformed length.
+        #: (IO socket) -> Hash[untyped, untyped]?
+        def read(socket)
+          length_line = socket.gets
```
Morriar: Should we timeout on the read? If the client sends an incorrectly sized query the server will block forever

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+        request = Request.read(client)
+        return unless request
+
+        unless authorized?(request["token"])
+          Request.write(client, response(stderr: "rdx server: unauthorized request\n", status: 1))
+          return
+        end
+
+        dispatch(request, client)
+      rescue Errno::EPIPE, Errno::ECONNRESET
```
Morriar, line 81: Should we also rescue on `JSON::ParseError`?

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+        # restart it before sending any request.
+        client.puts(@cache.expected_version)
+
+        request = Request.read(client)
+        return unless request
+
+        unless authorized?(request["token"])
+          Request.write(client, response(stderr: "rdx server: unauthorized request\n", status: 1))
+          return
+        end
```
Morriar: Do we even need that?

lib/rubydex/server/request.rb
```diff
@@ -0,0 +1,51 @@
...
+          body = socket.read(length)
+          return unless body
+
+          JSON.parse(body)
+        end
+
+        # Reads a single line from `socket`, raising `ServerReadTimeout` if nothing arrives within
+        # `timeout` seconds. Used for the handshake where a wedged server must never hang the client.
+        #: (IO socket, Float timeout) -> String
+        def read_line_with_timeout(socket, timeout)
```
Morriar: Ah I'm confused why we need both `read` and `read_line_with_timeout`?

lib/rubydex/server/request.rb
```diff
@@ -0,0 +1,51 @@
...
+          socket.puts(data.bytesize)
+          socket.write(data)
+          socket.flush
+        end
+
+        # Reads a framed JSON payload from `socket`. Returns `nil` on EOF / malformed length.
+        #: (IO socket) -> Hash[untyped, untyped]?
+        def read(socket)
+          length_line = socket.gets
+          return unless length_line
```
Morriar: Should we enforce a higher limit before trying to read terabytes from the socket?

lib/rubydex/server/request.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          length = length_line.to_i
```
Morriar:
```ruby
   "100garbage".to_i # => 100
 ```

are we okay with this?

File: lib/rubydex/server/request.rb [hunk 21]
```diff
@@ -0,0 +1,51 @@
...
+          return if length <= 0
+
+          body = socket.read(length)
+          return unless body
+
+          JSON.parse(body)
```
Morriar, line 178: Should we also make sure the thing we parsed is an object?

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+      # `glob` crate's `Pattern::matches_path` semantics.
+      #: (String path, Array[String] patterns) -> bool
+      def excluded_by_patterns?(path, patterns)
+        patterns.any? { |pattern| File.fnmatch?(pattern, path, File::FNM_PATHNAME) }
+      end
+
+      #: (String path) -> String
+      def uri_for(path)
+        path = "/#{path}" if Gem.win_platform?
+        URI::File.build(path: path).to_s
```
Morriar: I think this will raise `URI::InvalidComponentError` but we never handle it and it will crash the server?

Morriar: Actually I think we need to expose the Rust behavior through the API so you can reuse it here.

Let's take a file at `/a b.rb` for example, Rust will return `file:///a%20b.rb` while Ruby will raise. Maybe we should share the same behavior.

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+          Request.write(client, response(stderr: "rdx server: unauthorized request\n", status: 1))
+          return
+        end
+
+        dispatch(request, client)
+      rescue Errno::EPIPE, Errno::ECONNRESET
+        nil
+      ensure
+        begin
+          client.close
```
Morriar, line 94: That means we'll kill the server for any kind of error (invalid query, invalid path, etc.)?

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+      #: (Hash[untyped, untyped] request) -> Hash[String, untyped]
+      def handle_query(request)
+        query = request["query"]
+        format = request["query_format"] || "table"
+
+        @mutex.synchronize do
+          refresh_if_stale
+          response(stdout: Rubydex::Query.parse(query).render(@graph, format))
+        end
+      rescue ArgumentError => e
```
Morriar: I think this is both too broad and too narrow, many other kind of errors should be logged while at the same time you may rescue a random argument error from a down the stack?

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+          full = File.join(dir, entry)
+          next if excluded_by_patterns?(full, patterns)
+
+          if File.directory?(full)
+            collect_files(full, manifest, patterns)
+          elsif Rubydex::Graph::INDEXABLE_EXTENSIONS.include?(File.extname(entry))
+            manifest[full] = File.mtime(full).to_f
+          end
+        end
+      rescue Errno::ENOENT, Errno::EACCES
```
Morriar: Should we rescue inside the look and call `next`? If one entry is missing we'll discard all the remaining entries

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+      # Detects workspace files that changed since the graph was built and applies incremental
+      # updates before answering. Always correct, occasionally slow (Phase 1 freshness model).
+      #: -> void
+      def refresh_if_stale
+        current = workspace_manifest
+        changed = current.select { |path, mtime| @manifest[path] != mtime }.keys
+        deleted = @manifest.keys - current.keys
+        return if changed.empty? && deleted.empty?
+
+        @graph.index_all(changed) unless changed.empty?
```
Morriar: I think it's supposed to return an array of errors, you may want to check it before overwriting the whole manifest

lib/rubydex/server/core.rb
```diff
@@ -0,0 +1,249 @@
...
+        manifest
+      end
+
+      #: (String dir, Hash[String, Float] manifest, Array[String] patterns) -> void
+      def collect_files(dir, manifest, patterns)
+        Dir.each_child(dir) do |entry|
+          full = File.join(dir, entry)
+          next if excluded_by_patterns?(full, patterns)
+
+          if File.directory?(full)
```
Morriar: I _think_ this behavior is also different form the Rust one. Do we follow symlinks with the Rust file collect?

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+      end
+
+      # Creates the runtime directory (and its parent) with restrictive permissions.
+      #: -> void
+      def ensure_dir!
+        FileUtils.mkdir_p(runtime_root, mode: 0o700)
+        FileUtils.mkdir_p(dir, mode: 0o700)
+        # mkdir_p won't tighten perms on a pre-existing dir; enforce them explicitly.
+        File.chmod(0o700, runtime_root)
+        File.chmod(0o700, dir)
```
Morriar: Should we create a subdirectories instead to avoid applying chmod to the directory the user pass through `RDX_SERVER_DIR`? Could be something like `$RDX_SERVER_DIR/rubydex-<uid>/<app-id>/`

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+        pid.zero? ? nil : pid
+      end
+
+      # Whether a process with the recorded pid is alive.
+      #: -> bool
+      def server_alive?
+        pid = server_pid
+        return false unless pid
+
+        Process.kill(0, pid)
```
Morriar: I could be any random process behind this pid though

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+      #: -> bool
+      def version_compatible?
+        File.exist?(version_path) && File.read(version_path).chomp == expected_version
+      end
+
+      #: -> Integer?
+      def server_pid
+        return unless File.exist?(pid_path)
+
+        pid = File.read(pid_path).to_i
```
Morriar: Let's hope nobody is ending up with `-1` in this file 😂

lib/rubydex/server/cache.rb
```diff
@@ -0,0 +1,200 @@
...
+      #: (?workspace_path: String) -> void
+      def initialize(workspace_path: Dir.pwd)
+        @workspace_path = File.expand_path(workspace_path)
+      end
+
+      # Stable identity for this workspace's server.
+      #: -> String
+      def app_id
+        @app_id ||= Digest::SHA256.hexdigest(
+          [@workspace_path, RUBY_VERSION, Rubydex::VERSION, ext_fingerprint].join("\0"),
```
Morriar: Maybe this is where you wanted to use `PROTOCOL`?

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+
+      #: -> String
+      def version_path
+        File.join(dir, "version")
+      end
+
+      #: -> String
+      def lock_path
+        File.join(dir, "lock")
+      end
```
Morriar, line 68: This makes the server's metadata non-atomic which means concurrent clients may observe the intermediate state (`pid` written but `version` missing). wdyt of writing all matadata at once in one file?

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+
+      # Returns the existing token, generating and persisting one if absent.
+      #: -> String
+      def token
+        return @token if @token
+
+        @token = if File.exist?(token_path)
+          File.read(token_path).chomp
+        else
+          generate_token
```
Morriar: Concurrent clients may be trouble here, maybe you should do both the check and write in one shot:

```rb
File.open(token_path, File::WRONLY | File::CREAT | File::EXCL, 0o600) do |file|
  file.write(value)
end
```

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,200 @@
...
+        Process.kill(0, pid)
+        true
+      rescue Errno::ESRCH, Errno::EPERM
+        false
+      end
+
+      # Removes stale artifacts left behind by a dead server.
+      #: -> void
+      def clean!
+        [socket_path, pid_path, version_path].each do |path|
```
Morriar: Should we also clean `lock`? 🤔

lib/rubydex/server/cache.rb
```diff
@@ -0,0 +1,200 @@
...
+module Rubydex
+  module Server
+    # Owns the per-workspace runtime directory and the metadata files inside it (socket, pid, token,
+    # version, lock).
+    #
+    # The directory is keyed by an "app id" derived from the workspace path, the Ruby version, the
+    # rubydex version and a fingerprint of the loaded native extension. A gem upgrade (or a different
+    # Ruby) therefore maps to a different directory and forces a fresh server, since the loaded C
+    # extension cannot be hot-swapped.
+    class Cache
```
Morriar: I think the name is confusing it's more a server `state` than a server `cache`?

lib/rubydex/server/commands.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+
+        #: (Cache cache, ?stdout: IO, ?stderr: IO) -> Integer
+        def status(cache, stdout: $stdout, stderr: $stderr)
+          unless cache.server_alive? && File.socket?(cache.socket_path)
+            stdout.puts("rdx server: not running for #{cache.workspace_path}")
+            return 0
+          end
+
+          Client.request(cache, { "command" => "status" }, stdout: stdout, stderr: stderr)
```
Morriar: Doesn't that start a server? So checking the status _always_ starts a server?

lib/rubydex/server/commands.rb
```diff
@@ -0,0 +1,54 @@
...
+        #: (Cache cache, ?detach: bool, ?stdout: IO) -> Integer
+        def restart(cache, detach: true, stdout: $stdout)
+          Client.restart(cache, detach: detach)
+          stdout.puts("rdx server restarted (pid=#{cache.server_pid}, socket=#{cache.socket_path})")
+          0
+        end
+
+        #: (Cache cache, ?stdout: IO, ?stderr: IO) -> Integer
+        def status(cache, stdout: $stdout, stderr: $stderr)
+          unless cache.server_alive? && File.socket?(cache.socket_path)
```
Morriar: Should we also check version compatibility?

lib/rubydex/server/commands.rb
```diff
@@ -0,0 +1,54 @@
...
+      class << self
+        #: (Cache cache, ?detach: bool, ?stdout: IO) -> Integer
+        def start(cache, detach: true, stdout: $stdout)
+          if cache.server_alive? && cache.version_compatible? && File.socket?(cache.socket_path)
+            stdout.puts("rdx server already running (pid=#{cache.server_pid}, socket=#{cache.socket_path})")
+            return 0
+          end
+
+          Client.ensure_server(cache, detach: detach)
+          # With --no-detach the server runs in the foreground and only returns here once stopped.
```
Morriar: Which means we log that the server started before it's true?

Morriar: Do we even need to log again? It's done in `core` already?

lib/rubydex/server/commands.rb
```diff
@@ -0,0 +1,54 @@
...
+          Client.restart(cache, detach: detach)
+          stdout.puts("rdx server restarted (pid=#{cache.server_pid}, socket=#{cache.socket_path})")
+          0
+        end
+
+        #: (Cache cache, ?stdout: IO, ?stderr: IO) -> Integer
+        def status(cache, stdout: $stdout, stderr: $stderr)
+          unless cache.server_alive? && File.socket?(cache.socket_path)
+            stdout.puts("rdx server: not running for #{cache.workspace_path}")
+            return 0
```
Morriar, line 52: Should we return something else than `0` here?

lib/rubydex/server/commands.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Rubydex
+  module Server
+    # Implementations for the server-control CLI flags (`--start-server`, `--stop-server`,
+    # `--restart-server`, `--server-status`). Each returns a process exit status.
```
Morriar: You eman `rdx server start|stop|restart|status`?

lib/rubydex/cli.rb
paracycle: Yes, I split it out as #971, which is merged now: one class per command under `lib/rubydex/cli/command/`. 

Now, this PR only adds `command/server.rb` and the `--server` flag in `command/query.rb`.

lib/rubydex/server.rb
paracycle: Done, the comment does not point at a non-existent file any more.

lib/rubydex/server.rb
paracycle: Moved. It is `Client::HANDSHAKE_TIMEOUT` now.

lib/rubydex/server.rb
paracycle: Same here. It is `Client::BOOT_TIMEOUT` now, and `Client::STOP_TIMEOUT` moved along with it.

lib/rubydex/server.rb
paracycle: Actually, timeouts belong to the protocol rather than to the client (since the server can also time out waiting to read from the client, for example), so it is `Frame::ReadTimeout` now, defined next to `Frame::Malformed` and the code that
raises both.

lib/rubydex/server.rb
paracycle: It is now `Rubydex::Progress.with_timer` in `lib/rubydex/progress.rb`, called by both
`CLI::Command#build_graph` and `Server.build_graph`.

lib/rubydex/server.rb
paracycle: Nowhere at the time, you were right. `PROTOCOL` is an input to `app_id` now, so bumping it changes the runtime directory and the socket: a new client cannot reach an old server, because it does not address it. There is also a test now that asserts that the id changes with the protocol.

lib/rubydex/server/client.rb
paracycle: Yes, I also think it should error, and it does now. A response whose `status` is not an `Integer` is rejected as a malformed frame instead of being treated as success. Same for a server that closes without sending a frame at all.

lib/rubydex/server.rb
paracycle: No, we don't, and it is gone now. `--server` is opt-in and defaults to off, so the env flag only did something when you passed `--server` and then set a variable telling us not to use the server. 🙈 

If we ever make the server the default, an opt-out might make sense.

lib/rubydex/server/client.rb
paracycle: No rule, AFAIK. `Layout/LineLength` is disabled by `rubocop-shopify`, so nothing enforces a width. Happy to use something else if you prefer.

lib/rubydex/server/client.rb
paracycle: Good catch. I removed the signal entirely. Liveness is controlled by `flock` on the lock file now, which the kernel releases it when the holder dies. So a recycled pid can never look alive and `rdx` never signals a recorded pid. The pid is now only in the report so a human can find the process.

lib/rubydex/server/client.rb
paracycle: Implemented. The lock proves a live process owns the workspace. The handshake proves
identity: the server writes a version line, the client compares it to its own expectation, and then sends a per-boot token that the server checks before it dispatches anything. A wrong token gets refused with status `1`.

lib/rubydex/server/client.rb
paracycle: Removed, and the call sites use `Process.clock_gettime(Process::CLOCK_MONOTONIC)`
directly now.

lib/rubydex/server/request.rb
paracycle: Yeah. Every read has a deadline now: `REQUEST_TIMEOUT` (10s) for a whole request, `IDLE_TIMEOUT` (10s) between chunks of a response, and a bounded `read_line` for the version line. A client that announces a size and then stalls gets dropped.

lib/rubydex/server/core.rb
paracycle: Yeap. Bad JSON, a bad length line, an oversized frame and a truncated read all raise
`Frame::Malformed` now, and the connection handler answers with an error and keeps serving. A `JSON::ParserError` never gets raised past the `Frame` class.

lib/rubydex/server/core.rb
paracycle: `read_frame` returns `nil` when the peer closes before it sends a length line, which is exactly what a client that connects and hangs up does. The branch around it was the confusing part, so the authorization check is an explicit `if`/`else` now, instead of a second bare return.

lib/rubydex/server/request.rb
paracycle: Sorry, the first one was doing two jobs badly. There is one deadline-driven loop now, `read_frame`, with three entry points on top of it: `read_line` for the version line, `read_request` and `read_response`. The duplicated timeout logic is gone.

Btw, while cleaning this up I renamed the module too. It reads and writes in both directions, so `Request` named one of the four things it does. It is `Frame` now, in
`lib/rubydex/server/frame.rb`, and `MalformedFrame` became `Frame::Malformed`.

lib/rubydex/server/request.rb
paracycle: Added, and they are checked before any bytes are read: `MAX_LINE_BYTES` 1 KiB for the length line, `MAX_REQUEST_BYTES` 1 MiB for a request, `MAX_RESPONSE_BYTES` 512 MiB for a response.

lib/rubydex/server/request.rb
paracycle: No, not ok. The length line has to match `/\A\d+\n?\z/` now, otherwise the frame is
malformed. `to_i` is not used on it at all.

lib/rubydex/server/request.rb
paracycle: Yes, a payload that is not a JSON object is a malformed frame. The server also checks the type of `command` and `query` before it uses them.

lib/rubydex/server/core.rb
paracycle: Agreed, this is not great. I removed file addition/change/deletion related partial updates from this PR, and extracted them to #979 so that we can implement this more cleanly.

lib/rubydex/server/core.rb
paracycle: That was bad, yeah. Each connection is wrapped now: `StandardError` gets logged with its backtrace, answered as status `1`, and the accept loop continues. A query error never ends the process.

lib/rubydex/server/core.rb
paracycle: Agreed on both counts. The rescue is narrowed to the parse and the render, which are the only places that produce a use visible error. Everything else is an internal error: logged with class and backtrace, returned as status `1`. So a stray `ArgumentError` from further down does not look like a bad query any more.

lib/rubydex/server/core.rb
paracycle: Agreed but postponed to #979. In that PR, each entry is handled on its own now. A failure records the error for a file and continues, so one unreadable entry does not discard the rest.

lib/rubydex/server/core.rb
paracycle: You are right, but that moved to #979. There the indexer's errors are read and
attributed to the file that failed, so a file that could not be indexed is retried on the next update instead of being recorded as fresh.

lib/rubydex/server/core.rb
paracycle: Good catch, the Ruby walk did differ. It lives in #979 now, where it follows symlinked directories the way the Rust collector does, records the path it was reached by, and refuses to loop. There are tests for a cycle, a top-level symlinked directory, and a symlink to a file.

lib/rubydex/server/state.rb
paracycle: Good call. The layout is `$RDX_SERVER_DIR/rubydex-<uid>/<app-id>/` now, as you suggested, and only those two levels get created and chmod'ed.

Following that up turned up more than the chmod, so this ended up being the biggest change of the round:

- The base is canonicalized with `realpath` before use. A symlink component inside a sticky directory can be swapped by whoever owns the link, even when its target is perfectly safe, and a plain `stat` would follow it and approve.
- The base has to exist already, and every directory in its canonical chain has to be owned by this user or root, and either closed to other writers or sticky like `/tmp`. A missing base is refused rather than created, because creating it means racing anyone else who can write in its parent.
- The two directories we do create are opened with `O_NOFOLLOW`, then validated and chmod'ed through that descriptor, so the inode cannot change between the check and the change.
- The lock file and the state file are opened the same way, and `server_running?`, `clean!` and `read` verify the directories before they touch anything. They run before `ensure_dir!`, so `O_NOFOLLOW` on the final component alone would not have covered them.
- A platform without `O_NOFOLLOW` fails closed: `Server.supported?` reports it and the directory code refuses, rather than falling back to a path-based `chmod`.

There are tests for a world-writable base, a sticky one, a missing one, a symlinked base with a safe target, a runtime root left open to others, a swapped runtime root, and a symlink planted where the state file and the lock go. I checked each one by putting the old behaviour back and confirming that the test fails.

lib/rubydex/server/state.rb
paracycle: Exactly, so the pid is not the liveness test any more. Instead, we use `flock` which the kernel releases when the holder dies, so a recycled pid cannot look alive. The pid is printed only so a human can find the process.

lib/rubydex/server/state.rb
paracycle: Haha. `server_pid` now returns the value only when it is a positive Integer, and `nil` otherwise. Moreover, it is now a display-only value anyway.

lib/rubydex/server/cache.rb
paracycle: Yeap, that was the intent. `app_id` now hashes `PROTOCOL` along with the workspace path, the Ruby version, the gem version and the extension fingerprint. So bumping the protocol gives a different directory and a fresh server. There is also a test for it.

lib/rubydex/server/state.rb
paracycle: Agreed, and I did it the way you suggest. Pid, token, version and start time are all one `state.json` now, written to a temporary file in the same directory and renamed into place. A reader sees the old record or the new one, never a half-written one.

lib/rubydex/server/state.rb
paracycle: There is no separate token file any more, so the race is gone. The token is one
field of the atomic `state.json`, written by the server under its exclusive lock at boot. A fresh token per boot also means a client holding an old token cannot talk to a new server.

lib/rubydex/server/cache.rb
paracycle: Yeah, agreed. It is `Server::State` now. It owns the socket, the lock and the recorded metadata, so `state` is a better name.

lib/rubydex/server/commands.rb
paracycle: It did, but not any more: `status` now goes through `Client.probe`, which never starts, stops or restarts anything. Only `start` and `restart` create a server.

lib/rubydex/server/commands.rb
paracycle: It is checked by two different mechanisms. The gem version and the fingerprint of the compiled extension are checked as part of handshake (`state.expected_version`). The protocol and the Ruby version are checked as part of `app_id`, which now picks
the runtime directory and the socket, so a different protocol version or Ruby addresses a different socket and an incompatible server is never reached at all.

lib/rubydex/server/commands.rb
paracycle: Both are fixed. `ensure_server` blocks until the server answers, so the message follows readiness now instead of the fork.

On logging it twice: the two writes have different audiences, so I kept one of each and removed the overlap. `Core` logs the boot and any failure to the server log, which is for troubleshooting. `Commands` prints one line to the user's stdout, however that line does not repeat what the log holds any more, it is just `rdx server started`, with no pid and no socket path.

lib/rubydex/server/commands.rb
paracycle: I split it. `0` means the question got answered successfully. Either no server is running or a healthy server exists. `1` means a server holds the
lock but did not answer, which is more of a failure situation where we can't properly answer the `status` question.

lib/rubydex/server/commands.rb
paracycle: Yeah, the PR had a lot of churn as the base branch changed shape. The doc comment names the four subcommands now.

lib/rubydex/server/state.rb
paracycle: No, and deliberately so. `flock` binds to the inode, so unlinking the lock would let a second server lock a fresh inode while the first still holds the old one, and both would think they own the workspace. `clean!` takes the lock, removes only the socket, and leaves the lock file. It is one empty file per workspace.

lib/rubydex/cli.rb
```diff
@@ -26,6 +26,14 @@ def start(argv = ARGV)
         require "rubydex"
```
Morriar, line 26: Doesn't this defeat the `--server` fast path? `CLI.start` still requires `rubydex` before dispatch, so `rdx query --server ...` loads the native extension in the client before `Query#server_available?` runs. That means the client keeps paying the extension load cost and can still fail when the local extension is incompatible, even if a server could answer. I think we need to move the native require into the inline commands/paths (`query --schema`, `query` without `--server`, console/mcp as needed) and keep the top-level dispatcher Ruby-only.

lib/rubydex/server/state.rb
```diff
@@ -0,0 +1,389 @@
...
+        Process.respond_to?(:uid) ? Process.uid : "nobody"
+      end
+
+      # A recompiled extension invalidates a running server, because Ruby cannot reload it in place.
+      #: -> String
+      def ext_fingerprint
+        @ext_fingerprint ||= begin
+          lib_dir = File.expand_path("../..", __dir__)
+          artifacts = Dir.glob(File.join(lib_dir, "**", "rubydex.{bundle,so}")) +
+            Dir.glob(File.join(lib_dir, "librubydex_sys.*"))
```
Morriar, line 374: Looks like this misses the dylib we actually load. `lib_dir` is `.../lib`, but extconf copies `librubydex_sys.dylib` into `lib/rubydex/`, so this glob is empty. A Rust-only rebuild can leave `rubydex.bundle` unchanged while changing the dylib, and then `app_id`/`expected_version` won't change, so clients can keep talking to a server with stale native code. Can we make this recursive too?

```suggestion
            Dir.glob(File.join(lib_dir, "**", "librubydex_sys.*"))
```

PR #870, opened 2026-06-19
Add `Location.from_prism` helper

### LINKED ISSUE #849: Add a `Prism::Location` → `Rubydex::Location` conversion helper
Converting a `Prism::Location` to a `Rubydex::Location` has to be done by hand because of a convention mismatch: `Prism` uses 1-based lines / 0-based columns, while `Rubydex::Location` is 0-based on both. So lines need - 1 and columns pass through:

```ruby
Rubydex::Location.new(
  uri: uri,
  start_line: prism_location.start_line - 1,
  start_column: prism_location.start_column,
  end_line: prism_location.end_line - 1,
  end_column: prism_location.end_column,
)
```

Proposing a constructor that encapsulates this:

```ruby
Rubydex::Location.from_prism(prism_location, uri:)
```

Equality is already handled by `Comparable`, so with shared construction, comparing a converted location to a `Rubydex::Location` is just `==`.

alexcrocha:
Closes #849

`Prism::Location` and `Rubydex::Location` use different line conventions: Prism lines are 1-based, while Rubydex internal locations are 0-based. Callers that need to compare Prism-derived locations with Rubydex locations currently have to repeat the same conversion by hand.

This PR adds `Rubydex::Location.from_prism(prism_location, uri:)` so the conversion lives behind the `Location` API:

```ruby
Rubydex::Location.from_prism(prism_location, uri:)
```

The helper subtracts 1 from Prism's line values and keeps Prism's default column values as-is. `DisplayLocation.from_prism` is intentionally unavailable; callers that need a display location can convert explicitly with `Rubydex::Location.from_prism(...).to_display`.

### DIFF
--- a/lib/rubydex/location.rb
+++ b/lib/rubydex/location.rb
@@ -14,6 +14,19 @@ module Rubydex
     #: Integer
     attr_reader :start_line, :end_line, :start_column, :end_column
 
+    class << self
+      #: (Prism::Location prism_location, uri: String) -> Location
+      def from_prism(prism_location, uri:)
+        Location.new(
+          uri: uri,
+          start_line: prism_location.start_line - 1,
+          start_column: prism_location.start_column,
+          end_line: prism_location.end_line - 1,
+          end_column: prism_location.end_column,
+        )
+      end
+    end
+
     #: (?uri: String, ?start_line: Integer, ?end_line: Integer, ?start_column: Integer, ?end_column: Integer) -> void
     def initialize(uri:, start_line:, end_line:, start_column:, end_column:)
       @uri = uri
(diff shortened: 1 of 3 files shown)


lib/rubydex/location.rb
```diff
@@ -14,6 +14,19 @@ class NotFileUriError < StandardError; end
     #: Integer
     attr_reader :start_line, :end_line, :start_column, :end_column
 
+    class << self
+      #: (untyped prism_location, uri: String) -> Location
```
vinistock:
```suggestion
      #: (Prism::Location prism_location, uri: String) -> Location
```

lib/rubydex/location.rb
```diff
@@ -68,6 +81,10 @@ def to_s
   # A one based location intended for display purposes. This is what should be used when displaying a location to users,
   # like in CLIs
   class DisplayLocation < Location
+    class << self
+      undef_method :from_prism
```
vinistock: If we only undefine the method, then consumers of the API will get a `NoMethodError` if they try to call `from_prism` on a `DisplayLocation`, but it might not be obvious why that's not possible.

What do you think about re-defining the method to raise with a message explaining why this isn't available? Something like this:

```ruby
def from_prism(location, uri)
  raise NotImplementedError, <<~MESSAGE
    Cannot convert Prism::Location directly to a Rubydex::DisplayLocation.
    Start with `Prism::Location.from_prism(...)` and then convert the resulting
    location with `to_display`
  MESAGE
end
```

alexcrocha: Ah, yes. I like that, thanks!

vinistock (approved):

PR #873, opened 2026-06-23
Return Cypher query results as graph objects

paracycle:
## Goal

Build on the Cypher query engine from #868 so callers get **matched graph nodes back as first-class objects**, not as formatted text they have to re-parse or iterate over.

#868 added `Rubydex::Query#render(graph, format)`, which runs a query and returns a `table`/`json` **string** — great for the CLI and humans, but a dead end for programmatic callers: to actually *use* a matched class or method you'd have to parse the formatted output, then look the node back up in the graph by name. This PR adds an object-returning sibling, `Query#run(graph)`, that hands you the matched nodes directly.

## What changes

- **`Rubydex::Query#render(graph, format = :table)`** — unchanged: rows → formatted `String`.
- **`Rubydex::Query#run(graph)`** — new: rows → `Array<Hash>` (`T::Array[T::Hash[String, T::untyped]]`), where the values are real Ruby objects:
  - a column that binds a node (e.g. `RETURN c`) comes back as a live `Rubydex::Declaration` / `Definition` / `Document` handle;
  - scalar columns (`RETURN c.name`, `count(c)`) come back as plain Ruby values (`String`, `Integer`, `true`/`false`, `nil`);
  - a map projection (`RETURN c { .name, .kind }`) comes back as a Ruby `Hash`, and a list comes back as a Ruby `Array` (elements decoded recursively).

So instead of:

```ruby
# Before: query returns a string; you re-parse it, then re-find the node.
text = query.render(graph, :json)
JSON.parse(text).each do |row|
  decl = graph.find_declaration(row["c.name"])  # extra lookup just to get an object
  decl.definitions.each { ... }
end
```

you write:

```ruby
# After: the matched node IS the object. No iteration to re-resolve it.
query = Rubydex::Query.parse("MATCH (c:Class)-[:HAS_PARENT]->(:Class {name: 'Animal'}) RETURN c")
query.run(graph).each do |row|
  decl = row["c"]            # => Rubydex::Declaration, already resolved
  decl.definitions           # navigate the graph directly
end
```

The query language does the matching and filtering; you get the resulting nodes back ready to use, without writing your own traversal/iteration to turn names back into graph objects.

## How it works

- **`cypher-parser` 0.8.1**: the generic executor carries node identity and structured values. `CypherValue` covers `Null`/`Bool`/`Int`/`Str` scalars, `Node { id, name, .. }` for a bound graph node, and `List` / `Map` containers (so map projections like `c { .name, .kind }` survive execution). `GraphProvider` gains:
  - `node_id(node)` — encodes a node to an opaque id, so a bound node survives execution as an id rather than being flattened to text;
  - `expand_in(node, rel_type)` — walks a relationship **backwards** (incoming edges), enabling queries like `(:Definition)<-[:DEFINES]-(:Document)`.

  The crate stays dependency-free and rubydex-agnostic.
- **rubydex schema** (`query::cypher::schema`): implements `node_id` and `NodeRef::decode` over the existing `decl:/def:/doc:<u64>` id scheme, and implements `expand_in` for the reversible relationships (`DEFINES`, `DECLARES`) so incoming traversals resolve to the right nodes.
- **FFI** (`rubydex-sys`): all Cypher-related code lives in a new `cypher_api.rs` module — `CQueryResult`, `CParseResult`, `rdx_cypher_parse`, `rdx_cypher_query_free`, `rdx_cypher_schema`, and the structured-result types. Only `rdx_query_run` (the formatted-string runner) remains in `graph_api.rs`. `rdx_query_run_rows` returns a `CRunRows` wrapping a `CRowsIter` opaque iterator (following the existing `DeclarationsIter` / `DocumentsIter` pattern), walked via `rdx_rows_iter_{column_count,columns,len,next,free}`. `CCell` is a tagged union (`CCellTag`: `Null`/`Bool`/`Int`/`Str`/`Node`/`List`/`Map`); `Node` cells carry a category (declaration/definition/document) plus the entity id, and `Map` stores parallel key/value arrays. A `cstring_raw` helper was added to `utils.rs` for allocating owned C strings across the FFI boundary.
- **Gem** (`ext/rubydex/query.c`, extending the file added in #868): `Query#run` walks the `CRowsIter` iterator into an `Array<Hash>`, decoding each cell — node cells into the appropriate `Declaration` / `Definition` / `Document` handle, lists into `Array`, maps into `Hash`, scalars into plain values. The iterator walk is wrapped in `rb_ensure` so `rdx_rows_iter_free` runs even if Ruby raises during cell conversion. Both `render` and `run` use a new `rdxi_graph_from_object` helper in `graph.h` to extract the graph pointer from the Graph VALUE.

## Why this matters for both APIs

- **Ruby API**: queries become a graph-navigation tool, not just a reporting tool. Match with Cypher, then call methods on the returned handles — no manual name-based re-lookup, no iterating the whole graph to find what the query already found.
- **Rust API**: the executor now preserves node identity end-to-end (`CypherValue::Node`) and supports incoming-edge traversal (`expand_in`), so any `GraphProvider`-backed consumer — Rust callers, the FFI layer, future language servers/tools — can get matched nodes back by id and resolve them to their own representation, rather than being limited to formatted strings.

In short: **the query does the matching; callers get the nodes, not a transcript of them.**

## Verification

- `cargo build` / `cargo test` green; clippy clean (`-D warnings`); `cargo fmt --check` clean.
- `rake compile` green; `bundle exec ruby -Itest test/graph_test.rb` → 110 runs, 0 failures (includes new object-result, map-projection, aggregation, reverse-traversal, and label-disjunction tests).

### DIFF
--- a/ext/rubydex/query.c
+++ b/ext/rubydex/query.c
@@ -1,4 +1,7 @@
 #include "query.h"
+#include "declaration.h"
+#include "definition.h"
+#include "document.h"
 #include "graph.h"
 #include "rustbindings.h"
 #include "utils.h"
@@ -77,8 +80,7 @@ static VALUE rdxr_query_render(int argc, VALUE *argv, VALUE self) {
     void *query;
     TypedData_Get_Struct(self, void *, &query_type, query);
 
-    void *graph;
-    TypedData_Get_Struct(graph_obj, void *, &graph_type, graph);
+    void *graph = rdxi_graph_from_object(graph_obj);
 
     struct CQueryResult result = rdx_query_run(query, graph, rdxi_symbol_or_string_cstr(format, "table"));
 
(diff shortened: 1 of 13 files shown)


ext/rubydex/query.c
```diff
@@ -96,10 +99,102 @@ static VALUE rdxr_query_render(int argc, VALUE *argv, VALUE self) {
...
+ * keyed by RETURN column name. Scalar cells become String/Integer/true/false/nil, lists become
+ * Arrays, and node cells become Declaration / Definition / Document handles. Raises ArgumentError
+ * on an execution error.
+ */
+static VALUE rdxr_query_run(VALUE self, VALUE graph_obj) {
+    void *query;
+    TypedData_Get_Struct(self, void *, &query_type, query);
+
+    void *graph;
+    TypedData_Get_Struct(graph_obj, void *, &graph_type, graph);
```
vinistock: Could we use `rdxi_graph_from_handle` here?

rust/rubydex-sys/src/graph_api.rs
```diff
@@ -1183,6 +1185,381 @@ pub unsafe extern "C" fn rdx_query_run(
     })
 }
 
+/// Tag for a structured result cell.
+#[repr(C)]
```
vinistock: Can we move all of these into a `cypher_api.rs` file and leave only the graph run method here? That way we can keep all cypher related things in a single place and just connect to the graph.

rust/rubydex-sys/src/graph_api.rs
```diff
@@ -1183,6 +1185,381 @@ pub unsafe extern "C" fn rdx_query_run(
...
+    /// The `CDeclarationKind`/`DefinitionKind` value (ignored for documents).
+    pub kind: u32,
+    /// The entity id to build the handle from.
+    pub id: u64,
+}
+
+/// `List` cell payload: a heap array of nested cells (freed by `rdx_result_set_free`).
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub struct CList {
```
vinistock: Should this use an [iterator](https://github.com/Shopify/rubydex/blob/55c571f9ae323c3c7f66860ced53b92b676aeb98/rust/rubydex-sys/src/lib.rs#L18) instead?

rust/rubydex-sys/src/graph_api.rs
```diff
@@ -1183,6 +1185,381 @@ pub unsafe extern "C" fn rdx_query_run(
...
+pub struct CRunRows {
+    /// Non-null on success; free with `rdx_result_set_free`.
+    pub result: *mut CResultSet,
+    /// Non-null on error; free with `free_c_string`.
+    pub error: *const c_char,
+}
+
+fn cstring_raw(value: &str) -> *const c_char {
+    CString::new(value).map_or(ptr::null(), |s| s.into_raw().cast_const())
+}
```
vinistock: This should go in `utils.rs`. Also, two things:

1. We do this in multiple places in the FFI, so it would be nice to follow up with a PR to use the new helper everywhere
2. We should be able to safely do `unwrap` instead of `map_or`. The only time `CString::new` fails is if there's a null byte in the middle of the string (e.g.: `MyString\0something_else`), which shouldn't really happen

ext/rubydex/query.c
paracycle: `rdxi_graph_from_handle` is for handle objects (Declaration/Definition/Document) that store a back-reference to their parent Graph via `HandleData.graph_obj`. In `Query#run`, `graph_obj` is the Graph itself, not a handle, so it can't be used directly.                                                                                                                   

I added `rdxi_graph_from_object` in `graph.h` as the Graph-level equivalent, and used it in both `render` and `run`.

rust/rubydex-sys/src/graph_api.rs
paracycle: Done. Moved everything Cypher-related to `rust/rubydex-sys/src/cypher_api.rs` which includes `CQueryResult`, `CParseResult`, `rdx_cypher_parse`, `rdx_cypher_query_free`, `rdx_cypher_schema`, all structured-result types, and `rdx_query_run_rows`. Only `rdx_query_run` (the formatted-string runner) stays in `graph_api.rs`, importing `CQueryResult` from the new module.

rust/rubydex-sys/src/graph_api.rs
paracycle: Good call. I switched to a `CRowsIter` iterator following the existing `DeclarationsIter` / `DocumentsIter` pattern. The C side now walks rows via `rdx_rows_iter_next` / `rdx_rows_iter_free` instead of indexing into a `CResultSet`.

Cell conversion still happens eagerly inside `with_graph` since `build_node_cell` needs `&Graph` for kind lookups, and `with_graph` only borrows for the closure duration. This is the same tradeoff the existing iterators make (e.g. `rdx_graph_declarations_search` collects inside `with_graph` then returns an iterator over the boxed slice). 

`CResultSet` and `rdx_result_set_free` are gone; the iterator owns all allocations and frees them in `rdx_rows_iter_free`.

rust/rubydex-sys/src/graph_api.rs
paracycle: Moved `cstring_raw` to `utils.rs` with `unwrap` — null bytes shouldn't occur in rubydex data, matching the existing pattern used throughout the FFI (e.g. `declaration_api.rs:137`, `definition_api.rs:103`, `document_api.rs:58`).

One exception: `CQueryResult::success` keeps its explicit `CString::new` match with a graceful fallback, since the render path (`Query#render`) takes arbitrary formatted output where an interior NUL is plausible. `CQueryResult::error` and all other call sites now use `utils::cstring_raw`.

I have not done the retrofit of the existing call sites across the FFI to use the new method, though.

vinistock (approved):

ext/rubydex/query.c
```diff
@@ -96,10 +98,103 @@ static VALUE rdxr_query_render(int argc, VALUE *argv, VALUE self) {
...
+    while (rdx_rows_iter_next(iter, &row)) {
+        VALUE hash = rb_hash_new();
+        for (size_t c = 0; c < row.len && c < column_count; c++) {
+            VALUE key = rb_utf8_str_new_cstr(columns[c]);
+            rb_hash_aset(hash, key, cypher_cell_to_value(graph_obj, &row.cells[c]));
+        }
+        rb_ary_push(rows, hash);
+    }
+
+    rdx_rows_iter_free(iter);
```
Morriar: Should we wrap this conversion in `rb_ensure` so `rdx_rows_iter_free(iter)` runs even when Ruby raises?

paracycle: Done! Thanks for catching this.

PR #423, opened 2026-01-08
Restructure phase documentation

vinistock:
This PR does 3 things:

- Restructures our architecture doc to split phase documentation into their own files
- Documents the resolution phase design
- Updates our graph image to match the current state

I made a few conceptual name changes:

- Indexing -> Extraction (instead of discovery). Based on the discussion on Stan's PR
- Extraction + Resolution -> Indexing

I believe that makes it a bit clearer. We have indexing as a "stage" combining two phases. Let me know what you think!

PR #874, opened 2026-06-24
Add RDoc documentation site for Rubydex APIs

### LINKED ISSUE #843: Create Ruby API documentation for Rubydex
To make sure teams can create and maintain their own Rubydex linter rules, Rubydex needs to have a API documentation website. We should default to use RDoc but can use YARD or other things if our needs can't be supported.

st0012:
## Problems

- Rubydex does not have generated API documentation that can be published as a static site.
- The native extension exposes Ruby APIs from C, but most public C-backed methods are not documented in a format RDoc can capture.

## Solutions

- Add RDoc configuration, a documentation rake task, and a GitHub Pages workflow using the current RDoc commit.
- Include both Ruby and C extension sources in the generated site.
- Convert public C-backed API comments to RDoc `call-seq:` blocks with RBS-ish return shapes.

Screenshot 2026-06-24 at 21 07 02

(The `workspace_path` attribute's type label is weird but it's a RDoc issue)


ext/rubydex/declaration.c
```diff
@@ -65,7 +72,12 @@ static VALUE rdxr_declaration_name(VALUE self) {
     return str;
 }
 
-// Declaration#unqualified_name -> String
+/*
+ * call-seq:
+ *   unqualified_name -> String?
+ *
+ * Returns the declaration name without namespace qualification.
+ */
```
st0012, line 83: RDoc's C parser only recognizes this style of C comments.

ext/rubydex/declaration.c
```diff
@@ -65,7 +72,12 @@ static VALUE rdxr_declaration_name(VALUE self) {
     return str;
 }
 
-// Declaration#unqualified_name -> String
+/*
+ * call-seq:
+ *   unqualified_name -> String?
```
st0012, line 80: RDoc currently doesn't have a way to read RBS sigs from C files yet. So let's use call-seq for now.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Setup Pages
+        uses: actions/configure-pages@v6.0.0
+
+      - name: Build documentation
+        run: bundle exec rake rdoc
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v5.0.0
+
```
vinistock: I assume `upload-pages` will commit or somehow upload the `docs` directory, but what does `configure-pages` do?

.github/workflows/pages.yml
```diff
@@ -0,0 +1,49 @@
...
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v5.0.0
```
vinistock, line 45: Doesn't GH pages deploy automatically if `docs` is modified? Maybe something changed, but I remembered the process being a bit simpler.

For example, the Ruby LSP GH pages website just builds everything in `docs` and pushes a commit to a specific branch.

ext/rubydex/declaration.c
```diff
@@ -5,6 +5,8 @@
 #include "rustbindings.h"
 #include "utils.h"
 
+/* RDoc parses C files independently: mRubydex = rb_define_module("Rubydex") */
```
vinistock: Is this comment required for the docs? I'm not sure I understand what it does.

Gemfile
```diff
@@ -8,6 +8,7 @@ gemspec
 gem "rake", "~> 13.3"
 gem "rake-compiler"
 gem "minitest"
+gem "rdoc", git: "https://github.com/ruby/rdoc.git", ref: "daa7768b81c703b84c796bf60c27977c62cd4aa1"
```
vinistock: Why this specific commit and not the latest release?

.document
```diff
@@ -0,0 +1,5 @@
+README.md
+CONTRIBUTING.md
+docs/*.md
```
vinistock, line 3: After this, we really ought to give some love to the architecture docs, which are really behind and lacking at the moment. I ended up never finishing #423

Gemfile
st0012: Because RDoc 7.2.0 is much inferior compared to the current master. For example, it doesn't have the server mode, which is very handy for live preview doc edits.

.github/workflows/pages.yml
st0012: No I plan to deploy docs using the workflow approach (vs branch approach).
With this approach, html files are generated at `_site`, which then gets uploaded (what this action does). There will be no separate branch.

I chose this approach because:
1. I've done this setup with RDoc for irb, reline, and rdoc. And it's simple to maintain
2. The branch approach uses `/docs` by default, which is already taken in Rubydex. To use that approach, we first need to rename our `/docs` folder to something else.

.github/workflows/pages.yml
st0012: That's the branch-based deployment approach. I chose a different approach here as explained in https://github.com/Shopify/rubydex/pull/874#discussion_r3477118860

ext/rubydex/declaration.c
st0012: It's a workaround for https://github.com/ruby/rdoc/issues/1744

I'll update the comment to mention the issue

Gemfile
st0012: Added a todo comment so we won't leave it like this

.github/workflows/pages.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Shopify/rubydex' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@9eb537ca036ebaed86729dcb9309076e4c5c3b74 # v1.314.0
+        with:
+          ruby-version: "4.0.0"
```
vinistock: Let's remove this and so that it picks up `.ruby-version`.

ext/rubydex/location.c
```diff
@@ -1,5 +1,10 @@
 #include "location.h"
 
+/*
+ * RDoc parser workaround for https://github.com/ruby/rdoc/issues/1744:
```
vinistock, line 4: It's not the best that we need to spread this in almost all of our C files. Is there already a solution in RDoc? If there is, let's create a ticket to get rid of these so we don't forget.

vinistock (approved):

ext/rubydex/location.c
st0012: I asked Codex to prototype a solution but I haven't looked into it yet.

st0012: https://github.com/Shopify/rubydex/issues/885

Gemfile
st0012: Managed to cut 8.0 in time. Using it now

Issue #864, opened 2026-06-16
Track total memory used by the graph


amomchilov:
There are several popular crates that let you check the transitive size of an entire object graph, e.g.:

1. https://crates.io/crates/allocative
2. https://crates.io/crates/deepsize
3. https://crates.io/crates/get-size2

We should apply one of these to be able to query memory statistics about the whole index.

Then, we can implement the `dsize` function, to expose this memory usage to Ruby, so it can be queried e.g. via [`ObjectSpace.memsize_of`](https://docs.ruby-lang.org/en/master/ObjectSpace.html#method-c-memsize_of).

> [!NOTE]  
> Whichever crate we end up using, we should make sure it can deal with shared objects (like `Rc`/`Arc`) and doesn't double-count them.

PR #875, opened 2026-06-24
Add memory benchmark through jemalloc allocation trace

vinistock:
Not exactly #864, but a decent step towards it if we decide to move forward

I've been suspecting that our memory measurements using RSS are not super precise. Based on my understanding, max RSS represents the peak memory used by the process. Since we spawn multiple threads during indexing, I assumed that the peak would probably occur during that phase, but that's not what we're interested in. Instead, we want to know the amount of memory that the final graph consumes.

This PR adds a memory benchmark. The way it works is:

1. Use jemalloc as the global allocator in a benchmark target. Jemalloc allows tracing allocations across epochs, which gives us the opportunity to analyze the diff of bytes between epochs
2. Add the benchmark target. Using `[[bench]]` means that the code is always built in release mode, but can use development-only dependencies. The executable builds the graph and compares the memory being used before and after dropping it

I also added a convenience script that runs the benchmark for a given path and optionally accepts a `--compare` argument. When using `--compare`, it just runs the benchmark against `main` first to allow comparing the results.

In all workspaces I tried, the max RSS approach overestimates the amount of memory being used:

- Core: 46%
- Ruby LSP: 195%
- Rails: 52%

### DIFF
--- a/utils/bench-graph-memory
+++ b/utils/bench-graph-memory
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# Measure the total memory used by the graph. Allows comparing a branch against main
+#
+# Usage:
+#   utils/bench-graph-memory <path>             # current branch only
+#   utils/bench-graph-memory --compare <path>   # main, then current branch
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+RUST_DIR="$PROJECT_ROOT/rust"
+COMPARE=0
+TARGET=""
+BRANCH="$(git branch --show-current)"
+
+for arg in "$@"; do
+    case "$arg" in
+        --compare) COMPARE=1 ;;
+        *) TARGET="$arg" ;;
+    esac
+done
+
+if [ -z "$TARGET" ]; then
+    echo "Usage: $0 [--compare] <path>" >&2
+    exit 1
+fi
+
+# Expand the target based on Rubydex's root directory
+case "$TARGET" in
+    /*) ;;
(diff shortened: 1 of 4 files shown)


utils/bench-graph-memory
```diff
@@ -0,0 +1,45 @@
...
+    esac
+done
+
+if [ -z "$TARGET" ]; then
+    echo "Usage: $0 [--compare] <path>" >&2
+    exit 1
+fi
+
+run_bench() {
+    cd "$RUST_DIR" && cargo bench --bench graph_memory -- "$TARGET"
```
st0012, line 37: If we're `cd`ing into `rust/`, what should be the base for the target path?

utils/bench-graph-memory
```diff
@@ -0,0 +1,45 @@
...
+}
+
+if [ "$COMPARE" -eq 1 ]; then
+    git stash push --include-untracked --quiet
+
+    git checkout --quiet main
+    echo "########### Measuring main ############"
+    run_bench
+
+    git checkout --quiet "$BRANCH"
```
st0012, line 58: Should we add a step to make sure the current branch is based on the current `main`?
It's not uncommon for me to have a branch basing on a stale main.

utils/bench-graph-memory
```diff
@@ -0,0 +1,45 @@
...
+    git stash pop --quiet
```
st0012: Can we only `pop` if we did stash things in the beginning of the block? Otherwise, we could be popping previous, unrelated stash.

utils/bench-graph-memory
vinistock: I started expanding `TARGET` against Rubydex's root.

utils/bench-graph-memory
vinistock: Added a check to fail and prompt the person to rebase first.

st0012 (approved):

PR #877, opened 2026-06-25
Make definition nesting usable for constant resolution

soutaro:
## Summary

Fix `Graph#resolve_constant` for lexical nesting stacks containing rooted or fully qualified namespace names.

## Background

When callers construct a nesting stack from `Definition#lexical_nesting`, explicit qualification affects constant lookup. For example:

```ruby
module Foo
  module ::Bar
    CONST
  end
end
```

The inner frame must remain `::Bar`; treating it as `Bar` would incorrectly resolve it relative to `Foo`.

`Graph#resolve_constant` creates a temporary `Name` chain from the supplied strings. Rooted and fully qualified frames can introduce parent-scope and lexical-nesting dependencies that have not yet been resolved. Previously, only the final name was resolved, so lookup could fail even when the corresponding declarations existed in the graph.

## Changes

- Add recursive constant resolution that resolves a name's lexical nesting and parent scope before resolving the name itself.
- Use recursive resolution from the Ruby `Graph#resolve_constant` API.
- Add `source_name` to class, module, constant, and constant-alias definitions. Unlike `name`, it preserves explicit source-level qualification such as `Bar`, `Foo::Bar`, or `::Bar`, allowing callers to construct constant-resolution nesting stacks without inferring qualification from resolved declarations.

### DIFF
--- a/ext/rubydex/definition.c
+++ b/ext/rubydex/definition.c
@@ -155,6 +155,26 @@ static VALUE rdxr_definition_name(VALUE self) {
 
 /*
  * call-seq:
+ *   source_name -> String
+ *
+ * Returns the class, module, or constant source name, preserving explicit parent scopes like "Foo::Bar" and rooted
+ * paths like "::Bar".
+ */
+static VALUE rdxr_definition_source_name(VALUE self) {
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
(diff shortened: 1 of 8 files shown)


soutaro:
Closing this draft. The reported issue can be addressed with a smaller change that exposes source-level definition names; the broader recursive resolution change is not required.

PR #883, opened 2026-06-25
Run Cypher queries from the rdx console

paracycle:
## Goal

Let `rdx console` run Cypher directly, addressing the "query mode" idea from #868's review.

## What it does

The session is a stock IRB workspace with `graph` in scope, so plain Ruby works as always (`graph["Foo"]`). On top of that it registers three things through IRB's own extension APIs:

```
# `run(<CYPHER>)` — a helper method, so the result is a value you can assign and navigate.
rubydex(main):001> rows = run("MATCH (c:Class)-[:HAS_PARENT]->(p) RETURN c, p")
rubydex(main):002> rows.first["c"].name

# `query <CYPHER>` — a command taking the rest of the line verbatim; prints a formatted table.
rubydex(main):003> query MATCH (n:Class|Module) RETURN n.name ORDER BY n.name

# `schema` — prints the queryable schema.
rubydex(main):004> schema
```

`query` takes the rest of the line verbatim, so the Cypher needs neither quoting nor to be valid Ruby. Rather than swapping the REPL's evaluator for a modal toggle, Cypher lines are simply prefixed with `query` — the clean, idiomatic IRB realization of a query mode. `run` is the programmatic entry point and `query` the interactive quick-look; they are deliberately separate rather than one delegating to the other.

## How it hooks into IRB

IRB has two distinct registration APIs, and this PR uses each for what it's for:

- `IRB::Command.register` — for `query`/`schema`, which consume the rest of the line verbatim.
- `IRB::HelperMethod.register` — for `run`, which must stay an ordinary expression so its value can be assigned.

The session workspace is IRB's own `IRB::WorkSpace.new`, not a hand-rolled binding. That matters for three reasons:

- `self` is `main` and the cref is `Object`, so `class Foo; end` at the prompt defines `::Foo`. A `binding` captured inside `class << self` would define `#<Class:Rubydex::Console>::Foo` instead.
- No caller locals leak into the prompt. Deriving the binding from the CLI's `TOPLEVEL_BINDING` would inherit every top-level local of `exe/rdx` — including `query` and `schema`, and a local shadows the IRB command of the same name.
- `_`, `help` and the rest of IRB's conveniences behave exactly as in stock `irb`.

Only `graph` is injected, via `IRB::WorkSpace#local_variable_set`.

## Notes

- Console logic lives in `lib/rubydex/console.rb`. `Console.run` (objects), `Console.render` (formatted string) and `Console.describe_schema` are plain module methods, so they are testable without driving IRB.
- Both IRB APIs used here landed in IRB 1.13, so the file activates `gem "irb", ">= 1.13"` at load. A missing or older IRB raises `LoadError` (`Gem::MissingSpecVersionError` is one), which `exe/rdx` turns into `Interactive mode requires \`irb\` >= 1.13 to be in the bundle` rather than a backtrace. IRB is not declared in the gemspec, which carries no runtime dependencies.
- `run` returns the rich objects from `Query#run` (#873): node columns come back as live `Declaration`/`Definition`/`Document` handles.

## Verification

- `test/console_test.rb` — 8 runs, 29 assertions, 0 failures. Includes a regression test that the session workspace evaluates at top level (`self == main`, only `graph` injected, `class Probe; end` defines `::Probe`); it fails if a hand-rolled binding is reintroduced.
- `rubocop` clean.
- Live smoke test of `rdx console` against this repo: `self` → `main`, `local_variables` → `[:_, :__, :graph]`, `Foo.name` → `"Foo"`, `run("MATCH (c:Class {name: 'Rubydex::Graph'}) RETURN c.name")` → `[{"c.name" => "Rubydex::Graph"}]`, and `query`/`schema` both print their tables.
- Degradation checked both ways: with no `irb` installed, and with only `irb` 1.12 installed, `rdx console` aborts cleanly with exit status 1.

### DIFF
--- a/lib/rubydex/console.rb
+++ b/lib/rubydex/console.rb
@@ -1,0 +1,141 @@
+# frozen_string_literal: true
+
+# The console is built on IRB's command and helper-method registration APIs, both introduced in IRB
+# 1.13. Requiring an older or absent IRB raises `LoadError` (`Gem::LoadError` is one), which
+# `exe/rdx` turns into a friendly message rather than a backtrace.
+gem "irb", ">= 1.13"
+require "irb"
+require "irb/command"
+require "irb/helper_method"
+
+module Rubydex
+  # Interactive console backing `rdx console`.
+  #
+  # On top of a normal IRB session (with `graph` in scope so you can call Ruby directly, e.g.
+  # `graph["Foo"]`), it offers two complementary ways to run Cypher against the same graph:
+  #
+  #   # `run("...")` — a method that returns rich Ruby objects for programmatic use: an Array of
+  #   # Hashes keyed by RETURN column, where node columns are live Declaration/Definition/Document
+  #   # handles. Because it's a plain method, the result can be assigned and navigated.
+  #   rubydex(main):001> result = run("MATCH (c:Class)-[:HAS_PARENT]->(p) RETURN c, p")
+  #   rubydex(main):002> result.first["c"].name
+  #
+  #   # `query <CYPHER>` — a command that takes the rest of the line verbatim (no quotes / valid
+  #   # Ruby needed) and prints a formatted table for a quick look.
+  #   rubydex(main):003> query MATCH (n:Class|Module) RETURN n.name ORDER BY n.name
+  #
+  #   # `schema` — prints the queryable schema.
+  #   rubydex(main):004> schema
+  #
+  # So `run` is the programmatic entry point and `query` is the interactive quick-look; they are
+  # deliberately separate rather than one delegating to the other. This is the clean, idiomatic
+  # "query mode": rather than swapping the REPL's evaluator, Cypher lines are simply prefixed with
+  # `query`.
+  #
+  # The `run` helper and the `query`/`schema` commands are registered through IRB's own extension
+  # APIs, so the session itself is a stock IRB workspace (see {.start}).
+  module Console
+    class << self
+      # The graph that `run` and the `query`/`schema` commands operate on. Set by {.start}.
+      attr_accessor :graph
+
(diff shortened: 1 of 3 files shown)


lib/rubydex/console.rb
```diff
@@ -0,0 +1,125 @@
...
+      # Returns the queryable Cypher schema description as a formatted String.
+      def describe_schema(format: :table)
+        Rubydex::Query.schema(format)
+      end
+
+      private
+
+      # A binding that exposes `graph` as a local variable for the Ruby side of the console.
+      def session_binding(graph)
+        binding
```
Morriar: This means in the console if I do `class Foo; end` I'm actually creating `Rubydex::Console::Foo` instead of `::Foo`?

paracycle: Yes, it was causing weird scoping issue indeed. I've reworked the registration of `run` to use the new(er) [IRB Helper methods](https://ruby.github.io/irb/EXTEND_IRB_md.html#helper-methods) instead.

Morriar (commented):
Found one CI issue to fix before this lands.

lib/rubydex/console.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+# The console is built on IRB's command and helper-method registration APIs, both introduced in IRB
+# 1.13. Requiring an older or absent IRB raises `LoadError` (`Gem::LoadError` is one), which
+# `exe/rdx` turns into a friendly message rather than a backtrace.
+gem "irb", ">= 1.13"
+require "irb"
```
Morriar, line 7: You may need to `require "fiddle"` to fix the Windows build?

Morriar (approved):

PR #884, opened 2026-06-25
Return Diagnostic#rule as Symbol

st0012:
## Problems

- `Rubydex::Diagnostic#rule` is typed as a `Symbol`, but diagnostics were built with string rule values.
- That mismatch makes runtime behavior disagree with the Ruby-facing API contract.

## Solutions

- Intern native diagnostic rule strings before constructing `Rubydex::Diagnostic` objects.
- Update the parse-error diagnostic test to assert symbol rule values.


vinistock (commented):
What do you think of changing the code to return a symbol instead? It seems nicer for these unique labels

vinistock (approved):

PR #886, opened 2026-06-25
Extract Rubydex native adapter helpers

st0012:
## Problems

- Several native extension adapters repeated the same handle-to-graph unpacking sequence.
- Rust-owned C string conversion repeated the same `NULL` handling, Ruby string creation, and `free_c_string` ownership pattern.

## Solutions

- Add `rdxi_graph_from_handle` to centralize handle and graph typed-data unpacking while preserving the existing Ruby graph object ownership.
- Add `rdxi_owned_c_string_to_ruby` for Rust-owned strings and use it only at existing `free_c_string` callsites.

### DIFF
--- a/ext/rubydex/handle.h
+++ b/ext/rubydex/handle.h
@@ -1,6 +1,7 @@
 #ifndef RUBYDEX_HANDLE_H
 #define RUBYDEX_HANDLE_H
 
+#include "graph.h"
 #include "ruby.h"
 
 typedef struct {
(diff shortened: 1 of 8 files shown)


ext/rubydex/handle.h
```diff
@@ -34,6 +35,18 @@ static const rb_data_type_t handle_type = {
...
+    HandleData *data;
+    TypedData_Get_Struct(self, HandleData, &handle_type, data);
+
+    void *graph;
+    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);
+
+    *out_data = data;
+
+    return graph;
+}
```
vinistock, line 48: Instead of accepting a double pointer, can't we do this?


```suggestion
static inline void *rdxi_graph_from_handle(VALUE self, HandleData *data) {
    TypedData_Get_Struct(self, HandleData, &handle_type, data);

    void *graph;
    TypedData_Get_Struct(data->graph_obj, void *, &graph_type, graph);

    return graph;
}
```

vinistock (approved):

ext/rubydex/handle.h
st0012: That doesn't work if the helper takes `HandleData *data` and callers pass `data`.
Because the function receives a copy of the pointer value, `TypedData_Get_Struct`
will update that local copy, and when the function returns, the caller's `data`
variable is unchanged.

When the helper takes `HandleData **data` and callers pass `&data`, the pointer variable's address gets copied, and `TypedData_Get_Struct` can then update the pointer value stored in the caller's `data` variable.

PR #888, opened 2026-06-26
Introduce configuration file support

vinistock:
While Rubydex is a framework that can be used to build all sorts of tools, in most cases we want consistency in the base analysis.

For example, imagine you have an LSP and a linter based on Rubydex, but because they each configure the framework differently, you get different results in the editor vs in the CLI. That would be a pretty poor experience.

I believe we can provide an easy way to ensure consistency while still maintaining flexibility so that the odd case can load custom configuration files for a random tool.

## Implementation

I recommend reviewing per commit:

1. Allow loading configuration from a file. This adds the infrastructure to read, parse and apply a configuration file, hooking it up to the graph. See the example configuration file below for the format
2. Exposes `load_config` in the Ruby API
3. Adds `rubydex.toml` default configuration for this repo itself

```toml
# rubydex.toml
# Comments
single_value = true

multi_value = ["one", "two"]

[something.nested]
enabled = false
```

### DIFF
--- a/exe/rdx
+++ b/exe/rdx
@@ -35,6 +35,8 @@ def __with_timer(message, &block)
 end
 
 graph = Rubydex::Graph.new
+graph.load_config
+
 __with_timer("Indexing workspace...") { graph.index_workspace }
 __with_timer("Resolving graph...") { graph.resolve }
 
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -865,6 +865,40 @@ static VALUE rdxr_graph_set_workspace_path(VALUE self, VALUE path) {
 
 /*
  * call-seq:
+ *   load_config(config_path = nil) -> void
+ *
+ * Loads a configuration file for the graph. If `config_path` is nil, loads the default configuration file at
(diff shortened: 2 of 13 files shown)


st0012:
The implementation looks good but I'm not sure about the format we use for `.rubydex`.
I personally have NEVER used the INI format in any tools before, nor have seen it being used in the Ruby ecosystem. So I think using it would add additional friction to properly configure Rubydex for users.
I think `yml` is still the best choice for user adoption. `toml` will be fine for me too.

vinistock:
Switched to TOML according to our discussions.

st0012 (commented):
In the PR description, we still mention `.rubydex` and INI format. Let's update it
Also, should we use `.rubydex.toml`? Is it intentional to drop `.` prefix with switching to toml?

vinistock:
I removed the leading dot intentionally since we now have a file extension. Just `rubydex` (instead of `.rubydex`) felt odd, but `rubydex.toml` seems fine to me.

st0012 (approved):

Issue #826, opened 2026-05-26
Add interface for collecting performance metrics


amomchilov:
E.g. to emit to StatsD / OpenTelemetry how long each pipeline stage took.

vinistock:
Worth checking out https://crates.io/crates/tracing

st0012:
I had a [quick investigation with River](https://shopify.slack.com/archives/C0AHGMHM0BB/p1782402911734039) about OTel integration with Observe and here's a quick summary why it's not worth the effort for now:

- **Integration.** Ruby LSP and rubydex (in-process Rust extension) each need
  their own OTel SDK, exporter, and flush-on-exit — they can't share one. That's
  two instrumentation stacks plus per-call `traceparent` propagation across the
  FFI boundary to keep both halves in one trace.
- **Emitting to Observe from local machines.** Production exports to an in-cluster
  collector; a laptop can't reach it. Traces must flow through the `observe-dev`
  local collector (which most devs won't have running), the `ruby-lsp` service
  name needs allowlisting in the ingestion pipeline, and the Observability team
  has to confirm laptop-originated LSP traces are supported. Until that's settled,
  the data may not even land.
- **Extra dependencies.** Adds the opentelemetry-shopify gem on the Ruby side and
  the upstream `opentelemetry` / `opentelemetry_sdk` / `opentelemetry-otlp` /
  `tracing-opentelemetry` crates on the Rust side, all to maintain and keep
  version-aligned.

It'd be amazing to be able to see a waterfall breakdown for Rubydex indexing traces like below, but it's just too much work.

<img width="80%" alt="Image" src="https://github.com/user-attachments/assets/ed7e1320-20f2-4e7c-98d2-d8b4eb927a0c" />

PR #892, opened 2026-06-30
Compute descendants as the inverse of linearized ancestors

soutaro:
## Summary

- Rebuild `descendants` once at the end of `Resolver::resolve()` by inverting finalized `Complete` ancestor chains.
- Remove the per-linearization descendant bookkeeping this replaces, avoiding stale descendant edges from abandoned tentative parents.
- Batch descendant insertion through `Namespace::extend_descendants` so hot ancestors do not repeatedly grow and rehash their descendant sets.
- Add `--stats` detail breakdowns for `resolve()` and `compute_descendants()`.

## Performance notes

This PR is intentionally focused on the initial/full resolution path. It does not try to optimize incremental resolution.

On local Core Shopify runs the branch reduced resolution time versus main in repeated runs, and `compute_descendants` itself was around `0.16s` after batching.

The timing breakdowns are compiled into release builds, but detailed timing is only recorded when `--stats` initializes the global timer. Normal non-`--stats` runs pay only the small enabled-check cost around the measured resolution subphases.

## Validation

```sh
cargo test -p rubydex
```

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -76,6 +76,10 @@ pub struct Resolver<'a> {
     unit_queue: VecDeque<Unit>,
     /// Whether we made any progress in the last pass of the resolution loop
     made_progress: bool,
+    /// Declarations whose ancestor chain this resolver (re-)wrote via `set_ancestors`. A fresh resolver is created for
+    /// every `resolve()`, so this only ever holds the declarations linearized by the current run. Used to update the
+    /// `descendants` relation from just these declarations instead of rebuilding it from scratch.
+    linearized_declarations: IdentityHashSet<DeclarationId>,
 }
 
 impl<'a> Resolver<'a> {
@@ -84,6 +88,7 @@ impl<'a> Resolver<'a> {
             graph,
             unit_queue: VecDeque::new(),
             made_progress: false,
+            linearized_declarations: IdentityHashSet::default(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -134,6 +128,69 @@ impl<'a> Resolver<'a> {
...
+    /// is a clear win on full builds (it replaces the costlier per-linearization bookkeeping) but
+    /// adds a fixed cost to small incremental updates. Narrowing the rebuild to only the
+    /// declarations re-linearized in the current pass is a future optimization.
+    fn compute_descendants(&mut self) {
+        // Clear any previously recorded descendants.
+        for declaration in self.graph.declarations_mut().values_mut() {
+            if let Some(namespace) = declaration.as_namespace_mut() {
+                namespace.clear_descendants();
+            }
+        }
```
vinistock: This is heavy handed. We should only clear descendants when the invalidation logic identifies that it's necessary to do so, otherwise incremental resolution is going to be super slow.

soutaro: Sure. I'm trying to support incremental resolution here.

One note: the current implementation is not *super* slow I think. It takes `0.5s` for full resolution of our monolith, while I agree it's noticeable delay for user interaction -- editing files.

I will try incremental resolution performance profiling first.

rust/rubydex/src/stats/timer.rs
```diff
@@ -111,17 +154,53 @@ macro_rules! make_timer {
...
+    details "resolve" {
+        resolve_prepare, "prepare_units";
+        resolve_converge, "convergence loop";
+        resolve_remaining, "remaining definitions";
+        resolve_descendants, "compute_descendants";
+    }
+    details "compute_descendants" {
+        descendants_clear, "Clear";
+        descendants_edges, "Build edges";
+        descendants_insert, "Insert";
```
st0012: Let's separate these timer changes from the PR so it's easier to review

rust/rubydex/src/resolution.rs
```diff
@@ -104,47 +99,125 @@ impl<'a> Resolver<'a> {
...
+        // Descendants are derived from the finalized ancestor chains, so they must be computed
+        // after all linearization has settled.
+        time_it!(resolve_descendants, {
+            self.compute_descendants();
+        });
+    }
+
+    /// Recomputes the `descendants` relation as the inverse of the linearized ancestors.
+    ///
+    /// After resolution, `descendants(D)` must contain exactly the declarations `x` such that
```
st0012: Can we expand some "variables" here or just use direct examples?
I don't fully understand what `D` means here, for example.

rust/rubydex/src/resolution.rs
```diff
@@ -104,47 +99,125 @@ impl<'a> Resolver<'a> {
...
+    /// `Ancestor::Complete(D)` appears in `x`'s linearized ancestors. Because a declaration is its
+    /// own ancestor (see [`Self::linearize_ancestors`]), this relation is reflexive.
+    ///
+    /// Maintaining the relation while linearizing ancestors is error-prone: a declaration may be
+    /// linearized under a tentative parent in one pass and a different parent in a later pass, and
+    /// the stale descendant edges from the abandoned parent are never cleared, which inflates the
```
st0012: I don't think this type of "hypothetical flaws" comments is useful without concrete examples or past issues we can point to.

rust/rubydex/src/resolution.rs
```diff
@@ -104,47 +99,125 @@ impl<'a> Resolver<'a> {
...
+    /// relation. Computing it once here, by inverting the final ancestor chains, keeps it exact.
+    ///
+    /// This rebuilds the entire relation in one `O(total ancestors)` pass. That matches the
+    /// full-resolution path this code is optimized for and replaces the costlier per-linearization
+    /// descendant bookkeeping.
+    fn compute_descendants(&mut self) {
+        // Clear any previously recorded descendants.
+        time_it!(descendants_clear, {
```
st0012: I don't think clearing descendants will ever be a hotpath that we need to measure performance for 🤔

st0012 (commented):
> the branch reduced resolution time versus main in repeated runs

Can you include some before/after numbers?

rust/rubydex/src/resolution.rs
```diff
@@ -104,47 +99,125 @@ impl<'a> Resolver<'a> {
...
-
-            // Loop through the current length of the queue, which won't change during this pass. Retries pushed to the back
-            // are only processed in the next pass, so that we can assess whether we made any progress
-            for _ in 0..self.unit_queue.len() {
-                let Some(unit_id) = self.unit_queue.pop_front() else {
-                    break;
-                };
+        let other_ids = time_it!(resolve_prepare, { self.prepare_units() });
+
+        time_it!(resolve_converge, {
```
vinistock: I think this PR is trying to address #826 at the same time as changing how descendants are computed. We should split the work.

Also, timers weren't designed to be used outside of `main.rs`. For granular performance timers like these, we need some solution that allows us to drain the events from the Ruby side, so that we can easily send the data to a dashboard and track performance over time.

vinistock (commented):
I think the fundamental problem with this attempt to split things is that we're trying to separate the moment when we record the data from its discovery.

For incremental resolution, tracking descendants simply works out of the box. If a declaration needs to have its ancestors linearized again due to an edit, we invalidate the descendants and it naturally gets computed as part of linearizing.

This separation is trying to move the recording of descendants to the end, but then you lose the trigger point that is attached to the ancestor linearization. This is the reason why we're having to clear all descendants every single time resolution runs, even if they weren't invalidated by an incremental edit.

Whatever we do, it needs to fit the overall structure of the algorithm and naturally do the minimum amount of work necessary on each resolution run.

soutaro:
@vinistock Thanks, I see what you mean.

In practice, though, I think this is still okay to merge: incremental resolution also gets faster with this patch (slightly) and the follow-up in #915 (4 times faster with the simplest `--incremental_cycle` benchmark).

That said, I agree that it is not completely obvious yet how this approach should fit with a more incremental descendants update strategy. I will take another look and check whether we can combine this with incremental resolution cleanly.

(I think keeping track of namespaces that was linearized during resolution would be the good key to minimize descendants clear and calculation.)

I opened one more PR #916, but this is different. The `--incremental_cycle` benchmark got slower with the PR. I will take a look at it.

vinistock (changes requested):
I don't believe we can move forward with the changes in this current state and some of the reasons are what I tried to express in the documentation PR #913.

The fundamental problem with this change is that it doesn't fit the overall algorithm. If we move descendant tracking after the convergence loop and decoupled from ancestor linearization, then we disrespect the flow of incremental resolution updates. We currently don't minimize the ancestor work incremental resolution does, but that's the direction we want to go, so the optimization would be overfitting to the current state of the code instead of the desired state.

To try to get around this, we are now introducing a second source of truth: another piece of state in resolution to track which ancestors get linearized. However, which ancestors have to be re-computed are already stored in the algorithms primary source of truth: the unit queue. The only reason we need this secondary structure is because we are trying to decouple two things that are inherently connected: ancestor linearization and descendant tracking.

I understand that this makes things a bit faster for the current state of the codebase and benchmarks, but as we make incremental resolution more intelligent we would have to undo the changes.

PR #894, opened 2026-06-30
Handle `extend` inside `class << self` for rank-N singletons

soutaro:
- Linearize: drop the `!is_singleton_class` guard so an `extend M` written inside a `class << self` always creates that singleton's own singleton (the rank-2 `Foo::<Foo>::<<Foo>>`), with `M` mixed in as an ancestor. It was previously skipped for singleton classes.
- Seed rank-N materialization from any singleton whose own `class << self` carries an `extend`: the extend mixes the module into that singleton's own singleton as an ancestor (mixins are never members), so the definition/member checks that detect explicit singletons miss it. Seeding here materializes the rank-`(N + 1)` singleton for every subclass too.
- Add tests: rank-2 singleton for the extended class itself, and for a subclass.

### DIFF
--- a/rust/rubydex-mcp/src/server.rs
+++ b/rust/rubydex-mcp/src/server.rs
@@ -1119,12 +1119,15 @@ mod tests {
         let res = parse(&s.codebase_stats());
 
         assert_eq!(res["files"], 3);
-        assert_json_int!(res, "declarations", 7);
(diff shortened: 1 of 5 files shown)


soutaro:
Closing because the base PR is closed. I still want to have this, but it requires a lot of work to fix the implementation.

PR #895, opened 2026-06-30
Use jemalloc as the global allocator

vinistock:
Using jemalloc provides a 20% max RSS reduction and 15% speed up in total indexing + resolution time. It seems like an easy performance win and I think we should adopt it as the allocator on the Rust side.

Some relevant bits of the implementation:

- Rubydex is meant to be used as a crate library as well and setting the global allocator can only happen during linking. In other words, there's no way to expose a function that someone can invoke to change the allocator, it has to happen at the top of the file statically. It wouldn't be conventional to have a library crate changing the default allocator since the consumer may want to decide what's the best allocator for their use case, so I gated this in a cargo feature
- I did not go all the way to implement #864. We already have the new graph memory benchmark and tracking allocation stats is not free, so I decided against making the production build include it
- Jemalloc is used in two different ways: for the Rust crate, it is statically linked. For the Ruby gem, since it uses dlopen to load the shared rubydex_sys object, we cannot do so and must dynamically load it as well. I gated this in a feature, so that direct consumers of the crate don't have to pay the performance cost of dynamic jemalloc load

### DIFF
--- a/ext/rubydex/extconf.rb
+++ b/ext/rubydex/extconf.rb
@@ -78,7 +78,7 @@ cargo_command = if ENV["SANITIZER"]
   ENV["RUSTFLAGS"] = "-Zsanitizer=#{ENV["SANITIZER"]}"
   "cargo +nightly build -Zbuild-std #{cargo_args.join(" ")}".strip
 else
-  "cargo build #{cargo_args.join(" ")}".strip
+  "cargo build --features rubydex/jemalloc_dylib #{cargo_args.join(" ")}".strip
 end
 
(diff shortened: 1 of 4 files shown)


st0012 (approved):
With this change, will we be able to directly use jemalloc for more accurate memory measurement in `--stats` too?

vinistock:
Only if we include the `stats` feature in the jemalloc crate, but my understanding is that it's not completely free as it takes work to track all allocations and we'd not be able to turn it on only for `main.rs` and not for the rest.

My suggestion would be to rely on `utils/bench-graph-memory` and avoid enabling `stats` on the main crate.

PR #896, opened 2026-07-02
Use Box<str> instead of String for immutable interned strings

vinistock:
In Rust `String` is simply a `Vec<u8>` (a vector or UTF-8 bytes), which consumes 24 bytes (pointer, capacity and length) + more memory for however long the string ends up being. The `String` object allows for mutation, you can push more content into the same string, which is why it needs to track capacity.

In our use cases, we never have a need to mutate an existing string pushed to the graph. We only ever create or discard, there's never a case where we would mutate a fully qualified name after having already computed it.

For these cases, we can use a `Box<str>`, which is basically an immutable string (we put the byte slice into a boxed pointer). Since the string cannot be mutated, we save the 8 bytes for capacity, reducing the amount of memory we use.

It's a modest reduction since the size of the strings still matter more than the container where they are stored, but it still amounts for about 100MB in Core and it's so simple to apply I figured we go forward with it.

### DIFF
--- a/rust/rubydex/src/model/string_ref.rs
+++ b/rust/rubydex/src/model/string_ref.rs
@@ -9,15 +9,23 @@ use std::ops::Deref;
 /// graph when its count reaches zero.
 #[derive(Debug)]
 pub struct StringRef {
-    value: String,
+    value: Box<str>,
     ref_count: u32,
 }
-assert_mem_size!(StringRef, 32);
+assert_mem_size!(StringRef, 24);
 
 impl StringRef {
     #[must_use]
     pub fn new(value: String) -> Self {
-        Self { value, ref_count: 1 }
+        Self {
+            value: value.into_boxed_str(),
+            ref_count: 1,
+        }
+    }
+
+    #[must_use]
+    pub fn as_str(&self) -> &str {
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/string_ref.rs
```diff
@@ -44,7 +52,7 @@ impl StringRef {
 }
 
 impl Deref for StringRef {
-    type Target = String;
+    type Target = str;
```
st0012, line 55: Not related: what's this and the `&Self::Target` below for?

st0012 (approved):

rust/rubydex/src/model/string_ref.rs
vinistock: Implementing the `Deref` trait allows us to do `*string_ref` and get the underlying string back. Also, the Rust compiler often figures out on its own when to deref something, which makes implementing `Deref` quite convenient.

Regarding the type, the `Deref` traits declare the underlying type that can be de-referenced, so when implementing you have to declare it.

Essentially, we're saying, this `Deref` is for an `str` value and when it gets de-referenced you'll return a reference to the str (&str).

PR #901, opened 2026-07-02
Support glob patterns in the `exclude` config

alexcrocha:
This PR lets `exclude` entries be glob patterns (e.g. `**/fixtures`, `**/*_pb.rb`), not just literal paths. Every entry is compiled to a glob and matched during file discovery, against both directories and nested files.

A glob is not a real path and cannot be canonicalized, so matching compares paths as-is. Discovery does not canonicalize or resolve symlinks.

Two notes on symlinks:
 - A symlinked directory passed as a root is still traversed (the follow comes from `read_dir`), but its files are reported under the symlink path, not the resolved destination.
 - Nested symlinks are not followed: a symlinked directory is not descended, and a symlinked file is indexed at its own path.
 
 ## Breaking change (v0.3.0)

Exclude entries are now glob patterns rather than literal paths, so the API is renamed to match:
- `Graph#exclude_paths` becomes `Graph#exclude_patterns`
- `Graph#excluded_paths` becomes `Graph#excluded_patterns`

(along with the underlying config, `graph.rs`, and FFI names).

## Performance

No listing regression. Averaged over `utils/bench` runs:

| Ref | Listing average | Min | Max |
| --- | ---: | ---: | ---: |
| this branch | 0.676667s | 0.656s | 0.718s |
| `main` | 0.677333s | 0.674s | 0.683s |

Branch delta: -0.10% versus `main`.

### DIFF
--- a/rust/rubydex/src/config.rs
+++ b/rust/rubydex/src/config.rs
@@ -23,7 +23,7 @@ pub const DEFAULT_EXCLUDED_DIRECTORIES: &[&str] = &[
 struct ConfigFile {
     /// Paths to exclude from file discovery during indexing.
     #[serde(default)]
-    exclude: Vec<PathBuf>,
+    exclude: Vec<Box<str>>,
 }
 
 /// Project configuration
@@ -32,7 +32,7 @@ pub struct Config {
     /// Path to the workspace being analyzed
     workspace_path: Box<Path>,
     /// Paths to exclude from file discovery during indexing.
-    excluded_paths: HashSet<PathBuf>,
+    excluded_paths: HashSet<Box<str>>,
 }
 assert_mem_size!(Config, 64);
 
@@ -50,7 +50,7 @@ impl Config {
             workspace_path: std::env::current_dir()
                 .unwrap_or_else(|_| PathBuf::from("."))
                 .into_boxed_path(),
-            excluded_paths: DEFAULT_EXCLUDED_DIRECTORIES.iter().map(PathBuf::from).collect(),
+            excluded_paths: DEFAULT_EXCLUDED_DIRECTORIES.iter().map(|&dir| Box::from(dir)).collect(),
         }
     }
 
@@ -67,16 +67,22 @@ impl Config {
 
     /// Adds paths to exclude from file discovery during indexing. Excluded directories will be skipped entirely during
     /// directory traversal.
-    pub fn exclude_paths(&mut self, paths: impl IntoIterator<Item = PathBuf>) {
+    pub fn exclude_paths(&mut self, paths: impl IntoIterator<Item = Box<str>>) {
         self.excluded_paths.extend(paths);
     }
 
     /// Returns the set of paths excluded from file discovery
     #[must_use]
-    pub fn excluded_paths(&self) -> HashSet<PathBuf> {
+    pub fn excluded_paths(&self) -> HashSet<Box<str>> {
         self.excluded_paths
             .iter()
-            .map(|path| self.workspace_path.join(path))
+            .map(|entry| {
+                self.workspace_path
+                    .join(&**entry)
+                    .to_string_lossy()
+                    .replace(std::path::MAIN_SEPARATOR, "/")
+                    .into_boxed_str()
+            })
             .collect()
     }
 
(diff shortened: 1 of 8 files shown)


rust/rubydex/src/listing.rs
```diff
@@ -379,4 +397,21 @@ mod tests {
...
+    #[test]
+    fn collect_files_excludes_glob_patterns() {
+        let context = Context::new();
+        let kept = PathBuf::from("lib").join("foo.rb");
+        let nested = PathBuf::from("lib").join("test/fixtures").join("bar.rb");
+        context.touch(&kept);
+        context.touch(&nested);
+
+        let mut excluded = HashSet::new();
+        excluded.insert(context.absolute_path_to("**/fixtures"));
```
st0012: Does this insert a pattern or a path 🤔 

And should we have 2 tests exercising both branches of

```rs
Ok(canonical) => Pattern::escape(&canonical.to_string_lossy()),
                    Err(_) => path.to_string_lossy().into_owned(),
```

alexcrocha: > Does this insert a pattern or a path 🤔

Yeah, it was ambiguous, the type was lying to us. Exclude entries can be glob patterns so they are `Box<str>` now instead of `PathBuf`.

> should we have 2 tests exercising both branches

I changed the logic so this is a bit different now, but both branches were covered. The `Err` branch was the only one needing a test, as the `Ok` branch was the pre-existing behaviour.

rust/rubydex/src/config.rs
```diff
@@ -67,16 +67,22 @@ impl Config {
...
-    pub fn excluded_paths(&self) -> HashSet<PathBuf> {
+    pub fn excluded_paths(&self) -> HashSet<Box<str>> {
         self.excluded_paths
             .iter()
-            .map(|path| self.workspace_path.join(path))
+            .map(|entry| {
+                self.workspace_path
+                    .join(&**entry)
+                    .to_string_lossy()
+                    .replace(std::path::MAIN_SEPARATOR, "/")
```
vinistock: Why are we substituting the separator here? Note that this PR changes `excluded_paths` into `excluded_glob_patterns`. For file paths, backslashes are valid (on Windows), but not for glob patterns (which always use forward slashes).

rust/rubydex/src/listing.rs
```diff
@@ -154,14 +159,29 @@ impl Job for FileDiscoveryJob {
...
+    let excluded_patterns: Arc<Vec<Pattern>> = Arc::new(
+        excluded
+            .iter()
+            .filter_map(|entry| {
+                let entry: &str = entry;
+
+                if entry.contains(['*', '?', '[']) {
+                    Pattern::new(entry).ok()
+                } else {
+                    let canonical = fs::canonicalize(entry).ok()?;
```
vinistock: Did you find that treating all entries as glob patterns was too slow? Do we need the two separate paths?

Also, I think there's potential for this to be slightly confusing. Consider this example:

```
project_path/
  my_symlink -> some_other_path
```

Trying to exclude the same directory using a pattern vs a path results in different behaviour:

```
# rubydex.toml

# This config path gets canonicalized to `some_other_path`, which then
# gets excluded since we also canonicalize the paths being compared
exclude = ["my_symlink"]

# This pattern does not get canonicalized since you cannot do it for a glob
# pattern, which means we compare if `some_other_path` matches `**/my_symlink`.
#
# It does not and so we fail to exclude
exclude = ["**/my_symlink/"]
```

I would vote for us to do the following:

1. Stop following symlinks (both in the config and in the listing). If someone created a symlink in their project, I would assume they'd prefer dealing with the symlink they created rather than the destination path
2. Treat everything as glob patterns

rust/rubydex/src/config.rs
alexcrocha: Honestly this was a quick fix after seeing Windows CI failing. I made changes to handle this properly and we no longer need to replace the separator.

rust/rubydex/src/listing.rs
alexcrocha: Thanks for the extra context. Everything is a glob now and we dropped the two separate paths, stopped resolving symlinks and canonicalizing.

`read_dir` still reads and follows symlinks, and preventing it was introducing a lot of complexity, essentially "handling symlinks to prevent us from handling symlinks". So a symlinked top-level dir is still traversed, though files are reported under the symlink path and not the destination (we no longer canonicalize). This only happens at the root, so nested symlinks are not followed.

Unsure if this is a good compromise, curious what you think.

vinistock (approved):
I think the symlink behaviour is fine as long as it's consistent.

Some other things:

1. After this change, `exclude_paths` and `excluded_paths` no longer makes sense, so we'll have to update the fields and APIs in `config.rs`, `graph.rs` and on the Ruby API. It's a breaking change, but that's okay, we can ship v0.3.0
2. Can you run `utils/bench` and compare main against this branch? Run it a few times (3-5), get the average listing time and compare. As long as there's no major performance degradation, it should be fine

alexcrocha:
> 2. Can you run `utils/bench` and compare main against this branch? Run it a few times (3-5), get the average listing time and compare

| Ref | Listing average | Min | Max |
| --- | ---: | ---: | ---: |
| `ar-glob-pattern-config-support` | 0.676667s | 0.656s | 0.718s |
| `main` | 0.677333s | 0.674s | 0.683s |

Average branch delta: `-0.000667s` (`-0.10%`) versus `main`.
🙌

vinistock (approved):
Nice

PR #903, opened 2026-07-03
Add CLI options to benchmark incremental resolution

soutaro:
## Summary

Adds `--incremental_cycle` / `--incremental_files` to measure incremental resolution cost (the re-resolve after an edit) separately from the initial full build. Dev/benchmark tooling only — no library behavior change.

After the initial build, each cycle re-indexes a rotating window of files (parse + the same invalidation the LSP runs on save) and re-runs `resolve()`, printing per-cycle and aggregate `resolve()` timings. Reindex/invalidation time is reported separately so it doesn't inflate the resolve numbers.

## Usage

```
rubydex_cli <path> --incremental_cycle=10 --incremental_files=1 --stats
```

- `--incremental_cycle=N` — run N incremental cycles after the initial build (default: off)
- `--incremental_files=M` — files re-indexed per cycle (default: 1)

## Notes

- Files are chosen by index over the listing order, not a curated sample — sufficient for the current measurements, where the cost is largely delta-independent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust/rubydex/src/main.rs
+++ b/rust/rubydex/src/main.rs
@@ -270,6 +270,9 @@ fn run_incremental_resolution(
 
     println!();
     println!("Incremental resolution ({cycles} cycle(s), {files_per_cycle} file(s)/cycle)");
+    println!("  Scenario: no-op reindex; files are indexed again without changing their contents.");
+    println!("  This should be the fastest incremental resolution path.");
+    println!("  Add scenario-based benchmarks before optimizing incremental resolution.");
 
     for cycle in 0..cycles {
         // Re-index the window of files for this cycle. Rotating the window across cycles samples
--- a/rust/rubydex/src/main.rs
+++ b/rust/rubydex/src/main.rs
@@ -1,9 +1,14 @@
 use clap::{Parser, ValueEnum};
-use std::{collections::HashSet, mem};
+use std::{
+    collections::HashSet,
+    fs, mem,
+    path::{Path, PathBuf},
+    time::{Duration, Instant},
+};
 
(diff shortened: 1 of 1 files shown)


rafaelfranca (approved):

vinistock:
Currently, we don't skip unnecessary work in incremental resolution, which is one of the reasons why it is slow. In the future, we want to be able to do nothing if all file states match exactly what the graph currently holds.

We will also have different performance profiles according to the nature of the change. For example

1. Adding blank lines or changing comments only requires updating definitions, but no resolution
2. Creating new definitions may require a certain amount of resolution, but the blast radius is generally not very large
3. Changing ancestors impacts a lot of stuff and requires re-resolving a all constants inside of the namespace + re-linearizing all ancestors for every descendants recursively

If we're going to add a benchmark for incremental resolution, we need to be able to describe the scenarios in terms of the code changes so that we can understand the performance of best/worst case scenarios.

What do you think about having a dedicated benchmark executable (like `benches/graph_memory`) that allows us to describe all of these scenarios and then prints the performance for each one?

soutaro:
@vinistock I agree that the long-term benchmark should be scenario-based and live under `benches/`.

For this PR, I added a note to the incremental output clarifying what this benchmark is measuring: it reindexes the same file contents, so it represents a no-op incremental change and should be the fastest case. My goal is to keep this as the first baseline while we work on making incremental resolution faster.

When we start optimizing incremental resolution, we should add proper scenario-based benchmarks first, covering location-only changes, new definitions, and ancestor changes separately.

PR #905, opened 2026-07-06
Deduplicate queued ancestor resolution work

st0012:
## Problems

- Resolution can enqueue the same ancestor-linearization declaration multiple times while definitions and references wait for that ancestor chain to become available.
- On large graphs, those duplicate `Unit::Ancestors` entries repeat the same linearization attempts before dependency state has changed.

## Solutions

- Track ancestor declarations that already have pending work in the resolver queue.
- Route ancestor enqueueing through a helper that keeps at most one pending `Unit::Ancestors` per declaration, then clears that marker when the unit is dequeued.
- In a config-respected Core benchmark, this reduced average resolution time from `41.641s` to `20.447s`, saving `21.194s`, with identical indexed-file, name, and definition counts.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -69,6 +69,8 @@ pub struct Resolver<'a> {
     graph: &'a mut Graph,
     /// Contains all units of work for resolution, sorted in order for resolution (less complex constant names first)
     unit_queue: VecDeque<Unit>,
+    /// Ancestor declarations already present in `unit_queue`.
+    queued_ancestors: IdentityHashSet<DeclarationId>,
     /// Whether we made any progress in the last pass of the resolution loop
     made_progress: bool,
 }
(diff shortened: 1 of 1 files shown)


rafaelfranca (approved):

jesse-shopify (approved):

vinistock (commented):
Did you do a review of every place where we are in queuing ancestor linearization? If we can avoid an extra structure and instead ensure that we never push duplicate ancestor units, that would fit the algorithm more naturally.

Or maybe the answer is that we shouldn't try to enqueue ancestor linearization eagerly and should instead only enqueue if we detect that they are missing.

rust/rubydex/src/resolution.rs
Morriar, line 139: Should we also drain `queued_ancestors` here?

st0012:
> Or maybe the answer is that we shouldn't try to enqueue ancestor linearization eagerly and should instead only enqueue if we detect that they are missing.

@vinistock Wouldn't that be a more expensive and error-prone than simple dedup?

st0012:
Closing in favor of #906

PR #906, opened 2026-07-07
Reduce duplicate ancestor work enqueueing

vinistock:
Alternative to #905

The essence of the problem is that we were enqueueing the same ancestor units multiple times. The approach on this PR is to simply remove the duplicate enqueueing, so that we can reduce the size of the queue and still achieve the same results.

I recommend reviewing by commit:

1. Stop re-enqueuing ancestor linearization units on definitions and constant references outcomes. These ancestor units are already enqueued in other parts of the pipeline, which means we push the same work more than once
2. After 1, the concept of `id_pending_linearization` is no longer necessary. All we need to know is whether a resolution outcome is partial or not, which allows us to use a boolean and reduce the size of `AncestorOutcome`

On my machine, this reduces resolution time in Core from `74.770s` -> `28.159s` (x2.65 speed up). It also ended up fixing a bug that got surfaced in a query test.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -28,7 +28,7 @@ enum Outcome {
     /// must be placed back in the queue and retried once we have progressed further. `partial` is `true` when the
     /// blocker was a still-partial ancestor chain (the member may appear once the chain completes); `false` when it was
     /// a plain missing dependency (an unresolved parent scope, alias, or lexical nesting).
-    Retry { partial: bool },
+    Retry { partial_ancestors: bool },
 }
 
 impl Outcome {
@@ -195,7 +195,9 @@ impl<'a> Resolver<'a> {
                         None => Outcome::Unresolved,
                     },
                     // Owning class not resolved yet — retry next pass
-                    None => Outcome::Retry { partial: false },
+                    None => Outcome::Retry {
+                        partial_ancestors: false,
+                    },
                 }
             }
             _ => panic!("Expected constant or singleton method definitions"),
@@ -1233,7 +1235,9 @@ impl<'a> Resolver<'a> {
             // `set_singleton_class_id`, not `add_member`, so a TODO receiver would never gain a
             // member. Emit Retry so the unit is preserved for a later resolve where the receiver
             // may exist.
-            Outcome::Unresolved if singleton => Outcome::Retry { partial: false },
+            Outcome::Unresolved if singleton => Outcome::Retry {
+                partial_ancestors: false,
+            },
             Outcome::Unresolved => Outcome::Resolved(self.create_todo_for_parent(name_id)),
             other => other,
         };
@@ -1320,7 +1324,9 @@ impl<'a> Resolver<'a> {
                 // The parent scope is genuinely unknown — not a circular alias or pending
                 // linearization, but a name that doesn't exist anywhere in the graph.
                 Outcome::Unresolved => Outcome::Unresolved,
-                Outcome::Retry { partial: false } if !preserve_retry => Outcome::Unresolved,
+                Outcome::Retry {
+                    partial_ancestors: false,
+                } if !preserve_retry => Outcome::Unresolved,
                 retry @ Outcome::Retry { .. } => retry,
             }
         } else if let Some(nesting_id) = name_ref.nesting()
@@ -1336,7 +1342,9 @@ impl<'a> Resolver<'a> {
                 NameRef::Unresolved(_) => {
                     // The only case where we wouldn't have the nesting resolved at this point is if it's available through
                     // inheritance or if it doesn't exist, so we need to retry later
-                    Outcome::Retry { partial: false }
+                    Outcome::Retry {
+                        partial_ancestors: false,
+                    }
                 }
             }
         } else {
@@ -1405,7 +1413,9 @@ impl<'a> Resolver<'a> {
 
         // Get the primary (first) resolved target
         let Some(&primary_id) = resolved_ids.first() else {
-            return Outcome::Retry { partial: false };
+            return Outcome::Retry {
+                partial_ancestors: false,
+            };
         };
 
         // Check if the primary result is still an unresolved alias
@@ -1413,7 +1423,9 @@ impl<'a> Resolver<'a> {
             self.graph.declarations().get(&primary_id),
             Some(Declaration::ConstantAlias(_))
         ) {
-            return Outcome::Retry { partial: false };
+            return Outcome::Retry {
+                partial_ancestors: false,
+            };
         }
 
         Outcome::Resolved(primary_id)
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -191,40 +191,31 @@ impl<'a> Resolver<'a> {
...
+            Outcome::Retry { .. } | Outcome::Unresolved { partial: true } => {
                 self.unit_queue.push_back(unit_id);
             }
-            Outcome::Unresolved(None) => {
+            Outcome::Unresolved { partial: false } => {
                 // We couldn't resolve this name. Emit a diagnostic
             }
-            Outcome::Retry(Some(id_needing_linearization)) | Outcome::Unresolved(Some(id_needing_linearization)) => {
-                self.unit_queue.push_back(unit_id);
-                self.unit_queue.push_back(Unit::Ancestors(id_needing_linearization));
```
st0012, line 216: FWIW, in https://github.com/Shopify/rubydex/pull/908 we found that this line along takes about 10s when resolving Core.

rust/rubydex/src/resolution.rs
```diff
@@ -19,23 +19,23 @@ use crate::model::{
...
-    Unresolved(Option<DeclarationId>),
-    /// We couldn't resolve this constant right now because certain dependencies were missing. For example, a constant
-    /// reference involved in computing ancestors (like an include) was found, but wasn't resolved yet. We need to place
-    /// this back in the queue to retry once we have progressed further. The optional declaration ID is an ancestor that
-    /// needs to be linearized before we can retry.
-    Retry(Option<DeclarationId>),
+    /// The constant was successfully resolved to the given declaration ID.
+    Resolved(DeclarationId),
+    /// We had everything we needed to resolve this constant, but we couldn't find it. When `partial` is `true`, the
+    /// lookup stopped at a still-partial ancestor chain, so the member may yet appear once that chain completes and the
```
st0012: From reading this, it feels like `Unresolved { partial: true }` is the same as `Retry { partial : true }`?

Why do we need both?

st0012 (commented):
Are the numbers in `--stats` from indexing Core change because of this? Are they expected?

vinistock:
> Are the numbers in --stats from indexing Core change because of this? Are they expected?

Where are you seeing this? Running on my machine `main` vs this branch, the numbers are the same.

rust/rubydex/src/resolution.rs
vinistock: Good catch. We still need to differentiate between `Unresolved` and `Retry`, but the former doesn't need the `partial` state.

`Unresolved` means: we had all dependencies figured out, but still couldn't resolve this constant. It's terminal and it means an undefined constant diagnostic should be emitted.


`Retry` means: we did not have the required dependencies to resolve this constant, so we will try again later in case the dependencies get solved and we can go further in the analysis.

I removed the `partial` state in the latest commit.

st0012:
> Where are you seeing this? Running on my machine main vs this branch, the numbers are the same.

Sorry I wasn't clear: I wasn't seeing this, just asking if this is changing because I see query test's assertion changed and I'm not entirely sure why

rust/rubydex/src/resolution.rs
```diff
@@ -19,23 +19,21 @@ use crate::model::{
...
+    /// The constant was successfully resolved to the given declaration ID.
+    Resolved(DeclarationId),
+    /// We had everything we needed to resolve this constant (all relevant ancestor chains are complete), but the member
+    /// isn't there. This is definitive: the unit can be dropped and a diagnostic emitted.
+    Unresolved,
+    /// We couldn't resolve this constant right now because a dependency was missing, so it may succeed later. The unit
+    /// must be placed back in the queue and retried once we have progressed further. `partial` is `true` when the
+    /// blocker was a still-partial ancestor chain (the member may appear once the chain completes); `false` when it was
+    /// a plain missing dependency (an unresolved parent scope, alias, or lexical nesting).
+    Retry { partial: bool },
```
st0012: I'd like to name the flag more explicitly, like `with_partially_resolved_ancestor`.

rust/rubydex/src/resolution.rs
```diff
@@ -191,40 +189,31 @@ impl<'a> Resolver<'a> {
...
 
         match outcome {
-            Outcome::Retry(None) => {
-                // There might be dependencies we haven't figured out yet, so we need to retry
+            Outcome::Retry { .. } => {
                 self.unit_queue.push_back(unit_id);
             }
-            Outcome::Unresolved(None) => {
+            Outcome::Unresolved => {
                 // We couldn't resolve this name. Emit a diagnostic
```
st0012, line 211: Is this comment still relevant?

rust/rubydex/src/resolution.rs
```diff
@@ -233,24 +222,12 @@ impl<'a> Resolver<'a> {
...
+            Outcome::Resolved(declaration_id) => {
                 self.graph.record_resolved_reference(id, declaration_id);
                 self.made_progress = true;
             }
-            Outcome::Resolved(resolved_id, Some(id_needing_linearization)) => {
-                self.graph.record_resolved_reference(id, resolved_id);
-                self.made_progress = true;
-                self.unit_queue.push_back(Unit::Ancestors(id_needing_linearization));
+            Outcome::Retry { .. } | Outcome::Unresolved => {
+                self.unit_queue.push_back(unit_id);
```
st0012, line 232: If `Unresolved` is definitive, is it necessary to requeue the unit?

rust/rubydex/src/resolution.rs
vinistock: I renamed to `partial_ancestors` to make it a bit shorter.

rust/rubydex/src/resolution.rs
vinistock: It's not relevant only because of todos, but I still want to revisit this. I really believe our resolution strategy is incompatible with todos and I'd like to propose a different approach that doesn't take `Unresolved` constant definitions and force them to be defined with a name that we can't work backwards from during incremental changes.

rust/rubydex/src/resolution.rs
vinistock: Great catch. Indeed, we should `push_work` for `Unresolved`, so that the next iteration can try again, but not this one.

vinistock:
> Sorry I wasn't clear: I wasn't seeing this, just asking if this is changing because I see query test's assertion changed and I'm not entirely sure why

The simplification fixed a bug for that scenario. The `class << Foo` was incorrectly enqueuing linearization for `Bar` instead of `Foo::<Foo>`, which means that ancestors were missing. The extra completion items are all constants accessible through `Object` inheritance.

st0012 (approved):

PR #907, opened 2026-07-07
Allow deferring singleton linearization

vinistock:
We were using a boolean flag to decide whether to push singleton ancestor linearization into the queue or do it immediately. However, it turns out that we can use a third mode: do nothing at all.

This is useful when we need to create singleton classes in the middle of the linearization process itself, which helps to avoid pushing duplicate work into the queue as the linearization is happening inline.

Combined with #906, this gets us down to `24.613s` for resolution (a 14% improvement).

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -37,6 +37,19 @@ impl Outcome {
     }
 }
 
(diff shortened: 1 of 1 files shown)


alexcrocha (approved):
Looks good to me. And I like how it makes the intent clearer and we don't have to recall what the bool is for.

st0012 (approved):

PR #909, opened 2026-07-07
Load default config before CLI listing

st0012:
## Problems

- The CLI loaded files before initializing the graph config, so `rubydex.toml` exclusions were ignored during listing.

## Solutions

- Initialize the graph before listing, derive a workspace root for single-path invocations, and load the default config before collecting file paths.
- Pass `graph.excluded_paths()` into file discovery so configured exclusions are respected.
- Warn when multiple paths are passed because `rubydex.toml` is intentionally ignored in that mode.
- Add CLI coverage for both single-workspace config loading and the multi-path warning.

### DIFF
--- a/rust/rubydex/src/main.rs
+++ b/rust/rubydex/src/main.rs
@@ -1,5 +1,5 @@
 use clap::{Parser, ValueEnum};
-use std::{collections::HashSet, mem};
+use std::{fs, mem, path::PathBuf};
 
 use rubydex::{
     dot,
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/main.rs
```diff
@@ -85,16 +85,48 @@ fn exit(print_stats: bool) {
...
 
+fn workspace_path_for(paths: &[String]) -> Option<PathBuf> {
+    match paths {
+        [path] => {
+            let canonicalized = fs::canonicalize(path).ok()?;
+
+            if canonicalized.is_dir() {
+                Some(canonicalized)
+            } else {
+                canonicalized.parent().map(PathBuf::from)
```
vinistock: This assumption may not always be right. For example, if someone passes `workspace_path/lib/foo.rb`, we'd consider the workspace path as `workspace_path/lib`.

We might need to separate the CLI arguments into an optional workspace path + additional paths. I can't think of a heuristic that would allow us to figure out which one of the provided paths is the workspace root.

st0012: So we assume first path is a workspace root, and the rest are not?

vinistock: We can try that, but we can only assume the first one is the workspace path if it's a directory. Someone could run the analysis in random Ruby files, in which case the workspace path would be missing (and that should be okay).

vinistock (approved):

PR #910, opened 2026-07-08
Expose resolved method calls as a CALLS edge

paracycle:
## Goal

Expose **method calls** in the Cypher graph, resolved to the method declarations they target — so the graph can answer call-graph questions, not just definition/inheritance ones.

## What it adds

A new relationship type:

```
CALLS: Document → Declaration (a Method)
```

It maps a file's method references to the method declaration each one resolves to, for calls whose **receiver type is statically known**:

- a **constant receiver** (`Foo.bar`), resolved through the receiver's singleton-class ancestry (where class methods live), and
- an **implicit `self`** call in a known scope (e.g. calling a sibling method from inside an instance method).

```cypher
-- Everything a file calls (resolved)
MATCH (d:Document)-[:CALLS]->(m:Method) RETURN d.name, m.name

-- Callers of a specific method
MATCH (d:Document)-[:CALLS]->(m:Method {unqualified_name: 'bar()'}) RETURN d.name
```

## How it resolves

The indexer already records each call with the *receiver type's* name (the singleton-class name for `Foo.bar`, the enclosing type for a `self` call). The schema:

1. resolves that name to a declaration (the same way constant references resolve), then
2. looks the method up along that declaration's ancestry, matching on the base name — method members are keyed `name()` while call sites record the bare `name`, so a direct id match doesn't work.

## Scope / non-goals

Calls with a **dynamic or unknown receiver** (`foo.bar` on a local, a method-chain result, etc.) carry no receiver from the indexer and are **intentionally not represented** — resolving them would require type inference. This edge covers the statically-resolvable subset only; it's best-effort (first match in ancestry order) rather than a guaranteed runtime target.

The resolver currently lives in the cypher schema (consistent with how it already maps the graph); a future refactor could share a `Graph`-level method resolver with `query.rs`'s completion path.

## Verification

`cargo test -p rubydex query::cypher` (20 tests, incl. resolution + reverse "who calls this" + that dynamic-receiver calls are excluded); clippy + fmt clean; `--schema` lists `CALLS`; full gem suite green.

### DIFF
--- a/rust/rubydex/src/query/cypher/schema.rs
+++ b/rust/rubydex/src/query/cypher/schema.rs
@@ -21,13 +21,15 @@
 //! - `HAS_ANCESTOR`: `Declaration` → `Declaration` (linearized ancestor chain, incl. modules)
 //! - `HAS_DESCENDANT`: `Declaration` → `Declaration` (reverse of `HAS_ANCESTOR`)
 //! - `REFERENCES`: `Document` → `Declaration` (constant references)
+//! - `CALLS`: `Document` → `Declaration` (a `Method`; a constant-receiver method call like `Foo.bar`
+//!   resolved to its method declaration)
 
 use std::collections::{HashSet, VecDeque};
 
-use crate::model::declaration::Declaration;
+use crate::model::declaration::{Declaration, Namespace};
 use crate::model::definitions::{Definition, Mixin};
 use crate::model::graph::Graph;
-use crate::model::ids::{ConstantReferenceId, DeclarationId, DefinitionId, UriId};
+use crate::model::ids::{ConstantReferenceId, DeclarationId, DefinitionId, NameId, StringId, UriId};
 
 use cypher_parser::{CypherValue, GraphProvider};
 
@@ -71,6 +73,11 @@ pub enum RelType {
     HasDescendant,
     /// `Document` → `Declaration`: a constant reference in the file resolves to a declaration.
     References,
+    /// `Document` → `Declaration` (a `Method`): a method call whose receiver *type* is statically
+    /// known — a constant receiver (`Foo.bar`, resolved via the receiver's singleton-class ancestry)
+    /// or an implicit `self` call in a known scope — resolved to the method declaration. Calls with
+    /// a dynamic/unknown receiver carry no receiver and are not represented.
+    Calls,
 }
 
 /// Catalog metadata for a relationship type: the type itself, its canonical name, endpoint labels,
@@ -163,6 +170,13 @@ const REL_SCHEMAS: &[RelSchema] = &[
         from: "Document",
         to: "Declaration",
         description: "A file references a constant declaration",
+    },
+    RelSchema {
+        rel: RelType::Calls,
+        name: "CALLS",
+        from: "Document",
+        to: "Method",
+        description: "A file calls a method whose receiver type is statically known",
     },
 ];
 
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -633,6 +650,80 @@ fn document_references(graph: &Graph, uri_id: UriId) -> Vec<NodeRef> {
     targets
 }
 
+/// The method declarations reached by the constant-receiver method calls in a document.
+///
+/// Only calls whose receiver is a constant (e.g. `Foo.bar`) are statically resolvable, so calls
```
Morriar, line 655: Looks like we do handle calls on implicit receivers? 🤔

rust/rubydex/src/query/cypher/tests.rs
```diff
@@ -226,3 +226,50 @@ fn document_uri_path_and_name_are_distinct() {
...
+        "file:///m.rb",
+        "
+            class Foo
+              def self.bar; end
+            end
+
+            class Client
+              def run
+                Foo.bar       # constant receiver -> resolvable
+                helper        # implicit receiver -> not resolvable
```
Morriar, line 242: The test seems to resolve it?

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -633,6 +650,80 @@ fn document_references(graph: &Graph, uri_id: UriId) -> Vec<NodeRef> {
...
+    // Call sites record the bare message (`bar`), while method declarations are keyed by
+    // `name()` in a namespace's members, so we match on the base name rather than the `StringId`.
+    let call_name = graph.strings().get(&method_str)?.as_str();
+
+    let receiver_decl_id = *graph.name_id_to_declaration_id(receiver)?;
+    let receiver_decl_id = resolve_to_namespace(graph, receiver_decl_id)?;
+    let namespace = graph.declarations().get(&receiver_decl_id)?.as_namespace()?;
+
+    for ancestor in namespace.ancestors() {
+        let Ancestor::Complete(ancestor_id) = ancestor else {
```
st0012, line 697: If we encounter partial ancestor here, we should exit the loop instead of continuing. Otherwise we could be returning the incorrect target.

rust/rubydex/src/query/cypher/schema.rs
```diff
@@ -633,6 +650,80 @@ fn document_references(graph: &Graph, uri_id: UriId) -> Vec<NodeRef> {
...
+/// Only calls whose receiver is a constant (e.g. `Foo.bar`) are statically resolvable, so calls
+/// with an implicit or non-constant receiver are skipped.
+fn document_method_calls(graph: &Graph, uri_id: UriId) -> Vec<NodeRef> {
+    let Some(document) = graph.documents().get(&uri_id) else {
+        return Vec::new();
+    };
+
+    let mut seen = HashSet::new();
+    let mut targets = Vec::new();
+    for ref_id in document.method_references() {
```
Morriar, line 664: Note that we add a few synthetic method references: https://github.com/Shopify/rubydex/blob/5d58c118d7aae37f4cb5456ee09258e9a26880a0/rust/rubydex/src/indexing/ruby_indexer.rs#L2233-L2236

You may want to add a test showing the behavior:

```rb
class Foo
  def self.>(other); true; end
  def self.<=>(other); 0; end
end

Foo > 1
```

 Observed Cypher result:

```rb
["<=>()", ">()"]
```

PR #911, opened 2026-07-08
Remove Rust MCP server

st0012:
## Problems

- Rubydex now exposes MCP through the Ruby `rdx --mcp` server.
- The old Rust `rubydex-mcp` crate and precompiled `rubydex_mcp` packaging path are redundant.

## Solutions

- Remove the `rust/rubydex-mcp` crate from the workspace and regenerate `Cargo.lock`.
- Stop copying and packaging the old `rubydex_mcp` binary during extension builds and releases.
- Update architecture and contributor docs to point at the Ruby MCP server.


test/mcp_server_test.rb
```diff
@@ -368,6 +368,8 @@ def test_mcp_server_can_be_required_directly
   end
 
   def test_mcp_server_e2e
+    skip("This test times out when running with Valgrind") if ENV["RUBY_MEMCHECK_RUNNING"]
```
alexcrocha, line 371: Do we know why this test times out when running with Valgrind?

alexcrocha (approved):

test/mcp_server_test.rb
st0012: When Ruby program runs with Valgrind, it becomes extremely slow as memory access/allocation now needs to be checked. So MCP server requests now easily runs over 30s and times out.
We could increase the timeout to a degree the test can still work with Valgrind, but then it'll make it hard to catch legit hanging issues locally (e.g. the request hangs, and you only get notified after 5mins). It's also not necessary IMO.

alexcrocha: I see, makes sense

PR #913, opened 2026-07-08
Update architecture documentation to reflect current state

vinistock:
This PR updates our architecture docs, which were a bit outdated. I tried to be thorough and explicit with the model, the phases, and especially the resolution algorithm, which is complex and critical for the performance of the analysis.

I removed the Excalidraw thing as it's indeed annoying to try to keep it up to date.


docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
+Rubydex represents Ruby code through a graph, which is essentially a collection of nodes (entities) and edges
+(relationships). For example, if a class `Foo` inherits from class `Bar`, we have two entities related to each other
+through inheritance. The complete structure of the graph can be found in `rust/rubydex/src/model/graph.rs`.
 
-A **Declaration** represents the global semantic concept of a name, combining all definitions that contribute to the same fully qualified name. Declarations are produced during resolution.
+A **Definition** represents a single source-level construct found at a specific location in the code. It captures
+exactly what was found in the file, with no attempts to derive semantic meaning from the data.
+
+A **Declaration** represents the global semantic concept of an entity, combining all definitions that contribute to the
+same fully qualified name. Declarations are produced during resolution.
```
alexcrocha, line 68: We mention Declarations are produced during resolution but don't mention where Definitions are produced. We should be explicit and also say that Definitions are produced during indexing.

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
-## Two-Stage Indexing Pipeline
+```
+    ┌─────────────┐
+    | Declaration |  one per fully qualified name, assembled during resolution
+    └─────────────┘
+           ▲
+           |   many definitions that contribute to the same fully qualified name -> one declaration
+           ▼
+    ┌─────────────┐
+    | Definition  |   a single source-level construct, captured exactly as written
```
alexcrocha:
```suggestion
    | Definition  |   a single source-level construct, captured during indexing exactly as written
```

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
-Discovery walks the AST and extracts definitions from source code. It captures **only what is explicitly written**, making no assumptions about runtime behavior.
+Names are unique constant name structures that connect all of the pieces required to resolve them against the graph
+(implemented in `rust/rubydex/src/model/name.rs`). They contain the unqualified name of the constant (interned string),
+an optional parent scope (an ID to another name) and an optional nesting (an ID to another name). It is the connection
+of the 3 pieces that allows resolution to determine what a constant refers to.
 
-**What Discovery does:**
+**Note**: names and constant references are intentionally separated. Code can refer to the same constant multiple times
+within the same namespace. While the location of each reference is different, their name object might be the same. By
+reusing the same name, we don't have to resolve each constant reference, but rather only each unique name reference.
```
alexcrocha: Here we start by comparing names to constant references, but we haven't defined what a constant reference is in the Rubydex context yet. The note talks about the difference between the two, but I believe for a new reader it would be clearer if they already had an understanding of the constant reference concept. 

Something like:

```suggestion
A constant reference is a specific occurrence where code refers to a constant. In the graph, each reference points to a
`Name` and stores the document and offset where that occurrence appeared. 

**Note**: names and constant references are intentionally separated. Code can refer to the same constant multiple times
within the same namespace. While the location of each reference is different, their name object might be the same. By
reusing the same name, we don't have to resolve each constant reference, but rather only each unique name reference.
```

alexcrocha (approved):
Looks good to me. Left a couple suggestions to improve clarity 👍

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
+Rubydex analyzes Ruby codebases in two distinct phases: **Indexing** and **Resolution**. Understanding this separation
+is crucial for working with the codebase.
 
-## Core Concepts: Definition vs Declaration
+## Core model
 
-A **Definition** represents a single source-level construct found at a specific location in the code. It captures exactly what the parser sees without making assumptions about runtime behavior.
+Rubydex represents Ruby code through a graph, which is essentially a collection of nodes (entities) and edges
+(relationships). For example, if a class `Foo` inherits from class `Bar`, we have two entities related to each other
+through inheritance. The complete structure of the graph can be found in `rust/rubydex/src/model/graph.rs`.
```
Morriar: > The complete structure of the graph can be found in `rust/rubydex/src/model/graph.rs`.

We also have edges hidden in `definitions.rs`, `declaration.rs` and `document.rs`

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
 ```ruby
-# foo.rb
-module Foo
-  class Bar; end
-end
-
-# other_foos.rb
-class Foo::Bar; end
-class Foo::Bar; end
+0. # foo.rb
```
Morriar: I'm guessing you feel the need to number the lines because starting at 0 is confusing.

Though the numbering will mess up the syntactic highlighting of the Ruby code block. I'd use 3 blocks and drop the numbering but explain they are 0-indexed.

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
...
-3. Class definition for `Foo::Bar` in `other_foos.rb`
-4. Class definition for `Foo::Bar` in `other_foos.rb`
+- module Foo (line 1 foo.rb)
+- class Bar (line 2 foo.rb)
+- class Foo::Bar (line 1 other_foos.rb)
+- class Foo::Bar (line 2 other_foos.rb)
 
-**Declarations** (2 total - what resolution produces):
+Which produce 2 declarations during resolution: `Foo` and `Foo::Bar`. The graph can be traversed from the bottom
+(documents) or from the top (declarations) and all connections happen through IDs defined in
```
Morriar: Saying top and bottom for a graph is a bit weird. I'd remove this part of the sentence it doesn't really bring anything. Keep the edges as IDs and make it a front matter.

docs/architecture.md
```diff
@@ -1,135 +1,174 @@
 # Architecture
```
Morriar, line 2: I wonder if we could start by introducing the pipeline like Sorbet does here: https://github.com/sorbet/sorbet/blob/master/docs/pipeline.md#pipeline

vinistock: Good idea. I added one, let me know what you think.

docs/architecture.md
vinistock: I tried better explaining what I meant rather than removing. I wanted to communicate that the graph can be queried from the bottom by accessing a document directly or from the top by accessing a declaration (and both have their own purposes).

docs/architecture.md
vinistock: Not really. I was trying to provide anchors for AI. I used comments instead.

docs/architecture.md
vinistock: I included that in the paragraphs before and cleaned up the diagram.

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
-**What Discovery does:**
+Rubydex represents Ruby code through a graph, which is essentially a collection of nodes (entities) and edges
+(relationships). For example, if a class `Foo` inherits from class `Bar`, we have two entities related to each other
+through inheritance.
 
-- Creates `Definition` objects for classes, modules, methods, constants, variables
-- Records source locations, comments, and lexical ownership (`owner_id`)
-- Captures unresolved constant references (e.g., `Foo::Bar` as a `NameId`)
-- Records mixins (`include`, `prepend`, `extend`) on their containing class/module
+All top level nodes and edges can be found in `rust/rubydex/src/model/graph.rs`. Note that certain nodes also store
```
st0012: Reading the previous paragraph and then this, I'd expect to find class related models and inheritance edge definitions inside `graph.rs`. But I don't think that's the case?

Should we rephrase this so it doesn't appear to be that way?

st0012: And I'm not sure what top level nodes and edges mean here

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
-
-## Graph Structure
+- module Foo (1 in foo.rb)
+- class Bar (2 foo.rb)
+- class Foo::Bar (3 in other_foos.rb)
+- class Foo::Bar (4 in other_foos.rb)
 
-Rubydex represents the codebase as a graph, where entities are nodes and relationships are edges. The visualization below shows the conceptual structure (implemented as an adjacency list using IDs).
+Which produce 2 declarations during resolution: `Foo` and `Foo::Bar`. All connections happen through IDs defined in
+`rust/rubydex/src/model/ids.rs`.
```
st0012: It's not super clear what IDs mean here and the section below.
I think we can explaining these without mentioning the id, which IMO is an implementation details for how we chain things together, and can be immediately explained in the next major section.
I think the flow could be:

1. Core model
    1. Definition
    2. Declration
    3. Document
    4. Example
    5. How they related to each other, conceptually
2. How things connect to each other: IDs
    1. Explain what are ids, and why we chose it
    2. How the Declaration - Definition - Document relationship described using ids
3. Names (which immediately can use and reinforce the id concept)
...

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
-- `model/definitions.rs`: Individual definitions discovered from source code
-- `model/declaration.rs`: Global declarations produced during resolution
-- `model/graph.rs`: The main graph structure containing all entities
+Names are unique constant name structures that connect all of the pieces required to resolve them against the graph
+(implemented in `rust/rubydex/src/model/name.rs`). They contain the unqualified name of the constant (interned string),
+an optional parent scope (an ID to another name) and an optional nesting (an ID to another name). It is the connection
+of the 3 pieces that allows resolution to determine what a constant refers to.
 
-### ID Types
+A constant reference is a specific occurrence where code refers to a constant. In the graph, each constant reference
```
st0012: Let's give constant references its own section.
I think in the Names section, it's important to more clearly explain `parent_scope` and `nesting` and why we need both.

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
+points to a `Name` and stores the document and offset where that occurrence appeared.
 
-Connections between nodes use hashed IDs defined in `ids.rs`:
+**Note**: names and constant references are intentionally separated. Code can refer to the same constant multiple times
```
st0012: While this explains the distinction, I don't think readers just reaching the `Name` concept can already grasp this.
We need to have an example like:

```rb
class Foo; end

Foo
Foo

# We have 1 Foo Name
# But we'll need 2 constant references to Foo, as the 2 Foo references are located in separate location.
# So the structure of a constant reference consists of
# 1. A name Name(str: StringId(Foo)
# 2. A location
```

Again, I think this should be explained in a separate section, not along with `Names`.

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
+  class Foo::Qux
+  end
+end
+```
 
-All collection-returning tools iterate over `IdentityHashMap` or `IdentityHashSet` structures. These use a deterministic hasher, so iteration order is fixed for a given map state.
+Since tracing autoload order in Ruby is not possible, we may process `file2.rb` before `file1.rb`. At indexing time, we
+wouldn't know what the `Foo` constant reference points to since we have not yet discovered its definition. The key
+insight regarding this initial phase is that we can only capture document-specific data and any understanding that
+requires cross-file knowledge can only be derived once indexing is complete.
```
st0012: When expressing files can be indexed at arbitrary orders, let's not use ordered file name? Also I think we can rephrase it a bit:


```suggestion
`foo.rb` and `bar.rb` can be indexed at arbitrary order, as we can't statically predict Ruby's file loading order due to `autoload`, conditional require...etc.

Because of this, we wouldn't know what the `Foo` constant reference in `Foo::Qux` points to when `bar.rb` is indexed:
- If `foo.rb` is indexed first, it **could be** module `Foo`
- If `bar.rb` is indexed first, it'd be undefined

This is why we only capture document-specific data at this stage. And any cross-file knowledge is constructed only after we've indexed all the files.
```

docs/architecture.md
```diff
@@ -1,136 +1,235 @@
...
+  processed as part of the fixed-point loop. Introducing competing data structures that track similar state is a smell
+  that the implementation is not fitting the algorithm.
+- Single convergence point: all constant related resolution and dependency solving must happen inside of the resolution
+  loop. If a change requires moving constant related solving out of the loop, it's a smell that the change does not fit.
+- Minimal work: resolution always consumes the least amount of work that needs to be solved. If a change is
+  requiring a full graph traversal, then it's a smell that it does not fit the algorithm.
+
+## Incremental invalidation
+
+One of the goals of Rubydex is to perform extremely fast incremental analysis. If a document is edited, we want to be
```
st0012: How about:

`````
Rubydex aims to provide extremely fast incremental analysis. When a document is changed, we want to determine the minimum set of graph state that needs to be invalidated and re-analyzed. The invalidation algorithm
lives in `rust/rubydex/src/model/graph.rs` and it produces the pending work that feeds the resolution pass.

In its ideal form, Rubydex should be able to perform layered incremental invalidation by asking layered questions to decide which level of invalidation is needed.

For example:

Let's say we have:

```rb
# Yo I'm comment
class Foo; end

class Bar < Foo; end
```

1. Does the edit requires resolution?
    - If an edit only adds a space to `Foo`'s comment, we just need to update the `Foo` definition. But no re-resolution is ever needed
    - If a re-resolution is required, we go to the next question
2. Does the edit changes an ancestor?
    - If the edit adds `class Baz < Foo; end`, then we only need to invalidate `Foo`'s descendants, but `Foo` and `Bar` declarations stays unaffected.
    - If an ancestor chain is affected, we go to the next question
3. How many things need to be invalidated:
    - If the edit adds `include MyModule` to `Bar`, only `Bar`'s ancestors changed, so only `Bar` (and its descendants) needs to be invalidated
    - If the edit adds `include MyModule` to `Foo`, it affects the ancestor chain of `Foo`, which means both `Foo` and `Bar` needs to be invalidated and re-resolved

This does NOT describe the ideal algorithm precisely, but it shows the granularity we want to achieve in the long term.

And that means, infrastructure that will help us achieve the goal should be reviewed carefully during refactoring/optimization.
`````

vinistock: Nice, I like the idea. I made some edits, but followed your structure.

docs/architecture.md
vinistock: I added a section for constant references, but I don't think we need to list every type of parent scope. I feel like the architecture docs explain big blocks, but if people want to dig into specifics they can just open the files referred to here.

docs/architecture.md
vinistock: I tried explaining the IDs a bit further. I think having a separate section might be a bit much, let me know what you find after the changes.

docs/architecture.md
vinistock: I tried better explaining. The storage of the top level nodes is all in the graph, which doesn't mean that the structs are defined in the same file.

For example, `Declaration`, `Definition` and `Document` are all top level nodes, which get stored directly in the graph. However, fully qualified name is not a top level node. It's an attribute of `Declaration`.

This might seem a bit silly, but I do find the explicitness to be quite important because in the early days we made modelling mistakes by putting data inside of a node when it actually needed to be a top level node itself.

st0012 (approved):

PR #916, opened 2026-07-09
Precompute integer sort keys in prepare_units

soutaro:
## What

`prepare_units` sorts definition and reference units by `(name depth, URI, offset)`. The comparator did a hashmap lookup for the name depth and a URI **string** comparison on every comparison, so on large graphs the sort dominates `prepare_units`.

This precomputes integer sort keys instead:

- Compute each document URI's lexicographic rank once into a `UriId -> u32` map.
- Store `(depth, uri_rank, offset)` as the sort key when classifying units, so the sort compares plain integers.

## Result

- The resulting queue order is unchanged, so resolution output is identical — verified by a full `--stats` run being byte-identical (except timing) on a large Ruby codebase.
- On a large Ruby codebase, resolution time drops from ~16.7s to ~14.2s (~15%, roughly 2.5s), with lower run-to-run variance. Indexing time is unaffected.

## Independence

This is independent of the descendants-computation change (#892) and the partial-ancestor-reuse change (#915): it only touches the sort-key computation in `prepare_units`, preserves the exact queue order, and does not touch `linearize_ancestors` or descendant computation. Based directly on `main`.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1884,21 +1884,6 @@ impl<'a> Resolver<'a> {
         depth
     }
 
-    /// Computes name depths for all names into a `NameId → depth` map via memoized recursion.
-    ///
-    /// `prepare_units` computes depths on demand for only the names it sorts; this eager variant exists so tests
-    /// can assert the depth of every name in a fixture.
-    #[cfg(test)]
-    pub(crate) fn compute_name_depths(names: &IdentityHashMap<NameId, NameRef>) -> IdentityHashMap<NameId, u32> {
-        let mut cache = IdentityHashMap::with_capacity_and_hasher(names.len(), IdentityHashBuilder);
-
-        for &name_id in names.keys() {
-            Self::name_depth(name_id, names, &mut cache);
-        }
-
-        cache
-    }
-
     /// Drains `pending_work` and classifies items into the resolution queue.
     /// Namespace definitions and constant references are sorted by name depth for deterministic
     /// resolution order. Non-namespace definitions (methods, attrs, variables) are returned
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -14,7 +14,7 @@ use crate::model::{
     definitions::{Definition, Mixin, Receiver},
     graph::{Graph, Unit},
     identity_maps::{IdentityHashBuilder, IdentityHashMap, IdentityHashSet},
-    ids::{ConstantReferenceId, DeclarationId, DefinitionId, NameId, StringId},
+    ids::{ConstantReferenceId, DeclarationId, DefinitionId, NameId, StringId, UriId},
     name::{Name, NameRef, ParentScope},
 };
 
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -1912,6 +1912,22 @@ impl<'a> Resolver<'a> {
         let names = self.graph.names();
         let depths = Self::compute_name_depths(names);
 
+        // Precompute the lexicographic rank of every document URI. Definitions and references are sorted by
+        // (name depth, URI, offset) below; storing precomputed integer sort keys instead of looking up name
+        // depths and comparing URI strings on every comparison makes sorting substantially cheaper on large graphs.
+        let mut uris: Vec<(&str, UriId)> = self
```
amomchilov, line 1910: What actually determines the order of these? Isn't it non-deterministic?

soutaro: It's sorted by the `uri()` string of the documents. It's deterministic and not changed by this PR.

rust/rubydex/src/resolution.rs
```diff
@@ -1884,9 +1884,11 @@ impl<'a> Resolver<'a> {
...
     }
 
-    /// Pre-compute name depths for all names into a `NameId → depth` map. Each name's depth is
-    /// computed once via memoized recursion, then used as an O(1) lookup key during sorting in
-    /// `prepare_units`.
+    /// Computes name depths for all names into a `NameId → depth` map via memoized recursion.
+    ///
+    /// `prepare_units` computes depths on demand for only the names it sorts; this eager variant exists so tests
+    /// can assert the depth of every name in a fixture.
+    #[cfg(test)]
```
vinistock: Instead of keeping this for the sake of the tests, let's re-work the tests and get rid of the method.

rust/rubydex/src/resolution.rs
```diff
@@ -1965,11 +1993,9 @@ impl<'a> Resolver<'a> {
                     let Some(constant_ref) = self.graph.constant_references().get(&id) else {
                         continue;
                     };
-                    let uri = self.graph.documents().get(&constant_ref.uri_id()).unwrap().uri();
-                    const_refs.push((
-                        Unit::ConstantRef(id),
-                        (*constant_ref.name_id(), uri, constant_ref.offset()),
-                    ));
+                    let uri_rank = *uri_ranks.get(&constant_ref.uri_id()).unwrap();
+                    let depth = Self::name_depth(*constant_ref.name_id(), names, &mut depths);
```
vinistock, line 1982: There are two changes mixed in this PR. One is the URI related and the second one is inlining the name depth computation right here and in each branch above.

Did you benchmark these separately? I would expect the name change to have higher benefit than the URI one. Depending on what we find, we can drop the URI changes and just inline the name depth.

vinistock (commented):
Let's please check if most of the gains come from inlining the name depth and, if so, remove the URI ranking changes

rust/rubydex/src/resolution.rs
vinistock: Talked to Soutaro. Both changes improve performance.

vinistock (approved):

PR #918, opened 2026-07-09
Avoid invalidate a document when the file's content is not changed

st0012:
If a document's content doesn't change, we don't need to invalidate and produce work for it. So in this PR, I added `content_hash` to `Document` and compare the value before invalidation happens, so we can skip unnecessary work.

This currently doesn't save much in a real-world scenario, but it will be the first layer of the layered invalidation we need when we start introducing cache or database storage.

Ideally we can skip file indexing if the content is the same as well. But because indexing happens in parallel and currently does not need to access the main graph, making such change will be architectural. Given that it won't give us much additional gain, I think we can defer that change for later.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -1044,7 +1044,17 @@ impl Graph {
     /// 3. `extend` -- merges the new `LocalGraph` into the now-clean graph
     pub fn consume_document_changes(&mut self, other: LocalGraph) {
         let uri_id = other.uri_id();
-        let old_document = self.documents.remove(&uri_id);
+
+        let old_document = match self.documents.entry(uri_id) {
+            Entry::Occupied(entry) => {
+                // No changes to the document, skip invalidation and merging
+                if entry.get().content_hash() == other.document().content_hash() {
+                    return;
+                }
+                Some(entry.remove())
+            }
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/model/graph.rs
```diff
@@ -1044,6 +1044,16 @@ impl Graph {
...
         let uri_id = other.uri_id();
+
+        if self
+            .documents
+            .get(&uri_id)
+            .is_some_and(|old| old.content_hash() == other.document().content_hash())
+        {
+            // No changes to the document, so no invalidation is necessary.
+            return;
+        }
```
vinistock: Would it be possible to avoid the double lookup with the remove below? Maybe something like this:

```rust
match self.documents.entry(&uri_id) {
  Entry::Occupied(entry) => {
    // check content hash and return or remove the document
  }
}
```

vinistock (approved):

st0012:
@soutaro With this change, the incremental resolution benchmark would stop providing useful information. Are you able to update it to cover more scenarios? Like removing & reindexing the same files would also give us more signals for improvement opportunities.

rust/rubydex/src/model/graph.rs

soutaro:
@st0012 Sure! I will remove the incremental option.

Issue #825, opened 2026-05-26
Strange: `uri` not a `URI`


amomchilov:
Contrary to its name, `Rubydex::Declaration#uri` returns a `String`. Real users need to use `URI.parse(x.uri).path` to get what they really wanted, which was the path.

I think we can both improve the ergonomics, and the naming (for [least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)).

Proposal:

```diff
- #: -> String
+ #: -> URI
  def uri; end

+ #: -> Pathname
+ def pathname; end # Perhaps just `path`, it'll be very commonly used

+ #: -> String
+ def uri_string # Returns the raw URI as a string, like `#uri` does today.
```

amomchilov:
Decision from our project sync: punt for now. Wait to see more usages, and use that to inform what would be most useful for real-world code.

Morriar:
We already have `Location#to_file_path`.

I looked into making `Location#uri` an actual `URI::Generic`, while it works it brings 2 pain points: 

1. We have to build URI objects for every location and this isn't free.
2. Second, all the other APIs need to use URI as well, which means the user can't just do `graph.index_file("foo.rb")`, they have to wrap the path in a URI.
We could make the `index_file` and friends take a `(URI | String)` but I'd suggest holding of on that until we have a bit more clients.

vinistock:
This is related to #313. I do think we need to always use URI objects for both inputs and outputs.

While it requires consumers to parse and produce objects, it also prevents common URI related mistakes by surfacing URI parsing errors on the consumer side.

For example, it's easy to do something like this
```ruby
uri = "file://#{file_path}"
graph.any_operation(uri)
```

and not realize that produces invalid URIs on Windows, which can end up having surprising behaviour.

PR #920, opened 2026-07-10
Expose `Definition#document` in the Ruby API

alexcrocha:
This PR exposes `Definition#document`, which returns the `Rubydex::Document` a definition belongs to. 

A definition already stores its document's URI id, the public API just didn't expose it so callers had no direct path from a definition to its document. This moves the public API toward the internal representation, where the URI belongs to the `Document` and `Location` holds only line and column positions.

This change is part of the URI rework (#825). That issue is about the URI currently being represented as a plain `String` rather than a dedicated URI type. Changing the type is a later step. The current step replaces `location.uri` with `document.uri`, and this PR adds `document.uri` for definitions.

### DIFF
--- a/rust/rubydex-sys/src/definition_api.rs
+++ b/rust/rubydex-sys/src/definition_api.rs
@@ -493,6 +493,25 @@ pub unsafe extern "C" fn rdx_definition_mixins(pointer: GraphPointer, definition
     })
 }
 
+/// Returns a pointer to the URI ID of the document a definition belongs to, or
+/// NULL if the definition cannot be found. Caller must free the returned pointer
+/// with `free_u64`.
+///
+/// # Safety
+/// - `pointer` must be a valid pointer previously returned by `rdx_graph_new`.
+/// - `definition_id` must be a valid definition id.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_definition_document(pointer: GraphPointer, definition_id: u64) -> *const u64 {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        if let Some(defn) = graph.definitions().get(&def_id) {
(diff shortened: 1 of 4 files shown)


rust/rubydex-sys/src/definition_api.rs
```diff
@@ -493,6 +493,27 @@ pub unsafe extern "C" fn rdx_definition_mixins(pointer: GraphPointer, definition
...
+///
+/// # Panics
+/// This function will panic if the definition cannot be found.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_definition_document(pointer: GraphPointer, definition_id: u64) -> *const u64 {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        let Some(defn) = graph.definitions().get(&def_id) else {
+            panic!("Definition not found: {definition_id:?}");
+        };
```
vinistock: I know this is already an established pattern in the codebase, but I think we should avoid panicking in the FFI when things are not found.

From the Ruby native extension, we know that the definition **must** exist or else there's some data corruption. But the FFI crate should be general and might be used by other consumers, so I think we should change the way we deal with these and avoid panicking completely.

How about we return a `ptr::null()` if the definition cannot be found and then we have the Ruby side raise a runtime error? That way, we move the crash to the component that exercises a fundamental assumption (that the definition already exists in the graph).

vinistock (approved):

Morriar (approved):

rust/rubydex-sys/src/definition_api.rs
alexcrocha: That makes sense, I amended the commit with your suggestion.

> I know this is already an established pattern in the codebase, but I think we should avoid panicking in the FFI when things are not found.

Let's change this pattern. I'll push an update as a follow up

PR #923, opened 2026-07-13
Emit constant references for RBS parent-scope segments

soutaro:
## Summary

The RBS indexer created a resolution unit only for the final segment of a
qualified name. Parent scopes (e.g. `Minitest` in `Minitest::Guard`) were
interned as names but had no resolution unit of their own, so they were never
resolved. Qualified mixin and superclass references in RBS therefore stayed
unresolved, and the affected ancestor chains stayed permanently partial — every
class using such a reference (directly or by inheritance) was re-linearized on
every resolution pass.

This mirrors what the Ruby indexer already does (`index_constant_reference`):
emit a constant reference for each parent-scope segment so each one gets its own
resolution unit.

## Changes

- `index_type_name` now adds a constant reference per parent-scope segment.
- Add an RBS resolution regression test.

## Impact

Beyond correctness, this removes a large class of references that could never
resolve, which cuts the work each incremental resolution pass repeats on
codebases that load RBS signatures.

### DIFF
--- a/rust/rubydex/src/resolution_tests.rs
+++ b/rust/rubydex/src/resolution_tests.rs
@@ -4883,6 +4883,51 @@ mod rbs_tests {
     }
 
     #[test]
+    fn rbs_qualified_mixin_resolves_through_parent_scope() {
+        let mut context = graph_test();
+        context.index_rbs_uri("file:///minitest.rbs", {
+            r"
+            module Minitest
+              module Guard
+              end
+
+              module Assertions
+              end
+
+              class Runnable
+              end
+            end
+            "
+        });
+        context.index_rbs_uri("file:///minitest/test.rbs", {
+            r"
+            class Minitest::Test < ::Minitest::Runnable
+              include Minitest::Assertions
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution_tests.rs
```diff
@@ -4882,6 +4882,51 @@ mod rbs_tests {
...
+              end
+            end
+            "
+        });
+        context.index_rbs_uri("file:///minitest/test.rbs", {
+            r"
+            class Minitest::Test < ::Minitest::Runnable
+              include Minitest::Assertions
+              include Minitest::Guard
+              extend Minitest::Guard
```
vinistock, line 4907: We're not asserting anything related to this extent. Can we add an assert?

vinistock (approved):

rust/rubydex/src/resolution_tests.rs
soutaro: Added another `assert_ancestors_eq!` for the singleton class.

PR #924, opened 2026-07-13
Include dependencies in graph memory bench

vinistock:
I just realized that all of our benchmarks don't match the exact usage of Rubydex in real codebases. Because they are completely just benchmarking the Rust analysis, we miss RBS core definitions and dependencies.

That means that fundamental classes like `Object` and `BasicObject` are missing and so are all definitions that belong to gems. This means we never succeed in fully linearizing ancestor chains since the top of the chain is always missing and that we fail to resolve constants more often than we would in reality.

Those aspects can skew benchmark results and lead to sub-optimal decisions. For example, we're not interested in optimizing the analysis for the case where `Object` is missing, since we never want it to be missing.

This PR changes the `graph_memory` benchmark to include RBS core definitions and dependencies, so that the graph we produce is more realistic. I also added a super simplistic timer, so that we can see the execution time in the same run. Finally, I started loading the target project's configuration, so that the benchmark is closer to reality.

**Caveat**: in the bash script, we need to run some sub-commands with `bash -lc`. This is because we need those commands to login to the user's shell and load the version manager, otherwise we miss the right environment variables to find dependencies.

### DIFF
--- a/rust/rubydex/benches/graph_memory.rs
+++ b/rust/rubydex/benches/graph_memory.rs
@@ -1,6 +1,6 @@
 #[cfg(all(feature = "jemalloc", not(target_os = "windows")))]
 mod imp {
-    use std::collections::HashSet;
+    use std::time::Instant;
 
     use rubydex::{
         indexing::{self, IndexerBackend},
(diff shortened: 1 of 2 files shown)


utils/bench-graph-memory
```diff
@@ -33,8 +33,23 @@ case "$TARGET" in
...
+    exit 1
+fi
+
+# Collect RBS and dependency paths, so that the graph we're using to benchmark has complete ancestors. Otherwise, we
+# benchmark with missing declarations, like `Object`, which leads to ancestor chains never being completed and several
+# constant references never resolving.
+INDEX_PATHS=()
+while IFS= read -r line; do
+    [ -e "$line" ] && INDEX_PATHS+=("$line")
+done < <(/bin/bash -lc 'cd "$1" && bundle exec ruby -e "require \"rubydex\"; puts Rubydex::Graph.new.workspace_paths" 2>/dev/null' bash "$TARGET")
```
st0012, line 49: Can we print how many paths are added, as a simple way to check if there's potential issues?

rust/rubydex/benches/graph_memory.rs
```diff
@@ -18,13 +18,31 @@ mod imp {
...
 
     pub fn run() {
-        let paths: Vec<String> = std::env::args().skip(1).collect();
-        let paths = if paths.is_empty() { vec![".".to_string()] } else { paths };
-        let (file_paths, _) = listing::collect_file_paths(paths, &HashSet::new());
+        let mut args = std::env::args().skip(1);
+
+        let workspace = args
+            .next()
+            .expect("usage: cargo bench --bench graph_memory -- <workspace path> [path to index...]");
```
st0012: Should anyone ever use this command directly, instead of using `utils/bench-graph-memory`?

utils/bench-graph-memory
vinistock: Good idea.

rust/rubydex/benches/graph_memory.rs
vinistock: Honestly, I don't think so. You can, but then you need to list all paths manually.

st0012: Then I'd say it's better to include this context here. Something like: "Incorrect usage of `cargo bench --bench graph_memory`. Please use `utils/bench-graph-memory` for memory benchmark instead of invoking this command directly"

st0012 (approved):

PR #926, opened 2026-07-13
Reduce MCP server integration tests

st0012:
We do not need that many integration tests for the MCP server. Most behavior can be tested directly at the tool or server level, and a small subprocess smoke test is enough to cover the full path.

- Move the subprocess tests into `test/integration` and reduce the end-to-end server case to a small smoke test.
- Test each MCP tool directly against an indexed graph instead of routing every case through `Server#handle`.
- Exclude `test/integration/**/*_test.rb` from `ruby_test:valgrind`. Keeping the integration tests in one place lets memory checks ignore them entirely instead of starting child processes under Valgrind and skipping at runtime.
- Keep the stderr reader tolerant of teardown closing the pipe while it is blocked, since the subprocess exit status is still checked separately.
- Allow the Memleak workflow to be triggered manually when we need to debug it.


test/mcp_server_test.rb
```diff
@@ -368,7 +368,7 @@ def test_mcp_server_can_be_required_directly
   end
 
   def test_mcp_server_e2e
-    skip("This test times out when running with Valgrind") if ENV["RUBY_MEMCHECK_RUNNING"]
+    skip("This external process e2e times out when ruby_memcheck traces child processes under Valgrind") if ENV["RUBY_MEMCHECK_RUNNING"]
```
KaanOzkan: Unless I'm misunderstanding
```suggestion
    skip("This external process e2e test times out when ruby_memcheck traces child processes under Valgrind") if ENV["RUBY_MEMCHECK_RUNNING"]
```

KaanOzkan (approved):

vinistock (commented):
Because the memory leak checks are expensive and slow, maybe we shouldn't include integration tests at all in the Rakefile definition instead of skipping.

That said, similar to my previous feedback about the MCP tests, the majority of tests should be unit tests with only a few integration ones.

rafaelfranca (approved):

st0012:
@vinistock I reduced the number of true integration tests and moved the integration test file out of the memcheck task.

test/integration/mcp_server_test.rb
```diff
@@ -0,0 +1,223 @@
...
+    MAX_INDEXING_RETRIES.times do
+      parsed = call_tool(stdin, stdout, request_id, "codebase_stats", {})
+      return [parsed, request_id] unless parsed.key?("error")
+
+      assert_equal("indexing", parsed.fetch("error"))
+      request_id += 1
+      sleep(0.05)
+    end
+
+    flunk("Timed out waiting for indexing to complete")
```
vinistock, line 221: We can start out like this, but we need a better signal that indexing finished. Does the protocol support any form of notification from server to client that we can use to indicate completion?

vinistock (approved):

test/integration/mcp_server_test.rb
st0012: We may be able to implement this [progress flow](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress). But it's another feature to improve user (agent?) experience. So I'll leave open an issue for it and leave it to another PR.

https://github.com/Shopify/rubydex/issues/930

PR #927, opened 2026-07-13
Return NULL for missing entities in required FFI accessors

alexcrocha:
This PR follows up on a review comment from https://github.com/Shopify/rubydex/pull/920#discussion_r3559481317.

A Ruby object can outlive the graph entity it references. Accessing that stale object caused Rust to panic inside `extern "C"`. Since Rust cannot unwind through C, that aborted the Ruby process.

The affected FFI accessors now return `NULL` when the entity is missing. Ruby turns that result into an entity-specific `RuntimeError`.

This applies to definition comments and locations, constant and method reference names and locations, and declaration owners. Accessors where `NULL` or `false` already has a valid meaning need a different result type, so they remain follow-up work.

### DIFF
--- a/ext/rubydex/declaration.c
+++ b/ext/rubydex/declaration.c
@@ -236,7 +236,7 @@ static VALUE rdxr_declaration_owner(VALUE self) {
     const CDeclaration *decl = rdx_declaration_owner(graph, data->id);
 
     if (decl == NULL) {
-        rb_raise(rb_eRuntimeError, "owner can never be nil for any declarations");
+        rb_raise(rb_eRuntimeError, "Declaration must exist for a valid id");
     }
 
     VALUE decl_class = rdxi_declaration_class_for_kind(decl->kind);
--- a/rust/rubydex-sys/src/declaration_api.rs
+++ b/rust/rubydex-sys/src/declaration_api.rs
@@ -291,7 +291,8 @@ pub unsafe extern "C" fn rdx_declaration_singleton_class(pointer: GraphPointer, 
     })
 }
 
-/// Returns the owner of the declaration (attached object in the case of singleton classes)
+/// Returns the owner of the declaration (attached object in the case of singleton classes), or NULL if the declaration
+/// cannot be found.
 ///
 /// # Safety
 ///
@@ -299,11 +300,13 @@ pub unsafe extern "C" fn rdx_declaration_singleton_class(pointer: GraphPointer, 
 ///
 /// # Panics
 ///
-/// Will panic if invoked on a non-existing or non-namespace declaration
+/// This function will panic if the owner declaration cannot be found.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn rdx_declaration_owner(pointer: GraphPointer, decl_id: u64) -> *const CDeclaration {
     with_graph(pointer, |graph| {
-        let declaration = graph.declarations().get(&DeclarationId::new(decl_id)).unwrap();
+        let Some(declaration) = graph.declarations().get(&DeclarationId::new(decl_id)) else {
+            return ptr::null();
+        };
         let owner_id = *declaration.owner_id();
         Box::into_raw(Box::new(CDeclaration::from_declaration(
             owner_id,
--- a/ext/rubydex/definition.c
+++ b/ext/rubydex/definition.c
@@ -89,6 +89,9 @@ static VALUE rdxr_definition_location(VALUE self) {
     void *graph = rdxi_graph_from_handle(self, &data);
 
     Location *loc = rdx_definition_location(graph, data->id);
+    if (loc == NULL) {
+        rb_raise(rb_eRuntimeError, "Definition must exist for a valid id");
+    }
     VALUE location = rdxi_build_location_value(loc);
     rdx_location_free(loc);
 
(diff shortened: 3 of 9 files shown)


rust/rubydex-sys/src/declaration_api.rs
```diff
@@ -291,19 +291,22 @@ pub unsafe extern "C" fn rdx_declaration_singleton_class(pointer: GraphPointer,
...
+/// cannot be found.
 ///
 /// # Safety
 ///
 /// Assumes pointer is valid
 ///
 /// # Panics
 ///
-/// Will panic if invoked on a non-existing or non-namespace declaration
+/// This function will panic if the owner declaration cannot be found.
```
st0012, line 303: Why do we still want to panic in this case? It's not clear to me how we decide whether to panic or not in cases like this 🤔

alexcrocha: Fair question, and one I also asked myself while writing this. The boundary I decided on was, we can't guarantee the data coming from Ruby still exists (could be stale), so we don't panic and let Ruby handle it. Otherwise, if we own the data, I left the panic in place.

ext/rubydex/declaration.c
```diff
@@ -236,7 +236,7 @@ static VALUE rdxr_declaration_owner(VALUE self) {
     const CDeclaration *decl = rdx_declaration_owner(graph, data->id);
 
     if (decl == NULL) {
-        rb_raise(rb_eRuntimeError, "owner can never be nil for any declarations");
+        rb_raise(rb_eRuntimeError, "Declaration not found");
```
vinistock: I think this message change makes us lose information. `Declaration not found` seems like something plausible that could actually happen, but that indicates corruption in our Ruby data model.

No declaration can exist without an owner in the graph. The issue is that the FFI has no way of knowing that the Ruby VM side is searching for a declaration ID that is guaranteed to exist.

I would keep the original message.

rust/rubydex-sys/src/declaration_api.rs
vinistock: A good mental model is this: 

- Incorrect usage of an FFI API should not panic and crash, but instead return a result that communicates to the consumer that something went wrong
- Corrupt graph data should panic. No declarations can exist without an owner. If we registered an ID as an owner and there's no declaration associated to the ID, then there's a very significant bug on the core logic that we need to fix immediately

vinistock (approved):
The changes look good to me, except for the Ruby error messages, like I explained in my other comment.

`Something not found` seems like a possible outcome that may happen. In reality, these situations indicate a bug that we need to fix and not feedback to the API consumer.

Let's please ensure that the messages communicate that these errors mean the error was on **our side** and not on how the consumer used the API.

rust/rubydex-sys/src/declaration_api.rs
st0012: But how is "Ruby objects outliving graph entity it references, so now this pointer can't find the underlying data" an incorrect usage of the consumer?

To be clear I agree that in this case Rubydex should not crash. But the rule is still not clear to me.

> Corrupt graph data should panic.

Not for this PR, but instead of directly panic, is it possible to run integrity check and dump more information?

vinistock: > But how is "Ruby objects outliving graph entity it references, so now this pointer can't find the underlying data" an incorrect usage of the consumer?

It is incorrect usage by the consumer. The FFI provides consumers with a boxed graph pointer that they own and are responsible for freeing. If the underlying data gets freed while the Ruby object still exists, the entire mistake is on the Ruby VM side. The FFI doesn't own any objects, it only connects the Rust core logic with objects that it transfers ownership to the consumer.

> Not for this PR, but instead of directly panic, is it possible to run integrity check and dump more information?

What information would the integrity check provide that the panic doesn't?

st0012: > The FFI doesn't own any objects, it only connects the Rust core logic with objects that it transfers ownership to the consumer.

I feel this can be documented in architecture doc too. I know it's not Rubydex specific, but the concept is important to setting the correct boundaries.

> What information would the integrity check provide that the panic doesn't?

I could be wrong, but I thought integrity check can flag multiple corruptions, which I assume is likely to happen when there's any corruption? If not, yeah it won't be more helpful.

vinistock: > I feel this can be documented in architecture doc too. I know it's not Rubydex specific, but the concept is important to setting the correct boundaries.

Agreed. However, I would create a separate documentation file for the FFI and avoid mixing it with the architecture and design of the analysis.

> I could be wrong...

I see what you mean. The idea would be not to report just the one corruption we identified, but whenever there's a corruption we would scan the entire graph to try to report as many possible issues to the user.

We could try. Two concerns that come to mind is that we would need to basically add an integrity check guard to every single FFI API, which may be a bit tedious. Also, it could be confusing for API consumers to see a bunch of data integrity issues unrelated to the API they actually called, so we would need a good explanation about why they are seeing it.

ext/rubydex/declaration.c
alexcrocha: That makes sense. I changed the message here to `Declaration must exist for a valid id`. We can't go back to the original since we're now returning a null pointer before getting the owner id.

```rust
let Some(declaration) = graph.declarations().get(&DeclarationId::new(decl_id)) else {
    return ptr::null();
};
let owner_id = *declaration.owner_id();
```

I applied the same change to the other Ruby error messages in this PR.

PR #928, opened 2026-07-14
Remove incremental CLI benchmark options

soutaro:
## Summary
- Remove the no-op incremental resolution CLI benchmark options.
- Drop the reindex/resolution timing path that was only used by those flags.

## Test plan
- cargo fmt --check
- cargo test -p rubydex --bin rubydex_cli
- cargo test -p rubydex --test cli

### DIFF
--- a/rust/rubydex/src/main.rs
+++ b/rust/rubydex/src/main.rs
@@ -1,13 +1,9 @@
 use clap::{Parser, ValueEnum};
-use std::{
-    fs, mem,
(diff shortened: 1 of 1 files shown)


soutaro:
https://github.com/Shopify/rubydex/pull/918 implemented an optimization for no-op edit for incremental resolution. The CLI option doesn't make any sense now.

st0012 (approved):

vinistock (approved):

PR #932, opened 2026-07-15
Add an incremental resolution benchmark

soutaro:
## Summary

This adds a benchmark for incremental resolution. It covers both the lightest case—a no-op resolve—and one of the heaviest incremental operations—an ancestor change.

## Ancestor changes

The benchmark defines two empty modules:

```ruby
module BenchmarkMixinA
end

module BenchmarkMixinB
end
```

It then repeatedly replaces a virtual source file between the following two versions:

```ruby
class ExampleService
  include ::BenchmarkMixinA
end
```

```ruby
class ExampleService
  include ::BenchmarkMixinB
end
```

Replacing the included module invalidates the target class's ancestor chain and triggers ancestor recalculation for its transitive descendants.

Target classes can be selected by descendant count, allowing light and heavy ancestor changes to be compared.

## Results

On a monolith containing approximately 1.6 million declarations:

| Scenario | Merge median | Resolution median |
| --- | ---: | ---: |
| No-op | — | 933 ms |
| 1 descendant | 0.02–0.03 ms | 600–800 ms |
| Approximately 10,500 descendants | 63–75 ms | 720–790 ms |

Merge time increases substantially with the number of descendants, although it remains small in absolute terms. Resolution time stays within approximately the same range regardless of descendant count.

This does not necessarily mean that ancestor recalculation takes constant time. After resolution, the graph still contained 37,040 pending work items:

- 34,092 constant references
- 2,935 ancestor calculations
- 13 definitions

Unresolved work is retained so that it can be retried after future edits. Consequently, even a no-op resolve revisits a large amount of pending work. This fixed cost dominates the additional ancestor recalculation and makes differences between targets difficult to observe in the overall resolution time.

## How to run

First, find suitable target classes:

```sh
utils/bench-incremental-resolution --candidates /path/to/workspace
```

Run the benchmark for one or more classes reported by that command:

```sh
utils/bench-incremental-resolution \
  --target LightTarget \
  --target HeavyTarget \
  /path/to/workspace
```

Inspect the unresolved and pending graph state:

```sh
utils/bench-incremental-resolution \
  --resolution-stats \
  /path/to/workspace
```

The workspace can also be configured once through `DEFAULT_BENCH_WORKSPACE`:

```sh
export DEFAULT_BENCH_WORKSPACE=/path/to/workspace

utils/bench-incremental-resolution --candidates
utils/bench-incremental-resolution --target HeavyTarget
utils/bench-incremental-resolution --resolution-stats
```

### DIFF
--- a/rust/rubydex/benches/incremental_resolution.rs
+++ b/rust/rubydex/benches/incremental_resolution.rs
@@ -1,0 +1,674 @@
+use std::{
+    collections::HashSet,
+    error::Error,
+    io,
+    path::PathBuf,
+    process::ExitCode,
+    time::{Duration, Instant},
+};
+
+use clap::Parser;
+use rubydex::{
+    indexing::{self, IndexerBackend, LanguageId},
+    integrity, listing,
+    model::{
+        declaration::{Ancestor, Ancestors, Namespace},
+        graph::{Graph, Unit},
+        ids::{ConstantReferenceId, DeclarationId, DefinitionId},
+        name::NameRef,
+    },
+    resolution::Resolver,
+};
+
+const MIXIN_A: &str = "RubydexIncrementalBenchmarkMixinA";
+const MIXIN_B: &str = "RubydexIncrementalBenchmarkMixinB";
+const MIXIN_URI: &str = "file:///__rubydex_incremental_benchmark__/mixins.rb";
+
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+
+#[derive(Debug, Parser)]
+#[command(about = "Benchmark incremental resolution against an existing workspace graph")]
+struct Args {
+    /// Added by `cargo bench` for harness-less benchmarks.
+    #[arg(long = "bench", hide = true)]
+    _bench: bool,
+
+    /// Workspace root. Used for rubydex.toml and reporting.
+    workspace: PathBuf,
+
+    /// File or directory to index. May be repeated. Defaults to the workspace root.
+    #[arg(long = "index-path")]
+    index_paths: Vec<PathBuf>,
+
+    /// Top-level class name to benchmark. May be repeated.
+    #[arg(long = "target")]
+    targets: Vec<String>,
+
+    /// Print light/typical/heavy class candidates and exit unless --target is also supplied.
+    #[arg(long)]
+    list_candidates: bool,
+
+    /// Print unresolved graph state after the initial resolution and exit.
+    #[arg(long)]
+    resolution_stats: bool,
+
+    /// Number of unmeasured A -> B -> A warmup cycles.
+    #[arg(long, default_value_t = 5)]
+    warmup: usize,
+
+    /// Number of measured A -> B -> A cycles.
+    #[arg(long, default_value_t = 30)]
+    cycles: usize,
+}
+
+#[derive(Clone)]
+struct Candidate {
+    id: DeclarationId,
+    name: String,
+    descendants: usize,
+    ancestor_depth: usize,
+    definitions: usize,
+    direct_references: usize,
+}
+
+#[derive(Clone, Copy)]
+struct Sample {
+    build: Duration,
+    merge: Duration,
+    resolve: Duration,
+    total: Duration,
+}
+
+struct PendingWorkSummary {
+    total: usize,
+    raw_definitions: usize,
+    raw_constant_references: usize,
+    raw_ancestors: usize,
+    definitions: HashSet<DefinitionId>,
+    constant_references: HashSet<ConstantReferenceId>,
+    ancestors: HashSet<DeclarationId>,
+}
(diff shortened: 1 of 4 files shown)


rust/rubydex/benches/incremental_resolution.rs
```diff
@@ -0,0 +1,674 @@
...
+
+    Sample {
+        build,
+        merge,
+        resolve,
+        total: total_started.elapsed(),
+    }
+}
+
+fn assert_mixin(graph: &Graph, target: &str, target_id: DeclarationId, expected: &str, absent: &str) -> Result<()> {
```
st0012: Should we also check descendants are rebuild successfully here?

rust/rubydex/benches/incremental_resolution.rs
```diff
@@ -0,0 +1,674 @@
...
+
+    let uri = format!("file:///__rubydex_incremental_benchmark__/target_{index}.rb");
+    let source_a = format!("class {target}\n  include ::{MIXIN_A}\nend\n");
+    let source_b = format!("class {target}\n  include ::{MIXIN_B}\nend\n");
+    debug_assert_eq!(source_a.len(), source_b.len());
+
+    apply_source(graph, &uri, &source_a);
+    assert_mixin(graph, target, target_id, MIXIN_A, MIXIN_B)?;
+
+    for _ in 0..warmup {
```
st0012: What's warmup for? Like, what step could produce noise/inaccurate data if we don't warmup here?

rust/rubydex/benches/incremental_resolution.rs
```diff
@@ -0,0 +1,674 @@
...
+        )));
+    }
+
+    Ok(())
+}
+
+fn apply_source(graph: &mut Graph, uri: &str, source: &str) -> Sample {
+    let total_started = Instant::now();
+
+    let build_started = Instant::now();
```
st0012: Looks like this is to measure the indexing time. If so, can we just call it `index` instead of `build`?

rust/rubydex/benches/incremental_resolution.rs
soutaro: Added a check if descendants are rebuild successfully.

rust/rubydex/benches/incremental_resolution.rs
soutaro: Good point. I removed the warmup because this path doesn’t involve disk I/O, GC, or JIT compilation.

rust/rubydex/benches/incremental_resolution.rs
```diff
@@ -0,0 +1,686 @@
...
+fn subtree_reference_count(graph: &Graph, candidate: &Candidate) -> usize {
+    let Some(target) = graph
+        .declarations()
+        .get(&candidate.id)
+        .and_then(|declaration| declaration.as_namespace())
+    else {
+        return 0;
+    };
+
+    std::iter::once(&candidate.id)
```
Morriar: I think `descendants()` already includes the namespace itself. For example, in a tiny `Foo < Bar < Baz` workspace, a leaf candidate reports `descendants=1`, so adding `once(&candidate.id)` here double-counts the target's direct refs in `subtree_refs`.

Should we rely on `target.descendants()` as the affected set here, or rename the printed column if we intentionally want the self-inclusive count?

st0012 (commented):
I pair-reviewed this PR with @vinistock and I'll try my best to summarize our feedback:

- Can you provide a diagram to explain how the benchmark process works? Like why would it create pending work?
- Each benchmark should be self-contained and doesn't involve external workspace (like ruby-bench's benchmarks)
  - We can then print less information about the benchmark code's resolution as the code itself would be predictable
- We shouldn't assert things in benchmark. Having assertions make benchmark creation and related logic much more complicated
- Ideally we want to have a framework to build MANY small benchmark scenarios, instead of a benchmark that only runs a few of them
  - To achieve this, we can have a corpus to provide basic setup, like:

```rb
class ClassWithoutMixin; end
class IncludedModule; end
# class_with_mixin.rb
class ClassWithMixin
  include IncludedModule
end
# ...
```

And then we use macros to create each scenarios with ONLY 1 step:

```rust
incremental_scenario!(
  "Remove included module from a class",
  {
    "file:///class_with_mixin.rb" => "class ClassWithMixin; end",
  }
)
// Each scenario runs 10 times after warmup and we take the average
```

And the script could be: `rubydex_resolution_bench <target_sha|branch> [<optional_base>] -- <args passed to Rust>` 
For args to Rust, we should allow specifying scenarios:
- `--scenario all` (default)
- `--scenario <name1> <name2>`

Both Vini and I will be happy to pair to explain the idea more!

soutaro:
Closed because we are not working for incremental resolution right now.

PR #933, opened 2026-07-15
Improve superclass retrieval API

st0012:
- Let's call it `get_superclass` instead, to match Ruby's `superclass` API name
- `get_superclass` should handle `Object` and `BasicObject`'s superclass correctly
- Caller of `get_superclass` should not need to pass definitions separately

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1067,27 +1067,25 @@ impl<'a> Resolver<'a> {
         declaration_id: DeclarationId,
         context: &mut LinearizationContext,
     ) -> Option<Vec<Ancestor>> {
-        if declaration_id == *BASIC_OBJECT_ID {
-            return None;
-        }
-
         let declaration = self.graph.declarations().get(&declaration_id).unwrap();
 
         match declaration {
             Declaration::Namespace(Namespace::Class(_)) => {
                 let definition_ids = declaration.definitions().to_vec();
 
-                Some(match self.linearize_parent_class(&definition_ids, context) {
-                    Ancestors::Complete(ids) => ids,
-                    Ancestors::Cyclic(ids) => {
-                        context.cyclic = true;
-                        ids
-                    }
-                    Ancestors::Partial(ids) => {
-                        context.partial = true;
-                        ids
-                    }
-                })
+                Some(
+                    match self.linearize_parent_class(declaration_id, &definition_ids, context)? {
+                        Ancestors::Complete(ids) => ids,
+                        Ancestors::Cyclic(ids) => {
+                            context.cyclic = true;
+                            ids
+                        }
+                        Ancestors::Partial(ids) => {
+                            context.partial = true;
+                            ids
+                        }
+                    },
+                )
             }
             Declaration::Namespace(Namespace::SingletonClass(_)) => {
                 let owner_id = *declaration.owner_id();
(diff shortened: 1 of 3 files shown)


vinistock (commented):
I'm not sure I understand the motivation of this change. Don't we already handle object and basic object superclasses correctly in main?

st0012:
I was reviewing https://github.com/Shopify/rubydex/pull/893 and noticed that `get_parent_class` didn't handle the `BasicObject` case correctly when reading [here](https://github.com/Shopify/rubydex/pull/893/changes#diff-2c1da190ea79ca7c9b449ed8efe95bb004edd2618254fbcc08aafc3f7fe051f9R221). I know behaviour-wise it's handled in `linearize_parent_ancestors`, but IMO `get_parent_class` should already handle that case.

rust/rubydex/src/resolution_tests.rs
```diff
@@ -1944,6 +1944,32 @@ mod mixin_dedup_tests {
 mod object_ancestors_tests {
     use super::*;
 
+    #[test]
+    fn core_class_parent_hierarchy() {
```
vinistock: Do we need this test? I think we already have examples covering everything asserted here.

rust/rubydex/src/resolution_tests.rs
```diff
@@ -2164,6 +2190,31 @@ mod singleton_ancestors_tests {
...
+            "BasicObject::<BasicObject>",
+            [
+                "BasicObject::<BasicObject>",
+                "Class",
+                "Module",
+                "Object",
+                "Kernel",
+                "BasicObject"
+            ]
+        );
```
vinistock: Similarly here, these assertions are duplicate. The assertion below can only succeed if the linearization of `Object::<Object>` and `BasicObject::<BasicObject>` is correct, so we're basically asserting the same thing 3 times.

rust/rubydex/src/resolution.rs
```diff
@@ -2060,11 +2050,18 @@ impl<'a> Resolver<'a> {
...
-        let mut explicit_parents = Vec::new();
-        let mut unresolved_parent = None;
+    /// Returns the selected superclass and any unresolved explicit superclass. The top-level `BasicObject` declaration
+    /// is Ruby's root class and is the only class without a superclass.
+    fn get_superclass(&self, declaration_id: DeclarationId) -> Option<(DeclarationId, Option<NameId>)> {
+        if declaration_id == *BASIC_OBJECT_ID {
+            return None;
+        }
+
+        let declaration = self.graph.declarations().get(&declaration_id).unwrap();
```
vinistock, line 2060: The original intent of passing the definition IDs was to avoid looking up the same declaration we already have in `linearize_ancestors` a second time.

Lookups are generally fast, but if this is purely cosmetic, I would vote for passing both the `declaration_id` and `definition_ids` to avoid the second hash lookup.

vinistock (commented):
The refactor is good.

Did you benchmark to see if this provides any speed ups?

rust/rubydex/src/resolution.rs
st0012: I ran benchmark and there's no difference (improvement/regression) compared to `main`. 

Metric | Base average | Current average | Difference
-- | -- | -- | --
Resolution | 17.732s | 17.069s | 0.663s faster
Total | 26.875s | 26.072s | 0.803s faster
Execution | 27.330s | 26.498s | 0.832s faster

So I'd actually vote for the opposite: since reusing the same definitions list doesn't improve performance, let's make the method more intuitive to use by using the declaration_id as the only parameter instead.

vinistock: Fair.

vinistock (approved):

PR #934, opened 2026-07-16
Mark RBS constants as promotable

soutaro:
Resolving `ENV.fetch` left the synthetic `ENV::<ENV>` namespace unresolved. This occurs only for constants declared in RBS because they are not promotable.

This PR marks RBS constants as *promotable*, like dynamically assigned Ruby constants, so the namespace can be resolved.

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -495,7 +495,10 @@ impl Visit for RBSIndexer<'_> {
             self.uri_id,
             offset,
             comments,
-            Self::flags(&constant_node.annotations()),
+            // RBS establishes that the constant exists, but its value type is intentionally not
+            // represented in the graph. Treat it like a dynamic Ruby assignment so resolution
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -495,7 +495,10 @@ impl Visit for RBSIndexer<'_> {
             self.uri_id,
             offset,
             comments,
-            Self::flags(&constant_node.annotations()),
+            // RBS establishes that the constant exists, but its value type is intentionally not
+            // represented in the graph. Treat it like a dynamic Ruby assignment so resolution
+            // may promote it when a namespace or singleton receiver is required.
+            Self::flags(&constant_node.annotations()) | DefinitionFlags::PROMOTABLE,
```
soutaro, line 501: We may want to keep the constant non-promotable when it has literal types.
Note that we cannot make Ruby literal constants, like `Integer`, non-promotable, because the RBS type names are not resolved yet here.

st0012 (approved):

vinistock (approved):

PR #935, opened 2026-07-16
Preserve singleton hierarchies when concretizing Todos

soutaro:
When a `Todo` namespace is later replaced by an explicit `class` or `module`, preserve its singleton hierarchy and relinearize ancestor chains.

This covers the direct singleton class and any higher-order singleton classes.

### DIFF
--- a/rust/rubydex/src/model/graph.rs
+++ b/rust/rubydex/src/model/graph.rs
@@ -206,7 +206,15 @@ impl Graph {
                 if should_promote {
                     let mut new_declaration = constructor(fully_qualified_name);
                     let removed_declaration = occupied_entry.remove();
-                    new_declaration.as_namespace_mut().unwrap().extend(removed_declaration);
+                    let singleton_class_id = removed_declaration
+                        .as_namespace()
+                        .and_then(Namespace::singleton_class)
+                        .copied();
+                    let new_namespace = new_declaration.as_namespace_mut().unwrap();
+                    new_namespace.extend(removed_declaration);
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/model/declaration.rs
```diff
@@ -134,6 +134,7 @@ macro_rules! namespace_declaration {
                     Declaration::Namespace(namespace) => {
                         self.members.extend(namespace.members());
                         self.references.extend(namespace.references());
+                        self.singleton_class_id = namespace.singleton_class().copied();
```
st0012: I think singleton class reassignment doesn't really fit in `extend`. Unlike other extending operations, we actually lose the original value here. Why do we need this change and can we do it in another location?

st0012 (commented):
The behaviour makes sense to me. But the way it fills singleton class id doesn't look safe to me.

rust/rubydex/src/model/declaration.rs
soutaro: Thanks, you're right. The singleton class ID needs to be retained because promotion replaces the existing `Todo` declaration with a new class or module declaration. If the `Todo` already has a singleton class, losing the ID would prevent us from relinearizing that singleton hierarchy.

However, `extend` is the wrong place for that transfer. I moved it into the promotion path in `Graph::add_declaration`, where the `Todo` is explicitly replaced with the concrete namespace.

Morriar (approved):

PR #937, opened 2026-07-16
Propagate singleton class creation to descendants

vinistock:
## Background

We were creating broken descendant chains for singleton classes. The essence of the issue is that when a parent singleton class was created, it did not propagate the creation of singletons to its descendants, resulting in missing data in the graph and non-sensical results.

For example:

```ruby
class Parent
  # Singleton class created because it is explicitly opened here
  def self.foo; end
end

# This class has no explicit opening of its singleton. Because we didn't
# propagate, it would end up never being created
class Intermediate < Parent; end

class Child < Intermediate; end
```

That would result in `graph["Parent"].descendants` to return `Intermediate` and `Child` (correct), while `graph["Parent::<Parent>"].descendants` would return nothing, which is incorrect and inconsistent.

## The fix

When we are linearizing singleton ancestors, we need to figure out what is the parent class of the singleton, for which we have a dedicated method `singleton_parent_id`. That method recurses and determines the parent class based on the entity we are looking at. Eventually, we reach the ultimate attached object, read its parent class and use it as the ancestor.

That is the point where we need to grab the attached descendants and ensure that we're creating singletons for all of them, matching all depth levels so that `Parent::<Parent>::<<Parent>>` (and any other level) is also handled correctly and lazily (without creating unnecessary singleton classes).

The major gotcha that makes this change require context to apply is that simply propagating to descendants immediately results in infinite recursion. The reason is not very trivial to trace, but I tried to document it as best as I could in a comment (please let me know if it's not clear).

Finally, to ensure proper behaviour in incremental resolution scenarios, we also need a tiny modification to the parent ancestor linearization helper, so that when a class appears later, we correctly create all necessary singleton levels.

## Impact

Naturally, this change is not 100% free. We were "avoiding" work by building an inconsistent graph, which is faster, but not correct. That said, the impact is acceptable and we need the correctness (+ ~2s resolution + ~200MB graph). I'm sure we can shave this off with more optimizations.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1077,22 +1077,24 @@ impl<'a> Resolver<'a> {
             Declaration::Namespace(Namespace::Class(_)) => {
                 let definition_ids = declaration.definitions().to_vec();
 
-                Some(match self.linearize_parent_class(&definition_ids, context) {
-                    Ancestors::Complete(ids) => ids,
-                    Ancestors::Cyclic(ids) => {
-                        context.cyclic = true;
-                        ids
-                    }
-                    Ancestors::Partial(ids) => {
-                        context.partial = true;
-                        ids
-                    }
-                })
+                Some(
+                    match self.linearize_parent_class(declaration_id, &definition_ids, context) {
+                        Ancestors::Complete(ids) => ids,
+                        Ancestors::Cyclic(ids) => {
+                            context.cyclic = true;
+                            ids
+                        }
+                        Ancestors::Partial(ids) => {
+                            context.partial = true;
+                            ids
+                        }
+                    },
+                )
             }
             Declaration::Namespace(Namespace::SingletonClass(_)) => {
                 let owner_id = *declaration.owner_id();
 
-                let (singleton_parent_id, partial_singleton) = self.singleton_parent_id(owner_id);
+                let (singleton_parent_id, partial_singleton) = self.singleton_parent_id(owner_id, 1);
                 if partial_singleton {
                     context.partial = true;
                 }
@@ -2019,7 +2021,7 @@ impl<'a> Resolver<'a> {
     /// - Module: parent is the `Module` class
     /// - Class: parent is the singleton class of the original parent class
     /// - Singleton class: recurse as many times as necessary to wrap the original attached object's parent class
-    fn singleton_parent_id(&mut self, attached_id: DeclarationId) -> (DeclarationId, bool) {
+    fn singleton_parent_id(&mut self, attached_id: DeclarationId, depth: u16) -> (DeclarationId, bool) {
         // Base case: if we reached `BasicObject`, then the parent is `Class`
         if attached_id == *BASIC_OBJECT_ID {
             return (*CLASS_ID, false);
@@ -2034,7 +2036,7 @@ impl<'a> Resolver<'a> {
                 // object
                 let owner_id = *decl.owner_id();
 
-                let (inner_parent, partial) = self.singleton_parent_id(owner_id);
+                let (inner_parent, partial) = self.singleton_parent_id(owner_id, depth + 1);
                 (
                     self.get_or_create_singleton_class(inner_parent, SingletonAncestors::Deferred)
                         .expect("singleton parent should always be a namespace"),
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -2045,6 +2047,51 @@ impl<'a> Resolver<'a> {
...
+                    let descendants = decl
+                        .as_namespace()
+                        .unwrap()
+                        .descendants()
+                        .iter()
+                        .copied()
+                        .collect::<Vec<_>>();
+
+                    for descendant in descendants {
+                        if descendant == attached_id {
```
vinistock, line 2081: Add comment here explaining

rust/rubydex/src/resolution.rs
```diff
@@ -2045,6 +2047,51 @@ impl<'a> Resolver<'a> {
...
+                            continue;
+                        }
+
+                        let mut needs_singleton_id = descendant;
+
+                        for _ in 0..depth {
```
vinistock, line 2087: I think I can remove the `depth` by leaning more correctly into the recursion

rust/rubydex/src/resolution_tests.rs
vinistock, line 1: Add test case for switching parent classes.

soutaro:
@vinistock This looks good overall, but the reported `+200 MB` increase in graph size feels unexpectedly large. I think we should validate it against the number of additional singleton-class declarations. It may ultimately be acceptable, but I’m concerned the increase could be disproportionate to the additional graph structure required.

For reference, my previous PR (#893), which I believe should produce a broadly comparable graph shape, added `17,858` declarations for a `+25 MB` increase.

rust/rubydex/src/resolution_tests.rs
```diff
@@ -1185,6 +1185,292 @@ mod superclass_tests {
...
+
+        assert_declaration_exists!(context, "Bar::<Bar>");
+        assert_declaration_exists!(context, "Qux::<Qux>");
+
+        context.index_uri("file:///sub.rb", "class NewSub < Bar; end");
+        context.resolve();
+
+        assert_declaration_exists!(context, "NewSub::<NewSub>");
+        assert_descendants!(context, "Bar::<Bar>", ["NewSub::<NewSub>"]);
+    }
```
Morriar, line 1323:
```rb
# resolution 1

class Bar
  def self.foo; end
end

class Other
  def self.foo; end
end

class Child < Bar; end
```

```rb
# resolution 2

class Child < Other; end
```

Do we update the parent of `Child::<Child>`?

rust/rubydex/src/resolution.rs
```diff
@@ -2045,6 +2047,51 @@ impl<'a> Resolver<'a> {
...
+                        .as_namespace()
+                        .unwrap()
+                        .descendants()
+                        .iter()
+                        .copied()
+                        .collect::<Vec<_>>();
+
+                    for descendant in descendants {
+                        if descendant == attached_id {
+                            continue;
```
Morriar, line 2082: Can we add a test like this:

```
# First resolution
class Child; end

# Later resolution
def Object.foo; end
```

I'm not sure we're properly updating the `Child::<Child>` parent.

rust/rubydex/src/resolution.rs
```diff
@@ -2096,12 +2143,15 @@ impl<'a> Resolver<'a> {
...
     fn linearize_parent_class(
         &mut self,
+        declaration_id: DeclarationId,
         definition_ids: &[DefinitionId],
         context: &mut LinearizationContext,
     ) -> Ancestors {
         let (picked_parent, unresolved_parent) = self.get_parent_class(definition_ids);
         let mut result = self.linearize_ancestors(picked_parent, context);
 
+        self.ensure_matching_singleton_class_depth(declaration_id, picked_parent);
```
Morriar, line 2153: Can you try this:

```rb
# resolution 1

# parents.rb
class OldParent
  def self.old_method; end
end

class NewParent
  def self.new_method; end
end

# child_a.rb
class Child < OldParent; end

# child_b.rb
class Child < OldParent; end
```

then

```rb
# resolution 2

# child_a.rb
class Child < NewParent; end

# child_b.rb
class Child < NewParent; end
```

Is the parent of `Child::<Child>` updated?

rust/rubydex/src/resolution.rs
```diff
@@ -2045,6 +2047,51 @@ impl<'a> Resolver<'a> {
...
+                if !class_ancestors.contains(attached_id) {
+                    let descendants = decl
+                        .as_namespace()
+                        .unwrap()
+                        .descendants()
+                        .iter()
+                        .copied()
+                        .collect::<Vec<_>>();
+
+                    for descendant in descendants {
```
Morriar, line 2080: Since `descendants` is already transitive, aren't we repeating a lot of work here? Creating the singleton for `A` visits all of its descendants, and then each singleton we enqueue does another overlapping scan when it is linearized. For a deep hierarchy, this looks quadratic.

Could we track the singleton depth already propagated for each declaration during this resolution and skip these repeated scans?

PR #941, opened 2026-07-17
Make Graph Ractor-shareable via freeze

dersam:
## Summary

`Ractor.make_shareable(@graph)` raises `Ractor::Error` because `Rubydex::Graph` is a TypedData object whose `rb_data_type_t` lacks `RUBY_TYPED_FROZEN_SHAREABLE`. This PR makes a resolved graph safe to share across Ractors without copying, while preserving the ability to mutate the graph for interactive use cases (LSP/MCP incremental edits).

## Approach

The graph is wrapped in `Arc<RwLock<Graph>>` at the FFI boundary. The `RwLock` provides thread-safe concurrent access: read operations acquire a read lock, write operations acquire a write lock. `Ractor.make_shareable` "freezes" the Ruby object as part of making it shareable (this doesn't change the underlying Rust structure - we need to do the ceremonial freeze so that the object is not copied for the Ractor). Thread safety comes entirely from the `RwLock` on the Rust side - mutable graphs can still be acquired.

This covers all use cases:
- **Batch/read-only**: index once, share across Ractors for parallel queries.
- **Interactive**: keep the graph mutable for incremental edits while still allowing parallel reads from worker Ractors. Ordering guarantees are the consumer's responsibility, as with any shared mutable state across Ractors.

## Usage

```ruby
# Batch/read-only
graph = Rubydex::Graph.new(workspace_path: path)
graph.index_workspace
graph.resolve
Ractor.make_shareable(graph)
# Safe to share into worker Ractors for parallel queries

# Interactive (LSP/MCP)
graph = Rubydex::Graph.new(workspace_path: path)
graph.index_workspace
graph.resolve
Ractor.make_shareable(graph)
# Worker Ractors can read; the main Ractor can still mutate
# (e.g. index_source + resolve for incremental edits)
```

### DIFF
--- a/rust/rubydex-sys/src/graph_api.rs
+++ b/rust/rubydex-sys/src/graph_api.rs
@@ -21,42 +21,45 @@ use rubydex::resolution::Resolver;
 use rubydex::{indexing, integrity, listing, query};
 use std::ffi::CString;
 use std::path::{Path, PathBuf};
-use std::{mem, ptr};
+use std::{mem, ptr, sync::RwLock};
 
 pub type GraphPointer = *mut c_void;
 
-/// Creates a new graph within a mutex. This is meant to be used when creating new Graph objects in Ruby
+/// Creates a new graph wrapped in a `Box<RwLock<Graph>>` for thread-safe shared access across Ractors.
 #[unsafe(no_mangle)]
 pub extern "C" fn rdx_graph_new() -> GraphPointer {
-    Box::into_raw(Box::new(Graph::new())) as GraphPointer
+    Box::into_raw(Box::new(RwLock::new(Graph::new()))) as GraphPointer
 }
 
 /// Frees a Graph through its pointer
 #[unsafe(no_mangle)]
 pub extern "C" fn rdx_graph_free(pointer: GraphPointer) {
     unsafe {
-        let _ = Box::from_raw(pointer.cast::<Graph>());
+        let _ = Box::from_raw(pointer.cast::<RwLock<Graph>>());
     }
 }
 
+/// Runs `action` against the graph referenced by `pointer` under a read lock.
+///
+/// # Panics
+///
+/// Panics if the `RwLock` is poisoned (a writer panicked while holding the lock).
 pub fn with_graph<F, T>(pointer: GraphPointer, action: F) -> T
 where
     F: FnOnce(&Graph) -> T,
 {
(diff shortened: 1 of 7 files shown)


rust/rubydex-sys/src/graph_api.rs
```diff
@@ -43,10 +43,8 @@ pub fn with_graph<F, T>(pointer: GraphPointer, action: F) -> T
 where
     F: FnOnce(&Graph) -> T,
 {
-    let mut graph = unsafe { Box::from_raw(pointer.cast::<Graph>()) };
-    let result = action(&mut graph);
-    mem::forget(graph);
-    result
+    let graph = unsafe { &*pointer.cast::<Graph>() };
```
Morriar: Since we're now allowing the same `Graph` pointer to be dereferenced by multiple Ractors, let's make the `Sync` invariant explicit in Rust. The compiler can't see the cross-Ractor sharing through the Ruby typed-data flag, so a future non-`Sync` field on `Graph` would keep compiling but make this unsafe block unsound.

Could we add a compile-time assertion around here?

```rust
const _: () = {
    fn assert_sync<T: Sync>() {}
    let _ = assert_sync::<Graph>;
};
```

vinistock (commented):
Thanks, this is great. A few things I just realized when reviewing:

1. I actually don't think we need an `Arc` here because we have a single owner (the Ruby VM object). Can you please try using `Box<RwLock<Graph>>`? If that works, let's use `Box` as it's cheaper
2. Let's implement `initialize_copy` to raise, mentioning that we intentionally don't allow cloning the graph as it can balloon the memory use
3. Let's add a note about Ractor safety in our readme. The reality is that we're leaning into Rust's safety and disrespecting the frozen semantics a bit (since a frozen graph can actually be mutated). It's intentional, but it might be surprising to users, so we should document it
4. Since we now require the `Graph` to always be shareable across threads, let's add a compile-time macro to enforce it (otherwise we could add fields that aren't `Send` and `Sync`)

```rust
// compile_assertions.rs
#[macro_export]
macro_rules! assert_send_sync {
    ($struct:ident) => {
        const _: fn() = || {
            fn assert_send_sync<T: ?Sized + Send + Sync>() {}
            assert_send_sync::<$struct>();
        };
    };
}

// graph.rs

struct Graph {
  // ...
}
assert_mem_size!(Graph, 352);
assert_send_sync!(Graph);
```

vinistock (approved):
Amazing!

PR #942, opened 2026-07-17
Add `Location#short_path`

vinistock:
A small helper to return a "short path" (or maybe short name?) for a URI. When it's a file URI, we use the path relative to the PWD. If it's an `untitled` URI, we use the opaque (e.g.: `untitled:Untitled-1` -> `Untitled-1`).

### DIFF
--- a/lib/rubydex.rb
+++ b/lib/rubydex.rb
@@ -2,6 +2,7 @@
 
 require "bundler"
 require "uri"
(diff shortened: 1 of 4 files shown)


vinistock:
Actually, this is not an adequate solution as the graph might be used from a location that isn't the pwd.

PR #944, opened 2026-07-17
Extract declaration core

vinistock:
To move forward with our constant representation re-modelling, in a way that makes promotion with namespace storage easier, we first need an easy way to share common data across all declarations. We initially started using macros, but it makes for pretty weird, hard to share code.

This PR extracts a `DeclarationCore<T>`, which holds all fields common to every declaration. In the next stage of the refactor, constants will become something like this:

```rust
pub struct Constant {
  // the core data
  core: DeclarationCore<ConstantReferenceId>,
  // optional namespace store for ancestors, descendants, singleton classes. We create this
  // lazily when needed (which is the new "promotion" path)
  namespace_store: Option<NamespaceStore>,
  // the type of the constant (an enum with ConstantAlias, Module, Class...)
  constant_type: ConstantType
}
```

### DIFF
--- a/rust/rubydex/src/model/declaration.rs
+++ b/rust/rubydex/src/model/declaration.rs
@@ -1,3 +1,5 @@
+use std::hash::Hash;
+
 use crate::assert_mem_size;
 use crate::model::ids::{
     ClassVariableReferenceId, GlobalVariableReferenceId, InstanceVariableReferenceId, MethodReferenceId,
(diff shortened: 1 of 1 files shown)


st0012 (approved):

rust/rubydex/src/model/declaration.rs
```diff
@@ -604,17 +594,23 @@ namespace_declaration!(SingletonClass, SingletonClassDeclaration);
...
+
+pub type InstanceVariableDeclaration = DeclarationCore<InstanceVariableReferenceId>;
 assert_mem_size!(InstanceVariableDeclaration, 80);
-simple_declaration!(ClassVariableDeclaration, ClassVariableReferenceId);
+
+pub type ClassVariableDeclaration = DeclarationCore<ClassVariableReferenceId>;
 assert_mem_size!(ClassVariableDeclaration, 80);
-simple_declaration!(ConstantAliasDeclaration, ConstantReferenceId);
+
+pub type ConstantAliasDeclaration = DeclarationCore<ConstantReferenceId>;
```
Morriar, line 613: So this is valid:

```rust
let constant: ConstantDeclaration = ConstantDeclaration::new("Foo".into(), owner_id);
let alias: ConstantAliasDeclaration = constant;
```

are we okay with this?

vinistock: It's only valid at this point of the refactor. The final state merges all constants into a single `ConstantDeclaration` with an enum for its type that remembers whether it's an alias or not. I'm just trying to make small incremental PRs, it's an intermediate state.

Morriar (approved):

PR #946, opened 2026-07-21
Move existing singleton check earlier

vinistock:
A micro performance change with good impact: our check for an existing singleton class was not early enough in the method, which means we performed unnecessary work before returning.

By moving the check a few lines above, we save about 12% resolution time.


st0012 (approved):
Good catch

PR #948, opened 2026-07-22
Eagerly compute name depths into `Name`

vinistock:
Instead of computing name depths during resolution, we can actually cache it inside the `Name` structure during indexing. Because of struct alignment, we pay no memory price for adding the extra field and move a portion of the work to the parallelized indexing workers.

Additionally, computing the depths becomes easier because we can simply sum the depths of the dependent names.

This change gives us a 6% speedup for resolution and no significant differences in any other dimension.

```
BEFORE

Listing: 0.77
Indexing: 5.44
Resolution: 14.374
Total graph memory: 3889.036

AFTER

Listing: 0.796
Indexing: 5.366
Resolution: 13.53
Total graph memory: 3889.05
```

### DIFF
--- a/rust/rubydex/src/model/name.rs
+++ b/rust/rubydex/src/model/name.rs
@@ -4,6 +4,7 @@ use crate::{
     assert_mem_size,
     model::{
         id::id_from_parts,
+        identity_maps::IdentityHashMap,
         ids::{DeclarationId, NameId, StringId},
     },
 };
@@ -93,29 +94,55 @@ pub struct Name {
     /// list of names to represent the nesting
     nesting: Option<NameId>,
     ref_count: u32,
+    /// The depth of the name, which combines the depths of the parent scope and nesting
+    depth: u16,
 }
+assert_mem_size!(Name, 40);
 
 impl PartialEq for Name {
     fn eq(&self, other: &Self) -> bool {
         self.str == other.str && self.parent_scope == other.parent_scope && self.nesting == other.nesting
     }
 }
-assert_mem_size!(Name, 40);
 
 impl Name {
     #[must_use]
-    pub fn new(str: StringId, parent_scope: ParentScope, nesting: Option<NameId>) -> Self {
+    pub fn new(str: StringId, parent_scope: ParentScope, nesting: Option<NameId>, depth: u16) -> Self {
         Self {
             str,
             parent_scope,
             nesting,
             ref_count: 1,
+            depth,
         }
+    }
+
+    #[must_use]
+    pub(crate) fn name_depth(
+        names: &IdentityHashMap<NameId, NameRef>,
+        parent_scope: ParentScope,
+        nesting: Option<NameId>,
+    ) -> u16 {
+        if parent_scope.is_top_level() {
+            return 1;
+        }
+
+        let parent_depth = parent_scope
+            .as_ref()
+            .map_or(0, |id| names.get(id).map_or(0, NameRef::depth));
+        let nesting_depth = nesting.map_or(0, |id| names.get(&id).map_or(0, NameRef::depth));
+
+        parent_depth.saturating_add(nesting_depth).saturating_add(1)
     }
 
     #[must_use]
     pub fn str(&self) -> &StringId {
         &self.str
+    }
+
+    #[must_use]
+    pub fn depth(&self) -> u16 {
+        self.depth
     }
 
     #[must_use]
@@ -227,6 +254,14 @@ impl NameRef {
         }
     }
 
+    #[must_use]
+    pub fn depth(&self) -> u16 {
+        match self {
+            NameRef::Unresolved(name) => name.depth(),
+            NameRef::Resolved(resolved_name) => resolved_name.name.depth(),
+        }
+    }
+
     /// # Panics
     ///
     /// Panics if we exceed the maximum size of the reference count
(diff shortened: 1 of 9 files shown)


rust/rubydex/src/model/name.rs
```diff
@@ -93,31 +94,57 @@ pub struct Name {
     /// list of names to represent the nesting
     nesting: Option<NameId>,
     ref_count: u32,
+    /// The depth of the name, which combines the depths of the parent scope and nesting
+    depth: u16,
```
Morriar, line 98: This will saturate rather quickly. Can you try with this example?

 ```ruby
   module Root
     module A0; end
     module A0::B0
       module A1; end
       module A1::B1
         module A2; end
         module A2::B2
           module A3; end
           module A3::B3
             module A4; end
             module A4::B4
               module A5; end
               module A5::B5
                 module A6; end
                 module A6::B6
                   module A7; end
                   module A7::B7
                     module A8; end
                     module A8::B8
                       module A9; end
                       module A9::B9
                         module A10; end
                         module A10::B10
                           module A11; end
                           module A11::B11
                             module A12; end
                             module A12::B12
                               module A13; end
                               module A13::B13
                                 module A14; end
                                 module A14::B14
                                 end
                               end
                             end
                           end
                         end
                       end
                     end
                   end
                 end
               end
             end
           end
         end
       end
     end
   end
 ```

rust/rubydex/src/model/name.rs
```diff
@@ -93,31 +94,57 @@ pub struct Name {
...
     fn eq(&self, other: &Self) -> bool {
         self.str == other.str && self.parent_scope == other.parent_scope && self.nesting == other.nesting
     }
 }
-assert_mem_size!(Name, 40);
 
 impl Name {
     #[must_use]
-    pub fn new(str: StringId, parent_scope: ParentScope, nesting: Option<NameId>) -> Self {
+    pub fn new(str: StringId, parent_scope: ParentScope, nesting: Option<NameId>, depth: u16) -> Self {
```
Morriar: Why do we expose `depth` in the constructor instead of computing it inside of it and avoid mistakes?

rust/rubydex/src/model/name.rs
vinistock: Is this code anyone would actually write though? We can bump it to u32, but, realistically, do we really have to worry about a name with a depth higher than 60 thousand?

rust/rubydex/src/model/name.rs
vinistock: The name doesn't have the data it needs to compute it. That would require us passing a reference to the graph. The graph already knows all of the name parts to compute the depth on addition, so it felt like the natural place.

rust/rubydex/src/model/name.rs
Morriar: I can imagine some generated Ruby for Protobuf doing things a developer would not normally write.

I don't think the issue is the variable size but how the depth grows.

I wonder if we could use `d(name) = max(d(parent_scope), d(nesting)) + 1` instead of `d(name) = d(parent_scope) + d(nesting) + 1`?

rust/rubydex/src/model/name.rs
Morriar: Could we just take `names: &IdentityHashMap<NameId, NameRef>` instead of the graph?

rust/rubydex/src/model/name.rs
vinistock: I should also mention: we don't crash if we exceed the max depth because we're using saturating addition. If there's really a case like this, all it means is it will be ordered for resolution at the end.

If you feel strongly about this, we can bump to u32, but considering we don't crash and this scenario is not common, my vote is to keep u16 and just order these at the end.

Morriar: Look at my previous answer https://github.com/Shopify/rubydex/pull/948#discussion_r3685710454

vinistock: That approach currently breaks certain scenarios related to ancestors. It's not an inherent problem of using `max`, but rather a todo problem (which violates the name resolution rules of Rubydex to approximate using Sorbet rules). It does become worse by using `max` though.

Consider this example, which we have as a test in the codebase:

```ruby
# file1.rb
module Foo
  module Bar
    class Baz
    end
  end
end

# file2.rb
module Foo
  include Bar

  class Baz::Qux
  end
end
```

When we sum nesting + parent scope, `Foo::Bar::Baz` has a depth of 3 and `Foo::Bar::Baz::Qux` (resolve through inheritance) has a depth of 4. This ensures that we create `Foo::Bar::Baz` before it is used as a constant reference to define `Qux`.

If we use `max` instead, both `Foo::Bar::Baz` and the `Qux` reference have a depth of 3. This makes us try to process `Qux` before `Baz` is created.

That would be fine if todos didn't exist. We would simply enqueue a retry and succeed in resolving on the next pass. However, because of todos, we incorrectly create `Foo::Baz` when in fact it should resolve to `Foo::Bar::Baz`.

We can correctly make `max` work, but only if we change the strategy related to todos. Since their naming resolution doesn't match the overall algorithm, maybe we should move it to a separate method that consumers can use if they really need it, but we will keep hitting scenarios like these while we try to process them as part of the convergence loop. E.g.:

```ruby
# Rubydex based name resolution
graph.resolve

# Take any definitions we failed to process and approximate them
# Sorbet-style outside of the loop to prevent incorrectly making eager
# approximations for names that would be correctly resolved after more passes
graph.approximate_unresolved_namespaces
```

rust/rubydex/src/model/name.rs
vinistock: Done that in the latest commit.

rust/rubydex/src/model/name.rs
Morriar: As discussed offline, this is fine for now but we should look into a `depth` score that doesn't grow so fast.

Morriar (approved):

PR #949, opened 2026-07-22
Remove Mixin vec clones from ancestor linearization

vinistock:
Micro-performance: cloning the mixin vector every time we need to read it is more expensive than using a slice and cloning the individual mixins that we can to include in the compiled list.

This speeds up resolution by ~3%.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1019,9 +1019,9 @@ impl<'a> Resolver<'a> {
                 attached_decl
                     .definitions()
                     .iter()
-                    .filter_map(|definition_id| self.mixins_of(*definition_id))
-                    .flatten()
-                    .filter(|mixin| matches!(mixin, Mixin::Extend(_))),
+                    .flat_map(|definition_id| self.mixins_of(*definition_id))
+                    .filter(|mixin| matches!(mixin, Mixin::Extend(_)))
+                    .cloned(),
             );
         }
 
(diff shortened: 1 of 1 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -38,6 +38,17 @@ impl Outcome {
     }
 }
 
+macro_rules! mixins_of {
```
Morriar: I wonder if we can keep this as a regular helper returning a slice instead of introducing a macro. We still avoid cloning the `Vec`, but the call sites stay normal Rust and we don't have the helper accepting both `DefinitionId` and `&DefinitionId` through macro expansion.

```rust
fn mixins_of(&self, definition_id: DefinitionId) -> &[Mixin] {
    match self.graph.definitions().get(&definition_id).unwrap() {
        Definition::Class(class) => class.mixins(),
        Definition::SingletonClass(class) => class.mixins(),
        Definition::Module(module) => module.mixins(),
        _ => &[],
    }
}
```

rust/rubydex/src/resolution.rs
```diff
@@ -1019,34 +1030,32 @@ impl<'a> Resolver<'a> {
...
-                    }
+            for mixin in mixins_of!(self, definition_id) {
+                match mixin {
+                    Mixin::Prepend(_) | Mixin::Include(_) => mixins.push(mixin.clone()),
+                    Mixin::Extend(_) => has_extends = true,
                 }
             }
         }
 
+        let (linearized_prepends, linearized_includes) =
```
Morriar: Can we keep this call after creating/enqueuing the singleton in this PR?

rust/rubydex/src/resolution.rs
vinistock: Yeah, we can definitely do that 👍

rust/rubydex/src/resolution.rs
vinistock: Combined with the other change, yes. Otherwise, we hit a borrow issue.

Morriar (approved):

PR #952, opened 2026-07-23
Remove unnecessary context allocation

vinistock:
Currently, we are allocating a new `LinearizationContext` each time for `ancestors_of`. This is completely unnecessary because the linearization recursion already cleans up the context object.

We can simply reuse the same object stored in the `Resolver`, avoid the allocation entirely and remove `ancestors_of` while we're at it. This provides a 12% speed up for resolution, bringing us to ~11s.

### DIFF
--- a/rust/rubydex/src/resolution.rs
+++ b/rust/rubydex/src/resolution.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{HashSet, VecDeque, hash_map::Entry},
-    hash::BuildHasher,
-};
+use std::collections::{HashSet, VecDeque, hash_map::Entry};
 
 use crate::diagnostic::{Diagnostic, Rule, Severity};
 use crate::model::{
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/resolution.rs
```diff
@@ -85,6 +82,8 @@ impl LinearizationContext {
     fn finalize(&mut self, declaration_id: DeclarationId) {
         self.descendants.remove(&declaration_id);
         self.seen_ids.remove(&declaration_id);
+        self.cyclic = false;
+        self.partial = false;
```
Morriar: Let's take this example:

```rb
module M
end

class C
  include Missing
  include M
end
```

Are we going to first set partial to `true` while processing `Missing` then reset `partial` to `false` after finalizing `M`?

vinistock: Great catch. I fixed this by using a `ChainState`, which we use as part of the recursion. So the state moves from the `LinearizationContext` into just a local variable.

soutaro:
I find the lifecycle of `LinearizationContext` harder to follow when it is mutated implicitly through `Resolver`. Could we keep the reuse as an implementation detail of `ancestors_of`, while making the active traversal state explicit?

For example, `ancestors_of` could take the reusable context out, initialize it, pass it explicitly through the traversal, validate that it has been cleaned up, and then put it back:

```rust
let mut context = self
    .reusable_linearization_context
    .take()
    .expect("ancestor linearization must not be re-entered");

context.begin();
let result = self.linearize_ancestors(declaration_id, &mut context);
debug_assert!(context.is_idle());

self.reusable_linearization_context = Some(context);
result
```

This retains the allocation benefit, while making the context’s lifetime and ownership at a traversal call site easier to reason about. Using `Option` would also make accidental re-entry fail explicitly, instead of silently taking a new default context.

---

By the way, a 12% speedup from allocation reduction alone seems a little surprising. What is actually driving the improvement?

Morriar (approved):

vinistock:
@soutaro I agree that the flow of the code becomes a bit harder to read, but the performance gains are real when benchmarking against large codebases.

Allocating a new `LinearizationContext` involves allocating two hashmaps and we cannot skip allocating the context on already computed ancestor chains. By reusing the same context across everything, we eliminate millions of hashmap allocations and that pays dividends for performance.

PR #959, opened 2026-07-28
Add initial control-flow graph model and Prism builder

Morriar:
## Why

Rubydex needs a control-flow graph before it can model instruction semantics, local-variable flow, reachability, and type inference.

This PR introduces the minimum CFG topology needed to start that work. The representation follows Sorbet’s non-SSA CFG structure but does not attempt full Sorbet compatibility yet.

## What changed

Added a CFG model with:

- Stable `BlockId`s
- Dedicated entry and exit blocks
- Ordered instructions within each `BasicBlock`
- `Jump`, `Branch`, and `Exit` terminators
- Source locations for instructions and control-flow transfers
- Successor traversal and block lookup APIs

Added a Prism visitor that builds CFGs for:

- Sequential statements
- `if` and `unless`
- `while` and `until`
- `return`
- Parenthesized expressions
- Nested branching predicates

The builder preserves Ruby evaluation order and creates explicit branch, merge, loop-header, loop-exit, and unreachable blocks. These structural blocks may be empty; a future CFG simplification pass can remove unnecessary forwarding blocks.

Instructions currently contain only source locations. Instruction semantics, locals, block arguments, and type information will be introduced separately.

## Testing

Builder tests render the complete CFG into a canonical textual representation. Each test verifies:

- Block allocation
- Instruction ordering
- True and false successors
- Merge edges
- Loop backedges
- Return-to-exit edges
- Nested control flow

## Follow-up work

This PR intentionally does not include:

- Instruction lowering
- Local variables or temporary values
- `break`, `next`, `redo`, or `retry`
- Exception, `rescue`, or `ensure` control flow
- Method arguments or default arguments
- CFG simplification
- Sorbet fixture compatibility tests

### DIFF
--- a/rust/rubydex/src/cfg.rs
+++ b/rust/rubydex/src/cfg.rs
@@ -1,0 +1,173 @@
+//! Control-flow graph topology for Ruby method bodies.
+//!
+//! A control-flow graph (CFG) models every path execution can take through a
+//! method body. Each node is a [`BasicBlock`]: a straight-line sequence of
+//! [`Instruction`]s that always execute together. A block ends with one
+//! [`Terminator`], which transfers control to its successor blocks.
+
+pub mod builder;
+
+use std::fmt;
+
+use crate::offset::Offset;
+
+/// Index of a basic block inside a [`ControlFlowGraph`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BlockId(u32);
+
+impl BlockId {
+    /// The unique entry block. Method execution starts here.
+    pub const ENTRY: Self = Self(0);
+    /// The unique exit block. Every terminating path eventually reaches it.
+    pub const EXIT: Self = Self(1);
+
+    #[must_use]
+    pub const fn new(value: u32) -> Self {
+        Self(value)
+    }
+
+    #[must_use]
+    pub const fn get(self) -> u32 {
+        self.0
+    }
+
+    fn index(self) -> usize {
+        self.0 as usize
+    }
+}
+
+impl fmt::Display for BlockId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "bb{}", self.0)
+    }
+}
+
+/// An instruction whose semantics have not been lowered yet.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Instruction {
+    /// The corresponding byte range in the Ruby source.
+    location: Offset,
+}
+
+impl Instruction {
+    #[must_use]
+    pub const fn new(location: Offset) -> Self {
+        Self { location }
+    }
+
+    #[must_use]
+    pub const fn location(&self) -> &Offset {
+        &self.location
+    }
+}
+
+/// The control-flow instruction that ends a basic block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Terminator {
+    /// Construction placeholder for a block whose terminator has not been set.
+    /// The builder replaces it before returning a completed graph.
+    Todo,
+    /// Transfers control unconditionally to one successor.
+    Jump {
+        /// The block that executes next.
+        target: BlockId,
+        /// The Ruby source range responsible for the transfer.
+        location: Offset,
+    },
+    /// Selects one of two successors. The branch condition will be introduced
+    /// with instruction semantics in a later layer.
+    Branch {
+        /// The successor used when the condition is truthy.
+        then_block: BlockId,
+        /// The successor used when the condition is false or `nil`.
+        else_block: BlockId,
+        /// The Ruby source range responsible for the branch.
+        location: Offset,
+    },
+    /// Ends method execution. Only [`BlockId::EXIT`] may use this terminator.
+    Exit,
+}
+
+impl Terminator {
+    pub fn successors(&self) -> impl Iterator<Item = BlockId> {
+        let successors = match self {
+            Self::Todo | Self::Exit => [None, None],
+            Self::Jump { target, .. } => [Some(*target), None],
+            Self::Branch {
+                then_block, else_block, ..
+            } => [Some(*then_block), (*then_block != *else_block).then_some(*else_block)],
+        };
+        successors.into_iter().flatten()
+    }
+}
+
+/// A maximal straight-line instruction sequence with one terminator.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BasicBlock {
+    /// Position of this block in its CFG's block array.
+    id: BlockId,
+    /// Instructions that execute in source order when control enters the block.
+    instructions: Vec<Instruction>,
+    /// The control-flow instruction that ends the block.
+    terminator: Terminator,
+}
+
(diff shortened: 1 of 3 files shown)


rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub mod builder;
+
+use std::fmt;
+
+use crate::offset::Offset;
+
+/// Index of a basic block inside a [`ControlFlowGraph`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BlockId(u32);
```
vinistock, line 16: Could we get away with u16? How large does a method need to be to produce so many blocks that we exceed the 65k?

Maybe we can save a bit of memory.

rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl BlockId {
+    /// The unique entry block. Method execution starts here.
+    pub const ENTRY: Self = Self(0);
+    /// The unique exit block. Every terminating path eventually reaches it.
+    pub const EXIT: Self = Self(1);
```
vinistock, line 22: Question for my understanding: if a method has multiple return statements, do their blocks all point to the same exit?

Also, this might be worth documenting.

rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[must_use]
+    pub const fn new(location: Offset) -> Self {
+        Self { location }
+    }
+
+    #[must_use]
+    pub const fn location(&self) -> &Offset {
+        &self.location
+    }
+}
```
vinistock, line 62: Not necessary for this PR, but I imagine instructions are going to grow a lot, so I'd move it to its own file under `cfg/`.

rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "bb{}", self.0)
+    }
+}
+
+/// An instruction whose semantics have not been lowered yet.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Instruction {
+    /// The corresponding byte range in the Ruby source.
+    location: Offset,
```
vinistock, line 49: Let's call this offset. Location is a different concept so I'm worried it might cause confusion.

rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+pub enum Terminator {
+    /// Construction placeholder for a block whose terminator has not been set.
+    /// The builder replaces it before returning a completed graph.
+    Todo,
+    /// Transfers control unconditionally to one successor.
+    Jump {
+        /// The block that executes next.
+        target: BlockId,
+        /// The Ruby source range responsible for the transfer.
+        location: Offset,
```
vinistock, line 75: Two things: first is the same as my previous comment, let's call this offset.

Second: are we duplicating the offsets for the same instruction? I'm trying to build my mental model of how the blocks connect exactly.

Consider this

```ruby
def foo
  blah = true

  # The if statement here causes the branching. Is this if statement
  # stored exclusively as a block terminator or is it also an instruction
  # that belongs to the first block of `foo`?

  if blah
  else
  end
end
```

rust/rubydex/src/cfg.rs
```diff
@@ -0,0 +1,173 @@
...
+            Self::Jump { target, .. } => [Some(*target), None],
+            Self::Branch {
+                then_block, else_block, ..
+            } => [Some(*then_block), (*then_block != *else_block).then_some(*else_block)],
+        };
+        successors.into_iter().flatten()
+    }
+}
+
+/// A maximal straight-line instruction sequence with one terminator.
```
vinistock, line 104: I think the `maximal straight-line` part may not be clear to everyone, although I do like the concise way of describing it. Can we expand a bit on it?

rust/rubydex/src/cfg/builder.rs
```diff
@@ -0,0 +1,682 @@
...
+    let end_offset = node
+        .location()
+        .end_offset()
+        .try_into()
+        .expect("source offset must fit in u32");
+    builder.finish(end_offset)
+}
+
+struct Builder {
+    blocks: Vec<BasicBlock>,
```
vinistock, line 30: Here we're building the blocks vector and then instantiating the Cfg with the complete list. No strong preference, but what do you think about having the builder hold a `CFG` instance that gets mutated instead?

Basically, something like this
```rust
self.cfg.add_block(...)
```

rust/rubydex/src/cfg/builder.rs
```diff
@@ -0,0 +1,682 @@
...
+            current
+        } else {
+            let block = self.fresh_block();
+            self.current = Some(block);
+            block
+        }
+    }
+
+    /// Appends an empty block with the next sequential ID.
+    fn fresh_block(&mut self) -> BlockId {
```
vinistock, line 192: Nit:


```suggestion
    fn new_block(&mut self) -> BlockId {
```

rust/rubydex/src/cfg/builder.rs
```diff
@@ -0,0 +1,682 @@
...
+        let id = BlockId::new(self.blocks.len().try_into().expect("block count must fit in u32"));
+        self.blocks.push(BasicBlock::new(id, Vec::new(), Terminator::Todo));
+        id
+    }
+
+    /// Adds an instruction to the current block.
+    fn emit(&mut self, location: Offset) {
```
vinistock, line 199: Nit:


```suggestion
    fn emit_instruction(&mut self, location: Offset) {
```

rust/rubydex/src/cfg/builder.rs
```diff
@@ -0,0 +1,682 @@
...
+    fn finish(mut self, end: u32) -> ControlFlowGraph {
+        for block in &mut self.blocks {
+            if matches!(block.terminator, Terminator::Todo) {
+                block.terminator = Terminator::Jump {
+                    target: BlockId::EXIT,
+                    location: Offset::new(end, end),
+                };
+            }
+        }
+        ControlFlowGraph::new(self.blocks)
```
vinistock, line 57: If my understanding is correct, we build one CFG per file that we visit, which may contain multiple methods.

Considering we want to parallelize the analysis of the CFG, should we instead produce a new CFG for each statements node (a.k.a. method body)?

rust/rubydex/src/cfg.rs
soutaro, line 49: Nitpick:

`Offset` alone cannot uniquely identify the originating Prism AST node because multiple nodes can have exactly the same location. For example:

```ruby
> Prism.parse("def foo = nil")
...
              ├── body:
              │   @ StatementsNode (location: (1,10)-(1,13))
              │   ├── flags: ∅
              │   └── body: (length: 1)
              │       └── @ NilNode (location: (1,10)-(1,13))
              │           └── flags: static_literal
...
```

I’m not sure yet how `Instruction` will be used, but storing the node type as well would provide a more complete reference to the originating Prism node.

PR #961, opened 2026-07-29
Implement initial complexity analysis pass and CLI command

rafaelfranca:
## Summary

Adds an ABC complexity analysis pass to rubydex: a standalone Prism-based scorer that
computes assignment, branch, and call complexity for Ruby source, exposed through both
the Rust CLI (`rubydex_cli --complexity`) and the Ruby gem (`rdx complexity` /
`Rubydex::Complexity`). It runs as its own analysis pass — no graph or resolution
required — so it is fast enough to run repeatedly over a large codebase.

The scoring rules are modeled on [flog](https://github.com/seattlerb/flog); thanks to
[Ryan Davis](https://github.com/zenspider) for the original.

A score is reported per method as `sqrt(a² + b² + c²)`, where `a`, `b`, and `c`
accumulate assignment, branch, and call weight respectively (each scaled by the current
nesting multiplier). `.rbs` files are excluded; everything else the indexer treats as
Ruby (`.rb`, `.rake`, `.ru`, …) is scored.

## CLI

```bash
# Compute a report for the current directory (top 25 methods by default)
bundle exec rdx complexity

# Scope to specific paths and show more entries
bundle exec rdx complexity app/models lib/services --top 50

# Full report as JSON (--top is ignored for JSON, so every method is included)
bundle exec rdx complexity app/models --format json > report.json
```

### Text output

```
$ bundle exec rdx complexity app/models

    17.2: total complexity
     5.7: average complexity

    10.3: Rubydex::Complexity.analyze              app/models/rubydex/complexity.rb:8-12
     5.0: Rubydex::Complexity#none                 app/models/rubydex/complexity.rb:5-19
     1.9: Rubydex::Complexity.diff                 app/models/rubydex/complexity.rb:15-17
```

### JSON output

```bash
bundle exec rdx complexity app/models --format json > report.json
```

The JSON report (`schema_version: 1`) carries `total`, `average`, `methods_count`, and
a `methods` array of entries with per-bucket breakdowns and `start_line`/`end_line`
locations (`--top` is ignored for JSON, so all methods are emitted):

```json
{
  "schema_version": 1,
  "total": 17.2,
  "average": 5.7,
  "methods_count": 3,
  "methods": [
    {
      "name": "Rubydex::Complexity.analyze",
      "file": "app/models/rubydex/complexity.rb",
      "start_line": 8,
      "end_line": 12,
      "assignments": 1.0,
      "branches": 1.5,
      "calls": 10.1,
      "score": 10.3
    }
  ]
}
```

### Diffing against a baseline

Diff a fresh report against a baseline JSON report to track drift over time. Both
reports must be generated with the same `--methods-only` setting.

```bash
bundle exec rdx complexity app/models --diff baseline.json
```

```
Total:   1200.0 -> 1250.3 (+50.3)
Average:    8.0 ->    8.2 (+0.2)
Methods:    150 ->    155 (+5)

Regressions:
   +12.3: Foo#bar (24.0 -> 36.3)                 app/models/foo.rb
Improvements:
    -5.0: Baz#qux (15.0 -> 10.0)                 app/models/baz.rb
Added:
    20.0: New#thing                              app/models/new.rb
Removed:
     8.0: Old#gone                               app/models/old.rb
```

### Per-construct breakdown, methods-only, and grouping

```bash
# Show what drives each method's score: per-construct contributions
bundle exec rdx complexity app/models --details --top 10

# Skip code outside methods: drops top-level `#none` noise
bundle exec rdx complexity app/models --methods-only

# Group entries by class with per-class subtotals
bundle exec rdx complexity app/models --group
```

`--details` adds a per-method breakdown:

```
    10.3: Rubydex::Complexity.analyze              app/models/rubydex/complexity.rb:8-12
     1.8:   class
     1.7:   block_pass
     1.7:   map
     1.6:   raise
     1.5:   branch
     0.4:   magic_number
```

`--group` is text-only (rejected with `--format json`); neither `--details` nor
`--group` applies to `--diff` (both are rejected with it).

## Ruby API

```ruby
# Compute a report: returns the text or JSON string ready to print
puts Rubydex::Complexity.analyze(["app/models"], format: :text, top: 25)
json = Rubydex::Complexity.analyze(["app/models"], format: :json)

# Per-construct breakdown, methods-only, and grouping
puts Rubydex::Complexity.analyze(["app/models"], details: true, top: 10)
puts Rubydex::Complexity.analyze(["app/models"], methods_only: true)
puts Rubydex::Complexity.analyze(["app/models"], group: true)

# Diff two JSON reports (baseline against a fresh run).
# Both reports must share the same methods_only setting or diff raises ArgumentError.
puts Rubydex::Complexity.diff(baseline_json, json, format: :text, top: 25)
```

## Configured exclusions

Exclusions are configured in `rubydex.toml` and are **decoupled from indexing**: the
top-level `exclude` key affects indexing only, while a separate `[complexity]` table
controls what `rdx complexity` skips. This lets you keep a file in the graph but out of
the complexity report. Both share the default skipped directories (`.git`,
`node_modules`, `tmp`, …).

```toml
# rubydex.toml
exclude = ["vendor/**"]              # indexing only; complexity still scores these

[complexity]
exclude = ["app/assets/**", "**/*_spec.rb"]   # complexity only; still indexed
```

Malformed globs in either set are reported as a config error at load time rather than
silently dropped.

## Implementation

- New `complexity` module in the Rust crate: a Prism `Visit` implementation that scores
  each method into `a`/`b`/`c` buckets with compounding nesting penalties, plus
  `Report`/`DiffReport` with `render_text` / `to_json` / `from_json` / `diff`.
- FFI plumbing (`rdx_complexity_analyze` / `rdx_complexity_diff`) and a C extension
  (`Rubydex::Complexity.native_analyze` / `native_diff`).
- Parallel `analyze` mirrors the indexer's job queue; per-file non-fatal errors are
  surfaced as warnings.
- Scoring rules and deliberate deviations are documented in the module doc comment;
  behavior is pinned by inline regression tests.

## Test plan

- Rust: 1163 lib + 19 CLI tests, `cargo clippy --all-targets --all-features -- -D warnings` clean.
- Ruby: 282 runs / 1219 assertions, `bundle exec rake lint` clean.


rust/rubydex/src/listing.rs
```diff
@@ -4,13 +4,22 @@ use crate::{
...
     path::{Path, PathBuf},
     sync::Arc,
 };
 
+/// Returns the workspace root for the given paths: the first path, when it canonicalizes to a
+/// directory. Used to locate `rubydex.toml` and resolve exclusion patterns relative to the root.
+#[must_use]
+pub fn workspace_path_for(paths: &[String]) -> Option<PathBuf> {
+    let first_path = paths.first()?;
+    fs::canonicalize(first_path).ok().filter(|path| path.is_dir())
```
st0012: I'm not sure if we can do `paths.first` and treat it as workspace path, especially if we want to support complexity check on specific paths only. 
Let's say I use `rdx complexity lib/`, this will use `lib/` as the workspace, but then miss the `./rubydex.toml` file.

st0012 (commented):
There is a lot of complexity in `complexity.rs` especially on receiver/namespace handling, that we already went through in https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/indexing/ruby_indexer.rs. Is it possible to build part of the complexity detection using `Definition` instead?

It may look like:
1. For class & modules....etc. that we create definition for, we already handled a lot of the context building, and we can use definition attributes to calculate the penalty
2. For things that we don't index, we keep the existing visiting logic. Yes we double index, but it's also fast

And then we merge the results from 1 and 2. As time goes on and Rubydex index more and more stuff, we'll gradually expand 1 and shrink 2, until hopefully 1 day the indexing phase's result & CFG would already be enough for complexity analysis?

rust/rubydex/src/complexity.rs
```diff
@@ -0,0 +1,2114 @@
...
+
+    fn visit_def_node(&mut self, node: &ruby_prism::DefNode<'pr>) {
+        let method_name = String::from_utf8_lossy(node.name().as_slice()).to_string();
+        let namespace = self.namespace_string();
+        let separator = if node.receiver().is_some() || self.in_singleton() {
+            '.'
+        } else {
+            '#'
+        };
+        let key = format!("{namespace}{separator}{method_name}");
```
Morriar, line 465: We're losing the receiver information here. Consider this:

```rb
class Foo
  def Bar.bar; end
end
```

the key becomes `Foo.bar`.

rust/rubydex/src/complexity.rs
```diff
@@ -0,0 +1,2114 @@
...
+            return Err(Errors::ReportError(format!(
+                "cannot diff reports with mismatched scoring modes: baseline methods_only={}, current methods_only={} \
+                 (regenerate the baseline with the same `--methods-only` setting)",
+                baseline.methods_only, current.methods_only
+            )));
+        }
+        let baseline_map: HashMap<(String, String), &MethodEntry> = baseline
+            .methods
+            .iter()
+            .map(|entry| ((entry.name.clone(), entry.file.clone()), entry))
```
Morriar, line 1438: I think `entry.file` is an absolute path? Which means we can't compare reports from different machines as the absolute path may differ. Should we use the relative path to the workspace instead?

lib/rubydex/complexity.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module Rubydex
+  # ABC complexity reports. Partially defined in C (native_analyze / native_diff).
+  module Complexity
+    class << self
+      #: (Array[String] paths, ?format: String | Symbol, ?top: Integer, ?methods_only: bool, ?details: bool, ?group: bool) -> String
+      def analyze(paths, format: :text, top: 25, methods_only: false, details: false, group: false)
+        raise TypeError, "no implicit conversion of #{paths.class} into Array" unless paths.is_a?(Array)
+
+        native_analyze(paths.map(&:to_s), format, top, methods_only, details, group)
```
Morriar, line 11: Why the `to_s`, isn't `paths` an `Array[String]` from the signature?

exe/rdx
```diff
@@ -145,6 +146,54 @@ operation =
...
+      parser.on("--top N", Integer, "Max entries in text output, 0 = all (default 25)") { |value| top = value }
+      parser.on("--diff FILE", "Diff against a baseline JSON report") { |value| diff_file = value }
+      parser.on("--methods-only", "Skip code outside methods") { methods_only = true }
+      parser.on("--details", "Show per-construct score breakdown under each method") { details = true }
+      parser.on("--group", "Group and sort by class with subtotals") { group = true }
+      parser.on("-h", "--help", "Show this help") do
+        puts parser
+        exit
+      end
+    end.parse!
```
Morriar, line 168: Should we `rescue OptionParser::ParseError`?

exe/rdx
```diff
@@ -145,6 +146,54 @@ operation =
...
+      abort("`--group` does not apply to diff output; remove it or drop `--diff`")
+    end
+    if group && format == "json"
+      abort("`--group` only affects text output; use `--format text` or drop `--group`")
+    end
+    paths = ARGV.empty? ? [Dir.pwd] : ARGV.dup
+    if diff_file
+      baseline = begin
+        File.read(diff_file)
+      rescue Errno::ENOENT
```
Morriar, line 183: You could broaden to `SystemCallError` and say `can't read baseline report` so you also cover `EACCESS`, `EISDIR` and friends.

exe/rdx
```diff
@@ -145,6 +146,54 @@ operation =
...
+    format = "text"
+    top = 25
+    diff_file = nil
+    methods_only = false
+    details = false
+    group = false
+    OptionParser.new do |parser|
+      parser.banner = "Usage: rdx complexity [PATH ...] [options]"
+      parser.on("--format FORMAT", ["text", "json"], "Output format (text or json)") { |value| format = value }
+      parser.on("--top N", Integer, "Max entries in text output, 0 = all (default 25)") { |value| top = value }
```
Morriar, line 159: Should we check if `top` is negative?

README.md
```diff
@@ -143,6 +143,113 @@ puts query.render(graph, "json")
...
+
+```bash
+# Compute a report for the current directory (top 25 methods by default)
+bundle exec rdx complexity
+
+# Scope to specific paths and show more entries
+bundle exec rdx complexity app/models lib/services --top 50
+
+# Render the full report as JSON (use --top 0 for every method)
+bundle exec rdx complexity app/models --format json --top 0 > report.json
```
Morriar, line 184: Do we even consider `top` for `json` output?

ext/rubydex/complexity.c
```diff
@@ -0,0 +1,84 @@
...
+static VALUE rdxr_complexity_analyze(VALUE self, VALUE paths, VALUE format, VALUE top, VALUE methods_only, VALUE details, VALUE group) {
+    rdxi_check_array_of_strings(paths);
+
+    long length = RARRAY_LEN(paths);
+    char **paths_array = rdxi_str_array_to_char(paths, (size_t)length);
+
+    struct CComplexityResult result = rdx_complexity_analyze(
+        (const char *const *)paths_array,
+        (size_t)length,
+        rdxi_symbol_or_string_cstr(format, "text"),
```
Morriar, line 25: Could we use a C enum instead so we don't have to keep a pointer to the Ruby memory during the whole analysis?

ext/rubydex/complexity.c
```diff
@@ -0,0 +1,84 @@
...
+ * maximum number of entries in text output (0 prints all), ignored for JSON. +methods_only+
+ * skips code outside methods; +details+ collects the per-construct breakdown;
+ * +group+ groups text output by class with subtotals (ignored for JSON). Raises
+ * ArgumentError on a fatal config or format error.
+ */
+static VALUE rdxr_complexity_analyze(VALUE self, VALUE paths, VALUE format, VALUE top, VALUE methods_only, VALUE details, VALUE group) {
+    rdxi_check_array_of_strings(paths);
+
+    long length = RARRAY_LEN(paths);
+    char **paths_array = rdxi_str_array_to_char(paths, (size_t)length);
```
Morriar, line 20: This is going to leak if `rdxi_symbol_or_string_cstr` or `NUM2SIZET` raise

rust/rubydex-sys/src/complexity_api.rs
```diff
@@ -0,0 +1,162 @@
...
+            .join("\n")
+    };
+
+    let output = match format_str.as_str() {
+        "text" => report.render_text(top, details, group),
+        "json" => report.to_json(),
+        other => {
+            return CComplexityResult::error(&format!(
+                "unknown complexity format `{other}` (expected `text` or `json`)"
+            ));
```
Morriar, line 101: Let's validate the output format before doing the expensive computation in `analyze`

rust/rubydex/src/complexity.rs
```diff
@@ -0,0 +1,2114 @@
...
+            .collect();
+        if parts.is_empty() {
+            "main".to_string()
+        } else {
+            parts.join("::")
+        }
+    }
+
+    fn in_singleton(&self) -> bool {
+        self.scope_stack.iter().any(|frame| frame.is_singleton)
```
Morriar: Should we limit this to the next scope?

Consider this:

```rb
class Foo
  class << self
    class Bar
      def run
        work
      end
    end
  end
end
```

rust/rubydex/src/complexity.rs
```diff
@@ -0,0 +1,2114 @@
...
+
+    let total = methods.iter().map(|entry| entry.score).sum::<f64>();
+    let methods_count = methods.len();
+    let average = if methods_count == 0 {
+        0.0
+    } else {
+        total / f64::from(u32::try_from(methods_count).unwrap_or(u32::MAX))
+    };
+    Report {
+        schema_version: 1,
```
Morriar, line 1250: Do we ever enforce this?

rust/rubydex/src/complexity.rs
```diff
@@ -0,0 +1,2114 @@
...
+        // created; a real `def none` (whose key is also `Foo#none`) is scored normally because
+        // it is visited with `method_stack` non-empty.
+        if self.methods_only && self.method_stack.is_empty() {
+            return;
+        }
+        let value = weight * self.multiplier;
+        let (key, start_line, end_line) = if let Some((k, s, e)) = self.method_stack.last() {
+            (k.clone(), *s, *e)
+        } else {
+            let k = format!("{}#none", self.namespace_string());
```
Morriar, line 300: Won't this collide with an actual `#none` method _if_ `--methods-only=false`? 🤔

rafaelfranca:
@st0012 I tried that approach right now and it instead of deleting code required me to add way more code, not to speak on the duplicated indexing for each file. I'm not sure it is worth it

basicBrogrammer:
@rafaelfranca I'm curious why rewrite it instead of just using flog? Also, what would be your take on the rest of the ruby critic gang (reek and flay)?

rafaelfranca:
Speed mainly. This one runs in 15s against our codebase. Flog runs in 5m:38s.

PR #965, opened 2026-07-30
Make `Config` a proper object

vinistock:
This PR makes `Config` a proper object with a Ruby equivalent and allows us to easily add more config sections for each new tool.

### Implementation

There are a few relevant points:

1. The config is now composed of global fields (like workspace_path) and tool-specific fields which correspond to a section in the config file. For example, `[graph] -> GraphSettings`, [linter] -> LinterSettings`
2. All tools need to know what is the workspace path they are operating on. In certain tools, the workspace path isn't always pwd (e.g.: language servers that are spawned somewhere different than the workspace by editors). So the workspace_path state moved into `Config` rather than being inside of `Graph`
3. We started out with more complexity than we really needed, supporting loading arbitrary config files. I believe that decision is premature and we should just support the default `workspace_path/rubydex.toml` for now
4. Loading the configuration is now performed independently and then loaded into the graph. We need an approach like this because the configuration is shared by both the analysis (Rust) and tools built on top of it (Ruby or Rust). We don't want each tool to re-parse and maintain another config in memory or need to add a toml parser as a dependency. This approach avoids all of those

Note: this PR doesn't yet include the parsing of the new `[linter]` section. I tried to minimize the changes for an easier review, but it's a rather large breaking change. I also recommend reviewing by commit.

### DIFF
--- a/ext/rubydex/config.c
+++ b/ext/rubydex/config.c
@@ -1,0 +1,78 @@
+#include "config.h"
+#include "rustbindings.h"
+#include "utils.h"
+
+static VALUE mRubydex;
+
+// Free function for Rubydex::Config: releases the parsed configuration allocated by Rust.
+static void config_free(void *ptr) {
+    if (ptr) {
+        rdx_config_free(ptr);
+    }
+}
+
+const rb_data_type_t config_type = {
+    .wrap_struct_name = "Rubydex::Config",
+    .function = {
+        .dmark = NULL,
+        .dfree = config_free,
+        .dsize = NULL,
+        .dcompact = NULL,
+    },
+    .parent = NULL,
+    .data = NULL,
+    .flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_FROZEN_SHAREABLE,
+};
+
+/*
+ * call-seq:
+ *   Rubydex::Config.load(workspace_path) -> Rubydex::Config
+ *
+ * Loads the configuration of the workspace rooted at +workspace_path+, which is where its `rubydex.toml` is expected to
+ * be. A workspace without a configuration file gets an empty configuration. Raises Rubydex::ConfigError if the file
+ * exists, but cannot be read or is malformed.
+ */
+static VALUE rdxr_config_load(VALUE klass, VALUE workspace_path) {
+    Check_Type(workspace_path, T_STRING);
+
+    struct CConfigResult result = rdx_config_load(StringValueCStr(workspace_path));
+    if (result.error != NULL) {
+        VALUE message = rb_utf8_str_new_cstr(result.error);
+        free_c_string(result.error);
+
+        VALUE config_error = rb_const_get(mRubydex, rb_intern("ConfigError"));
+        rb_exc_raise(rb_exc_new_str(config_error, message));
+    }
+
+    return TypedData_Wrap_Struct(klass, &config_type, result.config);
+}
+
+/*
+ * call-seq:
+ *   workspace_path -> String
+ *
+ * Returns the root directory of the workspace this configuration was loaded for.
+ */
+static VALUE rdxr_config_workspace_path(VALUE self) {
+    const char *result = rdx_config_workspace_path(rdxi_config_from_object(self));
+    // A configuration always has a workspace, so there is no absent case to hand back to Ruby. NULL only means the
+    // conversion itself failed, which is why this raises instead of returning nil.
+    if (result == NULL) {
+        rb_raise(rb_eRuntimeError, "Converting workspace path to Ruby string failed");
+    }
+
+    return rdxi_owned_c_string_to_ruby(result);
+}
+
+void rdxi_initialize_config(VALUE moduleRubydex) {
+    mRubydex = moduleRubydex;
+
+    VALUE cConfig = rb_define_class_under(mRubydex, "Config", rb_cObject);
+    rb_undef_alloc_func(cConfig);
+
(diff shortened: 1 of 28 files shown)


st0012 (commented):
Can we provide `Graph#load_config_from(path)` and use that in most of the cases instead? I think most use cases would just need that, and some fewer cases may need more customized Config objects, which can use `load_config`.

ext/rubydex/config.c
```diff
@@ -0,0 +1,78 @@
...
+    .wrap_struct_name = "Rubydex::Config",
+    .function = {
+        .dmark = NULL,
+        .dfree = config_free,
+        .dsize = NULL,
+        .dcompact = NULL,
+    },
+    .parent = NULL,
+    .data = NULL,
+    .flags = RUBY_TYPED_FREE_IMMEDIATELY,
```
Morriar: I think we can make it Ractor safe:

```suggestion
    .flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_FROZEN_SHAREABLE,
```

vinistock:
Done!

lib/rubydex/graph.rb
```diff
@@ -7,9 +7,16 @@ module Rubydex
...
 
-    #: (?workspace_path: String?) -> void
-    def initialize(workspace_path: nil)
-      self.workspace_path = workspace_path if workspace_path
+    class << self
+      # Creates a new graph with the loaded configuration. For use cases where the graph must be shared between
+      # different tools, do not use this. Create and own a `Config` object instead.
+      #
+      #: (String) -> instance
+      def with_config(workspace_path)
```
Morriar: This name is weird when we neither pass a config object or the path to a config file

rust/rubydex/src/config.rs
```diff
@@ -18,141 +18,207 @@ pub const DEFAULT_EXCLUDED_DIRECTORIES: &[&str] = &[
...
+impl Config {
+    /// Loads the configuration of the workspace rooted at `workspace_path`, which is where its `rubydex.toml` is
+    /// expected to be. The configuration file itself is optional: a workspace without one gets the default settings.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Errors::ConfigError`] if `workspace_path` is not a directory, or if the configuration file exists but
+    /// cannot be read or is invalid.
+    pub fn load(workspace_path: &Path) -> Result<Self, Errors> {
+        // Since loading a configuration is how a workspace is chosen, a root that cannot be analyzed has to be rejected
```
Morriar, line 115: Let's canonicalize the path in case the user changes working directory

lib/rubydex/graph.rb
st0012: I agree. How about `with_config_from_workspace` or `load_config_from_workspace`?
I know it's longer, but then it's self-evident that it takes a workspace path.

vinistock: How about `with_workspace_config`?

st0012: It's better but IMO `with_*_config` still implies we're taking a config. How about `configure_for_workspace`?

rafaelfranca: `Config.load_from`

st0012: @rafaelfranca That's the original API in the PR, but every caller is doing `graph.load_config(Config.load_from(workspace_path))`. We can still have `graph.load_config` where users want to construct the `Config` object manually, but if the goal is to just load the project's configuration file, it's very verbose.
So I proposed adding a graph level API that just take that path like: `graph.configure_for_workspace(workspace_path)`.

vinistock: I went with `configure_for_workspace`.

vinistock:
Alright, I addressed the comments. Canonicalizing the workspace path has other ramifications because on Windows it might expand to verbatim paths (e.g: `\\?\C:\workspace`). We cannot use that format because we concatenate the glob patterns for file exclusions, so we must simplify the path first.

I added a `path_helpers` module, so that we can handle this more generally in the entire codebase.

rbi/rubydex.rbi
```diff
@@ -381,9 +396,6 @@ class Rubydex::Query
 end
 
 class Rubydex::Graph
```
Morriar, line 398: Let's also add the RBI for `Graph.configure_for_workspace` please. This is now the public entry point used by `rdx` and the MCP server, but Sorbet consumers currently won't see it.

```ruby
class << self
  sig { params(workspace_path: String).returns(Rubydex::Graph) }
  def configure_for_workspace(workspace_path); end
end
```

lib/rubydex/graph.rb
```diff
@@ -7,9 +7,16 @@ module Rubydex
...
 
-    #: (?workspace_path: String?) -> void
-    def initialize(workspace_path: nil)
-      self.workspace_path = workspace_path if workspace_path
+    class << self
+      # Creates a new graph with the loaded configuration. For use cases where the graph must be shared between
+      # different tools, do not use this. Create and own a `Config` object instead.
+      #
+      #: (String) -> instance
+      def configure_for_workspace(workspace_path)
```
Morriar, line 15: Do we need to add this to the `rubydex.rbi` file?

Morriar (approved):

rafaelfranca (approved):

st0012 (approved):

rbi/rubydex.rbi
vinistock: I thought Tapioca would pull in signatures for regular Ruby files. Is that not the case? I added to the rbi regardless.

PR #966, opened 2026-07-30
Move DOT visualization to the Ruby CLI

Morriar:
## What

Move graph visualization from the internal Rust CLI to the public Ruby CLI.

- Expose `Rubydex::Graph#to_dot(show_builtins: false)` through the Rust/C/Ruby boundary.
- Add `rdx dot [PATH]`.
- Add `rdx dot --show-builtins [PATH]`.
- Remove `--dot` and `--show-builtins` from the Rust CLI.
- Keep the existing Rust `DotBuilder` as the underlying renderer.
- Add Ruby API and CLI integration coverage.
- Extract shared helpers for invoking `rdx` from integration tests.
- Document the Ruby API and CLI usage.

## Why

We want to remove the Rust CLI and keep the Ruby `rdx` executable as Rubydex's only CLI entry point. Before removing the Rust CLI, its user-facing features need to be available through `rdx`.

DOT visualization was only exposed by the Rust CLI. This PR moves that feature to the Ruby API and CLI while preserving the existing Rust renderer and output format.

## Usage

Render the current workspace:

```bash
bundle exec rdx dot > graph.dot
```

Render another workspace:

```bash
bundle exec rdx dot path/to/workspace > graph.dot
```

Include Rubydex's built-in declarations:

```bash
bundle exec rdx dot --show-builtins path/to/workspace > graph.dot
```

The same output is available through the Ruby API:

```ruby
graph.to_dot
graph.to_dot(show_builtins: true)
```

Indexing progress is written to stderr so stdout contains only DOT output.

## Testing

- `bundle exec rake test`
- `bundle exec rake lint`
- `cargo fmt --check`
- Ruby DOT API tests
- `rdx dot` integration tests
- MCP integration tests after extracting the shared executable helper

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,32 @@ puts query.render(graph, "json")
 puts Rubydex::Query.schema("table")
 ```
 
+## Visualizing the graph
+
+`rdx dot` renders the complete resolved graph as Graphviz DOT. It includes documents,
+definitions, declarations, nesting, inheritance, and mixin relationships. Built-in
+declarations are hidden by default.
+
+```bash
+# Index the current workspace and write its graph
+bundle exec rdx dot > graph.dot
+
+# Index another workspace and include Rubydex's built-in declarations
+bundle exec rdx dot --show-builtins path/to/workspace > graph.dot
+
+# Render the DOT file with Graphviz
+dot -Tsvg graph.dot -o graph.svg
+```
+
+The same output is available from a resolved graph:
+
+```ruby
+File.write("graph.dot", graph.to_dot)
+File.write("graph-with-builtins.dot", graph.to_dot(show_builtins: true))
+```
+
+DOT is a whole-graph visualization, not a `rdx query --format` option.
+
 ## MCP Server (Experimental)
 
 Rubydex can run as an MCP (Model Context Protocol) server, enabling AI assistants
--- a/exe/rdx
+++ b/exe/rdx
@@ -13,6 +13,7 @@ USAGE = <<~TEXT
                      Use `query --schema` to describe the queryable schema (labels,
                      relationships, properties) without indexing the workspace.
     console          Open an interactive session with a populated graph for the current workspace
+    dot [PATH]       Output a Graphviz DOT visualization (workspace defaults to the current dir)
     mcp [PATH]       Run the MCP server for AI assistants (workspace defaults to the current dir)
     help             Show this help message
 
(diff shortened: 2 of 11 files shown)


rust/rubydex/src/main.rs
```diff
@@ -2,7 +2,6 @@ use clap::{Parser, ValueEnum};
 use std::{fs, mem, path::PathBuf};
 
 use rubydex::{
-    dot,
```
st0012, line 5: Can we delete https://github.com/Shopify/rubydex/blob/main/rust/rubydex/src/dot.rs as well?

exe/rdx
```diff
@@ -106,6 +109,28 @@ operation =
...
+        exit
+      end
+    end
+    begin
+      parser.parse!
+    rescue OptionParser::ParseError => e
+      abort_with_usage(e.message)
+    end
+
+    workspace_path = ARGV.shift || Dir.pwd
```
vinistock, line 128: Note: passing an explicit argument here won't include that workspace's dependencies without setting up the Bundle based on its Gemfile, so this has difference behaviour than pwd.

test/helpers/executable.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "open3"
+require "rbconfig"
+
+module Test
```
vinistock, line 6: Testing behaviour through executables feels weird. I think this is a result of us not having a proper CLI object that can be unit tested yet, which forces us to reach into actually shelling out.

Since we already need to refactor the CLI for the linter, can we avoid adding this? We can add the dot scenarios as unit tests for the CLI as soon as we add it.

README.md
```diff
@@ -143,6 +143,32 @@ puts query.render(graph, "json")
...
+`rdx dot` renders the complete resolved graph as Graphviz DOT. It includes documents,
+definitions, declarations, nesting, inheritance, and mixin relationships. Built-in
+declarations are hidden by default.
+
+```bash
+# Index the current workspace and write its graph
+bundle exec rdx dot > graph.dot
+
+# Index another workspace and include Rubydex's built-in declarations
+bundle exec rdx dot --show-builtins path/to/workspace > graph.dot
```
vinistock, line 157: If we're really going to allow the path argument, we should be explicit that it cannot include dependencies.

rust/rubydex/src/main.rs
Morriar: It's still backing the dot generation

test/helpers/executable.rb
paracycle: This should be streamlined on latest `main` now, since we have the infrastructure to test `rdx` invocations through calls to the `CLI` class.

README.md
paracycle: I also find it interesting that we are indexing a workspace using the gemfile of some other workspace. Do we really need the workspace option?

PR #969, opened 2026-07-30
Fix leaks after string array conversion

Morriar:
Validate scalar string arguments with `StringValueCStr` before allocating converted nesting arrays in:

- `Graph#resolve_constant`
- `Graph#complete_method_argument`

This prevents an embedded-NUL exception from bypassing `rdxi_free_str_array` after ownership has returned to the caller.

### DIFF
--- a/ext/rubydex/utils.c
+++ b/ext/rubydex/utils.c
@@ -4,16 +4,23 @@
 #include "rustbindings.h"
 
 // Convert a Ruby array of strings into a double char pointer so that we can pass that to Rust.
-// This copies the data so it must be freed
+// This copies the data so it must be freed.
 char **rdxi_str_array_to_char(VALUE array, size_t length) {
+    // Validate every element before allocating so StringValueCStr cannot strand a partially built array.
+    for (size_t i = 0; i < length; i++) {
+        VALUE item = rb_ary_entry(array, i);
+        Check_Type(item, T_STRING);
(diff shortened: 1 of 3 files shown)


st0012:
@Morriar Is this PR on the right base branch?

Morriar:
> @Morriar Is this PR on the right base branch?

Rebased on `main` 👍

st0012 (approved):

ext/rubydex/utils.c
```diff
@@ -4,16 +4,23 @@
...
 
     for (size_t i = 0; i < length; i++) {
         VALUE item = rb_ary_entry(array, i);
-        const char *string = StringValueCStr(item);
+        size_t string_length = (size_t)RSTRING_LEN(item);
 
-        converted_array[i] = malloc(strlen(string) + 1);
-        strcpy(converted_array[i], string);
+        converted_array[i] = malloc(string_length + 1);
+        memcpy(converted_array[i], RSTRING_PTR(item), string_length + 1);
```
vinistock, line 23: Why do we need to use `memcpy` instead of `strcpy`?

vinistock (approved):

ext/rubydex/utils.c
Morriar: Because we already have `string_length` so we can just copy the bytes instead of rescanning the string looking for the null terminator

PR #970, opened 2026-07-31
Index RBS attribute methods

soutaro:
Index RBS `attr_reader`, `attr_writer`, and `attr_accessor` members.

RBS attributes are represented as `MethodDefinition`s because the existing attribute-definition variants cannot represent singleton attributes such as `attr_reader self.foo: String`.

This includes instance and singleton attributes, visibility, comments, and `deprecated` annotations. Writers are indexed with one required positional parameter.

### DIFF
--- a/rust/rubydex/src/indexing/rbs_indexer.rs
+++ b/rust/rubydex/src/indexing/rbs_indexer.rs
@@ -3,8 +3,9 @@
 use core::panic;
 
 use ruby_rbs::node::{
-    self, AliasKind, ClassNode, CommentNode, ConstantNode, ExtendNode, FunctionTypeNode, GlobalNode, IncludeNode,
-    ModuleNode, Node, NodeList, PrependNode, TypeNameNode, Visit,
+    self, AliasKind, AttrAccessorNode, AttrReaderNode, AttrWriterNode, AttributeKind, AttributeVisibility, ClassNode,
+    CommentNode, ConstantNode, ExtendNode, FunctionTypeNode, GlobalNode, IncludeNode, ModuleNode, Node, NodeList,
+    PrependNode, TypeNameNode, Visit,
 };
 
 use crate::diagnostic::Rule;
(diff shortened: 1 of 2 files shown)


rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -223,6 +224,111 @@ impl<'a> RBSIndexer<'a> {
...
+            .intern_string(format!("{name}{}()", if writer { "=" } else { "" }));
+        let signatures = if writer {
+            let parameter_name = self.local_graph.intern_string("arg0".to_string());
+            let parameter = Parameter::RequiredPositional(ParameterStruct::new(name_offset.clone(), parameter_name));
+            Signatures::Simple(vec![parameter].into_boxed_slice())
+        } else {
+            Signatures::Simple(Box::new([]))
+        };
+
+        let definition = Definition::Method(Box::new(MethodDefinition::new(
```
vinistock, line 338: We have specific definitions for `AttrWriter`, `AttrReader` and `AttrAccessor`. Also, attributes don't carry a name offset and regular offset, since there's no distinction.

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -223,6 +224,111 @@ impl<'a> RBSIndexer<'a> {
...
+        flags: DefinitionFlags,
+        lexical_nesting_id: Option<DefinitionId>,
+        visibility: Visibility,
+        receiver: Option<Receiver>,
+    ) {
+        let str_id = self
+            .local_graph
+            .intern_string(format!("{name}{}()", if writer { "=" } else { "" }));
+        let signatures = if writer {
+            let parameter_name = self.local_graph.intern_string("arg0".to_string());
```
vinistock: Could we use the name of the attribute as the parameter name? E.g.:

```ruby
def value=(value); end
```

rust/rubydex/src/indexing/rbs_indexer.rs
```diff
@@ -223,6 +224,111 @@ impl<'a> RBSIndexer<'a> {
...
+            self.register_attribute_method(
+                name,
+                false,
+                offset.clone(),
+                name_offset.clone(),
+                comments.clone(),
+                flags.clone(),
+                lexical_nesting_id,
+                visibility,
+                receiver.clone(),
```
vinistock: Here, I assume we are cloning because we need to reuse this below for the writer. However, if there's no writer, we don't really need to clone.

Is there a way to structure this so that we avoid cloning when no writer is required?

vinistock (commented):
Just a few things, but it looks good. I can't believe we forgot attributes on RBS!

rust/rubydex/src/indexing/rbs_indexer.rs
Morriar: Can't you just move the `if writer` over this?

rust/rubydex/src/indexing/rbs_indexer.rs
soutaro: The `Attr*` definitions do not support singleton attribute syntax in RBS, such as `attr_reader self.name: String`. I can extend them if it's better, but I don't think it would be worthwhile.

vinistock: Ah, I see. Yeah, we would need to add a receiver to attributes, but indeed it's not worth it at this time.

I also think that the DSL work might render the `Attr*` definitions unnecessary, so let's not do it right now 👍

vinistock (approved):

alexcrocha (approved):

Morriar (approved):

PR #971, opened 2026-07-31
Extract the rdx CLI into one class per command

paracycle:
## Goal

@Morriar asked on #869 whether that PR needs to split the CLI. He also said that a split should put each command in its own file. This PR does that split on its own. It contains no server work from #869.

Comment: https://github.com/Shopify/rubydex/pull/869#discussion_r3677371112

## What changes

`exe/rdx` contained 150 lines of option parsing and command dispatch. This PR moves that logic into `lib/rubydex/cli.rb`. Each subcommand has its own file.

| File | Content |
|---|---|
| `exe/rdx` | 8 lines. It calls `Rubydex::CLI.start`. |
| `lib/rubydex/cli.rb` | The top-level flags, the dispatch, and the help text. |
| `lib/rubydex/cli/command.rb` | The base class, the DSL, the discovery, and the shared helpers. |
| `lib/rubydex/cli/command/query.rb` | `rdx query` |
| `lib/rubydex/cli/command/console.rb` | `rdx console` |
| `lib/rubydex/cli/command/mcp.rb` | `rdx mcp` |

A subcommand declares itself with a small DSL:

```ruby
class Command
  class Query < Command
    command "query"
    arguments "<CYPHER>"
    summary "Run a Cypher query against the workspace graph and print the result."
  end
end
```

`Command.all` globs `cli/command/*.rb`. It then collects the classes from `Class#subclasses` and sorts them by declared name. `CLI` uses that list for two purposes:

- `CLI` maps a command name to a class.
- `CLI` renders the `rdx help` command list.

A developer adds a subcommand with one new file. This PR removes the dispatch table, the load list, and the manual usage text. Each command also builds its `--help` banner from its own declaration.

`CLI` loads the command files only when it needs them. Therefore `rdx --version` loads no command file and no native extension.

## Behaviour

The commands behave as before, except for three deliberate differences.

### 1. The help list is alphabetical

```diff
 Commands:
+  console          Open an interactive session with a populated graph for the current workspace
+  mcp [PATH]       Run the MCP server for AI assistants (workspace defaults to the current dir)
   query <CYPHER>   Run a Cypher query against the workspace graph and print the result.
                    Use `query --schema` to describe the queryable schema (labels,
                    relationships, properties) without indexing the workspace.
-  console          Open an interactive session with a populated graph for the current workspace
-  mcp [PATH]       Run the MCP server for AI assistants (workspace defaults to the current dir)
   help             Show this help message
```

The `help` entry stays last. The words, the alignment, and the column width do not change. The three `rdx <command> --help` banners are identical.

### 2. `rdx console` no longer exposes the local variables of `exe/rdx`

The old session used a `binding` from the top level of `exe/rdx`. That binding exposed the option parsing variables of the executable:

```
before: [:_, :__, :command, :e, :format, :graph, :handled, :operation,
         :parsed, :parser, :path, :query, :schema]
after:  [:_, :__, :graph]
```

The command now uses `IRB::WorkSpace.new` and adds `graph` as a local variable. The session keeps these properties:

- `self` is `main`.
- `class Foo; end` at the prompt defines `::Foo`.
- The session provides `graph`.

The old exposure was accidental. If a class contains the old `binding` expression, `self` becomes the command object instead of `main`. `IRB::WorkSpace.new` keeps `self` as `main` and drops only the extra variables.

### 3. Every command rejects a bad option the same way

`mcp` reported a bad option with the usage text. `query` and `console` raised `OptionParser::InvalidOption` and printed a Ruby backtrace. `Command#parse_options!` now handles `OptionParser::ParseError` for every command:

```
$ rdx query --bogus-flag
invalid option: --bogus-flag
```

This also covers a bad value for `--format`.


lib/rubydex/cli/command.rb
```diff
@@ -0,0 +1,156 @@
...
+          yield(p) if block_given?
+          p.on("-h", "--help", "Show this help") do
+            puts(p)
+            exit
+          end
+        end
+
+        parser.parse!(argv)
+      rescue OptionParser::ParseError => e
+        abort_with_usage(e.message)
```
Morriar, line 134: I wonder if we should print the command-specific parser here instead of routing through `CLI.abort_with_usage`.

At this point the command has already been resolved and `parser` knows about flags like `--schema` / `--format`, but `abort_with_usage` rebuilds the top-level usage. So `rdx query --format yaml` tells the user to run `rdx <command> --help`, but does not show the valid formats/options that failed. It also forces `usage` to load every command file for a subcommand parse error.

Maybe we can keep this local to the parser:

```ruby
rescue OptionParser::ParseError => e
  warn(e.message)
  warn("")
  warn(parser)
  exit(1)
end
```

st0012 (approved):

st0012:
For the CI failure, it looks like we want to add `fiddle` to Gemfile too?

paracycle:
> For the CI failure, it looks like we want to add `fiddle` to Gemfile too?

Honestly, I am not sure if we should add a hard dependency on `irb` for Rubydex for a use case that might not ever be exercised. I am not sure why only the Windows 4.0 tests are failing, btw.

test/cli_test.rb
```diff
@@ -0,0 +1,280 @@
...
+  end
+
+  private
+
+  #: (Array[String] args, ?chdir: String?, ?stdin: String?) -> [String, String, Process::Status]
+  def rdx(args, chdir: nil, stdin: nil)
+    options = {} #: Hash[Symbol, untyped]
+    options[:chdir] = chdir if chdir
+    options[:stdin_data] = stdin if stdin
+    Open3.capture3(RbConfig.ruby, "-I", LIB_DIR, EXE, *args, **options)
```
vinistock: I suspect the Windows failure is related to not including everything for `-I`. We had similar pain in the Ruby LSP.

Try adding the entire `$LOAD_PATH` of the parent process manually for the child. Like [this](https://github.com/Shopify/ruby-lsp/blob/218030210e7cad6c0e4841dcc684849dad8bc279/exe/ruby-lsp-launcher#L52) and [this](https://github.com/Shopify/ruby-lsp/blob/218030210e7cad6c0e4841dcc684849dad8bc279/exe/ruby-lsp-launcher#L61).

That said, if we can use unit tests that don't shell out, I think that would be better. The executable is now trivial: it just calls `start` on the CLI.

If there's a way to test the commands directly, even if we need to use `assert_output`, I think that's better than shelling out.

vinistock (approved):

st0012:
@paracycle I don't think I proposed adding `irb` as a hard dependency? It's already in `Gemfile` so we should be good as it's mostly for our own uses. 
For `fiddle`, we need it because `reline` relies on `fiddle` ONLY on windows, but does not declare it as a runtime dependency. So we only hit that warning on windows. Why 4.0 tho I think it may be related when gems are promoted to bundled gems. But I haven't checked

paracycle:
> @paracycle I don't think I proposed adding `irb` as a hard dependency? It's already in `Gemfile` so we should be good as it's mostly for our own uses. For `fiddle`, we need it because `reline` relies on `fiddle` ONLY on windows, but does not declare it as a runtime dependency. So we only hit that warning on windows. Why 4.0 tho I think it may be related when gems are promoted to bundled gems. But I haven't checked

Ah ok, sorry, I misunderstood what you wrote. Thanks for the explainer.

test/cli_test.rb
paracycle: Ok, I've done the refactor to call the `CLI` class in Ruby for testing CLI behaviour.

PR #973, opened 2026-08-03
Add severity and related information to diagnostics

st0012:
This is part of the preparation work to bring Rubydex linter into this project. This PR updates Diagnostic class to match the same fields as what the linter needs.


rust/rubydex/src/indexing/local_graph.rs
```diff
@@ -196,7 +196,7 @@ impl LocalGraph {
     }
 
     pub fn add_diagnostic(&mut self, rule: Rule, offset: Offset, message: String) {
-        let diagnostic = Diagnostic::new(rule, self.uri_id, offset, message);
+        let diagnostic = Diagnostic::new(rule, Severity::Information, self.uri_id, offset, message);
```
Morriar, line 199: Now that severity is part of the public diagnostic shape, I wonder if we should avoid hardcoding every Rust diagnostic to `Information`. `Rule::ParseError` and `Rule::ParseWarning` already imply different severity levels, and LSP consumers would render syntax errors as low-priority info diagnostics with this default.

Could we either make severity explicit at the diagnostic call sites, or add something like `Rule::default_severity()` and use it here and in `RubyOperationBuilder#add_diagnostic`?

rbi/rubydex.rbi
```diff
@@ -300,18 +300,58 @@ class Rubydex::Include < Rubydex::Mixin; end
...
 
+module Rubydex::Severity
+  sig { params(value: Symbol).returns(T.class_of(Rubydex::Severity::Base)) }
+  def self.from_value(value); end
+end
+
+class Rubydex::Severity::Base
+  sig { returns(Symbol) }
+  def self.value; end
+end
```
alexcrocha, line 313:
```suggestion
class Rubydex::Severity::Base
  abstract!
  
  sig { returns(Symbol) }
  def self.value; end
end
```

alexcrocha (approved):

soutaro (approved):
The code looks good overall.

I'm not sure whether reporting all existing graph diagnostics as `Information` is the right choice. Users may find it surprising that Ruby or RBS syntax errors are reported at `Information` severity.

(Perhaps this is something you're already considering for a follow-up PR—for example, defining a default severity for each diagnostic and making those defaults configurable?)

st0012:
@soutaro I'm not sure how to do with graph diagnostics tbh. My assumption is that projects don't usually have invalid code. And even if they do, it's unlikely it's in user's own program. So blocking them because of it would be silly.
Should we just not show them at all?

rbi/rubydex.rbi
st0012: Good catch, thanks!

rust/rubydex/src/diagnostic.rs
```diff
@@ -5,16 +5,18 @@ use crate::{model::ids::UriId, offset::Offset};
 #[derive(Debug)]
 pub struct Diagnostic {
     rule: Rule,
+    severity: Severity,
```
Morriar, line 8: I'd add a comment on the severity here explaining why it is not in the Rule: ultimately we want to make diagnostics severity configurable so the rule explains the problem, the "reporter" of the diagnostic choose how important it is.

rust/rubydex/src/resolution.rs
```diff
@@ -635,6 +635,7 @@ impl<'a> Resolver<'a> {
                     } else {
                         let diagnostic = Diagnostic::new(
                             Rule::UndefinedConstantVisibilityTarget,
+                            Severity::Information,
```
Morriar, line 638: I wonder if we shouldn't bump these diagnostics to `warning`

st0012:
@Morriar Addressed them in https://github.com/Shopify/rubydex/pull/985

PR #974, opened 2026-08-03
Add a configurable Rubydex linter with rule discovery

st0012:
Add an in-gem `Rubydex::Linter` framework and `rdx lint [PATH]` for running semantic rules against a workspace. The runner combines native and rule diagnostics, preserves related information, produces deterministic output, and fails only for error diagnostics.

The command discovers rules from the workspace and active Bundler dependencies, applies `rubydex.toml` settings, skips disabled rules, and warns about configured rules that were not loaded.

### DIFF
--- a/lib/rubydex/cli/command.rb
+++ b/lib/rubydex/cli/command.rb
@@ -135,21 +135,24 @@ module Rubydex
       end
 
       # Builds the workspace graph, sending progress messages to `progress_io`.
-      #: (IO progress_io) -> Rubydex::Graph
-      def build_graph(progress_io)
-        graph = Rubydex::Graph.configure_for_workspace(Dir.pwd)
-        with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+      #: (IO progress_io, ?workspace_path: String, ?fail_on_index_errors: bool) -> Rubydex::Graph
+      def build_graph(progress_io, workspace_path: Dir.pwd, fail_on_index_errors: false)
+        graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+        errors = with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+        abort(errors.join("\n")) if fail_on_index_errors && !errors.empty?
+
         with_timer(progress_io, "Resolving graph...") { graph.resolve }
         graph
       end
 
-      #: (IO io, String message) { -> void } -> void
+      #: [T] (IO io, String message) { -> T } -> T
       def with_timer(io, message)
         io.print(message)
         start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
-        yield
+        result = yield
         duration = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond) - start
         io.puts(" finished in #{duration.round(2)}ms")
+        result
       end
     end
   end
--- a/lib/rubydex/cli/command.rb
+++ b/lib/rubydex/cli/command.rb
@@ -135,9 +135,15 @@ module Rubydex
       end
 
       # Builds the workspace graph, sending progress messages to `progress_io`.
-      #: (IO progress_io, ?workspace_path: String, ?fail_on_index_errors: bool) -> Rubydex::Graph
-      def build_graph(progress_io, workspace_path: Dir.pwd, fail_on_index_errors: false)
-        graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+      #: (IO progress_io, ?workspace_path: String, ?config: Rubydex::Config, ?fail_on_index_errors: bool) -> Rubydex::Graph
+      def build_graph(
+        progress_io,
+        workspace_path: Dir.pwd,
+        config: Rubydex::Config.load(workspace_path),
+        fail_on_index_errors: false
+      )
+        graph = Rubydex::Graph.new
+        graph.load_config(config)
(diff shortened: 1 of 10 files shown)


exe/rdx
```diff
@@ -49,19 +50,56 @@ command = ARGV.shift
...
+def load_linter_rules(rule_files, workspace_path)
+  existing_rules = Rubydex::Linter::Rule.subclasses
+
+  rule_files.each do |rule_file|
+    require File.expand_path(rule_file, workspace_path)
+  rescue LoadError, SyntaxError => error
+    abort("Unable to load linter rules from #{rule_file}: #{error.message}")
+  end
+
+  Rubydex::Linter::Rule.subclasses - existing_rules
```
Morriar: I wonder if we should collect descendant rule classes here instead of only direct subclasses. `Class#subclasses` only returns immediate children, so a project-local shared base class would be treated as the loaded rule and the concrete rules below it would be ignored:

```ruby
class ApplicationRule < Rubydex::Linter::Rule
end

class NoFoo < ApplicationRule
end
```

In that setup the runner would instantiate `ApplicationRule` and hit the abstract `lint` / `severity` methods, so `rdx lint` aborts before running `NoFoo`. Could we recurse through subclasses, or make rule files register the concrete classes explicitly?

lib/rubydex/linter/runner.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        Result.new(diagnostics)
+      end
+
+      private
+
+      #: (Diagnostic) -> bool
+      def diagnostic_in_workspace?(diagnostic)
+        path = URI::RFC2396_PARSER.unescape(diagnostic.location.to_file_path)
+        relative_path = Pathname.new(File.realpath(path)).relative_path_from(Pathname.new(@graph.workspace_path))
```
Morriar: Should we compare the workspace-relative expanded path before resolving symlinks, or otherwise use the same path spelling that was used when listing/indexing the file?

The realpath comparison will drop diagnostics for files that Rubydex intentionally indexed through a symlinked directory under the workspace.

* `Graph#workspace_paths` adds symlinked directory children as roots, and the listing code indexes files under the requested symlink path. 
* `File.realpath(path)` resolves the diagnostic path to the symlink target, so a workspace path like `workspace/link/foo.rb` can become `/outside/foo.rb` and fail the `relative_path_from(@graph.workspace_path)` check.

lib/rubydex/cli/command.rb
```diff
@@ -135,21 +135,24 @@ def parse_options!(options: false)
...
+        graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+        errors = with_timer(progress_io, "Indexing workspace...") { graph.index_workspace }
+        abort(errors.join("\n")) if fail_on_index_errors && !errors.empty?
+
         with_timer(progress_io, "Resolving graph...") { graph.resolve }
         graph
       end
 
-      #: (IO io, String message) { -> void } -> void
+      #: (IO io, String message) { -> untyped } -> untyped
```
soutaro:
```suggestion
      #: [T] (IO io, String message) { -> T } -> T
```

soutaro:
It looks like `RuleConfig` isn’t consumed by the linter in this PR, and there doesn’t seem to be a standard way for a rule to access its configuration. Is that integration planned for a follow-up PR?

st0012:
@soutaro Yes I split the work into multiple small PRs to speed up review

st0012:
@soutaro I decided to expand a PR quite a bit so now config and automatic rule loading is part of the PR too.

lib/rubydex/cli/command/lint.rb
```diff
@@ -0,0 +1,99 @@
...
+            rule_files.concat(Gem.find_latest_files(RULE_GLOB))
+          end
+
+          rule_files.each do |rule_file|
+            require rule_file
+          rescue LoadError, SyntaxError => error
+            abort("Unable to load linter rules from #{rule_file}: #{error.message}")
+          end
+
+          Rubydex::Linter::Rule.subclasses - existing_rules
```
Morriar, line 70: I wonder if someone will one day try to create a custom rule by subclassing another rule? 🤔

st0012: The linter in Core currently does this as well as we haven't seen a need to create abstract rule classes. I think we can design this later and finish the porting first. Will create an issue for this.

Issue: https://github.com/Shopify/rubydex/issues/984

lib/rubydex/linter/runner.rb
st0012: Good catch, fixed 👍

alexcrocha (approved):
Both the rule discovery and config wiring look good to me 👍 

I do have a question about merging the graph diagnostics with the rule diagnostics, mostly to shape my mental model on the concepts of linting rules and diagnostics. I know for the output they're both diagnostics, but are they the same concept that surfacing both through the linter is the natural output?

st0012:
I think there's not a clear answer to it. But I think linter is where we should try surfacing them because:
- If you think about it, the graph diagnostics are the "rules" we made that we think the codebase should not follow.
- Among all the tools we've planned so far, linter is the only place that's suitable to surface them.

And if we found these diagnostics noisy later, we can:
- Review whether our diagnostics make sense at all. Perhaps we can remove some
- Provide a config to opt out

For now I'll keep them, but it'd be easy to drop as well.

PR #975, opened 2026-08-03
Improve linter infrastructure

st0012:
Most of this PR ports existing production-rule infrastructure from Core into `Rubydex::Linter`. It preserves the rule and path helpers, diagnostic output, severity overrides, exclusions, and project and in-repo-gem rule discovery that Core uses today. The main Rubydex-specific change adapts this configuration behavior to `rubydex.toml` and the existing Rubydex configuration API.

This PR adds shared infrastructure only. It does not port Core rules or add list and explain commands, rule-test helpers, or Ruby LSP integration.

### DIFF
--- a/ext/rubydex/config.c
+++ b/ext/rubydex/config.c
@@ -1,4 +1,5 @@
 #include "config.h"
+#include "diagnostic.h"
 #include "rustbindings.h"
 #include "utils.h"
 
@@ -29,9 +30,19 @@ static VALUE config_linter_build(VALUE opaque_rule_array) {
     for (size_t i = 0; i < rule_array->len; i++) {
         CLinterRule rule = rule_array->items[i];
         VALUE rule_name = rb_str_freeze(rb_utf8_str_new(rule.name, (long)rule.name_length));
-        VALUE argv[] = {rule_name, rule.enabled ? Qtrue : Qfalse};
+        VALUE exclude_patterns = rb_ary_new_capa((long)rule.exclude_patterns_length);
+        for (size_t j = 0; j < rule.exclude_patterns_length; j++) {
+            CConfigString pattern = rule.exclude_patterns[j];
+            rb_ary_push(exclude_patterns, rb_str_freeze(rb_utf8_str_new(pattern.data, (long)pattern.length)));
+        }
+        rb_obj_freeze(exclude_patterns);
 
-        rb_hash_aset(rules, rule_name, rb_class_new_instance(2, argv, cRuleConfig));
+        VALUE severity = rule.severity == NULL
+            ? Qnil
+            : rdxi_build_diagnostic_severity_value(mRubydex, *rule.severity);
+        VALUE argv[] = {rule_name, rule.enabled ? Qtrue : Qfalse, exclude_patterns, severity};
+
+        rb_hash_aset(rules, rule_name, rb_class_new_instance(4, argv, cRuleConfig));
     }
 
     return rb_class_new_instance(1, &rules, cLinterConfig);
--- a/lib/rubydex/cli/command/lint.rb
+++ b/lib/rubydex/cli/command/lint.rb
@@ -32,7 +32,12 @@ module Rubydex
 
           graph = build_graph($stderr, workspace_path:, config:, fail_on_index_errors: true)
           result = Rubydex::Linter::Runner.new(graph, rules:, config: config.linter).run
-          result.diagnostics.each { |diagnostic| puts(format_linter_diagnostic(diagnostic)) }
+          if result.diagnostics.empty?
+            print_summary(graph.documents.count, result.diagnostics)
+            return
+          end
+
+          print_offenses(result.diagnostics, graph)
           exit(1) unless result.success?
         end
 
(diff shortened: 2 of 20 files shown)


Morriar (commented):
Can we start by documenting the feature in the README? 

It will make it easier to connect all the pieces and review the code.

ext/rubydex/config.c
```diff
@@ -29,9 +30,19 @@ static VALUE config_linter_build(VALUE opaque_rule_array) {
     for (size_t i = 0; i < rule_array->len; i++) {
         CLinterRule rule = rule_array->items[i];
         VALUE rule_name = rb_str_freeze(rb_utf8_str_new(rule.name, (long)rule.name_length));
-        VALUE argv[] = {rule_name, rule.enabled ? Qtrue : Qfalse};
-
-        rb_hash_aset(rules, rule_name, rb_class_new_instance(2, argv, cRuleConfig));
+        VALUE exclude_patterns = rb_ary_new_capa((long)rule.exclude_patterns_length);
```
dersam, line 33: Does every config need to pass through to the C/Rust layer? For configs that are only used at the Ruby layer, why not have the Ruby layer be responsible for them?

I'm mainly thinking of when we start introducing tool specific configs that the indexing algorithms don't need to know about at all.

lib/rubydex/linter/helpers/path_helpers.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+require "pathname"
+require "uri"
+
+module Rubydex
+  module Linter
+    module Helpers
```
dersam, line 8: Nit: I really don't like calling modules "Helpers", it doesn't tell me much about what I find in them and it turns into a dumping ground. Maybe `PathMatchers` for this one?

lib/rubydex/linter/rule.rb
```diff
@@ -34,12 +36,67 @@ def severity
...
+      #: (String) -> String
+      def path_for_uri(uri)
+        file_location(uri).to_file_path
+      rescue Location::NotFileUriError
+        uri
+      end
+
+      # Returns every class inheriting from +base_name+, excluding the base class itself.
+      #: (String) -> Enumerable[Rubydex::Class]
+      def child_classes(base_name)
```
dersam, line 58: This (and some of the other util methods) seems strange to put in the rule class. Maybe this could be a module that a rule can choose to include if it's needed?

ext/rubydex/config.c
st0012: The `enabled` and `exclude` can also be used to graph level diagnostics later to give users the same level of control. So I'd not split configs at different language layers just yet.

lib/rubydex/linter/helpers/path_helpers.rb
st0012: This module is actually called `PathHelpers` and `Helpers` is a namespace to group helper modules. 
It's still not the best name but `PathMatchers` is too narrow as this module also has `display_path`.

lib/rubydex/linter/rule.rb
st0012: We can extract some of the methods out but not these graph related ones as 19/23 rules use one or more of these methods.
I'll extract methods like `source_for_location` instead.

dersam: Works for me. Nothing stopping us from extracting more in the future if the class gets fat.

lib/rubydex/cli/command/lint.rb
```diff
@@ -51,48 +54,102 @@ def warn_unknown_rules(config, known_rule_classes)
...
 
           "#{path}:#{display_location.start_line}:#{display_location.start_column}"
         end
 
-        #: (Diagnostic diagnostic) -> String
-        def format_linter_diagnostic(diagnostic)
-          content = +"#{format_linter_location(diagnostic.location)}: " \
+        #: (Diagnostic diagnostic, workspace_path: String) -> String
+        def format_linter_diagnostic(diagnostic, workspace_path:)
+          content = +"#{format_linter_location(diagnostic.location, workspace_path:)}: " \
```
dersam, line 88: Why is this string prefixed with a `+`?

lib/rubydex/linter/helpers/source_access_helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+        def path_for_uri(uri)
+          file_location(uri).to_file_path
+        rescue Location::NotFileUriError
+          uri
+        end
+
+        private
+
+        #: (Location) -> String?
+        def source_for_location(location)
```
dersam, line 23: This method is private but not used anywhere.

st0012: Yes, because we don't have any rules here 😅 It'll be used by Core in 3 rules.

lib/rubydex/cli/command/lint.rb
st0012: Because with `frozen_string_literal: true` at the top of the file, string literals are frozen. But in this case, we want to mutate it later, so `+` means this literal is not frozen.

dersam: TIL

dersam (approved):

Morriar (approved):

PR #976, opened 2026-08-03
Add linter rule test support

st0012:
Expose `Rubydex::Linter::RuleTestCase` for rule tests with file-backed graphs, multi-file sources, shared sources, and inline diagnostic annotations. Related information is matched at its own location, and missing graph dependencies can be asserted.

The helper loads only when required, so `rubydex/linter` does not load Minitest.


st0012:
This work is now folded into #988.

PR #980, opened 2026-08-05
Print command-specific usage for CLI errors

Morriar:
## Why

Follow-up to https://github.com/Shopify/rubydex/pull/971#discussion_r3693074796.

Once a subcommand has been resolved, option and argument errors should show that command's parser rather than rebuilding the top-level usage. This gives users the valid command-specific flags and avoids loading every command file while reporting a parse error.

## What changed

- Print each command's `OptionParser` for invalid options and arguments.
- Keep unknown-command errors on the top-level CLI usage path.
- Reuse the parsed command parser for validation after option parsing.

cc @paracycle

### DIFF
--- a/lib/rubydex/cli/command.rb
+++ b/lib/rubydex/cli/command.rb
@@ -106,7 +106,10 @@ module Rubydex
 
       #: (String message) -> void
       def abort_with_usage(message)
-        CLI.abort_with_usage(message)
+        warn(message)
+        warn("")
+        warn(@parser)
+        exit(1)
       end
 
       # Parses this command's options out of `argv`, with a banner derived from the command's own
@@ -120,7 +123,7 @@ module Rubydex
         banner = +"Usage: rdx #{self.class.usage_form}"
         banner << " [options]" if options
 
-        parser = OptionParser.new do |p|
+        @parser = OptionParser.new do |p|
           p.banner = banner
           yield(p) if block_given?
           p.on("-h", "--help", "Show this help") do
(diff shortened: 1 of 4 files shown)


lib/rubydex/cli/command.rb
```diff
@@ -130,8 +133,9 @@ def parse_options!(options: false)
         end
 
         parser.parse!(argv)
+        parser
```
paracycle: Can we not store the parser in an ivar? That way we don't have to return it to store in a local var just to pass it back to `abort_with_error` again. Since `Command` class is the only consumer of the parser, it can hold on to it in an instance variable.

Morriar: Updated: the command now stores its parser in `@parser`, and `parse_options!` no longer returns it for callers to pass back into `abort_with_usage`.

lib/rubydex/cli/command/query.rb
```diff
@@ -37,7 +37,8 @@ def run
             return
           end
 
-          abort_with_usage("`query` requires a Cypher query argument (or pass `--schema`)") if query.nil? || query.empty?
+          abort_with_usage("`query` requires a Cypher query argument (or pass `--schema`)") if
+            query.nil? || query.empty?
```
paracycle: Super nitpick: I'd rather we use the block from of the `if` if we are going to span 2 lines:

```suggestion
          if query.nil? || query.empty?
            abort_with_usage("`query` requires a Cypher query argument (or pass `--schema`)") 
          end
```

paracycle (approved):
Thanks

PR #982, opened 2026-08-05
Add Rubydex skill library with CLI command

dersam:
## Summary

Adds a skill system that lets the `rdx` CLI load and display skill files — Markdown documents with YAML frontmatter that encode remediation guidance for linter rules.

This is the first part of a system that will allow Rubydex tools to provide references to detailed skills for fixing issues, without having to load every skill. The linter finding is the trigger point for skill loading, rather than relying on the agent pulling from context.

Screenshot 2026-08-05 at 3 58 27 PM

Skill files live at `skills/<id>/SKILL.md` and use YAML frontmatter to declare a `name` and `description`. This is the standard agent skill format.

- `rdx skill` — lists available skills with id and description, aligned in columns
- `rdx skill <id>` — loads and prints the skill body
- `rdx skill <unknown id>` — reports `unknown skill: <id>` on stderr, exits non-zero

The skills directory is resolved relative to the source file (`__dir__`), so it works both in development and when installed as a gem. Skills are also added to the gemspec file list.

The diagram below illustrates the intended use of skill ids in a Rubydex tooling loop.

Screenshot 2026-08-05 at 3 58 52 PM

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,6 +32,9 @@ the Ruby VM to the Rust crate logic.
 - `ext/rubydex`: The C native extension that connects the Ruby VM with the Rust crate logic through FFI
 - `lib`: The rest of the Ruby code
 - `test`: Ruby test files
+- `skills`: the skill library for deterministic skill loading. These are NOT for developing on Rubydex and should not be loaded unless explicitly requested. Each `skills/<id>/SKILL.md` is a Markdown playbook with YAML frontmatter (`name`,
+`description`) that `rdx skill <id>` prints for an AI agent. The directory name is the id and the frontmatter `name`
+must repeat it. Add a skill by adding the directory and the file; nothing else registers it
 
 ### Naming Conventions
 
--- a/lib/rubydex/cli/command/skill.rb
+++ b/lib/rubydex/cli/command/skill.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require "rubydex/cli/command"
+
+module Rubydex
+  module CLI
+    class Command
+      # `rdx skill [ID]` — lists available skill ids, or loads and prints a skill's body by id.
+      class Skill < Command
+        command "skill"
+        arguments "[ID]"
+        summary <<~TEXT
+          Load a Rubydex skill from the library by id. With no id, list the available skills.
+        TEXT
+
+        #: -> void
+        def run
+          parse_options!
+
+          id = argv.shift
+          abort_with_usage("unexpected argument: #{argv.first}") unless argv.empty?
+
+          id ? print_skill(id) : list_skills
+        rescue Rubydex::SkillError => e
+          abort(e.message)
+        end
+
+        private
+
+        #: -> void
+        def list_skills
+          ids = registry.ids
+
+          if ids.empty?
+            puts(available_skills)
+            return
+          end
+
+          ids.each { |id| puts(id) }
+        end
+
+        #: (String id) -> void
+        def print_skill(id)
+          puts(registry.fetch(id).body)
+        rescue Rubydex::UnknownSkillError => e
+          abort("#{e.message}. #{available_skills}")
+        end
+
+        #: -> String
+        def available_skills
+          ids = registry.ids
+
+          ids.empty? ? "No skills are available." : "Available skills: #{ids.join(", ")}"
+        end
+
+        #: -> Rubydex::SkillRegistry
+        def registry
+          require "rubydex/skill_registry"
+          @registry ||= Rubydex::SkillRegistry.load(skills_directory)
+        end
+
+        #: -> String
+        def skills_directory
+          File.expand_path("../../../../skills", __dir__)
+        end
+      end
+    end
+  end
+end
--- a/lib/rubydex/errors.rb
+++ b/lib/rubydex/errors.rb
@@ -9,4 +9,16 @@ module Rubydex
   # Raised by `Config.load` when the workspace does not exist, or when its config file cannot be read or is malformed.
   # A workspace with no config file at all is not an error.
   class ConfigError < Error; end
+
+  # Raised by `Skill.load` when the file cannot be read, is missing frontmatter, has malformed YAML, or lacks the
+  # required `name` or `description` fields.
+  # Every failure to produce a Skill is a `SkillError`, so a caller rescues one class instead
+  # of the union of what `File.read` and Psych happen to raise.
+  class SkillError < Error; end
+
+  # Raised by `SkillRegistry#fetch` when the library has no skill under the requested id.
+  class UnknownSkillError < SkillError; end
+
+  # Raised by `SkillRegistry.load` when the skill directory doesn't exist.
+  class UnknownSkillDirectoryError < SkillError; end
 end
--- a/lib/rubydex/skill.rb
+++ b/lib/rubydex/skill.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+require "rubydex/errors"
+
+module Rubydex
+  # A skill file loaded from disk. Skill files are Markdown documents with YAML frontmatter
+  # that declare a `name` and `description`. The body is the Markdown content after the frontmatter.
+  class Skill
+    #: String
+    attr_reader :name
+
+    #: String
+    attr_reader :description
+
+    #: String
+    attr_reader :body
+
+    class << self
+      #: (String path) -> Rubydex::Skill
+      def load(path)
+        from_content(File.read(path), path:)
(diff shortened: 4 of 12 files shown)


lib/rubydex/cli/command/skill.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+require "rubydex/cli/command"
+require "rubydex/skill_registry"
```
Morriar: Let's keep this require lazy:

```ruby
require "rubydex/cli/command"

# ...

def registry
  require "rubydex/skill_registry"

  @registry ||= Rubydex::SkillRegistry.load(skills_directory)
end
```

lib/rubydex/skill.rb
```diff
@@ -0,0 +1,91 @@
...
+          raise SkillError, "Missing YAML frontmatter delimiter#{location(path)}"
+        end
+
+        closing_index = (1...lines.length).find { |i| lines[i].strip == "---" }
+        unless closing_index
+          raise SkillError, "Unterminated YAML frontmatter#{location(path)}"
+        end
+
+        frontmatter = lines[1...closing_index].join
+        body = lines[(closing_index + 1)..].join.strip
```
st0012, line 47: Should we strip the body? Indentation is common and helpful in markdown. And although I don't think our team uses it, you can create code blocks using indentation too

st0012 (commented):
I think we should validate all skill files (maybe via `skill list`) on CI?

lib/rubydex/cli/command/skill.rb
```diff
@@ -0,0 +1,103 @@
...
+          parse_options!
+
+          id = argv.shift
+          abort_with_usage("unexpected argument: #{argv.first}") unless argv.empty?
+
+          if id.nil?
+            list_skills
+          else
+            print_skill(id)
+          end
```
paracycle: This is my pet peeve, no need to use a `.nil?` call when the object itself is falsy or truthy:

```suggestion
          if id
            print_skill(id)
          else
            list_skills
          end
```

lib/rubydex/skill.rb
paracycle: I think this is just stripping the leading and trailing whitespace of the "whole" body, not each line, but maybe we don't really need it.

lib/rubydex/skill_registry.rb
```diff
@@ -0,0 +1,68 @@
...
+    #: (Hash[String, String] paths) -> void
+    def initialize(paths)
+      @paths = paths.freeze
+      # Deliberately mutable inside a frozen registry: `freeze` protects the set of skills, while
+      # the cache only remembers what was already read from those files.
+      @cache = {} #: Hash[String, Skill]
+      freeze
+    end
+
+    #: () -> Array[String]
```
Morriar:
```suggestion
    #: -> Array[String]
```

lib/rubydex/cli/command/skill.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        #: -> Rubydex::SkillRegistry
+        def registry
+          @registry ||= Rubydex::SkillRegistry.load(skills_directory)
+        end
+
+        #: -> String
+        def skills_directory
+          File.expand_path("../../../../skills", __dir__)
```
Morriar, line 64: This is going to be a problem if we expect people to create their own rules & skills?

Morriar (commented):
Let's document this skill approach in the README.

lib/rubydex/cli/command/skill.rb
dersam: For simplicity, this first iteration is restricted to skills provided by Rubydex. A followup would introduce config and rule definitions that can hook into user or gem provided skills.

lib/rubydex/skill.rb
dersam: Correct, this is only removing trailing whitespace. The body formatting should be otherwise untouched.

dersam:
Updated. I've added an experimental section for skills to the README.

I decided not to add the project/gem provided skills yet. I think we should experiment with this on the rdx linter before investing in extensibility - the current state doesn't close the door to it.

> I think we should validate all skill files (maybe via `skill list`) on CI?

I added a test that parses all the skill files.

lib/rubydex/cli/command/skill.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+
+          ids.each { |id| puts(id) }
+        end
+
+        #: (String id) -> void
+        def print_skill(id)
+          puts(registry.fetch(id).body)
+        rescue Rubydex::UnknownSkillError => e
+          # What to offer instead is this interface's answer to an unknown id, not the registry's.
```
st0012: This comment is not helpful IMO

lib/rubydex/cli/command/skill.rb
```diff
@@ -0,0 +1,71 @@
...
+          abort("#{e.message}. #{available}")
+        end
+
+        # Names the skills the library does have, for a caller that asked for one it does not.
```
st0012: What does the caller refer to here? The 2nd half of this is confusing

lib/rubydex/cli.rb
```diff
@@ -39,6 +39,16 @@ def abort_with_usage(message)
         exit(1)
       end
 
+      # Renders one entry of a command-style list, aligning any continuation lines of `summary`
+      # under the first. Public because a subcommand that prints its own list — `rdx skill` and its
+      # skill library — lines its entries up the same way `rdx help` does.
```
st0012: > Public because a subcommand that prints its own list — `rdx skill` and its
      # skill library — lines its entries up the same way `rdx help` does.

I don't get this 😅

lib/rubydex/errors.rb
```diff
@@ -9,4 +9,14 @@ class AliasCycleError < Error; end
   # Raised by `Config.load` when the workspace does not exist, or when its config file cannot be read or is malformed.
   # A workspace with no config file at all is not an error.
   class ConfigError < Error; end
+
+  # Raised by `Skill.load` when the file cannot be read, is missing frontmatter, has malformed YAML, or lacks the
+  # required `name` or `description` fields, and by `SkillRegistry` when the skill directory does not exist, or when
+  # a skill's `name` disagrees with the directory it lives in.
```
st0012: Reading the comment, I feel we probably want another error class for the `SkillRegsitry` cases?

lib/rubydex/skill.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    #: String
+    attr_reader :body
+
+    class << self
+      #: (String path) -> Rubydex::Skill
+      def load(path)
+        from_content(File.read(path), path:)
+      rescue SystemCallError, IOError => e
+        # Every failure to produce a Skill is a `SkillError`, so a caller rescues one class instead
```
st0012: This comment is more valuable directly above `SkillError` class IMO

lib/rubydex/skill_registry.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "rubydex/skill"
+
+module Rubydex
+  # The skills available on disk, keyed by id. `load` scans a directory for `<id>/SKILL.md` entries
+  # and records their paths; a skill file is only read when {#fetch} asks for it, so resolving one
+  # skill does not pay for the whole library. Listing every description does read them all, since a
+  # description only exists in the file.
```
st0012:
```suggestion
  # The skills available on disk, keyed by id. `load` scans a directory for `<id>/SKILL.md` entries
  # and records their paths. 
  # A skill file is only read when {#fetch} asks for it. But listing every description reads all skill files.
```

lib/rubydex/cli.rb
dersam: I had deleted this usage, forgot to remove the comment.

st0012 (approved):

AGENTS.md
```diff
@@ -32,6 +32,9 @@ the Ruby VM to the Rust crate logic.
 - `ext/rubydex`: The C native extension that connects the Ruby VM with the Rust crate logic through FFI
 - `lib`: The rest of the Ruby code
 - `test`: Ruby test files
+- `skills`: the skill library. Each `skills/<id>/SKILL.md` is a Markdown playbook with YAML frontmatter (`name`,
```
Morriar: Should we make it clear to the agents that those are **not** skills to use while working on Rubydex?

dersam: Yes, good idea.

dersam: Done.

Morriar (approved):

PR #986, opened 2026-08-06
Add `rdx lint explain` command to show rule documentation

st0012:
Add `rdx lint explain <Unqualified Rule Name>` for showing each Rule class's comments:

```
$ be rdx lint explain RuleStructure
Rubydex::Linter::Rules::RuleStructure

Ensures discovered workspace linter rules follow these conventions:

- A rule file does not define more than one linter rule.
- Each rule subclass outside a test directory is in a rule directory.
- Each checked rule subclass is in the `Rubydex::Linter::Rules` namespace.

The rule directories are `rubydex_linter/rules/` and `lib/rubydex_linter/rules/`.
This rule does not report files in those directories that define no rule subclass.
```

### DIFF
--- a/lib/rubydex/cli/command/lint/explain.rb
+++ b/lib/rubydex/cli/command/lint/explain.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+require "rubydex"
+require "rubydex/linter/rule_loader"
+
+module Rubydex
+  module CLI
+    class Command
+      class Lint
+        # `rdx lint explain <RULE>` — prints documentation for rules with a matching name.
+        class Explain < Command
+          BASE_RULE_NAME = "Rubydex::Linter::Rule" #: String
+          BASE_RULE_PATH = File.expand_path("../../../linter/rule.rb", __dir__) #: String
+
+          class << self
+            #: -> String
+            def usage_form
+              "lint explain <RULE>"
+            end
+          end
+
+          #: -> void
+          def run
+            parse_options!
+
+            rule_name = argv.shift
+            abort_with_usage("`lint explain` requires a rule name argument") unless rule_name
+            abort_with_usage("unexpected argument: #{argv.first}") unless argv.empty?
+
+            workspace_path = current_workspace_path
+
+            graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+            graph.index_all([BASE_RULE_PATH, *Rubydex::Linter::RuleLoader.paths(workspace_path)])
+            graph.resolve
+
+            base_rule = graph[BASE_RULE_NAME]
+            abort("Base rule class #{BASE_RULE_NAME} is not found. This is likely an issue in Rubydex itself") unless base_rule.is_a?(Rubydex::Class)
+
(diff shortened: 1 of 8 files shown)


lib/rubydex/cli/command/lint/explain.rb
```diff
@@ -0,0 +1,72 @@
...
+          end
+
+          #: -> void
+          def run
+            parse_options!
+
+            rule_name = argv.shift
+            abort_with_usage("`lint explain` requires a rule name argument") unless rule_name
+
+            workspace_path = File.expand_path(argv.shift || Dir.pwd)
```
vinistock: I noticed this in other commands and it's something I think we need to fix. Due to the way that Bundler connects with dependencies, it's not possible to simply pass a different workspace path as dependencies will automatically be missing.

Personally, I think that can be surprising behaviour. If you request the same actions from inside of a directory, the results are different than if you pass that directory as an argument while PWD is a different one.

I would completely disallow passing paths in the CLI and always default the workspace path to:

```ruby
begin
  Bundler.root
rescue Bundler::GemfileNotFound
  Dir.pwd
end
```

lib/rubydex/linter/rule_loader.rb
```diff
@@ -4,17 +4,24 @@ module Rubydex
...
+        def paths(workspace_path)
+          rule_files = Dir.glob(BUILT_IN_RULE_GLOB)
+          rule_files.concat(Dir.glob(RULE_GLOB, base: workspace_path).map do |rule_file|
             File.expand_path(rule_file, workspace_path)
-          end
+          end)
           rule_files.concat(Gem.find_latest_files(RULE_GLOB)) if ENV["BUNDLE_GEMFILE"]
 
-          rule_files.each do |rule_file|
+          rule_files.uniq
```
vinistock, line 20: Why do we need the `uniq`? Is there a scenario where there might be duplicates?

lib/rubydex/cli/command/lint/explain.rb
```diff
@@ -0,0 +1,72 @@
...
+            graph = Rubydex::Graph.configure_for_workspace(workspace_path)
+            graph.index_all([BASE_RULE_PATH, *Rubydex::Linter::RuleLoader.paths(workspace_path)])
+            graph.resolve
+
+            base_rule = graph[BASE_RULE_NAME]
+            abort("Base rule class #{BASE_RULE_NAME} is not found. This is likely an issue in Rubydex itself") unless base_rule.is_a?(Rubydex::Class)
+
+            matched_rule_names = base_rule.descendants.grep(Rubydex::Class).filter_map do |declaration|
+              next if declaration.name == BASE_RULE_NAME
+              next unless declaration.unqualified_name == rule_name
```
vinistock: I wonder if we should use this instead


```suggestion
              next unless declaration.name.end_with?(rule_name)
```

to allow explain to account for namespaces. For example, `explain Rails::Something`

vinistock (approved):

lib/rubydex/linter/rule_loader.rb
st0012: When bundler is used, because rubydex is in the dependencies too, it can return the built-in rules too.

PR #988, opened 2026-08-06
Add an in-project linter rule and rule test support

st0012:
Add a built-in rule that checks the structure of discovered linter rules. The rule also provides an end-to-end smoke test for the linter infrastructure.

This PR also adds test support for project rules. Missing gem load paths and diagnostics from installed gems no longer block `rdx lint`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,12 @@ jobs:
       - name: Lint
         run: bundle exec rake lint
 
+      - name: Compile
+        run: bundle exec rake compile
+
+      - name: Run Rubydex linter
+        run: bundle exec rdx lint .
+
       # Verify that licenses conform to the policy in rust/about.toml
       - name: Check licenses
         run: |
--- a/lib/rubydex/cli/command/lint.rb
+++ b/lib/rubydex/cli/command/lint.rb
@@ -28,8 +28,11 @@ module Rubydex
           config = Rubydex::Config.load(workspace_path)
           warn_unknown_rules(config.linter, rules)
 
-          graph = build_graph($stderr, workspace_path:, config:, fail_on_index_errors: true)
-          result = Rubydex::Linter::Runner.new(graph, rules:, config: config.linter).run
+          graph = build_graph($stderr, workspace_path:, config:)
+          runner = Rubydex::Linter::Runner.new(graph, rules:, config: config.linter)
+          rule_count = runner.rules.size
+          $stderr.puts("Running #{rule_count} #{pluralize("rule", rule_count)}...")
+          result = runner.run
           if result.diagnostics.empty?
             print_summary(graph.documents.count, result.diagnostics)
             return
@@ -57,6 +60,7 @@ module Rubydex
         #: (String workspace_path) -> Array[singleton(Rubydex::Linter::Rule)]
         def load_linter_rules(workspace_path)
           Rubydex::Linter::RuleLoader.load(workspace_path)
+          Rubydex::Linter::Rule.subclasses
         rescue Rubydex::Linter::RuleLoadError => error
           abort(error.message)
         end
--- a/lib/rubydex/linter/rule_loader.rb
+++ b/lib/rubydex/linter/rule_loader.rb
@@ -7,23 +7,18 @@ module Rubydex
       RULE_GLOB = "rubydex_linter/rules/**/*.rb" #: String
 
       class << self
-        #: (String workspace_path) -> Array[singleton(Rule)]
+        #: (String workspace_path) -> void
         def load(workspace_path)
-          existing_rules = Rule.subclasses
           rule_files = Dir.glob(RULE_GLOB, base: workspace_path).map do |rule_file|
             File.expand_path(rule_file, workspace_path)
           end
-          if ENV["BUNDLE_GEMFILE"]
-            rule_files.concat(Gem.find_latest_files(RULE_GLOB))
-          end
+          rule_files.concat(Gem.find_latest_files(RULE_GLOB)) if ENV["BUNDLE_GEMFILE"]
 
           rule_files.each do |rule_file|
             require rule_file
           rescue LoadError, SyntaxError => error
             raise RuleLoadError, "Unable to load linter rules from #{rule_file}: #{error.message}", cause: error
           end
-
-          Rule.subclasses - existing_rules
         end
       end
     end
--- a/lib/rubydex/linter/rule_test_case.rb
+++ b/lib/rubydex/linter/rule_test_case.rb
@@ -1,0 +1,343 @@
+# frozen_string_literal: true
+
+require "fileutils"
+require "minitest/test"
+require "pathname"
+require "tmpdir"
+require "uri"
+require "rubydex/linter"
+
+module Rubydex
+  module Linter
+    # Base test case for Rubydex linter rules.
+    #
+    # Add expected diagnostics below the source:
+    #
+    #     assert_diagnostics(<<~RUBY)
+    #       FOO = 123
+    #       ^^^ Failure: FOO
+    #     RUBY
+    #
+    # The carets mark the diagnostic range. Use `^{}` for a zero-width range.
+    # Pass a hash to test multiple files. Repeat an annotation line for each message line.
+    # By default, `FooTest` tests the `Foo` rule.
+    class RuleTestCase < Minitest::Test
+      DEFAULT_FILE = "test.rb" #: String
+
+      ANNOTATION_PATTERN = /\A(?<indent>\s*)(?<carets>(?<!\\)\^+|\^\{\})(?:\s(?<message>.*))?\z/ #: Regexp
+
+      #: type annotation = [Integer, Integer, Integer, String]
+
+      #: String
+      attr_reader :workspace_path
+
+      #: (String) -> void
+      def initialize(name)
+        super
+        @workspace_path = File.realpath(Dir.mktmpdir("rubydex-rule-test-")) #: String
+      end
+
+      #: -> void
+      def teardown
+        super
+      ensure
+        FileUtils.remove_entry(workspace_path)
+      end
+
+      #: -> singleton(Rule)
+      def rule_class
+        @rule_class ||= begin
+          name = self.class.to_s.delete_suffix("Test")
+          if name == self.class.to_s
+            raise "Could not infer rule class from #{self.class}; override `#rule_class`"
+          end
+
+          Object.const_get(name) #: as singleton(Rule)
+        end #: singleton(Rule)?
+      end
+
+      # Registers sources that are loaded into the graph for every assertion.
+      # Shared sources must not contain caret annotations. Test-specific
+      # sources win on filename collision.
+      #: (Hash[String, String]) -> void
+      def add_shared_source(sources)
+        shared_sources.merge!(sources)
+      end
+
+      # Returns absolute file paths whose diagnostics should be ignored.
+      #: -> Array[String]
+      def ignored_diagnostic_files
+        @ignored_diagnostic_files ||= [] #: Array[String]?
+      end
+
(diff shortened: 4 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -67,6 +67,10 @@ jobs:
       - name: Run Ruby tests
         run: bundle exec rake test
 
+      - name: Run Rubydex linter
+        if: matrix.os == 'ubuntu-latest' && matrix.ruby == '4.0'
```
Morriar: Why can't we test on all platforms?

lib/rubydex/graph.rb
```diff
@@ -62,7 +62,8 @@ def add_workspace_dependency_paths(paths)
           # descending them
           next if File.absolute_path?(path)
 
-          paths << File.join(spec.full_gem_path, path)
+          require_path = File.join(spec.full_gem_path, path)
+          paths << require_path if File.directory?(require_path)
```
alexcrocha: It's not clear to me why this change is needed. Can you give an example?

lib/rubydex/cli/command/lint.rb
```diff
@@ -29,7 +29,10 @@ def run
           warn_unknown_rules(config.linter, rules)
 
           graph = build_graph($stderr, workspace_path:, config:, fail_on_index_errors: true)
-          result = Rubydex::Linter::Runner.new(graph, rules:, config: config.linter).run
+          runner = Rubydex::Linter::Runner.new(graph, rules:, config: config.linter)
+          rule_count = runner.rules.size
+          $stderr.puts("Running #{rule_count} #{pluralize("rule", rule_count)}...")
```
vinistock, line 34: Fine this PR, but we need some form of output abstraction to turn off printing for certain tools. For example, the LSP doesn't really need to print everything to the output tab. Especially considering that the linter will run repeatedly.

lib/rubydex/linter/rule_loader.rb
```diff
@@ -5,25 +5,26 @@ module Linter
...
 
           rule_files.each do |rule_file|
             require rule_file
           rescue LoadError, SyntaxError => error
             raise RuleLoadError, "Unable to load linter rules from #{rule_file}: #{error.message}", cause: error
           end
 
-          Rule.subclasses - existing_rules
+          built_in_rules = [Rules::RuleStructure] #: Array[singleton(Rule)]
+          built_in_rules | (Rule.subclasses - existing_rules)
```
vinistock: It seems that we're trying to return the difference of rules loaded here. However, we would normally only load once per workspace. Why do we need to return the difference here rather than just reading `Rule.subclasses`?

lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -0,0 +1,125 @@
...
+            definitions.each do |name, definition|
+              validate_rule_definition(name, definition, path)
+            end
+          end
+        end
+
+        private
+
+        #: (Document, Hash[String, bool]) -> Hash[String, ClassDefinition]
+        def rule_definitions(document, rule_class_names)
```
alexcrocha: We might not need to do this work to get the rule definitions. Can't we get them through `rule_class_names = child_classes(BASE_RULE_NAME)` above?

lib/rubydex/linter/rule_test_case.rb
```diff
@@ -0,0 +1,343 @@
...
+
+      #: type annotation = [Integer, Integer, Integer, String]
+
+      #: String
+      attr_reader :workspace_path
+
+      #: (String) -> void
+      def initialize(name)
+        super
+        @workspace_path = File.realpath(Dir.mktmpdir("rubydex-rule-test-")) #: String
```
vinistock, line 37: Do we really need to create a new temporary directory for each example? Most linting rule tests only care about the actual code being analyzed, so if we can avoid IO on every single example, that would be better.

lib/rubydex/cli/command/lint.rb
st0012: The LSP addon operates on `Runner` directly and doesn't enter here, so it shouldn't be affected by this message.

lib/rubydex/linter/rule_test_case.rb
st0012: Yes it's a bit wasteful, but in Core's `RuleTestCase` we do the same for all Rule tests (through `Spoom::Context.mktmp!`). So for smooth adoption I think we should maintain the same behaviour and optimize test setup later. I'll create an issue after this PR is merged.

vinistock: Yeah, I know we're currently using the Spoom context, but it's completely unnecessary. As long as we provide helpers for `TestCase` that only create in-memory sources, the change should be mostly transparent for the adoption.

Can we please already start out without doing IO? I really think we should avoid this from the start.

lib/rubydex/cli/command/lint.rb
vinistock: Ah, right. Good point.

lib/rubydex/linter/rule_test_case.rb
st0012: I can do it before cutting a release but please let me do it in another PR as this one is already bloated

st0012: Fixed

st0012: Actually, this won't work for rules that use `Prism.parse` to find constant/method call nodes.

lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -0,0 +1,126 @@
...
+          end
+        end
+
+        private
+
+        #: -> Hash[String, Hash[String, ClassDefinition]]
+        def rule_definitions_by_uri
+          definitions_by_uri = {} #: Hash[String, Hash[String, ClassDefinition]]
+
+          child_classes(BASE_RULE_NAME).each do |rule_class|
```
Morriar: Can you add a test doing something like this?

```rb
def test_plain_workspace_without_project_rules
  write("app.rb", "class App; end\n")

  result = run_lint

  assert_predicate(result, :success?)
  assert_empty(result.diagnostics)
end
```

lib/rubydex/linter/rule_loader.rb
```diff
@@ -9,21 +9,18 @@ class RuleLoader
...
+          rule_files.concat(Gem.find_latest_files(RULE_GLOB)) if ENV["BUNDLE_GEMFILE"]
 
           rule_files.each do |rule_file|
             require rule_file
           rescue LoadError, SyntaxError => error
             raise RuleLoadError, "Unable to load linter rules from #{rule_file}: #{error.message}", cause: error
           end
 
-          Rule.subclasses - existing_rules
+          Rule.subclasses
```
Morriar: Let's add a test like this:

```ruby
def test_does_not_return_rules_loaded_from_another_workspace
  workspace_a = create_workspace(
    "rubydex_linter/rules/no_foo.rb" => <<~RUBY,
      class NoFoo < Rubydex::Linter::Rule
      end
    RUBY
  )
  workspace_b = create_workspace

  workspace_a_rules = Rubydex::Linter::RuleLoader.load(workspace_a)
  workspace_b_rules = Rubydex::Linter::RuleLoader.load(workspace_b)

  assert_includes(workspace_a_rules, NoFoo)
  refute_includes(workspace_b_rules, NoFoo)
end
```

lib/rubydex/linter/rule_test_case.rb
vinistock: Are the rules reading content from disk? That won't work for LSP scenarios either. The linter shouldn't be reading things from disk.

lib/rubydex/linter/rule_loader.rb
st0012: Why do we need this test? Are we joining to make the same linter process check against 2 different workspaces?
I deliberately simplified the code because I thought this would never happen.

lib/rubydex/linter/rule_test_case.rb
st0012: Discussed in call: in principle linter rules should not separately access the file on disk, but right now Rubydex doesn't have all the info some rules need (e.g. method call arguments), so they still need to read the files from disk and parse them with Prism.
So for now, tests need to support reading from disk for those rules' tests.

lib/rubydex/linter/rule_loader.rb
Morriar: That's how we run Spoom dead code detection at the moment. Spoom is installed in CodeDB's bundle then executed against each workspace.

st0012: But this is the linter and I don't think we'd build the dead code detection on top of it. IMO it's probably better to be a separate tool.

lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -0,0 +1,126 @@
...
+            rule_file = rule_file?(path)
+            next if !rule_file && test_file?(path)
+
+            definitions = definitions_by_uri.fetch(document.uri, {})
+            validate_rule_file(document, definitions) if rule_file
+
+            definitions.each do |name, definition|
+              validate_rule_definition(name, definition, path)
+            end
+          end
```
vinistock: This seems like a roundabout way of checking if all rules are defined in a given glob pattern. Is there a reason we can't find them through descendants?

```ruby
rules = graph["Rubydex::Linter::Rule"].descendants

rules.each do |rule|
  defs = rule.defs

  # check if single definition
  # check if definition uri matches
end
```

vinistock (approved):
Generally good, but I think the new rule can be simplified

lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -0,0 +1,121 @@
...
+
+          graph.documents.each do |document|
+            path = path_for_uri(document.uri)
+            next unless workspace_path?(path)
+            next unless rule_file?(path)
+
+            definitions = definitions_by_uri.fetch(document.uri, [])
+              .sort_by { |_, definition| definition.location }
+            validate_rule_file(document, definitions)
+          end
```
vinistock: I'm confused. Why do we still need to accumulate definitions by uri and then loop through all documents? Can't we achieve it all in a single loop?

Morriar (approved):

lib/rubydex_linter/rules/rule_structure.rb
st0012: We need 1) all Rule child classes and their locations and 2) all documents under rule paths so we can check if there are child classes placed **outside of** the rule paths.

If we were to use a single loop, we'd need nested loops for them, which is worse?

vinistock: I don't understand why we need 2. If you place a rule outside of the expected path, doesn't it still inherit from `Rule`?

Why doesn't this work?

```ruby
# This is every single rule regardless of where it is defined. I don't
# understand why this isn't enough
rules = graph["Rubydex::Linter::Rule"].descendants

rules.each do |rule|
  defs = rule.definitions

  unless defs.one?
    # add diagnostic
  end

  rule_def = defs.first
  path = path_for_uri(rule_def.uri)
  next unless workspace_path?(path)
  next unless rule_file?(path)

  # add diagnostic that the rule was defined outside of the expected glob
  # patterns
end
```

What file can you write that isn't captured by this approach?

st0012: But then you can't verify whether a rule file contains multiple rule classes.

I'll update the PR to document this but I want to check 3 things at once:

- Each rule file contains exactly one rule class definition that inherits the `Rule` base class
- Rule subclasses only exist in a rule OR a test directory
- Rule subclasses all have the right namespaces

And I can't think of a way to do this without some kind of intermediate map to avoid nested loops.

PR #990, opened 2026-08-06
Add graph listing commands

st0012:
Add `rdx list docs` and `rdx list roots` for inspecting the files and roots Rubydex uses to build a workspace graph. The command follows the existing CLI graph and configuration flow and keeps workspace paths relative in its output.


st0012:
Closing this PR because the list command is not needed now.

PR #991, opened 2026-08-06
Make Cypher query execution return a `Result` instance

paracycle:
This addresses two review comments on the Cypher blog post in [rails-at-scale#431](https://github.com/Shopify/rails-at-scale/pull/431):

- [`ArgumentError` on a syntax error should be a more specific error](https://github.com/Shopify/rails-at-scale/pull/431#discussion_r3729283405)
- [`render` mixes execution and rendering; the API should be `query.run(graph).render(:json)`](https://github.com/Shopify/rails-at-scale/pull/431#discussion_r3729343006)

## Execution and rendering are separate

`Rubydex::Query#render(graph, format)` is gone. The new shape is:

```ruby
query = Rubydex::Query.parse("MATCH (c:Class) RETURN c.name")
result = query.run(graph)

result.rows.each { |row| puts row["c.name"] }
puts result.render(:json)
```

`Query#run` returns a `Rubydex::Query::Result`. That object owns the executed result set, so `render` never runs the query again. It is `Enumerable` over its rows, and it answers `columns`, `rows`, `each`, `size`, `length`, `empty?`, and `render`.

## Specific errors

| Class | Raised by |
| --- | --- |
| `Rubydex::QueryError` | The base class. Rescue it to catch either failure. |
| `Rubydex::QuerySyntaxError` | `Query.parse`, on invalid Cypher. |
| `Rubydex::QueryExecutionError` | `Query#run`, when the query fails against the graph. |

`ArgumentError` remains for true argument faults, such as an unknown output format.

The Rust FFI reports the failure kind through a new `CQueryErrorKind`, so the C extension raises the matching class.

## Verification

- `bundle exec rake test`: 300 Ruby runs and the full Rust suite. 0 failures.
- `cargo fmt --check` and `rubocop`: clean. `cargo clippy` reports no new warnings.
- CLI smoke test: the table and JSON output are unchanged. A syntax error still fails before indexing. An execution error aborts with its message.
- Compaction stress: `Rubydex::Query::Result` marks `graph_obj` and `rows` movable and implements `dcompact`, so `GC.verify_compaction_references` followed by `GC.compact` keeps every reference valid.

## Follow-up

The blog post in rails-at-scale#431 still describes the old API. It needs the two snippets updated once this merges.

### DIFF
--- a/ext/rubydex/query.c
+++ b/ext/rubydex/query.c
@@ -7,6 +7,38 @@
 #include "utils.h"
 
 /*
+ * RDoc parser workaround for https://github.com/ruby/rdoc/issues/1744:
+ * mRubydex = rb_define_module("Rubydex")
+ */
+
+static VALUE mRubydex;
+static VALUE cQueryResult;
+
+// Raises the Ruby error that matches a Cypher failure reported by Rust and releases `message`.
+// Syntax and execution failures get a Rubydex error; everything else is a Ruby argument error.
+NORETURN(static void raise_query_error(const char *message, CQueryErrorKind kind));
+
+static void raise_query_error(const char *message, CQueryErrorKind kind) {
+    VALUE error_message = rb_utf8_str_new_cstr(message);
+    free_c_string(message);
+
+    VALUE error_class;
+    switch (kind) {
+    case CQueryErrorKind_Syntax:
+        error_class = rb_const_get(mRubydex, rb_intern("QuerySyntaxError"));
+        break;
+    case CQueryErrorKind_Execution:
+        error_class = rb_const_get(mRubydex, rb_intern("QueryExecutionError"));
+        break;
+    default:
+        error_class = rb_eArgError;
+        break;
+    }
+
+    rb_exc_raise(rb_exc_new_str(error_class, error_message));
+}
+
+/*
  * call-seq:
  *   Rubydex::Query.schema(format = :table) -> String
  *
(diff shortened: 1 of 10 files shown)


dersam:
Should this be two separate PRs (one for errors, one for render)?

paracycle:
> Should this be two separate PRs (one for errors, one for render)?

I could split, but it felt natural to add the new exceptions while making the API change.

lib/rubydex/cli/command/query.rb
```diff
@@ -37,32 +37,30 @@ def run
...
         def parse_query(query)
           Rubydex::Query.parse(query)
-        rescue ArgumentError => e
+        rescue Rubydex::QuerySyntaxError => e
           abort(e.message)
         end
 
-        #: (Rubydex::Query parsed, Rubydex::Graph graph, String format) -> void
+        #: (Rubydex::Query parsed, Rubydex::Graph graph, String format) -> String
         def render(parsed, graph, format)
```
st0012: Do we still need this method? I feel it can be folded into `run`.

rust/rubydex-sys/src/cypher_api.rs
```diff
@@ -227,13 +255,21 @@ pub struct CRowsIter {
     index: usize,
 }
 
-/// The result of running a query for structured rows: either a row iterator or an error message.
+/// An executed query's result set: the column names and rows that `rdx_query_execute` produced.
+/// Opaque from the C side — use `rdx_result_set_column_count`, `rdx_result_set_column`,
+/// `rdx_result_set_row_count`, `rdx_result_set_format`, `rdx_result_set_rows`, and
+/// `rdx_result_set_free`.
```
st0012: Maybe we list naming pattern `rdx_result_set_*` here instead of all functions? Feels easy to get out of sync and also not really necessary.

ext/rubydex/query.c
```diff
@@ -172,45 +251,156 @@ static VALUE query_run_yield(VALUE args) {
...
-    TypedData_Get_Struct(self, void *, &query_type, query);
+static VALUE rdxr_query_result_rows(VALUE self) {
+    QueryResultData *data = query_result_data(self);
 
-    void *graph = rdxi_graph_from_object(graph_obj);
+    if (!NIL_P(data->rows)) {
+        return data->rows;
+    }
+
+    struct CRowsIter *iter = rdx_result_set_rows(data->result_set, rdxi_graph_from_object(data->graph_obj));
```
Morriar, line 277: One concern with lazy materialization: `rows` converts node cells by looking them up in the current graph state. If the graph changes after `query.run(graph)` but before the first `result.rows`, the executed result can change shape. For example:

```ruby
result = Rubydex::Query.parse("MATCH (c:Class {name: 'Dog'}) RETURN c").run(graph)
graph.index_source("zoo.rb", "class Cat; end\n", "ruby")
graph.resolve

result.render(:json) # still has the Dog node from the executed result set
result.rows          # [{ "c" => "Dog" }], no longer a Declaration handle
```

Can we materialize the row cells at execution time, or have the result set carry the node category/kind/id captured during execution, so `Result` is independent of later graph mutations? I'd add a regression test for `run; mutate graph; rows` too.

lib/rubydex/cli/command/query.rb
paracycle: I thought about that but the need for a `begin`/`end` block with a `rescue` inside made me keep the method. Now that you also raise the point, maybe I should inline it.

ext/rubydex/query.c
```diff
@@ -172,45 +251,156 @@ static VALUE query_run_yield(VALUE args) {
...
+    /*
+     * The result of running a Rubydex::Query against a graph: the columns and rows produced by one
+     * execution. Enumerable over its rows.
+     */
+    cQueryResult = rb_define_class_under(cQuery, "Result", rb_cObject);
+    rb_undef_alloc_func(cQueryResult);
+
+    // A result can only be obtained from `Query#run`; `new` would create an object with no Rust
+    // data behind it.
+    rb_undef_method(rb_singleton_class(cQueryResult), "new");
```
st0012, line 396: Let's hide it in `rbi` too?

ext/rubydex/query.c
paracycle: I've gone the other direction and implemented proper lazy materialization [here](https://github.com/Shopify/rubydex/pull/992) which also checks for node invalidation too. Is that enough to address your concern?

lib/rubydex/cli/command/query.rb
paracycle: Done!

rust/rubydex-sys/src/cypher_api.rs
paracycle: Done!

ext/rubydex/query.c
paracycle: How? This isn't making the method `private`, it is completely removing it. How can we remove methods in RBI files?

st0012: I thought we can hide it by making `Result.new` private? Hide as using Sorbet to prevent users from calling it.

paracycle: I am not sure that I like that as a solution, since it only makes it `private` and doesn't mean it can't be called from other contexts:
```ruby
class Bar
  class << self
    private

    def new = super
  end
end

class Baz < Bar
  (Instance = new) rescue puts "Error"
  puts "No error when creating: #{Instance}" if defined?(Instance)
end

Bar.new rescue puts "Error"
Baz.new rescue puts "Error"
```
gives:

```
❯ ruby bar.rb
No error when creating: #<Baz:0x0000000121aec870>
Error
Error
```

On the other hand, `undef_method :new` completely removes it:
```ruby
class Bar
  class << self
    undef_method :new
  end
end

class Baz < Bar
  (Instance = new) rescue puts "Error"
  puts "No error when creating: #{Instance}" if defined?(Instance)
end

Bar.new rescue puts "Error"
Baz.new rescue puts "Error"
```
gives:
```
❯ ruby bar.rb
Error
Error
Error
```

ext/rubydex/query.c
Morriar: Looks like `Result#render` and `Result#rows` disagree based on graph mutations after execution. Do we need to rebase on #992?

paracycle: On the contrary, #992 is based on this branch, since it need this refactor to operate.

rbi/rubydex.rbi
```diff
@@ -529,11 +532,37 @@ class Rubydex::Query
...
+
+  Elem = type_member { { fixed: T::Hash[String, T.untyped] } }
+
+  sig { returns(T::Array[String]) }
+  def columns; end
+
+  sig { returns(T::Array[T::Hash[String, T.untyped]]) }
+  def rows; end
+
+  sig { override.params(block: T.proc.params(row: T::Hash[String, T.untyped]).void).returns(T.self_type) }
```
Morriar, line 560: The C method uses `RETURN_ENUMERATOR`, so `result.each` is valid and returns an `Enumerator`, but this sig requires a block and always returns `T.self_type`.

ext/rubydex/query.c
paracycle: Ok, finally I went for:
```ruby
class << self
  private

  # A result only comes from `Query#run`. The C extension undefines `new`, so a call raises.
  sig { returns(T.noreturn) }
  def new; end
end
```
so that a call to `new`, even in a private setting, always returns a type that isn't usable.

st0012 (approved):

PR #992, opened 2026-08-06
Build query result rows one row at a time

paracycle:
## TL;DR

`Rubydex::Query::Result#each` built every row before it yielded the first one. It converts one row at a time now, so `first` on a 300 row result went from 1516 allocations down to 19. Also, the walk shares one frozen key String per column instead of allocating a key per cell.

This stacks on #991, so it is based on `uk-cypher-api-refactors`. I recommend reviewing per commit.

## The problem

`#each` called `#rows`, which built the whole array first. A caller that read one row still paid for all of them, twice: once as `CCell`s in C memory, and once as Ruby objects. `first` and `find` were the worst case, since they break after one row and then throw the other 299 away.

## What changed

The row iterator is a cursor now. `rdx_result_set_rows` builds only the column strings, and `rdx_rows_iter_next` frees the cells of the previous row and converts one row under the graph read lock.

The lock is released before the call returns, so a block can run any code between two rows, including code that writes to the graph. Holding it across `rb_yield` is not an option, since a block that calls `graph.index_source` would deadlock the thread.

`rows` keeps its old behaviour and collects everything through the same cursor, so `each` still walks the memoized array once it exists.

Allocations for 300 rows and 2 columns, from `GC.stat(:total_allocated_objects)`, measured against a real build of each tree:

| Call | Before | After |
| --- | --- | --- |
| `result.first` | 1516 | 19 |
| `result.rows` | 1506 | 907 |

The first row is the streaming win. The second is the shared keys: 600 key Strings disappear.

## The stale node check

Since the graph can change mid-walk now, the second commit removes a fallback in `build_cell` that turned an unresolvable node into a plain String. I hit it while testing the cursor: after a `delete_document` inside an `each` block, the remaining rows came back as `"Cat"` and `"Dog"` instead of handles, with no signal at all. A column changing type halfway through a result is worse than an exception.

So `rdx_rows_iter_next` reports `MissingNode` and the extension raises `Rubydex::StaleQueryResultError`, which is a `Rubydex::QueryError`. The message names the node.

The check is narrow on purpose, and it is not mutation detection:

- It only fires when the graph no longer holds a node that the result returned.
- It does not fire for a re-index that keeps the ids, since a declaration id comes from the name.
- It says nothing about a handle that a walk already handed out. Those resolve against the graph on each call, exactly like the handles from `Graph#[]`.

`render`, `columns`, `size`, and `empty?` keep working after a change, since they read the executed result set and never touch the graph.

I did not add a graph revision counter for this. It would need a change to the graph pointer in `graph_api.rs`, and it would fire on `load_config`, `set_encoding`, and completion, none of which can drop a node.

## Verification

- `bundle exec rake test`: 356 Ruby runs, 1597 assertions, plus the Rust suite. 0 failures.
- `cargo fmt --check`, `cargo clippy`, and `rubocop` are all clean.
- Eight new tests cover the early exit, an `Enumerator` without a block, a `break` and a raising block, the shared frozen keys, and the four stale result cases.
- A retention probe streams 60 rows with node, string, list, and map columns, holds every value, churns the C heap with 20,000 strings, runs `GC.verify_compaction_references` and `GC.compact`, and then compares against an eager `rows` call. Everything matches, so nothing Ruby hands back points into the freed cells.

## Tophatting

```ruby
graph = Rubydex::Graph.new
graph.index_workspace
graph.resolve

query = Rubydex::Query.parse("MATCH (c:Class) RETURN c, c.name")
query.run(graph).first          # converts one row
query.run(graph).render(:json)  # no rows built at all
```

### DIFF
--- a/rust/rubydex-sys/src/cypher_api.rs
+++ b/rust/rubydex-sys/src/cypher_api.rs
@@ -246,6 +246,19 @@ pub struct CResultRow {
     pub len: usize,
 }
 
+/// The outcome of one `rdx_rows_iter_next` call.
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CRowsNextStatus {
+    /// `out` holds the next row.
+    Row,
+    /// The cursor reached the end of the result set.
+    Done,
+    /// The graph no longer holds a node that the row returned, so the row cannot be built.
+    /// `rdx_rows_iter_error` names the node.
+    MissingNode,
+}
+
 /// A cursor over an executed result set's rows. It converts one row per `rdx_rows_iter_next` call,
 /// so a caller can walk a large result set without a copy of every cell in memory at once. Opaque
 /// from the C side — use the `rdx_rows_iter_*` methods to work with it.
@@ -256,6 +269,8 @@ pub struct CRowsIter {
     columns: Box<[*const c_char]>,
     /// Cells of the row that the last `rdx_rows_iter_next` call produced.
     current: Vec<CCell>,
+    /// Name of the node that the last `rdx_rows_iter_next` call could not resolve.
+    error: Option<CString>,
     index: usize,
 }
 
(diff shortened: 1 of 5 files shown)


test/graph_test.rb
```diff
@@ -1753,6 +1753,163 @@ def test_result_reports_columns_rows_and_emptiness
...
+
+      graph = Rubydex::Graph.new
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      result = Rubydex::Query.parse(
+        "MATCH (c:Class) WHERE c.name IN ['Animal', 'Dog'] RETURN c.name ORDER BY c.name",
+      ).run(graph)
+
+      enumerator = result.each
```
Morriar, line 1788: Do we need to mark the block as optional here: https://github.com/Shopify/rubydex/blob/dd9d491abb2c93daaf3c32cfd31af141dcf50158/rbi/rubydex.rbi#L553-L554?

st0012 (approved):

rust/rubydex-sys/src/cypher_api.rs
```diff
@@ -270,41 +290,74 @@ pub struct CExecuteResult {
...
-        ),
+        )),
         CypherValue::List(items) => {
-            let cells: Vec<CCell> = items.iter().map(|item| build_cell(graph, item)).collect();
+            let mut cells: Vec<CCell> = Vec::with_capacity(items.len());
+
+            for item in items {
+                match build_cell(graph, item) {
+                    Ok(cell) => cells.push(cell),
+                    Err(node) => {
```
Morriar, line 318: Let's also add a regression test for stale nodes inside a list/map result please. The top-level stale cases are covered, but these `List` and `Map` branches have their own propagation and cleanup paths. A test with `RETURN [c]` and `RETURN c { .name, node: c }` after `delete_document` would make sure we don't regress to silently returning strings in nested values.

test/graph_test.rb
paracycle: I'll fix this in #991 since it belongs there

rust/rubydex-sys/src/cypher_api.rs
paracycle: I've added two new tests with `List` and `Map` containers.

Morriar (approved):

PR #995, opened 2026-08-07
Skip CLI tests in Memleak

st0012:
The CLI suite adds substantial time to the Memleak job as it grows. This change excludes `test/cli_test.rb` from `ruby_test:valgrind` and keeps it in the regular `ruby_test` task.

Date | Daily median
-- | --
Jul 24 | 36m
Jul 27–31 | 35m
Aug 3 | 37m
Aug 4 | 41m
Aug 5 | 57m
Aug 6 | 65m
Aug 7 | 65m

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,7 +25,7 @@ begin
   namespace(:ruby_test) do
     RubyMemcheck::TestTask.new(valgrind: :compile) do |task|
       test_config.call(task)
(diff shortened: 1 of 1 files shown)


alexcrocha (approved):
Do we know what exactly in the cli tests is making memleak jobs so slow?

st0012:
I know which individual CLI tests dominate yet. RubyMemcheck runs the test process under Valgrind, so tests that perform more work generally add more time. And in cli tests, we run several full index cycles, which can be the reason.

PR #1000, opened 2026-08-11
Allow linter configuration to filter graph diagnostics

st0012:
I decided to surface the graph rule diagnostics via linter because they are basically rules defined by us. But currently the linter configuration can't disable them, which is inconsistent and can create noise users don't want.

So in this PR, I implemented the plumbing to allow filtering graph diagnostics with linter configurations.
One difference is: when we disable a linter rule (say `MyLinterRule`), that linter rule is not executed at all. If we do the same for the graph rule (say `invalid-constant-visibility`), the diagnostics can still be accessed via graph, but won't show up in the linter result.

### DIFF
--- a/lib/rubydex/cli/command/lint.rb
+++ b/lib/rubydex/cli/command/lint.rb
@@ -24,7 +24,6 @@ module Rubydex
           require "rubydex/linter"
 
           rules = load_linter_rules(workspace_path)
-          abort("No Rubydex::Linter::Rule subclasses were loaded") if rules.empty?
           config = Rubydex::Config.load(workspace_path)
           warn_unknown_rules(config.linter, rules)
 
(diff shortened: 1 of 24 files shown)


lib/rubydex/cli/command/lint.rb
```diff
@@ -43,7 +43,7 @@ def run
 
         #: (Rubydex::LinterConfig config, Array[singleton(Rubydex::Linter::Rule)] known_rule_classes) -> void
         def warn_unknown_rules(config, known_rule_classes)
-          known_rule_names = known_rule_classes.map(&:rule_name).uniq.sort
+          known_rule_names = (known_rule_classes.map(&:rule_name) + Rubydex::Diagnostic.graph_rule_names).uniq.sort
```
Morriar: I don't think we ever reach this point from the CLI if `known_rule_classes` is empty?

Morriar: Let's add a test where the CLI without a rule should at least show the built-in diagnostics?

rust/rubydex-sys/src/diagnostic_api.rs
```diff
@@ -52,6 +52,39 @@ impl DiagnosticArray {
...
+///
+/// # Panics
+///
+/// Panics if a generated rule name contains `\0`, which C strings cannot represent.
+///
+/// # Safety
+///
+/// - `out_names` must be a valid, writable pointer.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_graph_diagnostic_names(out_names: *mut *const *const c_char) -> usize {
```
vinistock, line 66: This looks like it belongs to `Rule` and not `Diagnostic`. Conceptually, a `Rule` is a type of mistake someone can make and a `Diagnostic` is an occurrence of that mistake.

If we want the names of the rules, so that they can be filtered, then shouldn't this method be in `Rule`? Like, `Rule.all` or `Rule.built_in`?

PR #1005, opened 2026-08-14
Linearize ancestors for implicitly created namespaces

HoneyryderChuck:
A namespace implied only by compact definitions (`class A::B::C`, with no `module A::B` anywhere) gets a `TodoDeclaration` from `create_todo_for_parent`, which was the one namespace-creating path that never enqueued a `Unit::Ancestors` for it. Its chain stayed at the `Ancestors::Partial([])` default, and since a namespace's own id is pushed onto that chain by `linearize_ancestors`, `find_member_in_ancestors` had nothing to walk: every member lookup returned `MemberNotFound` even though the member table was fully populated.

### DIFF
--- a/rust/rubydex/src/resolution_tests.rs
+++ b/rust/rubydex/src/resolution_tests.rs
@@ -3984,6 +3984,41 @@ mod todo_tests {
     }
 
     #[test]
+    fn have_linearized_ancestors() {
+        // Todos are namespaces, so they must be the first entry of their own ancestor chain. Member lookups walk that
+        // chain, so an empty one makes every member of the Todo unreachable
+        let mut context = graph_test();
+        context.index_uri("file:///a.rb", {
+            r"
+            class A::B::C
+              def foo; end
+            end
+            "
+        });
+        context.index_uri("file:///d.rb", "class A::B::D < A::B::C; end");
+        context.resolve();
+
+        assert_declaration_kind_eq!(context, "A", "<TODO>");
+        assert_declaration_kind_eq!(context, "A::B", "<TODO>");
+        assert_ancestors_eq!(context, "A", ["A"]);
+        assert_ancestors_eq!(context, "A::B", ["A::B"]);
+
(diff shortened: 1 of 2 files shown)


HoneyryderChuck:
CLA signed, but can't rerun it.

rust/rubydex/src/query.rs
```diff
@@ -3982,4 +3982,45 @@ mod tests {
             Err(FindMemberError::DeclarationNotFound),
         );
     }
+
+    #[test]
+    fn find_member_in_ancestors_returns_member_of_implicit_namespace() {
```
st0012: I don't think this test is needed. It's basically testing resolution behaviour in query tests.

test/declaration_test.rb
```diff
@@ -698,6 +698,28 @@ def initialize
     end
   end
 
+  def test_find_member_returns_members_of_implicit_namespaces
```
st0012: We should test this in the test in `resolution_tests`, and delete this one

rust/rubydex/src/query.rs
HoneyryderChuck: done

test/declaration_test.rb
HoneyryderChuck: done

rust/rubydex/src/resolution_tests.rs
```diff
@@ -3983,6 +3983,31 @@ mod todo_tests {
         assert_members_eq!(context, "A::B::D", vec!["d_method()"]);
     }
 
+    #[test]
+    fn have_linearized_ancestors() {
```
st0012, line 3987: This doesn't seem to test inheritance?

HoneyryderChuck: added test explicitly testing inheritance.

st0012 (approved):
Thanks for the fix!

PR #1007, opened 2026-08-14
Extract `RuleDefinition` as the base identity for rules

vinistock:
Third step towards #1000

We need to share the concept of a rule across Rust and Ruby, the graph and the linter. However, `Rule` currently enforces that each descendant must implement `lint`. This is an issue for unification because graph and linter diagnostics are collected differently.

In reality, the common concept is not `Rule`, but a `RuleDefinition`. All we need is the identity of the rule: a name and the default severity. With this, we can treat all rules similarly for the configuration as it applies to the `RuleDefinition` and not the `Rule`.

In more abstract terms, this PR makes `RuleDefinition` the identity of a rule, while `Rule` is the execution abstraction for the definition.


st0012:
IMO `RuleDefinition` is more like an interface then a base class, so using a module probably makes more sense?
And I think it should be kept under the `Linter` namespace.

vinistock:
I moved it under the linter. However, after trying out the module approach, I find it a bit strange. I would keep it as a parent class that defines the identity of a rule.

st0012 (approved):

PR #1008, opened 2026-08-14
Expose built-in rules in Ruby

vinistock:
Fourth step towards #1000

This PR makes sure we automatically create `RuleDefinition` classes for all built-in diagnostics. This allows us to treat rule definitions consistently between the linter and the graph, since they are all just classes that inherit from `RuleDefinition`.

The idea is basically to have an `all` method in Rust that returns all rules and we use the names to create classes like `class ParseError < RuleDefinition`, which registers the identity of that rule in the Ruby side consistently with linting rules.

### DIFF
--- a/ext/rubydex/diagnostic.c
+++ b/ext/rubydex/diagnostic.c
@@ -6,7 +6,10 @@
  * mRubydex = rb_define_module("Rubydex")
  */
 
+static VALUE mRubydex;
 VALUE cDiagnostic;
+static VALUE cRuleDefinition;
+static ID id_default_severity;
 
 VALUE rdxi_build_diagnostic_severity_value(VALUE mRubydex, DiagnosticSeverity severity) {
     VALUE mSeverity = rb_const_get(mRubydex, rb_intern("Severity"));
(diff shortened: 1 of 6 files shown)


ext/rubydex/diagnostic.c
```diff
@@ -27,4 +30,50 @@ VALUE rdxi_build_diagnostic_severity_value(VALUE mRubydex, DiagnosticSeverity se
...
+        VALUE name = rb_utf8_str_new(definition.name, (long)definition.name_length);
+        VALUE rule = rb_define_class_under(mRules, StringValueCStr(name), cRuleDefinition);
+        VALUE severity = rdxi_build_diagnostic_severity_value(mRubydex, definition.default_severity);
+
+        rb_ivar_set(rule, id_default_severity, severity);
+        rb_define_singleton_method(rule, "default_severity", rdxr_generated_rule_default_severity, 0);
+        rb_ary_push(definitions, rule);
+    }
+
+    rdx_rule_definitions_free(definition_array);
```
st0012, line 68: I think we need to add an `rb_ensure` here to free the array when the above code raises exceptions.

ext/rubydex/diagnostic.c
```diff
@@ -27,4 +30,50 @@ VALUE rdxi_build_diagnostic_severity_value(VALUE mRubydex, DiagnosticSeverity se
...
+// read, which is why Rubydex::Severity is loaded before this extension (see `lib/rubydex.rb`).
+static void define_generated_rules(VALUE mRules) {
+    CRuleDefinitionArray definition_array = rdx_rule_definitions();
+    VALUE definitions = rb_ary_new_capa((long)definition_array.len);
+
+    for (size_t i = 0; i < definition_array.len; i++) {
+        CRuleDefinition definition = definition_array.items[i];
+        VALUE name = rb_utf8_str_new(definition.name, (long)definition.name_length);
+        VALUE rule = rb_define_class_under(mRules, StringValueCStr(name), cRuleDefinition);
+        VALUE severity = rdxi_build_diagnostic_severity_value(mRubydex, definition.default_severity);
```
st0012, line 61: For built-in rules, should we already take linter config and apply the configured severity here?

vinistock: The configuration is not loaded at this time. This runs when rubydex is required.

ext/rubydex/diagnostic.c
vinistock: Similar to the other comment, this runs upon requiring Rubydex. If this code raises, then it's a bug we have to fix internally and don't want to swallow. There's no way for a consumer of Rubydex to make this code raise.

st0012: It's not to swallow the exception tho. It's so that we can still free the `definition_array` in that case.
I know it's not likely to happen but I think so far we always make sure free's called properly in all scenarios.

st0012 (approved):

ext/rubydex/diagnostic.c
vinistock: Yes, but in this case, we're not protecting ourselves from external code that might be incorrect. We'd be protecting ourselves against our own code, but we can ensure that is never raising on our side.

PR #1009, opened 2026-08-14
Add graph serialization

paracycle:
> [!NOTE] 
> This is a prototype of what a graph serialization for Rubydex might look like. 

After some investigation with Claude for potential opportunities for graph serialization, we settled in on using the `rkyv` (get it? pronounces `archive` when read 😂) for zero-copy serialization and cheap mmap'ed deserialization.

The integration with [`rkyv`](https://rkyv.org/) is mainly via deriving the crate's traits, so the intrusiveness is minimal, except for a few types that need special handling.

A report against Core is as following:

> ### Cache on disk
> 
> One file per root set, named by a hash of the indexed roots, under the user cache directory (`XDG_CACHE_HOME`, else `~/.cache`):
> 
> ```
> ~/.cache/rubydex/12cf5c75d1280a55.rdxsnap
> ```
> 
> | Region | Size | Share |
> |---|---:|---:|
> | Header | 32 B | ~0% |
> | Manifest archive (Merkle tree) | 22.8 MB | 0.94% |
> | Padding | 0 B | 0% |
> | Graph archive | 2,402.5 MB | 99.06% |
> | **Total** | **2,425.3 MB** | 100% |
> 
> Roughly **22 KB of cache per indexed file**. The Merkle tree is under 1% of the file, which is the point of keeping it in its own archive: a staleness check reads ~23 MB and never faults in the 2.4 GB graph.
> 
> The cache never lives inside the workspace. A file written into an indexed directory would bump that directory's mtime, and the tree would then report a change on every run.
> 
> ### Timings
> 
> Steady state, warm operating-system page cache.
> 
> | Scenario | Wall time | Peak RSS |
> |---|---:|---:|
> | `--no-cache` (previous behaviour) | 18.6 s | 4.6 GB |
> | Cold, no cache present (writes the snapshot) | 32.3 s | 6.2 GB |
> | Warm, workspace unchanged | **2.3 s** | 5.5 GB |
> | Warm, one file edited | 7.3 s | 5.6 GB |
> | Warm, `--verify-content` | 5.5 s | 5.5 GB |
> 
> `--no-cache` splits into 5.7 s indexing and 12.8 s resolution. Resolution is what the cache removes.
> 
> **Steady state is 8.1x faster than `--no-cache`.**
> 
> ### Cost of the first run
> 
> A cold run costs 32.3 s against 18.6 s uncached, so the snapshot costs about 14 s once: build the manifest (stat 110,621 files and their directories), serialize 2.4 GB, and write it. Every later run recovers that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@ foundation to power a variety of tools, such as type checkers, linters, language
 
 ## Usage
 
+Rubydex caches the index by default. The first run indexes all files and writes a snapshot file.
+Later runs read the snapshot and index only the files that changed since the last run.
+The snapshot file lives under the user cache directory, in a file named for the workspace.
+Set `XDG_CACHE_HOME` to change the user cache directory.
+Pass `--no-cache` on the command line, or `cache: false` to `index_workspace` in the Ruby API, to skip the cache.
+
 Both Ruby and Rust APIs are made available through a gem and a crate, respectively. Here's a simple example
 of using the Ruby API:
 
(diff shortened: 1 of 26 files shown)


Morriar (changes requested):
State that the PR is too broad to review safely as submitted.

PR #1010, opened 2026-08-14
Add dead code candidates endpoint

mathyasp:
- Adds `dead_code_candidates` endpoint 
- Adds 2 filters, intentionally as a separate commit so it can be removed if there's disagreement
  - Skip declarations with no definitions (avoids reporting phantom nodes the graph knows about but no source defines).
  - Skip declarations that have any rubydex-seeded built-in definition (avoids flagging `String`, `Array`, etc. even when user code reopens them).

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -260,14 +260,10 @@ static VALUE rdxr_graph_fuzzy_search(int argc, VALUE *argv, VALUE self) {
  * call-seq:
  *   dead_code_candidates -> Enumerator[Rubydex::Declaration]
  *
- * Returns an enumerator that yields constant-like declarations (namespace, constant, or constant alias) that have zero
- * resolved constant references and at least one definition. Declarations with a rubydex-seeded built-in definition are
- * excluded, even when user code reopens them. Requires a resolved graph and covers everything indexed, including
- * dependencies.
+ * Returns an enumerator over declarations that may be unused because the analysis could not find any references to
+ * them. This misses metaprogramming or untyped code, which is why the term candidates is used.
  *
- * The result is a working set, not an inventory. Deleting a reported declaration can leave its container unreferenced,
- * so re-index and call again until the result comes back empty. Accuracy follows the graph's reference data, so
- * whatever the indexer cannot see goes uncounted.
+ * Currently, only supports constants.
  */
 static VALUE rdxr_graph_dead_code_candidates(VALUE self) {
     if (!rb_block_given_p()) {
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -256,6 +256,32 @@ static VALUE rdxr_graph_fuzzy_search(int argc, VALUE *argv, VALUE self) {
     return rdxr_graph_yield_search_results(self, iter);
 }
 
+/*
+ * call-seq:
+ *   dead_code_candidates -> Enumerator[Rubydex::Declaration]
+ *
+ * Returns an enumerator that yields constant-like declarations (namespace, constant, or constant alias) that have zero
+ * resolved constant references. Requires a resolved graph and covers everything indexed, including dependencies.
+ *
+ * The result is a working set, not an inventory. Deleting a reported declaration can leave its container unreferenced,
+ * so re-index and call again until the result comes back empty. Accuracy follows the graph's reference data, so
+ * whatever the indexer cannot see goes uncounted.
+ */
+static VALUE rdxr_graph_dead_code_candidates(VALUE self) {
+    if (!rb_block_given_p()) {
+        return rb_enumeratorize(self, rb_str_new2("dead_code_candidates"), 0, NULL);
+    }
+
+    void *graph;
+    TypedData_Get_Struct(self, void *, &graph_type, graph);
+
+    void *iter = rdx_graph_dead_code_candidates(graph);
+    VALUE args = rb_ary_new_from_args(2, self, ULL2NUM((uintptr_t)iter));
+    rb_ensure(rdxi_declarations_yield, args, rdxi_declarations_ensure, args);
+
+    return self;
+}
+
 // Body function for rb_ensure in Graph#documents
 static VALUE graph_documents_yield(VALUE args) {
     VALUE self = rb_ary_entry(args, 0);
@@ -938,6 +964,7 @@ void rdxi_initialize_graph(VALUE moduleRubydex) {
     rb_define_method(cGraph, "[]", rdxr_graph_aref, 1);
     rb_define_method(cGraph, "search", rdxr_graph_search, -1);
     rb_define_method(cGraph, "fuzzy_search", rdxr_graph_fuzzy_search, -1);
+    rb_define_method(cGraph, "dead_code_candidates", rdxr_graph_dead_code_candidates, 0);
     rb_define_method(cGraph, "encoding=", rdxr_graph_set_encoding, 1);
     rb_define_method(cGraph, "resolve_require_path", rdxr_graph_resolve_require_path, 2);
     rb_define_method(cGraph, "require_paths", rdxr_graph_require_paths, 1);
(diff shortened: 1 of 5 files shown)


rust/rubydex/src/query.rs
```diff
@@ -833,6 +833,69 @@ pub fn follow_method_alias(graph: &Graph, alias_id: DefinitionId) -> Result<Decl
...
+/// Returns every constant-like declaration (namespace, constant or constant alias) that has zero resolved constant
+/// references and at least one definition. Declarations with a rubydex-seeded built-in definition are excluded, even
+/// when user code reopens them. Methods and variables are never included, they cannot be the target of a constant
+/// reference.
+///
+/// Expects a resolved graph, and covers everything indexed, including dependencies.
+///
+/// The result is a working set, not an inventory. Deleting a reported declaration can leave its container
+/// unreferenced, so re-index and call again until the result comes back empty. Accuracy follows the graph's reference
+/// data, so whatever the indexer cannot see (metaprogramming, autoloading) goes uncounted.
```
vinistock: Suggestion on how to simplify this:


```suggestion
/// Returns a list of declaration that may be unused in the codebase, which means
/// that the analysis could not find any references to them. This misses
/// meta-programming or untyped code, which is why the term candidates is used.
///
/// Currently, only supports constants.
```

rust/rubydex/src/query.rs
```diff
@@ -833,6 +833,69 @@ pub fn follow_method_alias(graph: &Graph, alias_id: DefinitionId) -> Result<Decl
...
+        let handles: Vec<_> = ids
+            .chunks(chunk_size)
+            .map(|chunk| {
+                s.spawn(|| {
+                    chunk
+                        .iter()
+                        .filter(|id| {
+                            let decl = declarations.get(id).unwrap();
+                            // Methods and variables are excluded here: rubydex cannot attribute method
+                            // calls to declarations, so called and uncalled both report zero references.
```
vinistock: Let's remove these comments.

rust/rubydex/src/query.rs
```diff
@@ -3982,4 +4045,181 @@ mod tests {
...
+        // Explicit singleton definitions are actionable, and class-level calls make their reference count non-zero.
+        let names = candidate_names(&context);
+        assert!(
+            names.contains(&"Unreferenced::<Unreferenced>".to_string()),
+            "dead explicit singleton missing from {names:?}"
+        );
+        assert!(
+            !names.contains(&"Referenced::<Referenced>".to_string()),
+            "called explicit singleton unexpectedly present in {names:?}"
+        );
```
vinistock: This test is encoding behaviour that's a bit weird. In the scenario, it's not necessarily the singleton class that's dead, but the method `never_called`.

Because singleton classes are a shadow hierarchy that exists for objects and the only way to directly refer to them is with `Foo.singleton_class`, it's a bit odd to consider them as part of dead code detection.

Also, they are often created synthetically by the internals to represent the code correctly, regardless of whether there was an opening of the singleton.

Let's instead always exclude singleton classes from the dead code candidates.

ext/rubydex/graph.c
```diff
@@ -256,6 +256,34 @@ static VALUE rdxr_graph_fuzzy_search(int argc, VALUE *argv, VALUE self) {
...
+ *   dead_code_candidates -> Enumerator[Rubydex::Declaration]
+ *
+ * Returns an enumerator that yields constant-like declarations (namespace, constant, or constant alias) that have zero
+ * resolved constant references and at least one definition. Declarations with a rubydex-seeded built-in definition are
+ * excluded, even when user code reopens them. Requires a resolved graph and covers everything indexed, including
+ * dependencies.
+ *
+ * The result is a working set, not an inventory. Deleting a reported declaration can leave its container unreferenced,
+ * so re-index and call again until the result comes back empty. Accuracy follows the graph's reference data, so
+ * whatever the indexer cannot see goes uncounted.
```
vinistock: Let's simplify this one too.

test/graph_test.rb
```diff
@@ -297,6 +297,81 @@ def test_graph_search_requires_at_least_one_query
     assert_raises(ArgumentError) { graph.fuzzy_search }
   end
 
+  def test_graph_dead_code_candidates
```
vinistock, line 300: There's no need to repeat the same scenarios that exist on the Rust side. The only Ruby specific portion is actually being able to call `dead_code_candidates` and getting an enumerable back.

Let's simplify these in a single test that just shows how the API can be used, but not create a matrix of scenarios (that belongs in the Rust side).

rust/rubydex/src/query.rs
```diff
@@ -833,6 +833,69 @@ pub fn follow_method_alias(graph: &Graph, alias_id: DefinitionId) -> Result<Decl
...
+            .chunks(chunk_size)
+            .map(|chunk| {
+                s.spawn(|| {
+                    chunk
+                        .iter()
+                        .filter(|id| {
+                            let decl = declarations.get(id).unwrap();
+                            // Methods and variables are excluded here: rubydex cannot attribute method
+                            // calls to declarations, so called and uncalled both report zero references.
+                            decl.constant_references().is_some_and(HashSet::is_empty)
```
Morriar: This also includes explicit singleton classes from `class << Foo` blocks: they are `Namespace` declarations with a definition and usually no constant references, so we end up reporting `Foo::<Foo>` even though singleton classes aren't constants. The actionable dead code there is the method inside the singleton class, which this endpoint intentionally doesn't handle.

Let's exclude `Namespace::SingletonClass(_)` from this query.

Suggested test: index `class Foo; end; class << Foo; def never_called; end; end`, resolve the graph, and assert `candidate_names` does not include `Foo::<Foo>`. That fails on this branch because `Foo::<Foo>` is currently returned.

test/graph_test.rb
mathyasp: That makes sense! Updated this

rust/rubydex/src/query.rs
mathyasp: Yeah I see! This has been updated 👍

rust/rubydex/src/query.rs
```diff
@@ -833,6 +833,63 @@ pub fn follow_method_alias(graph: &Graph, alias_id: DefinitionId) -> Result<Decl
     }
 }
 
+/// Returns `true` when any definition of `declaration` was seeded by rubydex rather than read from indexed source.
+///
+/// Candidates are reported at declaration granularity, so a seeded definition disqualifies the declaration even when
```
vinistock: Nit: this file is not exclusive to dead code detection, so let's be specific about the context in which built-in declarations can be disqualified.

```suggestion
/// Dead code candidates are reported at declaration granularity, so a seeded definition disqualifies the declaration even when
```

rust/rubydex/src/query.rs
```diff
@@ -833,6 +833,63 @@ pub fn follow_method_alias(graph: &Graph, alias_id: DefinitionId) -> Result<Decl
...
+fn is_built_in(graph: &Graph, declaration: &Declaration) -> bool {
+    declaration.definitions().iter().any(|definition_id| {
+        graph
+            .definitions()
+            .get(definition_id)
+            .is_some_and(|definition| definition.uri_id() == &*BUILT_IN_URI_ID)
+    })
+}
+
+/// Returns a list of declaration that may be unused in the codebase, which means
```
vinistock:
```suggestion
/// Returns a list of declarations that may be unused in the codebase, which means
```

rust/rubydex/src/query.rs
```diff
@@ -3982,4 +4039,166 @@ mod tests {
...
     }
+
+    fn candidate_names(context: &GraphTest) -> Vec<String> {
+        let mut names: Vec<String> = dead_code_candidates(context.graph())
+            .iter()
+            .map(|id| context.graph().declarations().get(id).unwrap().name().to_string())
+            .collect();
+        names.sort();
+        names
+    }
```
vinistock: We usually create assertion macros for these. How about we create an `assert_dead_code_candidates!(context, ["ExpectedDeadConstant"])`?

It will make all of the test cases a bit shorter.

vinistock (approved):
Very nice. Just a few improvements, but this is good to go

rust/rubydex/src/query.rs
mathyasp: Ah yes definitely! Looked at the existing ones and added the `assert_dead_code_candidates` macro and cleaned up the tests 👍

mathyasp:
image

vinistock (approved):

PR #1014, opened 2026-08-18
Expose source-level names for definitions

soutaro:
`Definition#name` returns an unqualified name, while `definition.declaration.name` returns its resolved fully qualified name. Neither preserves whether a definition was written as `Bar`, `Foo::Bar`, or `::Bar`.

This distinction matters when constructing a nesting stack for `Graph#resolve_constant`: a rooted name such as `::Bar` must remain rooted instead of being resolved relative to its lexical owner.

Add <s>`source_name`</s> `raw_name` to class, module, constant, and constant-alias definitions to expose this source-level qualification.

### DIFF
--- a/rust/rubydex-sys/src/definition_api.rs
+++ b/rust/rubydex-sys/src/definition_api.rs
@@ -6,7 +6,9 @@ use crate::location_api::{Location, create_location_for_uri_and_offset};
 use crate::reference_api::CConstantReference;
 use libc::c_char;
 use rubydex::model::definitions::{Definition, Mixin};
-use rubydex::model::ids::DefinitionId;
+use rubydex::model::graph::Graph;
+use rubydex::model::ids::{DefinitionId, NameId};
+use rubydex::model::name::ParentScope;
 use rubydex::query::AliasResolutionError;
 use std::ffi::CString;
 use std::ptr;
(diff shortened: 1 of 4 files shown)


rust/rubydex-sys/src/definition_api.rs
```diff
@@ -110,6 +112,64 @@ pub unsafe extern "C" fn rdx_definition_name(pointer: GraphPointer, definition_i
...
+///
+/// This function will panic if the definition cannot be found or if it is not a class, module, or constant definition.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn rdx_definition_source_name(pointer: GraphPointer, definition_id: u64) -> *const c_char {
+    with_graph(pointer, |graph| {
+        let def_id = DefinitionId::new(definition_id);
+        let definition = graph
+            .definitions()
+            .get(&def_id)
+            .unwrap_or_else(|| panic!("Definition not found: {definition_id:?}"));
```
st0012, line 166: Would we trigger this panic if:
1. We retrieved definitions at Ruby level with `foo_decl.definitions`
2. Something reindexed the codebase, and `foo_decl`'s definitions are changed
3. We then call `definitions.each(&:source_name)`

If it's possible, then we should avoid panic here as it'll directly crash the process. Perhaps we can return `null` here and raise exceptions at the C extension level?

st0012 (commented):
I think `source` is often referred to the "source code" of things. So it wasn't clear to me that it means "source level name". I think either `source_level_name` or `written_name` would be clearer? @vinistock @Morriar WDYT?

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -110,6 +112,64 @@ pub unsafe extern "C" fn rdx_definition_name(pointer: GraphPointer, definition_i
...
+
+fn definition_source_name_id(definition: &Definition) -> NameId {
+    match definition {
+        Definition::Class(definition) => *definition.name_id(),
+        Definition::Module(definition) => *definition.name_id(),
+        Definition::Constant(definition) => *definition.name_id(),
+        Definition::ConstantAlias(definition) => *definition.name_id(),
+        _ => panic!("source_name is only available for class, module, and constant definitions"),
+    }
+}
```
vinistock, line 145: This is an exact re-implementation of `Definition#name`, no? I think we can just do

```rust
definition.name_id().unwrap()
```

rust/rubydex-sys/src/definition_api.rs
vinistock: Yeah, this is also something I was discussing with Rocha. Currently, our FFI is too coupled with the Ruby VM, which provides us with some guarantees that allow us to panic.

However, `rubydex-sys` is a public crate and C API. We shouldn't panic all the time or couple the logic with the VM guarantees.

Let's return a `null` pointer if we fail to find the definition, so that the API consumer can decide how to handle it (and then we can raise on the Ruby side).

vinistock (commented):
I agree that `source_name` is not super clear. My vote would actually go to match Prism's naming `constant_path`. It avoids things derived from `name`, which can be confusing.

What do you think?

Morriar (commented):
Could also be called `written_name`

rust/rubydex-sys/src/definition_api.rs
soutaro: Addressed to return `null` pointer.

rust/rubydex-sys/src/definition_api.rs
soutaro: This function returns `None` for singleton class definitions, different from `Definition#name`.

rust/rubydex-sys/src/definition_api.rs
```diff
@@ -110,6 +112,72 @@ pub unsafe extern "C" fn rdx_definition_name(pointer: GraphPointer, definition_i
...
+    let name = graph
+        .names()
+        .get(&name_id)
+        .expect("name should exist while building raw_name");
+    let simple_name_id = *name.str();
+    let simple_name = graph
+        .strings()
+        .get(&simple_name_id)
+        .expect("string should exist while building raw_name")
+        .as_str();
```
vinistock:
```suggestion
    let simple_name = graph
        .strings()
        .get(name.str())
        .expect("string should exist while building raw_name")
        .as_str();
```

vinistock (approved):

PR #1018, opened 2026-08-19
Filter built-in and linter diagnostics uniformly

vinistock:
Closes #1000

This PR unifies the diagnostic filtering, so that built-in (graph) and linter diagnostics are treated consistently.

I also noticed that we were filtering based on path exclusions/inclusions separately 3 times, so I simplified the code to do it within the same loop, which is a bit easier to follow.

Lastly, I also renamed the glob flags constant. Not sure why we had a `RUBOCOP` prefix, but we can probably remove it.

### DIFF
--- a/lib/rubydex/linter/runner.rb
+++ b/lib/rubydex/linter/runner.rb
@@ -16,60 +16,40 @@ module Rubydex
         @graph = graph
         @config = config
         @custom_rules = custom_rules.select { |rule| config.rule_enabled?(rule) }.sort_by(&:rule_name)
-        @dependency_paths = Gem.path #: Array[String]
+        @dependency_patterns = Gem.path.map { |path| "#{path}/**/*" } #: Array[String]
       end
 
       #: () -> Array[Diagnostic]
       def run
-        rule_diagnostics = @custom_rules.flat_map do |rule_class|
+        diagnostics = @graph.diagnostics
+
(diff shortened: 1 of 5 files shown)


lib/rubydex/linter/runner.rb
```diff
@@ -16,60 +16,37 @@ def initialize(graph, custom_rules:, config:)
...
-        end
-      end
+          path = diagnostic.location.to_file_path
 
-      #: (Location, Array[String]) -> bool
-      def location_excluded?(location, patterns)
-        path = location.to_file_path
-        return true if @dependency_paths.any? do |dependency_path|
-          path == dependency_path || path.start_with?("#{dependency_path}/")
+          path.start_with?(workspace_path) &&
```
Morriar, line 46: Let's also apply `enabled = false` here. Right now custom rules are filtered out before they run, but built-in diagnostics still make it through this unified path, so this config keeps reporting parse errors:

```toml
[linter.rules.ParseError]
enabled = false
```

We can include `@config.rule_enabled?(diagnostic.rule)` in this predicate and add the matching `ParseError` regression test.

vinistock: Ah, great catch! I totally forgot about that.

st0012 (approved):

Morriar (approved):

PR #1019, opened 2026-08-20
Add complete linter documentation

### LINKED ISSUE #1002: [Documentation] Have a rule-author facing documentation on how to write linter rules with Rubydex
## Outline

- Introduction
    - It's currently only a basic infrastructure without built-in rules
- Related commands
    - `rdx lint`
    - `rdx lint explain`
- How are rules discovered
    - In app or a gem
- How to structure a rule class
- How to test a rule
- Ruby LSP addon for the linter

vinistock:
Closes #1002

This PR adds full documentation on how to use and write custom linting rules.

### DIFF
--- a/docs/linter.md
+++ b/docs/linter.md
@@ -1,0 +1,191 @@
+# Rubydex linter
+
+Rubydex ships with a built-in linter and infrastructure to write custom rules. It includes the built-in rules that are part of the core analysis by default, such as parse errors, dynamic constant references, and others that will be available in the future like undefined constant, cyclic ancestor, superclass mismatch and so on. The goal is to provide a unified interface for linting the code from both a base analysis and custom rule perspective.
+
+## Usage
+
+### Commands
+
+The linter includes the following commands:
+
+```bash
+# Lint the codebase.
+bundle exec rdx lint
+
+# Explain a given rule. This works for custom rules made available by dependencies or by
+# the current workspace. The command prints the documentation written on top of a rule definition,
+# which should provide longer explanations than a diagnostic message.
+bundle exec rdx lint explain MyRule
+```
+
+### Configuration
+
+All Rubydex tools are always configured in `rubydex.toml` within their specific sections. Linter configurations allow users to disable rules, change their severity or exclude certain paths.
+
+```toml
+[linter.rules.MyCustomRule]
+enabled = true
+exclude = ["path_to_skip/**"]
+severity = "warning" # Valid values: "hint" | "information" | "warning" | "error"
+```
+
+### Editor support
+
+Linting results are automatically surfaced by the [Ruby LSP](https://github.com/Shopify/ruby-lsp). The only caveat is that only the beta version of the Ruby LSP is currently supported due to the Rubydex requirement. We are working on stabilizing a release of the Ruby LSP to remove this requirement. You can enable the beta in VS Code with the following setting
+
+```jsonc
+{
+  // VS Code settings. Enable the beta version of the server
+  "rubyLsp.featureFlags": {
+    "betaServer": true
+  }
+}
+```
+
+For other editors, pass the `betaServer` flag through [initialization options](https://shopify.github.io/ruby-lsp/editors.html).
+
+## Creating custom rules
+
+Rubydex ships with the base infrastructure allowing developers to create their own rules, which can exist inside of a given project or even published as a gem. Rules are loaded automatically by Rubydex as long as they follow the expected path patterns:
+
+- gems: `lib/rubydex_linter/rules/**/*`
+- project: `rubydex_linter/rules/**/*`
+
+
+**IMPORTANT**: Rubydex's analysis is still being built and therefore it cannot reason about every kind of Ruby code. For example, local variables are not yet supported. Rule implementations are limited to what the graph can currently offer.
+
+Custom rules inherit from `Rubydex::Linter::CustomRule` and must define the `lint` method. All rules have access to a fully populated `Rubydex::Graph`, which can be inspected for finding possible mistakes, in addition to convenience helpers.
+
+Consider the following example, which is a rule to prohibit re-opening Rails models:
+```ruby
+module Rubydex
+  module Linter
+    module Rules
+      # Prohibits Rails models from being re-opened (more than one definition).
+      #
+      # Note: rule class names become the config key.
+      class ProhibitModelReopen < CustomRule
+        class << self
+          # The default severity of this rule, which can be overridden by the user's configuration.
+          #
+          # @override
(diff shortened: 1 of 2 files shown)


docs/linter.md
```diff
@@ -0,0 +1,180 @@
...
+
+# Explain a given rule. This works for custom rules made available by dependencies or by
+# the current workspace. The command prints the documentation written on top of a rule definition,
+# which should provide longer explanations than a diagnostic message.
+bundle exec rdx lint explain MyRule
+```
+
+### Configuration
+
+All Rubydex tools are always configured in `rubydex.toml` within their specific sections. Linter configurations allow users to disable rules, change their severity or exclude certain paths.
```
Morriar: Let's scope this to custom rules for now. The runner only applies `LinterConfig` to `custom_rules`; graph diagnostics (`@graph.diagnostics`, including parse errors, dynamic constant references, etc.) are always reported with their default severity and no per-rule excludes. As written, this suggests `[linter.rules.ParseError]` or `[linter.rules.DynamicConstantReference]` can be disabled or have their severity changed, which doesn't work today.

docs/linter.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Rules should not read information from disk. The design of the linter is to operate on a graph object. Trying to read from disk from the implementation of a rule not only introduces IO in the middle of the analysis, but it may operate on a stale version of the codebase for scenarios like language servers
+- Fewer generic rules are better than many highly specific ones. For example, instead of writing 2 separate rules to check if models inherit from `ApplicationModel` and controllers from `ApplicationController`, consider creating a `RequiredParentClass` rule that can be applied generically and configured
+
+### Testing a rule
+
+Rubydex provides the `RuleTestCase` parent class to conveniently test rules with a more visual approach.
+
+```ruby
+require "rubydex/linter/rule_test_case"
```
Morriar: If this is meant to be a copy-pasteable test file, let's require the rule under test as well. `RuleTestCase#rule_class` infers `Rubydex::Linter::Rules::ProhibitModelReopen`, but it doesn't load it, so this example will raise `NameError` unless the rule has already been required.

```suggestion
require "rubydex/linter/rule_test_case"
require "rubydex_linter/rules/prohibit_model_reopen"
```

docs/linter.md
```diff
@@ -0,0 +1,180 @@
...
+```toml
+[linter.rules.MyCustomRule]
+enabled = true
+exclude = ["path_to_skip/**"]
+severity = "warning" # Valid values: "hint" | "information" | "warning" | "error"
+```
+
+### Editor support
+
+Linting results are automatically surfaced by the [Ruby LSP](https://github.com/Shopify/ruby-lsp). The only caveat is that only the beta version of the Ruby LSP is currently supported due to the Rubydex requirement. We are working on stabilizing a release of the Ruby LSP to remove this requirement.
```
st0012, line 34: Is there a doc somewhere explaining how to use beta version of Ruby LSP?

docs/linter.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Creating custom rules
+
+Rubydex ships with the base infrastructure allowing developers to create their own rules, which can exist inside of a given project or even published as a gem. Rules are loaded automatically by Rubydex as long as they follow the expected path patterns:
+
+- gems: `lib/rubydex_linter/rules/**/*`
+- project: `rubydex_linter/rules/**/*`
+
+
+**IMPORTANT**: currently, Rubydex's analysis is limited to declarations. We are working on analyzing code flow and type inference, but that is not yet ready. Therefore, custom linting rules can only reason about declaration mistakes at the moment.
```
st0012, line 44: I'm not sure if most readers will already know what `declarations` means, and what code flow and type inference will unlock. Can we explain this a bit differently and with examples? Like:

`````
Currently, Rubydex's analysis only provide information on class,module,constants,and methods. It doesn't provide information on conditionals, begin blocks, method call arguments...etc.

For example:

```rb
class Foo # ✅
  def bar # ✅
    if is_that_true? # Rubydex sees is_that_true? is called, but it doesn't see the if statement
      return false # it doesn't see the return
    else 
      baz
    end
  rescue FooError # it sees FooError as a constant reference, but not the rescue
    qux
  end
end
```

`````

File: docs/linter.md [hunk 5]
```diff
@@ -0,0 +1,180 @@
...
+
+Custom rules inherit from `Rubydex::Linter::CustomRule` and must define the `lint` method. All rules have access to a fully populated `Rubydex::Graph`, which can be inspected for finding possible mistakes, in addition to convenience helpers.
+
+Consider the following example, which is a rule to prohibit re-opening Rails models:
```
st0012, line 48: Let's add a code example on what's prohibited and how the diagnostic can be surfaced.

docs/linter.md
```diff
@@ -0,0 +1,180 @@
...
+graph["Foo"].descendants
+
+# bad: loops through the entire graph to find all methods named `is_a?` declared on any type
+graph.declarations.select { |d| d.is_a?(Rubydex::Method) && d.name.end_with?("#is_a?()") }
+
+# good: uses the search API to find all matches directly
+graph.search("#is_a?()")
+```
+
+- Rules should not read information from disk. The design of the linter is to operate on a graph object. Trying to read from disk from the implementation of a rule not only introduces IO in the middle of the analysis, but it may operate on a stale version of the codebase for scenarios like language servers
```
st0012, line 124: But for many cases this is necessary now, and we do do it in Core too. I feel it's too early to advise against this.

docs/linter.md
vinistock: Not really, but I added a brief explanation.

docs/linter.md
vinistock: I re-wrote the phrase to be a bit more generic instead of using an example. Let me know what you think.

docs/linter.md
vinistock: There's already an example for the rule in the tests section. Also, I'd argue the concept of the rule here is quite simple and probably be understood from the description alone.

Do you find that people may not understand what we're trying to achieve?

docs/linter.md
vinistock: A linting rule reading from disk is never correct because it ties to the linting logic to the frontend where it is being surface (the CLI).

For the LSP, files need to be linted before they are committed to disk, so reading from disk is not just a "not ideal" because of the IO, it's actually incorrect behaviour (linting the stale version of the code).

I understand we took shortcuts to be able to express certain rules right now before we have support for introspecting a CFG in custom linting rules, but the advice in this point stands nonetheless and we should be explicit that it's a smell.

st0012 (approved):

PR #1020, opened 2026-08-24
Enforce that linter rules must be documented

vinistock:
I propose that we enforce custom rules are always documented. It's helpful for maintainers to know what the rule is intended for and powers the explain command.

This PR adds a check if any of the definitions includes comments and then adds a violation if the rule is undocumented.

### DIFF
--- a/lib/rubydex_linter/rules/rule_structure.rb
+++ b/lib/rubydex_linter/rules/rule_structure.rb
@@ -37,6 +37,8 @@ module Rubydex
           rule_definitions_by_file = {} #: Hash[String, Hash[Rubydex::Class, Definition]]
 
           rules.each do |rule|
+            rule_docs = []
+
             rule.definitions.each do |rule_definition|
               uri = rule_definition.document.uri
               path = path_for_uri(uri)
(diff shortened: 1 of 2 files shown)


lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -48,16 +50,28 @@ def lint
               elsif !test_file?(path)
                 report_wrong_rule_directory(rule.name, rule_definition)
               end
+
+              rule_docs.concat(rule_definition.comments)
```
Morriar: I think this should only count comments from definitions that are part of the checked rule surface. `rdx lint` indexes the whole workspace, so a documented test fixture or test-only reopening of the same rule can make `rule_docs` non-empty even when the actual rule file has no documentation. `lint explain` only indexes `RuleLoader.paths`, so that rule would still have no useful documentation there.

Maybe we can gate this on the same predicate used to pick `rule_definition`?

```suggestion
              rule_docs.concat(rule_definition.comments) if rule_file?(path) || !test_file?(path)
```

Let's also add a regression test with an undocumented rule under `rubydex_linter/rules/` and a commented reopening under `test/` so we don't accidentally count test comments as rule docs.

st0012 (approved):

PR #1022, opened 2026-08-24
Auto generate built in rules

vinistock:
The existing mechanism of generating the rule classes for built-in rules dynamically using the C extension didn't feel right. Not only does it feel a bit hacky, but it means that the `explain` command cannot provide long form documentation for built-in rules. That's definitely not what we want. If we add some resolution diagnostic, we want to be able to provide a clear explanation of why the code pattern is being flagged.

This PR shifts from the dynamic definition to compile-time auto generating the Ruby class definitions for each built-in rule. It's hooked to the compile task, so that releasing the gem always includes these classes.

I recommend reviewing by commit:

1. Add a Rust binary that extracts the documentation from the built-in rules and auto-generates the Ruby classes (single source of truth for the docs). **Note**: these rules are placed in `lib/rubydex/rules` (as opposed to `lib/rubydex_linter/rules`). The reason is because our API allows reading diagnostics from `graph` even when the linter is not loaded, which means these rule classes _must be loaded_ with the default Rubydex
2. Removed the previous dynamic generation code from C
3. Made a few tweaks to allow explaining built-in rules
4. This commit only has the auto-generated Ruby classes
5. Adds a CI step to ensure that any rule modifications are committed

image


lib/rubydex/rules/parse_error.rb
```diff
@@ -0,0 +1,25 @@
...
+module Rubydex
+  module Rules
+    # A parse error represents invalid Ruby syntax and a program that will fail to execute. For example, a missing
+    # `end`, an unterminated string, a missing parenthesis.
+    #
+    # ```ruby
+    # class Foo
+    # ^^^^^ Syntax error. Missing end token
+    # ```
+    class ParseError < Rule
```
Morriar: Since these classes are now regular Ruby files, I think we need to keep the RBI in sync too. `rbi/rubydex.rbi` currently removes all of the `Rubydex::Rules::*` declarations, so Sorbet consumers won't be able to reference `Rubydex::Rules::ParseError` even though the constants now exist at runtime. Could the generator also emit or update the RBI entries for the built-in rules?

st0012 (approved):
Can we also update AGENTS.md and/or CONTRIBUTING.md to mention that those generated rules should not be updated and the command needed to update them?

PR #1023, opened 2026-08-24
Recognize linter rules in nested gems

st0012:
`RuleLoader` can discover rules under the `lib` directory of a nested gem, but `RuleStructure` accepts that directory only at the workspace root. This change accepts `lib/rubydex_linter/rules/**/*.rb` at any depth in the workspace, so in-repo gems no longer need a `RuleStructure` exclusion.

### DIFF
--- a/lib/rubydex_linter/rules/rule_structure.rb
+++ b/lib/rubydex_linter/rules/rule_structure.rb
@@ -19,7 +19,6 @@ module Rubydex
         RULE_NAMESPACE = "Rubydex::Linter::Rules" #: String
         RULE_FILE_PATTERNS = [
           "rubydex_linter/rules/**/*.rb",
-          "lib/rubydex_linter/rules/**/*.rb",
           "**/lib/rubydex_linter/rules/**/*.rb",
         ].freeze #: Array[String]
         TEST_FILE_PATTERNS = ["test/**/*", "**/test/**/*"].freeze #: Array[String]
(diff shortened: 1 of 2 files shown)


lib/rubydex_linter/rules/rule_structure.rb
```diff
@@ -19,6 +20,7 @@ class RuleStructure < CustomRule
         RULE_FILE_PATTERNS = [
           "rubydex_linter/rules/**/*.rb",
           "lib/rubydex_linter/rules/**/*.rb",
+          "**/lib/rubydex_linter/rules/**/*.rb",
```
Morriar: I think we still need to keep test directories excluded before accepting this as a rule file. `**/lib/rubydex_linter/rules/**/*.rb` also matches paths like `test/fixtures/lib/rubydex_linter/rules/foo.rb`, and `lint` checks `rule_file?` before `test_file?`. That means test fixtures under a nested `lib/rubydex_linter/rules` path will now get namespace / multiple-rule diagnostics even though the rule says classes under test directories are ignored.

Could we check `test_file?` first, or add a regression test for this shape?

vinistock:
As spoken offline, rules defined by gems are already loaded via `Gem.find_latest_files`, even if the gem is inside of the workspace.

What we may want is avoiding loading the gem's own exported rules when running the linter inside of it since the rules may depend on definitions that don't exist in that context.

lib/rubydex_linter/rules/rule_structure.rb
vinistock, line 22: The pattern above is fully covered by this one. I think we can remove it.

vinistock (approved):

PR #1024, opened 2026-08-25
Accept definitions as constant resolution contexts

soutaro:
`Graph#resolve_constant` currently requires callers to reconstruct lexical nesting as an array of strings. Definitions already retain the exact `NameId` needed by the resolver, including explicit qualification and singleton class context.

The Ruby API now accepts `graph.resolve_constant("CONST", definition)` for `ClassDefinition`, `ModuleDefinition`, and `SingletonClassDefinition` while preserving the existing `Array[String]` API. Definitions from another graph, deleted definitions, and unsupported definition types are reported separately.

### DIFF
--- a/ext/rubydex/graph.c
+++ b/ext/rubydex/graph.c
@@ -1,6 +1,7 @@
 #include "graph.h"
 #include "config.h"
 #include "declaration.h"
+#include "definition.h"
 #include "diagnostic.h"
 #include "document.h"
 #include "location.h"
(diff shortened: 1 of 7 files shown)


rust/rubydex-sys/src/name_api.rs
```diff
@@ -47,8 +66,8 @@ pub fn nesting_stack_to_name_id(
 fn process_qualified_name(
     graph: &mut Graph,
     qualified_name: &str,
+    current_nesting: Option<NameId>,
```
soutaro, line 71: Changed this to pass by value, since `current_nesting` is not mutated by the function.

ext/rubydex/graph.c
```diff
@@ -510,26 +511,50 @@ static VALUE rdxr_graph_set_encoding(VALUE self, VALUE encoding) {
...
+
+        if (result.status == CDefinitionConstantResolution_DefinitionNotFound) {
+            rb_raise(rb_eRuntimeError, "Definition not found");
+        }
+        if (result.status == CDefinitionConstantResolution_InvalidDefinitionKind) {
+            rb_raise(
+                rb_eTypeError,
+                "context must be an Array of Strings, ClassDefinition, SingletonClassDefinition, or ModuleDefinition"
+            );
+        }
```
vinistock, line 547: It seems that we're also making this check with the `rb_obj_is_kind_of`. Do we need both? Maybe we can delegate to the FFI and just check the result?

rust/rubydex-sys/src/name_api.rs
```diff
@@ -13,26 +13,45 @@ pub fn nesting_stack_to_name_id(
...
+    graph: &mut Graph,
+    const_name: &str,
+    nesting: Option<NameId>,
+) -> Option<(NameId, Vec<NameId>)> {
+    let mut current_name = ParentScope::None;
+    let mut names_to_untrack = Vec::new();
+
+    process_qualified_name(graph, const_name, nesting, &mut current_name, &mut names_to_untrack);
+
+    let (ParentScope::Some(name_id) | ParentScope::Attached(name_id)) = current_name else {
```
vinistock, line 52: Let's add a comment explaining what this is doing. It wasn't immediately obvious that this is treating broken names.

test/graph_test.rb
```diff
@@ -423,6 +538,20 @@ def test_graph_resolve_with_invalid_argument
     assert_raises(ArgumentError) do
       graph.resolve_constant("CO\0NST", ["Foo"])
     end
+
+    other_graph = Rubydex::Graph.new
+    other_graph.index_source("file:///foo.rb", "class Foo; def target; end; end", "ruby")
+    other_graph.resolve
```
vinistock: Instead of creating another graph for this test, can we reuse the same one? It seems we just need to index a file and resolve. I think the other assertions should all pass too since they are making type-level verifications.

vinistock (approved):

test/graph_test.rb
```diff
@@ -401,6 +401,121 @@ class Bar::Baz
...
+      graph.index_all(context.glob("**/*.rb"))
+      graph.resolve
+
+      definitions = graph.document(context.uri_to("foo.rb")).definitions
+      rooted_bar_definition = definitions.find do |definition|
+        definition.is_a?(Rubydex::ClassDefinition) && definition.name == "Bar" && definition.lexical_owner&.name == "Foo"
+      end
+
+      assert_equal("Bar::Baz", graph.resolve_constant("Baz", rooted_bar_definition).name)
+      assert_equal("Foo::OUTER_CONST", graph.resolve_constant("OUTER_CONST", rooted_bar_definition).name)
```
paracycle: It would be really cool to have a case where we resolve the same `"Baz"` name but in two different contexts and expect them to resolve to two different constants.

Something like:
```ruby
irb(main):001* class Baz
irb(main):002*   class Foo
irb(main):003*     def foo = Baz;
irb(main):004*   end
irb(main):005> end
irb(main):006>
irb(main):007* class Bar
irb(main):008*   module Baz; end
irb(main):009*
irb(main):010*   module Foo
irb(main):011*     def foo = Baz;
irb(main):012*   end
irb(main):013*
irb(main):014*   include Foo
irb(main):015> end
=> Bar
irb(main):016> Baz::Foo.new.foo
=> Baz
irb(main):017> Bar.new.foo
=> Bar::Baz
```
so:
```ruby
baz_foo_foo_def = graph["Baz::Foo#foo()"]
bar_foo_def = graph["Bar#foo()"]

graph.resolve_constant("Baz", baz_foo_foo_def).name # => "Baz"
graph.resolve_constant("Baz", bar_foo_def).name # => "Bar::Baz"
```

ext/rubydex/graph.c
soutaro: Yeah, this is unreachable because of `rb_obj_is_kind_of` test. Changed the `rb_raise` message.

test/graph_test.rb
soutaro: This is a test that passing other graph's definition raises an error. No other tests verifies this, so simply skip this assertions.

Issue #1027, opened 2026-08-26
built-in core declarations carry no members


HoneyryderChuck:
This is a report generated by claude while investigating a surge of errors following the enablement of `UseProjectIndex` in rubocop 1.90 (rubydex integration) on the ArgumentMismatch cop and a misinterpretation of a redefinition of new (see [this](https://github.com/rubocop/rubocop/pull/15610) and [this](https://github.com/rubocop/rubocop/pull/15611) for more details). Not sure if the below is undefined or known-and-accepted behaviour, so posting it as an issue for now.

--------------------------

`add_built_in_data` indexes an RBS snippet that declares `BasicObject`, `Kernel`, `Object`, `Module` and `Class` with empty bodies. That is enough for what it was written for — ancestor linearization resolves correctly without the full RBS core — but it leaves the graph with no core methods at all. `Class` and `Module` have zero members.

A consumer cannot tell "this name does not exist" from "this name is core and therefore invisible", because both come back nil. Two consequences show up immediately in RuboCop's index-backed cops:

1. `Object.const_get` reads as an unknown name. `Lint/NameTypo` then spell-checks it against the members it can see and suggests whatever is closest — in a large project that is some unrelated private top-level `def`, so `const_get` is reported as a possible typo for `context`.

2. Lookup does not stop where Ruby stops. `Report.new` walks the singleton ancestry for `new`, finds nothing on `Class` or `Module`, and continues to `Object`'s instance methods, which sit *below* `Class#new` in the method resolution order. One bare `def new` at the top level of any file in the project — the shape every block-based DSL produces — then answers every constructor lookup in that project.

Reproduction [here](https://github.com/Shopify/rubydex/pull/1028)

vinistock:
Thank you for the report, however, there's no real bug here.

The base classes `Object`, `Kernel`, `BasicObject`, `Module` and `Class` are created built-in because they must exist ahead of time for correct membership association. It's not the right expectation that the built-ins contain the full information about those entities.

The complete information for those comes from the `rbs` gem, which gets indexed as part of invoking `Graph#index_workspace`. If you do not index the `rbs` files for core definitions, as presented in the test, then you will indeed not have the information.

HoneyryderChuck:
@vinistock thx for the feedback. Then if I understood correctly, rubocop has to do it in its integration? I ask this because we have the `rbs` in the bundle, so it should have picked it up if so, right?

vinistock:
[...]

PR #1029, opened 2026-08-28
Expose core RBS definition paths as public API

HoneyryderChuck:
Tools that build their own list of paths and index it with `index_all` instead of `index_workspace` (RuboCop's `AllCops/UseProjectIndex` does this, so that its own `Include`/`Exclude` configuration governs the index) ended up with a graph that has no definitions for Ruby's core classes. The core definitions come from indexing the `rbs` gem's `core` and `stdlib` signature directories, which only `index_workspace` did, through the private `add_core_rbs_definition_paths`.

Add `Rubydex::Graph.core_rbs_definition_paths`, which returns those paths, so callers can append them to their own list without reaching for a private method. `add_core_rbs_definition_paths` now delegates to it, keeping a single source of truth as well as the existing semantics: the latest installation of `rbs` in the system wins, `rbs` does not have to be part of the bundle and no installation being found is not an error.

-----

@vinistock this is to address what we discussed in #1027; rubocop can't really use `graph_workspace` as that'd mean indexing the 3rd party gems too; the easiest workaround I got to was to use this method to expose the rbs sigs there, which would be better served as a public API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,9 @@ graph = Rubydex::Graph.new
 graph.encoding = "utf16"
 # Index the entire workspace with all dependencies
 graph.index_workspace
-# Or index specific file paths
-graph.index_all(["path/to/file.rb"])
+# Or index specific file paths. Include the core and standard library RBS definition paths, otherwise the graph
+# will have no definitions for Ruby's own core classes
+graph.index_all(["path/to/file.rb", *Rubydex::Graph.core_rbs_definition_paths])
 # Transform the initially collected information into its semantic understanding by running resolution
 graph.resolve
 # Get all diagnostics acquired during the analysis
--- a/lib/rubydex/graph.rb
+++ b/lib/rubydex/graph.rb
@@ -17,6 +17,25 @@ module Rubydex
         graph.load_config(Config.load(workspace_path))
         graph
       end
+
+      # Returns the paths for the core and standard library RBS definitions of the latest installation of the `rbs` gem,
+      # which are the definitions for Ruby itself. Tools that build their own list of paths and index it with
+      # `index_all` must append these paths, otherwise the graph will have no definitions for core classes like `Object`
+      # or `Kernel`.
+      #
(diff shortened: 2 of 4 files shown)


vinistock (changes requested):
Thanks for the contribution. Exposing this method is not a problem, but I want to call attention that performing the analysis in an incomplete graph can have unexpected results and this is not going to fix the problem 100%.

Constants in Ruby are global and can be influenced by any code included in the application, which includes dependencies.

Consider this example:

```ruby
# rbs definition
class Object; end

# some_gem
class UsefulParentClass
  class Object; end
end

# your code
class MyImplementation < UsefulParentClass
  # This Object reference points to UsefulParentClass::Object and not Object.
  # By not including the full context of the application into the analysis, constants
  # may resolve differently and you might be inaccurate results.
  Object

  # Note that inaccuracies in constant resolution can easily propagate. For example:
  #
  # This class inherits from UsefulParentClass::Object and not Object. By incorrectly
  # understanding its parent class, we resolve every constant reference inside of this
  # namespace against the wrong ancestor chain, which produces more inaccuracies in the
  # analysis.
  class SomeOtherThing < Object
  end
end
```

Exposing the method you're proposing would make the analysis less incomplete by including the core definitions, but it will still be incomplete and subject to more inaccuracies if you don't take dependencies into account.

This is why in the [Rubydex linter](https://github.com/Shopify/rubydex/blob/main/docs/linter.md) included in this gem, we filter diagnostics from dependencies only at the end during presentation and never exclude anything from the analysis as it may impact the understanding of the code.

HoneyryderChuck:
@vinistock I understand. When it comes to rubocop though, it really depends on how they'd like the integration to go forward. [indexing the 3rd party gems is an opt-in](https://docs.rubocop.org/rubocop/latest/usage/project_index.html),in which case `graph_workspace` could be employed. But what you're implying leads me to question whether that opt-in should even be a thing. Perhaps something to take to rubocop team.

PR #1032, opened 2026-08-31
Require yaml in Rubydex::Skill

### LINKED ISSUE #1031: `rdx skill <id>` raises `NameError` — `Rubydex::Skill` uses `YAML`/`Psych` without requiring `yaml`
`rdx skill <id>` crashes on Ruby 4.0.3, so a skill can be listed but never loaded.

```
$ rdx skill send-private-method
lib/rubydex/skill.rb:70:in 'Rubydex::Skill.parse_frontmatter':
  uninitialized constant #<Class:Rubydex::Skill>::Psych (NameError)
      rescue Psych::Exception => e
lib/rubydex/skill.rb:69:in 'Rubydex::Skill.parse_frontmatter':
  uninitialized constant #<Class:Rubydex::Skill>::YAML (NameError)
        YAML.safe_load(frontmatter)
```

`lib/rubydex/skill.rb` calls `YAML.safe_load` (:69) and rescues `Psych::Exception` (:70), but its only require is `rubydex/errors` (:3). `grep -rn 'require "yaml"' lib/` returns nothing anywhere in the gem, and Ruby 4.0 does not autoload it — `ruby -e 'YAML'` raises `NameError`. The rescue clause fails to resolve first, hence the doubled error.

Fix — add to `lib/rubydex/skill.rb`:

```ruby
require "yaml"
```

Env: rubydex v0.4.0 (`arm64-darwin`), Ruby 4.0.3, macOS 15.6. Still present on `main` as of today.

Tests likely pass because the test process pulls `yaml` in transitively; `exe/rdx` loads a thinner chain.

Verified: with that one line added, `rdx skill send-private-method` prints the skill body.

deril:
Fixes #1031.

`Skill.parse_frontmatter` calls `YAML.safe_load` (`lib/rubydex/skill.rb:69`) and rescues `Psych::Exception` (:70), but nothing under `lib/` requires `yaml`. Ruby 4.0 no longer autoloads it, so `rdx skill <id>` aborts with a `NameError` on the rescue clause before it can print anything:

```
$ rdx skill send-private-method
lib/rubydex/skill.rb:70:in 'Rubydex::Skill.parse_frontmatter':
  uninitialized constant #<Class:Rubydex::Skill>::Psych (NameError)
```

The test suite doesn't catch it because the test process pulls `yaml` in transitively; `exe/rdx` loads a thinner chain.

With this line added, `rdx skill send-private-method` prints the skill body. Verified on rubydex v0.4.0 (`arm64-darwin`), Ruby 4.0.3, macOS 15.6.

### DIFF
--- a/lib/rubydex/skill.rb
+++ b/lib/rubydex/skill.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "yaml"
 require "rubydex/errors"
 
 module Rubydex


deril:
I have signed the CLA!

vinistock (approved):
Thank you for the fix.

One concern I have is that the skills implementation implicitly depends on yaml, but we want the rubydex gem to be dependency free. 

@st0012 do you know if there are plans to make yaml a bundled gem? That would require us to declare it as a runtime dependency.

PR #1036, opened 2026-09-03
Configure gem and crate release automation

alexcrocha:
This PR does 2 things: 
1. Centralizes gem and crate versions by making the Cargo workspace the source of truth. This allows us to automate releases to both RubyGems and crates.io. 
2. Extends the existing release workflow for trusted publishing on crates.io.

Before this PR merges, we will need to complete the setup of trusted publishing on crates.io before our next release. The steps needed for both `rubydex` and `rubydex-sys` are:
- Open Settings → Trusted Publishing
- Add a GitHub Actions publisher with:
  - Owner: `Shopify`
  - Repository: `rubydex`
  - Workflow: `release.yml`
  - Environment: `release`


.github/workflows/release.yml
```diff
@@ -90,9 +130,37 @@ jobs:
...
+        run: |
+          rustup update --no-self-update stable
+          rustup default stable
+      # The trusted publisher must match this repository, workflow, and environment
+      # for both crates.
+      - name: "Exchange the Actions identity token for a crates.io token"
+        id: "crates_io_auth"
+        uses: "rust-lang/crates-io-auth-action@c6f97d42243bad5fab37ca0427f495c86d5b1a18" # v1.0.5
+      - name: "Publish the crates"
+        run: cargo publish --workspace --manifest-path rust/Cargo.toml
```
Morriar: This fails today because `rubydex-sys` verifies before `rubydex = "=0.4.1"` exists in the crates.io index. I tried:

```sh
cargo publish --dry-run --workspace --manifest-path rust/Cargo.toml
```

and `rubydex-sys` fails in `build.rs` through `cbindgen` / `cargo metadata` with:

```text
failed to select a version for the requirement `rubydex = "=0.4.1"`
candidate versions found which did not match: 0.2.5
```

Let's publish these in order instead: publish `rubydex`, wait/retry until the version is available in the index, then publish `rubydex-sys`. Otherwise the tag workflow will fail after the RubyGems release job has already completed.

rust/Cargo.toml
```diff
@@ -6,6 +6,12 @@ members = [
 
 resolver = "3"
 
+[workspace.package]
+version = "0.4.1"
+
+[workspace.dependencies]
+rubydex = { version = "=0.4.1", path = "rubydex" }
```
vinistock: Why do we need this entry? I'm not sure I understand what it does.

CONTRIBUTING.md
```diff
@@ -57,11 +57,22 @@ To cut a new release:
...
-3. Refresh `Gemfile.lock` so the local `rubydex` spec version matches:
+   ```toml
+   [workspace.package]
+   version = "X.Y.Z"
+
+   [workspace.dependencies]
+   rubydex = { version = "=X.Y.Z", path = "rubydex" }
+   ```
+
+   The Ruby gem dynamically reads its version from this manifest, so do not edit `lib/rubydex/version.rb`. For pre-release versions, Cargo requires a SemVer prerelease identifier such as `X.Y.Z-beta.N`, which `lib/rubydex/version.rb` translates to `X.Y.Z.betaN` for RubyGems.
```
vinistock: I think a lot of this sentence is an implementation detail and not really instructions for releasing.

Let's tailor this section specifically for the instructions:

- Bump the version here. This controls the versions of all crates and the Ruby gem
- If you want to ship a beta release, use the format `1.2.3-beta.1`

CONTRIBUTING.md
```diff
@@ -92,7 +103,8 @@ To cut a new release:
...
 Pushing a tag matching `vX.Y.Z` or `vX.Y.Z.betaN` triggers the release workflow
-in `.github/workflows/cibuildgem.yaml`. That workflow cross-compiles the
-precompiled gems, runs install verification, publishes to RubyGems, and creates
-the GitHub release. Workflow dispatch can be used for a dry run; only tag pushes
-publish a release.
+in `.github/workflows/release.yml`. That workflow verifies that the tag matches
+the declared version, cross-compiles the precompiled gems, runs install
+verification, publishes to RubyGems, publishes the workspace crates to
+crates.io, and creates the GitHub release. You can trigger a dry run with
+workflow dispatch; only tag pushes publish a release.
```
vinistock: Similar here. Are these internal details or relevant instructions for someone trying to release?

I'd honestly not add anything here. If someone tries to push a tag that doesn't match the version, they will notice the mistake.

.github/workflows/release.yml
```diff
@@ -1,16 +1,56 @@
...
+            packages = metadata.fetch("packages")
+            rubydex = packages.find { |package| package.fetch("name") == "rubydex" }
+            rubydex_sys = packages.find { |package| package.fetch("name") == "rubydex-sys" }
+            requirement = rubydex_sys.fetch("dependencies").find { |dependency| dependency.fetch("name") == "rubydex" }.fetch("req")
+            expected = "=#{rubydex.fetch("version")}"
+
+            if requirement != expected
+              abort "rubydex-sys requires rubydex #{requirement.inspect}; expected #{expected.inspect}"
+            end
+          ' <<<"$metadata"
```
vinistock: What is this check protecting us from?

vinistock (commented):
Have we already configured trusted publishing in crates.io?

CONTRIBUTING.md
alexcrocha: 👍 updated as suggested

CONTRIBUTING.md
alexcrocha: I added a mention that it publishes to crates.io as well, and took the opportunity to simplified here

.github/workflows/release.yml
alexcrocha: This was a defensive check to make sure the dependency version was pinned to the crate version. 
It's overkill; removed.

rust/Cargo.toml
alexcrocha: This is a dependency of `rubydex-sys` that I moved here as a convenience so there was only one file to update when bumping the version. As discussed offline, I moved it back to `rubydex-sys/Cargo.toml`

.github/workflows/release.yml
```diff
@@ -1,16 +1,40 @@
...
+    timeout-minutes: 5
+    name: "Verify the tag matches the declared version"
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1"
+      - name: "Setup Ruby"
+        uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1.321.0
+        with:
+          ruby-version: "3.2.9"
```
vinistock, line 25: Does this need to run on Ruby `3.2.9`? If not, we can just remove this and it will pick up the version specified in `.ruby-version`.

.github/workflows/release.yml
```diff
@@ -1,16 +1,40 @@
...
+      - name: "Verify the tag matches the declared version"
+        if: startsWith(github.ref, 'refs/tags/')
```
vinistock, line 27: Should we match the other similar checks?


```suggestion
        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
```

.github/workflows/release.yml
```diff
@@ -1,16 +1,40 @@
...
+        run: |
+          declared="$(ruby -r./lib/rubydex/version -e 'print Rubydex::VERSION')"
+          tagged="${GITHUB_REF_NAME#v}"
+          if [ "$declared" != "$tagged" ]; then
+            echo "Tag ${GITHUB_REF_NAME} does not match the declared version ${declared}" >&2
+            exit 1
+          fi
   compile:
     timeout-minutes: 20
+    needs: verify_version
```
vinistock, line 37: Here we're adding a `needs` requirement to an optional step (since verify_version doesn't run if you trigger the workflow manually for debugging instead of releasing).

Will this work properly if `verify_version` is skipped by the conditional? Or will the entire workflow get stuck?

.github/workflows/release.yml
```diff
@@ -90,9 +114,37 @@ jobs:
         uses: "shopify/cibuildgem/.github/actions/cibuildgem@main"
         with:
           step: "release"
+  release_crates:
+    environment: release
+    permissions:
+      id-token: write
+      contents: read
+    timeout-minutes: 30
```
vinistock, line 122: 30 minutes seems like a lot for just publishing the crates. Maybe 5 is fine?

rust/rubydex-sys/Cargo.toml
```diff
@@ -15,7 +15,7 @@ categories = ["development-tools", "external-ffi-bindings"]
 crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
-rubydex = { version = "0.2.6", path = "../rubydex" }
+rubydex = { version = "=0.4.1", path = "../rubydex" }
```
vinistock, line 18: Did you confirm that if we bump the version in `rust/Cargo.toml` and forget to bump it here, it fails compilation?

vinistock (approved):
A few questions, but the general automation looks good. Let's get it configured on the crates io side and ship 🚀

.github/workflows/release.yml
alexcrocha: Makes sense to me 👍

rust/rubydex-sys/Cargo.toml
alexcrocha: Yes, it fails immediately with
```
error: failed to select a version for the requirement `rubydex = "=0.4.1"`
candidate versions found which didn't match: 0.4.2
```

.github/workflows/release.yml
alexcrocha: This is a fair question. The condition is on the step, not on the job. So the `verify_version` job will run and complete regardless of the conditional step being true or not. 

But now that you point it out, the previous steps will always run and that's wasteful.

I will look into an alternative approach for this

vinistock: No, it's fine. As long as it doesn't get stuck, this is good to ship.

lib/rubydex/version.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 module Rubydex
-  VERSION = "0.4.1"
+  # The gem and the crates are released under one version number, so it lives in rust/Cargo.toml under
+  # [workspace.package] and is read from there rather than declared here. Cargo requires its version to be a literal,
+  # so the manifest is the only place that can hold the single source of truth.
+  VERSION = File.read(File.expand_path("../../rust/Cargo.toml", __dir__))[/^version = "(.+)"/, 1]
```
rafaelfranca: Can we have a static version string please? Add a CI step to make sure the version string is the same and a rake task to make sure they are in sync, but having to look at another file to figure out the version is really bad.

rafaelfranca (changes requested):

lib/rubydex/version.rb
alexcrocha: Updated as suggested. The gem version is a static string again, and we'll catch in CI if there's any version drift

.github/workflows/ci.yml
```diff
@@ -24,6 +24,16 @@ jobs:
         with:
           bundler-cache: true
 
+      - name: Check versions are in sync
```
vinistock, line 27: Since we only bump versions when releasing, I think we can skip adding this to CI.

rakelib/cargo.rake
```diff
@@ -29,3 +32,34 @@ task :format_rust do
   sh "cargo clippy --all-targets --all-features --fix --allow-dirty", chdir: "rust"
   sh "cargo fmt", chdir: "rust"
 end
+
+desc "Sync the gem version and crate dependency with the Cargo workspace version"
+task :sync_versions do
```
vinistock, line 37: I'd move this to `Rakefile` since the version check is not exclusively about the Rust crates.

rakelib/cargo.rake
```diff
@@ -29,3 +32,34 @@ task :format_rust do
...
+  updated_files = replacements.to_h do |path, (pattern, replacement)|
+    contents = File.read(path)
+    abort("Could not find the version in #{path}") unless contents.sub!(pattern, replacement)
+
+    [path, contents]
+  end
+
+  updated_files.each do |path, contents|
+    File.write(path, contents)
+  end
```
vinistock, line 64: I think we can ship and give this a try, but I do wonder if it wouldn't be a bit more intuitive to have two tasks:

1. One for verifying that all versions match used by the release workflow
2. Another one for bumping the version across crates and the gem

rakelib/cargo.rake
alexcrocha: Good point, I moved it

rakelib/cargo.rake
alexcrocha: To move the current pr, let's explore this idea as a follow-up

PR #1047, opened 2026-09-20
Add typed namespace member lookup helpers

### LINKED ISSUE #1044: Type-safe member lookup
`Declaration#member` returns a `Rubydex::Declaration?` ([rbi](https://github.com/Shopify/rubydex/blob/d4b733d8848123d83ab244e435c071f9a0798c97/rbi/rubydex.rbi#L119-L120)). This is true in the general case, but in many real-world applications you're looking up specific hard-coded values, whose types are well-known:

E.g. from the Rubydex readme:

```ruby
declaration.member("bar()") # Should always be nil or `Rubydex::Method`
declaration.member("@ivar") # Should always be nil or `Rubydex::InstanceVariable`
```

I saw clunky code like this in the wild:

```ruby
initializer = state_class.member("initialize()")
if initializer.is_a?(Rubydex::Method)
  do_stuff(initailizer)
end
```

That `.is_a?(Rubydex::Method)` check is really intended to just check for `nil` or not. But redundantly checking the type is necessary to narrow static type from `Rubydex::Declaration` to `Rubydex::Method`, even though that'll always be the case at runtime.

I propose we add some type-specific accessors, like:

```ruby
class Rubydex::Declaration
  #: -> Rubdex::Method?
  def find_method(name)
    r = member("#{name}()") or return
    raise Rubdex::Error unless r.is_a?(Rubdex::Method)
    r
  end

  #: -> Rubdex::InstanceVariable?
  def find_instance_variable(name)
    r = member("@#{name}") or return
    raise Rubdex::Error unless r.is_a?(Rubdex::InstanceVariable)
    r
  end

  # ... and so on for the other types.
end
```

FranciscoMaxwell:
Closes #1044.

## Summary
- adds typed namespace member lookup helpers for methods, instance variables, and class variables
- normalizes both API-style names (`foo`, `name`, `class_var`) and internal member keys (`foo()`, `@name`, `@@class_var`)
- adds RBI signatures and regression coverage for direct and inherited lookup

## Validation
- `git diff --check`

I could not run the Ruby validation suite locally because Ruby and Bundler are not installed in this environment (`ruby` and `bundle` were not recognized). The project validation commands from `CONTRIBUTING.md` are `bundle exec rake compile`, `bundle exec rake ruby_test`, and `bundle exec rubocop`.


lib/rubydex/declaration.rb
```diff
@@ -25,6 +25,53 @@ class Namespace < Declaration
...
+      return unless member
+      return member if member.is_a?(Rubydex::ClassVariable)
+
+      raise Error, "Expected #{member_name} to resolve to Rubydex::ClassVariable, got #{member.class}"
+    end
+
+    private
+
+    #: (String) -> String
+    def format_method_name(name)
```
Morriar: Can we keep the input names exact here instead of normalizing them?

The design discussion in #1044 explicitly called out that this API should continue to accept Rubydex member names (`foo()`, `@name`, `@@class_var`) rather than alternate shapes like `foo` / `name`. If we support both forms here, consumers get a second naming convention that behaves differently from `member` / `find_member`.

I think these helpers should either delegate directly to `find_member(name, ...)` with the expected type check, or we should settle the API shape first.

rbi/rubydex.rbi
```diff
@@ -122,6 +122,15 @@ class Rubydex::Namespace < Rubydex::Declaration
   sig { params(name: String, only_inherited: T::Boolean).returns(T.nilable(Rubydex::Declaration)) }
   def find_member(name, only_inherited: false); end
 
+  sig { params(name: String, only_inherited: T::Boolean).returns(T.nilable(Rubydex::Method)) }
```
Morriar: Should this cover `member` too, or use the `expected_type:` shape discussed in #1044?

The motivating example there is direct lookup through `member("bar()")`, so adding only inherited `find_*` helpers still leaves that use case returning `Rubydex::Declaration?`. The issue discussion suggested narrowing the existing APIs, e.g. `member(name, expected_type: Rubydex::Method)` / `find_member(...)`, which would avoid adding one method per declaration kind while covering both direct and inherited lookups.

FranciscoMaxwell:
Assinei o CLA!

vinistock (changes requested):
Thank you for the PR, but as discussed by the team, we'd prefer [this approach](https://github.com/Shopify/rubydex/issues/1044#issuecomment-5687491719) to avoid having one API per each declaration type.

Also, we need that approach to be applied to both `member` and `find_member`.

ext/rubydex/declaration.c
```diff
@@ -177,29 +209,30 @@ static VALUE rdxr_declaration_find_member(int argc, VALUE *argv, VALUE self) {
...
+            only_inherited = RTEST(kwarg_vals[0]);
+        }
 
-        if (kwarg_val != Qundef) {
-            only_inherited = RTEST(kwarg_val);
+        if (kwarg_vals[1] != Qundef) {
+            expected_type = kwarg_vals[1];
+            Check_Type(expected_type, T_CLASS);
+            if (!RTEST(rb_class_inherited_p(expected_type, cDeclaration))) {
+                rb_raise(rb_eTypeError, "expected expected_type to be Rubydex::Declaration or a subclass");
```
vinistock: Passing `Declaration` as the expected type is the default behaviour, so let's tailor the error message for the use case of narrowing the type.
```suggestion
                rb_raise(rb_eTypeError, "expected expected_type to inherit from Rubydex::Declaration");
```

ext/rubydex/declaration.c
```diff
@@ -22,6 +22,46 @@ VALUE cMethod;
...
+static VALUE wrap_declaration(HandleData *data, const CDeclaration *decl, VALUE expected_type, const char *name) {
+    if (decl == NULL) {
+        return Qnil;
+    }
+
+    VALUE decl_class = rdxi_declaration_class_for_kind(decl->kind);
+    VALUE argv[] = {data->graph_obj, ULL2NUM(decl->id)};
+    free_c_declaration(decl);
+
+    if (!NIL_P(expected_type) && !RTEST(rb_class_inherited_p(decl_class, expected_type))) {
```
vinistock: Aren't we already checking the expected type earlier? Do we need to check here again?

Also, it seems like `validate_expected_type` is also performing the same check.

Let's unify the argument checking inside each function (for `find_member` and `member`), so that we fail there if necessary. Then `wrap_declaration` can be exclusively concerned with instantiating the Ruby object.

test/declaration_test.rb
```diff
@@ -698,6 +698,57 @@ def initialize
     end
   end
 
+  def test_member_lookup_accepts_expected_type
```
vinistock: Let's split this test into 2: one for `find_member` and one for `member`. Also, there's no need to verify that this works for methods, class variables, instance variables.

If it works for a single expected type, then the narrowing is working as expected. Let's simplify the example with a single case for the two APIs.

vinistock (commented):
This is the right direction. Just a few comments before shipping

FranciscoMaxwell:
Thanks for the review! I updated the PR based on your comments:

- adjusted the TypeError message to say that expected_type must inherit from Rubydex::Declaration;
- moved the expected_type argument validation into member and find_member, keeping wrap_declaration focused only on instantiating the Ruby object;
- split the regression coverage into separate tests for member and find_member, using a single expected type case to keep the examples smaller.

I also ran the native compile successfully and verified the focused tests for the updated behavior.

PR #1048, opened 2026-09-23
CLI and MCP for dead code detection

yaroslav-shopify:
CLI and MCP for dead code detection:

- Add `rdx dead-code` with table output via `terminal-table`, JSON output, `--path` filtering, and opt-in `--fail-on-candidates` for CI.
- Add the paginated `find_dead_code_candidates` MCP tool.
- Support `[dead-code].exclude` path patterns in `rubydex.toml` through the existing typed configuration and native bindings.

Filters and exclusions affect reported definitions while references across the full indexed graph still count. Exclusions apply before totals, pagination, and failure status.


vinistock:
The additions are generally good, but let's please split the work across a few PRs because there are too many decisions baked into the same change. It's easier to discuss each one of them with more focused PRs. For example, I'm not exactly sure if we need JSON output for the dead code CLI and I'd suggest we start out just printing a list of `file_path:line,col-line,col` with each candidate.

It may also be a bit early to introduce exclusion configuration before we gave the CLI a few tries to understand what's the most helpful way of interacting with this. Since the candidates are not 100% precise, we may need a way to exclude at a candidate level and not entire glob patterns.

Here's my breakdown suggestion for shipping this more quickly:

1. The CLI addition printing a plain text list of all identified candidates (`file_path:line,col-line,col`) with a summary at the end (similar to the linter, but without printing the source code). I'd also refrain from adding a `fail-on-candidates` at the moment since we don't have DSL support yet and that will result in lots of false positives. The command simply populates the graph, invokes `graph.dead_code_candidates` and formats the list
2. The MCP tool for reading dead code candidates. It basically does the same as the CLI (`graph.dead_code_candidates`), but then formats the output as JSON

After we learn a bit more from having those out, we can better shape what kinds of configurations we may need for this command. WDYT?

yaroslav-shopify:
@vinistock Gotcha, careful changes, smaller sets of changes, two PRs. Will do.

PR #1049, opened 2026-09-23
Add a dead-code CLI with plain-text output

yaroslav-shopify:
Add `rdx dead-code` to build the workspace graph, call `graph.dead_code_candidates`, and print each candidate definition as `file_path:line,col-line,col`. Results are sorted by location and end with `N files inspected, M offenses detected`, or `N files inspected, no offenses detected` when there are no candidates.

See also: https://github.com/Shopify/rubydex/pull/1048


lib/rubydex/cli/command/dead_code.rb
```diff
@@ -0,0 +1,56 @@
...
+        #: -> void
+        def run
+          parse_options!
+          abort_with_usage("unexpected argument: #{argv.first}") unless argv.empty?
+
+          require "rubydex/linter/helpers/path_helpers"
+
+          graph = build_graph($stderr, workspace_path: current_workspace_path)
+          candidates = graph.dead_code_candidates.to_a
+          locations = candidates.flat_map { |candidate| candidate.definitions.map(&:location) }
```
Morriar: Should we filter these to the workspace before flattening/counting? `build_graph#index_workspace` also indexes Bundler dependencies and core RBS paths, so this will report dead-code candidates from gems too. In a real app that can swamp the output with non-actionable offenses.

I’d expect `rdx dead-code` to only report definitions under `graph.workspace_path` and count candidates after that filter.

README.md
```diff
@@ -157,6 +157,18 @@ puts Rubydex::Query.schema("table")
...
+
+```bash
+bundle exec rdx dead-code
+```
+
+Builds the workspace graph and lists the definitions returned by `graph.dead_code_candidates` as
+`file_path:line,col-line,col`, followed by a summary of inspected files and offenses detected.
+Line and column numbers are one-based.
+Candidates are classes, modules, and constants with no detected references. DSL usage may result in false positives;
+finding candidates does not cause a nonzero exit status.
```
vinistock: These are user facing docs, so I'd remove some of the internal details and say something like this:


```suggestion
Prints a list of potentially dead code candidates. The list is composed of all declarations for which the analysis found no references for in the codebase.

Since some declarations may be used through meta-programming or otherwise untyped code, there may be false positives and the removal of candidates must be done carefully with context about the codebase.
```

lib/rubydex/cli/command/dead_code.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require "rubydex/cli/command"
+
+module Rubydex
+  module CLI
+    # `rdx dead-code` — lists dead code candidates in the current workspace.
+    class Command
+      class DeadCode < Command
+        command "dead-code"
```
vinistock: Nit

```suggestion
        command "dead_code"
```

lib/rubydex/cli/command/dead_code.rb
```diff
@@ -0,0 +1,56 @@
...
+          puts unless locations.empty?
+          print_summary(graph.documents.count, candidates.length)
+        end
+
+        private
+
+        #: (Integer file_count, Integer candidate_count) -> void
+        def print_summary(file_count, candidate_count)
+          if candidate_count.zero?
+            puts("#{file_count} #{pluralize("file", file_count)} inspected, no offenses detected")
```
vinistock:
```suggestion
            puts("#{file_count} #{pluralize("file", file_count)} inspected, no dead code candidates found")
```

lib/rubydex/cli/command/dead_code.rb
```diff
@@ -0,0 +1,56 @@
...
+            return
+          end
+
+          puts(
+            "#{file_count} #{pluralize("file", file_count)} inspected, " \
+              "#{candidate_count} #{pluralize("offense", candidate_count)} detected",
```
vinistock:
```suggestion
              "#{candidate_count} dead code #{pluralize("candidates", candidate_count)} found",
```

test/cli_test.rb
```diff
@@ -446,6 +450,116 @@ def test_lint_rejects_a_workspace_argument
...
+      result = rdx("dead-code")
+
+      assert_success_status(result)
+      assert_stdout_equals(<<~OUTPUT, result)
+        a.rb:1,1-2,4
+        a.rb:3,1-3,7
+        z.rb:3,1-4,4
+
+        3 files inspected, 3 offenses detected
+      OUTPUT
```
vinistock, line 484: Some ideas on the output:

1. Let's include an initial message first. Something like `Looking for potentially dead code...`
2. What do you think of including the declaration name and grouping the file paths under it? Something like this:

```
Foo
  first_foo.rb:1,1-2,4
  second_foo.rb:1,1-2,4

Bar
  only_bar.rb:1,1-2,4
```

lib/rubydex/cli/command/dead_code.rb
```diff
@@ -0,0 +1,56 @@
...
+
+          require "rubydex/linter/helpers/path_helpers"
+
+          graph = build_graph($stderr, workspace_path: current_workspace_path)
+          candidates = graph.dead_code_candidates.to_a
+          locations = candidates.flat_map { |candidate| candidate.definitions.map(&:location) }
+
+          locations.sort.each do |location|
+            display = location.to_display
+            path = Rubydex::Linter::Helpers::PathHelpers.display_path(display, workspace: graph.workspace_path)
```
vinistock: For a future PR: I think it makes sense to extract these path helpers as CLI helpers. They are not really linter specific, but rather CLI-presentation specific.

vinistock (commented):
A few adjustments, but this looks great

vinistock (approved):

PR #1050, opened 2026-09-23
Expose dead-code candidates through MCP

yaroslav-shopify:
Adds a `find_dead_code_candidates` MCP tool backed by `graph.dead_code_candidates`.

Each candidate includes its fully qualified name, kind, and all definition locations with paths and one-based line numbers. Results are sorted by name and support `limit` and `offset` pagination, with a total count.


README.md
```diff
@@ -198,9 +198,15 @@ like Claude to semantically query your Ruby codebase.
...
 | `find_constant_references` | All precise, resolved constant references across the codebase |
+| `find_dead_code_candidates` | Classes, modules, and constants with no detected references |
 | `get_file_declarations` | List declarations defined in a specific file |
 | `codebase_stats` | High-level statistics about the indexed codebase |
 
+`find_dead_code_candidates` returns `candidates` and `total`. Each candidate includes its `name`, `kind`, and all
+definition `locations` with a `path` and one-based `line`. Use `limit` (default 50, maximum 100) and `offset` to page
+through results. The tool reports all candidates returned by `graph.dead_code_candidates`, including dependencies.
+DSLs and metaprogramming can produce false positives, so candidates need review before removal.
+
```
vinistock: I think this part can be removed.

README.md
```diff
@@ -198,9 +198,15 @@ like Claude to semantically query your Ruby codebase.
 | `get_declaration` | Full details by fully qualified name with docs, ancestors, members |
 | `get_descendants` | What classes/modules inherit from or include this one |
 | `find_constant_references` | All precise, resolved constant references across the codebase |
+| `find_dead_code_candidates` | Classes, modules, and constants with no detected references |
```
vinistock: Currently, only constants are supported, but let's make the docs a bit more generic


```suggestion
| `find_dead_code_candidates` | List of potentially unused code |
```

test/mcp_server_tools_test.rb
```diff
@@ -89,6 +89,209 @@ def test_find_constant_references_tool
...
+            },
+          ],
+          "total" => 2,
+        },
+        result,
+      )
+    end
+  end
+
+  def test_find_dead_code_candidates_includes_ruby_and_rbs_definitions
```
vinistock: This test is a bit overkill. Dead code detection happens at the graph level, so the original file type where we encountered the definitions has no influence in the outcome.

test/mcp_server_tools_test.rb
```diff
@@ -89,6 +89,209 @@ def test_find_constant_references_tool
...
+            "candidates" => [
+              { "name" => "Unused", "kind" => "Class", "locations" => [{ "path" => "app.rb", "line" => 2 }] },
+            ],
+            "total" => 1,
+          },
+          result,
+        )
+      end
+    end
+  end
```
vinistock: Same for these 3, I don't think they add much in terms of verifying functionality specific to the tool.

vinistock (approved):
A few adjustments, but this looks good to me. @st0012, you have more context on how to tune these. WDYT?

st0012 (approved):
I think the tool description is too verbose and captures unnecessary information like input schema, which is already exposed in `tools/list` response.
BUT this pattern was actually introduced by my MCP server prototype, which means they're present in existing tools too. So let's merge this first and then I'll open a separate PR to address them.

vinistock (approved):

